PR #22, opened 2016-11-23
Glow flag/stop button when scripts are running

rschamp:
Requires https://github.com/LLK/scratch-vm/pull/348

For now this uses the CSS `filter` property to crudely "highlight" for the active state. This is a temporary solution until we have a real design for the active states.


CosmicWebServices:
Yay!

PR #23, opened 2016-11-29
Don't block keys intended for the browser

rschamp:
This is mostly because it was annoying that Cmd-Opt-J didn't work, but seems like a good thing to have in general.


tmickel:
Not sure about this:
https://scratch.mit.edu/projects/133279734/

rschamp:
Ah — maybe we should let it bubble if those keys are there?

Issue #14, opened 2016-10-31
Implement sprite selector


tmickel:
There is a `<CostumeCanvas />` component which should help with that. All the data from the VM should be present, except maybe the current costume per target.

rschamp:
Initial implementation is done, this will continue to get refined.

PR #25, opened 2016-12-07
Sprite selector, CSS and Redux overhaul

rschamp:
Believe it or not this a WIP PR. But it's getting out of hand so I figured I should submit it for review now before I do anything more.

Towards #14.
Requires https://github.com/LLK/scratch-vm/pull/353.

This is probably better suited for a walk-through so let me know if you want to do that instead of trying to figure out everything this does.

* Add `TargetSelector` component, comprised of `SpriteSelector` and `StageSelector`. Eventually it will also contain the "sprite info pane" in some form.
* Refactor the various media library modals/buttons to be the responsibility of the `TargetSelector`
* Display costumes/backdrops for each sprite and the stage in the `SpriteSelector` and `StageSelector`, respectively.
* Use redux to centralize the VM state for use in various components. In this case, the `StageSelector` and `SpriteSelector` should share the VM's target list state. This state requires listening to both `SPRITE_INFO_REPORT`s and `targetsUpdate` events from the VM and reducing them into a single object. After I wrote the logic for this in the component I realized I was just poorly re-inventing redux, so I switched to using redux for this. After doing that, I realized it could also take the responsibility of the "VM manager" object.
* Apply redux to the modals, just because it simplified the structure of the `TargetSelector` component.
* Stop using inline styles, use CSS modules instead.  I gave inline styles a shot, but it just seemed like it wasn't going to scale well, and it's not very user-friendly to have CSS written in JS.  The local scoping that CSS modules provide give the same benefits of inline styles without sacrificing ergonomics.
* Do a pass on the layout of the GUI: there just wasn't room for the sprite selector with the green flag and stop buttons where they were, so I moved things around.
* Give the `GUIComponent` responsibility over the overall layout. With the GUI container controlling the set of components, eventually I would need to be putting layout styles in the container, which isn't its responsibility. So move this to the GUI component instead.


src/components/gui/gui.jsx
```diff
@@ -0,0 +1,90 @@
...
+                {...stopAllProps}
+            />
+            <Stage
+                vm={vm}
+                {...stageProps}
+            />
+            <TargetPane
+                mediaLibrary={mediaLibrary}
+                vm={vm}
+                {... targetPaneProps}
```
rschamp: Nope, but I should make that consistent.

src/components/modal/modal.jsx
```diff
@@ -0,0 +1,34 @@
...
+                isOpen={this.props.visible}
+                overlayClassName={styles.modalOverlay}
+                ref={m => (this.modal = m)}
+                onRequestClose={this.props.onRequestClose}
+            >
+                <div
+                    className={styles.modalCloseButton}
+                    onClick={this.props.onRequestClose}
+                >
+                    {'x'}
```
rschamp: The only reasons are for editors, explained here https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-literals.md

No reason in this case. We may have had that rule turned on when @tmickel did the first pass of the modal.

rschamp:
> * How would you feel about going further in an event-driven (or Redux-style) direction? I'm looking at the big list of forwarded props in the `TargetPane` component and wishing that `TargetPane` didn't have to know about all of them...

Yeah this was something that was annoying me too. It would be nice if the containers/components could remain self-contained and only list the PropTypes for the props that they explicitly use. But because of the pattern where a container passes props to the component, the containers effectively require the props of their components.

If I blindly pass the `... rest` props to the component from the container, and don't list them in `propTypes` then the linter won't complain. If you then failed to pass all the required components to the container, the component would throw the error. That's more correct than the container throwing the error since the component is the one missing props. On the other hand, it would be harder to figure out why that's happening since the error would come from a deeper level than the source of the problem... So I'm not sure which is better.

src/components/sprite-selector/sprite-selector.jsx
```diff
@@ -0,0 +1,57 @@
...
+        ...componentProps
+    } = props;
+    return (
+        <div
+            className={styles.spriteSelector}
+            {...componentProps}
+        >
+            {Object.keys(sprites)
+                // Only render ready sprites
+                .filter(id => 'costume' in sprites[id] && 'order' in sprites[id])
```
tmickel: can you explain this line a little more? do you know why some sprites are missing these properties?

src/components/target-pane/target-pane.jsx
```diff
@@ -0,0 +1,116 @@
...
+            <SpriteSelectorComponent
+                selectedId={editingTarget}
+                sprites={sprites}
+                onSelectSprite={onSelectSprite}
+            />
+            <StageSelector
+                backdropCount={stage.costumeCount}
+                id={stage.id}
+                selected={stage.id === editingTarget}
+                url={stage.costume && stage.costume.skin}
```
tmickel: I know what this is going for (return stage.costume.skin if stage.costume is non-null) but it does read a little oddly :P

src/components/sprite-selector/sprite-selector.jsx
rschamp: Good point, this is unnecessary now. Previously the targetUpdates would only return name and id, so costume was missing until the sprite info report.

src/components/target-pane/target-pane.jsx
```diff
@@ -0,0 +1,116 @@
...
+                onSelect={onSelectSprite}
+            />
+            <p className={styles.targetPaneLibraryButtons}>
+                <button onClick={onNewSpriteClick}>New Sprite</button>
```
tmickel: Just a note, maybe soon is the time we might consider working on a localization plan for scratch-gui and scratch-blocks? doing it after the fact with ScratchJr was no fun :)

src/components/target-pane/target-pane.jsx
liam4: `stage.costume ? stage.costume.skin : ''` might read better :P

rschamp: For me this is kind of idiomatic React, e.g. https://facebook.github.io/react/docs/conditional-rendering.html#inline-if-with-logical-ampamp-operator. But I agree it's messy. It should probably just be handled better in the way that the data is passed to the component, so you only have the attributes you need. That or we just don't render things that don't have costume data yet.

src/components/target-pane/target-pane.jsx
rschamp: For scratch-gui at least, we can probably use the same system we use for www.

src/components/modal/modal.jsx
rschamp: Leaving this as-is for now, when I changed it I saw one other reason, which is that it's really easy to miss the stray `x` on that line. It's much more visible this way.

PR #26, opened 2016-12-15
Pass for overall layout

rschamp:
Move layout styles from component stylesheets to the GUI, or whichever component is placing the component.

Add a new `Box` component to control layouts, basically giving a convenient way to lay out with flexbox.

Use `Box` as the output component for any component that may need to be passed layout options. The props passing convention allows giving layout props to the final `Box` component, making wrapping with another `Box` unnecessary in some cases.


rschamp:
Reviewed by @cwillisf in #28

PR #28, opened 2016-12-16
Update in/active states for flag and stop

### LINKED ISSUE #15: Green flag/stop button
Copying this as it's a line item from the prototype plan. I'm assuming it just means getting the styling right.

rschamp:
Resolves #15. Depends on #26.


src/components/box/box.jsx
```diff
@@ -0,0 +1,119 @@
...
+        return function () {
+            mZ = ((36969 * (mZ & 65535)) + (mZ >> 16)) & mask;
+            mW = ((18000 * (mW & 65535)) + (mW >> 16)) & mask;
+            let result = ((mZ << 16) + mW) & mask;
+            result /= 4294967296;
+            return result + 1;
+        };
+    }(601));
+    return function () {
+        const r = Math.max(parseInt(random() * 100, 10) % 256, 1);
```
cwillisf: Not actually important: you can `& 255` instead of `% 256` and get a conversion to integer for free :)

src/components/blocks/blocks.jsx
```diff
@@ -1,25 +1,21 @@
...
+        ...componentProps
+    } = props;
+    return (
+        <Box
+            className={styles.blocks}
+            componentRef={componentRef}
+            {...componentProps}
+        />
+    );
+};
```
cwillisf: Just out of curiosity: what was the reason to convert the style of these components?

src/components/target-pane/target-pane.jsx
```diff
@@ -113,4 +132,8 @@ TargetPane.propTypes = {
     vm: React.PropTypes.instanceOf(VM)
 };
 
+TargetPane.defaultProps = {
+    mediaLibrary: new MediaLibrary()
```
cwillisf: What's the lifetime for this? Specifically, does it survive this component or the parent component re-rendering?

webpack.config.js
```diff
@@ -52,10 +53,15 @@ module.exports = {
             loader: 'json-loader'
         }]
     },
-    postcss: [autoprefixer],
+    postcss: [
+        autoprefixer({
+            browsers: ['last 3 versions', 'Safari >= 8', 'iOS >= 8']
```
cwillisf: This could be a discussion for outside this PR, but I have mixed feelings about putting `'last 3 versions'` in here. By my understanding, putting that here (and presumably the `webpack.config.js` for other modules) means that unless our various modules have the same version of `autoprefixer` they might target different browsers.

I guess that's "easy" to fix by making sure we standardize on a particular version of `autoprefixer`... is that the way to go in your opinion?

src/components/box/box.jsx
rschamp: Hah, good to know — in case you couldn't already tell, this was copy-pasted. I don't know how long it will stick around.

src/components/blocks/blocks.jsx
rschamp: This came from trying to adhere to the [react/prefer-stateless-function](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-stateless-function.md) warning, which in turn exists because the React team has suggested that stateless components written this way will receive performance enhancements: https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#stateless-functional-components in addition to other nice things about them being functions: https://medium.com/@housecor/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc

src/components/target-pane/target-pane.jsx
rschamp: Although I've never confirmed this assumption, I am pretty sure this survives rerendering of the component, since the VM is instantiated in the same way and doesn't get recreated on each render. I haven't confirmed this but it should act as a static, since it's only instantiated once for the class. If you don't override it, it should be reusing the same instance for each component.

webpack.config.js
rschamp: You're right that we probably should standardize on a version of autoprefixer across all the repos that use it. In addition to that, we should probably also define an explicit browser target rather than just "whatever `last 3 versions` means right now".

src/components/box/box.jsx
liam4: It'll stick around as long as the general surrounding code sticks around.. likely forever :)

PR #50, opened 2017-01-28
Add Tabs for Costumes and Sounds

TheBrokenRail:
### Resolves

_Tiny bit of #1_

### Proposed Changes

_Adds tabs above the categories menu like in the concept image. NOTE: They don't actually do anything._

### Reason for Changes

_Make it easier to implement the sound and costume tabs._


TheBrokenRail:
### Before
before
### After
after

src/components/gui/gui.jsx
```diff
@@ -46,15 +46,19 @@ const GUIComponent = props => {
...
-                    </Tab>
-                    <Tab>
-                        <TabText>Sounds</TabText>
-                    </Tab>
+                    <Box
+                        width={250}
+                    >
+                        <TabSelected>
+                            <TabText>Scripts</TabText>
+                        </TabSelected>
```
liam4: I'm not a React developer, let alone a scratch-gui dev, but wouldn't it make sense to have some kind of state and/or class to control whether a tab is selected or not, rather than a whole new tag? (Likely `Tab` and `TabSelected` will have nearly the same behaviour, aside from styling.)

TheBrokenRail:
I don't know enough about React to do it, and yes they basically are the same.


________________________________

TheBrokenRail:
Figured out how to do it with a attribute!

src/components/gui/gui.jsx
```diff
@@ -49,9 +49,11 @@ const GUIComponent = props => {
                     <Box
                         width={250}
                     >
-                        <TabSelected>
+                        <Tab
+                            selected={true}
```
liam4: Great! Now to make tab styles rely on the state, you just need to store some property on the state of the GUI that says which tab is selected.

```jsx
const selected = this.state.selected

// ..other tag things..

<Tab selected={selected === 'scripts'}>
  <TabText>Scripts</TabText>
</Tab>

<Tab selected={selected === 'costumes'}>
  <TabText>Costumes</TabText>
</Tab>

<Tab selected={selected === 'sounds'}>
  <TabText>Sounds</TabText>
</Tab>
```

liam4: ..and of course, you'd have to make it interactive, as well, but I'm not a Professional React Developer who would know how to (really) do that.

src/components/tab/tab.jsx
```diff
@@ -48,7 +48,7 @@ const Tab = props => {
         ...componentProps
     } = props;
     return React.createElement(element, {
-        className: classNames(styles.tab, className),
+        className: classNames(eval("styles.tab" + ((props.selected == true) ? "Selected" : "")), className),
```
liam4: Eval is generally a bad idea. I'm actually not sure why eval is even needed here?

You could do something like this:

```js
const style = (props.selected == true) ? styles.tabSelected : styles.tab

// or this, if you wanted (it's more similar to the way you used eval)
const style = styles['tab' + (props.selected == true ? 'Selected' : '')]
// but I'd go for the first.
```

src/components/gui/gui.jsx
```diff
@@ -12,6 +12,16 @@ const Box = require('../box/box.jsx');
...
 
+function tabOne {
+    this.setState({tabOne: true, tabTwo: false, tabThree: false});
+}
+function tabTwo {
+    this.setState({tabOne: false, tabTwo: true, tabThree: false});
+}
+function tabThree {
+    this.setState({tabOne: false, tabTwo: false, tabThree: true});
+}
```
liam4: This is a syntax error (should be `function foo() {`, not `function foo {`).

You should test your code on your own computer before you make a pull request.. :P

TheBrokenRail:
Yes! The tabs can now be switched, it doesn't actually do anything but a few box components would probally fix that!

TheBrokenRail:
Everything seems to be working now!

thisandagain:
Thanks for trying out a solution to this, but again we are in the process of implementing the GUI "chrome" now. It's great if you want to explore ideas like this in your fork, but creating PRs in this way creates a lot of extra work for everyone on the project. Continuing to ignore the contribution guidelines will result in your account being blocked.

PR #51, opened 2017-01-30
Make Compatibility Mode on By Default

TheBrokenRail:
### Resolves

_#41_

### Proposed Changes

_Make 30TPS on by default._

### Reason for Changes

_For compatibility testing._


thisandagain:
Thanks!

PR #61, opened 2017-02-02
Cache stage bounding box and avoid unnecessary library rendering

rschamp:
This gets scratch-gui a bit more in line performance-wise with the scratch-vm playground.

Caching the stage bounding box prevents layout recalculations on every mouse event.
Don't render the contents of the libraries until they're actually visible. This was especially expensive now that we're listening to every `spriteInfoReport`. Further optimizations to the VM event listeners should come later.

### DIFF
--- a/src/components/library/library.jsx
+++ b/src/components/library/library.jsx
@@ -22,6 +22,7 @@ class LibraryComponent extends React.Component {
         this.setState({selectedItem: id});
     }
     render () {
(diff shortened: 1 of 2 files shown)


rschamp:
To be honest it may not be, just something the profiler pointed out while I was looking for differences between the vm and the gui.  It complained about the forced reflow more than the function call itself.

PR #76, opened 2017-02-07
Monkeypatch dynamic sound menu into ScratchBlocks

ericrosenbaum:
/ht @rschamp

### Resolves

Addresses https://github.com/LLK/scratch-audio/issues/25

### Proposed Changes

Update the Blocks component to receive a patched version of ScratchBlocks. The patched version receives a reference to the VM from the Blocks component. 

### Reason for Changes

We use the reference to the VM to create a dynamic Blockly menu containing the names of the sounds of the editing target.

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -1,7 +1,7 @@
 const bindAll = require('lodash.bindall');
 const defaultsDeep = require('lodash.defaultsdeep');
 const React = require('react');
-const ScratchBlocks = require('scratch-blocks');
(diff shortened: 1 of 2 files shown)


TheBrokenRail:
This is a great way to implement it but new sounds won't display in the menus for blocks until you reload them (switch to another categories and back for toolbox, and switch to another sprite and back for workspace) which is weird.

rschamp:
Good catch @TheBrokenRail. @ericrosenbaum is already looking into the best way to do this refresh, because his previous method didn't seem to work with the GUI.

PR #81, opened 2017-02-11
Add sprites to drop-down menus

TheBrokenRail:
### Resolves

_Part of LLK/scratch-blocks#749_

### Proposed Changes

_Adds sprites to drop-down menus in the same way #80 did_

### Reason for Changes

_So you can actually select sprites in drop-down menus to interact with them._

### DIFF
--- a/src/lib/blocks.js
+++ b/src/lib/blocks.js
@@ -39,7 +39,7 @@ module.exports = function (vm) {
         var k = 0;
         for (i = 0; i < vm.runtime.targets.length; i++) {
             if (vm.runtime.targets[i].isOriginal == true) {
-                if (vm.runtime.targets[i].isStae == false) {
+                if (vm.runtime.targets[i].isStage == false) {
                     sprites[k] = [];
                     sprites[k][0] = vm.runtime.targets[k].sprite.name;
                     sprites[k][1] = vm.runtime.targets[k].sprite.name;
@@ -71,8 +71,8 @@ module.exports = function (vm) {
         this.jsonInit(json);
     };
     
-    ScratchBlocks.Blocks.motion_pointtowards_menu = function () {
-        const json = jsonForMenuBlock('TOWARDS', spriteMenu, motionColors, []);
+    ScratchBlocks.Blocks.motion_pointtowards_menu.init = function () {
+        const json = jsonForMenuBlock('TOWARDS', spriteMenu, motionColors, [['mouse-pointer', '_mouse_']]);
         this.jsonInit(json);
     };
 
--- a/src/lib/blocks.js
+++ b/src/lib/blocks.js
@@ -34,19 +34,15 @@ module.exports = function (vm) {
     };
     
     const spriteMenu = function () {
-        var i = 0;
         var sprites = [];
-        var k = 0;
-        for (i = 0; i < vm.runtime.targets.length; i++) {
-            if (vm.runtime.targets[i].isOriginal === true) {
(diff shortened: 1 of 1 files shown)


src/lib/blocks.js
```diff
@@ -32,23 +32,100 @@ module.exports = function (vm) {
...
+        for (i = 0; i < vm.runtime.targets.length; i++) {
+            if (vm.runtime.targets[i].isOriginal === true) {
+                if (vm.runtime.targets[i].isStage === false) {
+                    if (vm.runtime.targets[i].sprite.name === vm.editingTarget.sprite.name) {
+                        continue;
+                    }
+                    sprites[k] = [];
+                    sprites[k][0] = vm.runtime.targets[i].sprite.name;
+                    sprites[k][1] = vm.runtime.targets[i].sprite.name;
+                    k++;
```
rschamp: You can omit your `k` counter if you use this instead:
```
sprites.push([vm.runtime.targets[i].sprite.name, vm.runtime.targets[i].sprite.name]);
```

src/lib/blocks.js
```diff
@@ -32,23 +32,100 @@ module.exports = function (vm) {
     const backdropsMenu = function () {
         return vm.runtime.targets[0].sprite.costumes.map(costume => [costume.name, costume.name]);
     };
+    
+    const spriteMenu = function () {
+        var i = 0;
+        var sprites = [];
+        var k = 0;
+        for (i = 0; i < vm.runtime.targets.length; i++) {
+            if (vm.runtime.targets[i].isOriginal === true) {
```
rschamp: It's ok to just use `if (vm.runtime.targets[i].isOriginal)`

src/lib/blocks.js
```diff
@@ -32,23 +32,100 @@ module.exports = function (vm) {
...
+                if (vm.runtime.targets[i].isStage === false) {
```
rschamp: And here, `if (!vm.runtime.targets[i].isStage)`.

src/lib/blocks.js
```diff
@@ -32,23 +32,100 @@ module.exports = function (vm) {
...
+                    if (vm.runtime.targets[i].sprite.name === vm.editingTarget.sprite.name) {
```
rschamp: It would be safer to use the sprite `id` rather than its `name`, since that is guaranteed to be unique.

src/lib/blocks.js
```diff
@@ -32,23 +32,100 @@ module.exports = function (vm) {
     const backdropsMenu = function () {
         return vm.runtime.targets[0].sprite.costumes.map(costume => [costume.name, costume.name]);
     };
+    
+    const spriteMenu = function () {
+        var i = 0;
+        var sprites = [];
+        var k = 0;
+        for (i = 0; i < vm.runtime.targets.length; i++) {
```
rschamp: Please use `for (var targetId in vm.runtime.targets)` since then we won't need these iterator variables.

rschamp: I should add, if you use that method, then it should be followed by
```javascript
if (!vm.runtime.targets.hasOwnProperty(targetId)) continue;
let target = vm.runtime.targets[targetId];
```

TheBrokenRail:
I updated the code.

rschamp:
Thanks @TheBrokenRail!

TheBrokenRail:
This doesn't seem to be working now since the new changes, and I have no idea why.

TheBrokenRail:
Found the issue!

PR #114, opened 2017-02-20
Use large range on true "dev" dependencies

rschamp:
The exact versions of these dependencies are not important for building, so give them large ranges in order to quiet down Greenkeeper.

I did not split them between dependencies and devDependencies because a package's dependencies are installed whenever the package is installed as a dependency. Installing all of these would be obnoxious and unnecessary.

Also remove some unneeded dev dependencies.


thisandagain:
@rschamp I fixed up the merge conflicts for you. Sorry about that ... I should have waited before merging in some changes from Greenkeeper.

rschamp:
Hah I probably should have merged this a while ago, got behind.

PR #125, opened 2017-02-21
Use postcss-simple-vars for variables

rschamp:
Using postcss-modules-values with imported values failed on our current basic usage; it is apparently not mature enough to be used yet.

The syntax used by postcss-simple-vars is already familiar to most people familiar with modern CSS build systems.

Just a note - there were a lot of lines with trailing spaces in `gui.css` that my editor automatically fixed. To view the changes without those whitespace changes, add `?w=1` to the end of the URL.


lifeinchords:
Looks great, yay vars!
Will clean trailing spaces from now on.

Issue #100, opened 2017-02-15
Backdrop "+" should open backdrop library


thisandagain:
### Expected Behavior

- The backdrop library opens

### Actual Behavior

- The costume library opens

### Steps to Reproduce

- Load GUI
- Press the "+" button in the backdrop area (bottom right)

### Operating System and Browser

Mac OS 10.12.2 Chrome Version 56.0.2924.87 (64-bit)

PR #134, opened 2017-02-26
Stage/Backdrop: Design pass

### LINKED ISSUE #65: Stage/Backdrop Area
We need to refine the design of this pane

lifeinchords:
Related #65 
Related #100 

* Stage/Backdrop: design pass. Changed how selected state works to keep width of pane from shifting. Switched to fixed pixel width to match design spec
* SpriteSelector: removed flex basis width. Pane grows to expand the empty space, since Stage is now fixed width
* SpriteInfo: font sizing closer to spec

----

Questions:

* The header in this pane felt pretty simple, compared to the SpriteInfo pane. Ok to leave, or better to make it a component too?

* I had trouble rendering the `border-width` along with `border-radius` on the `canvas` element directly, without causing a few pixels of extra margin between the rendered image and the border. Any previous runins with that issue elsewhere, looks like this: 
screen shot 2017-02-26 at 2 07 50 pm

*  Then,`box-shadow: inset...` alone didn't work, as the image hides the inset. I used [this method](https://stackoverflow.com/questions/21414925/why-doesnt-inset-box-shadow-work-over-images) as a base, using a pseudo selector and overlapping the border over the image. That the h/w is related to what's set in the JSX is not great, but know of a better way? Worth adding this as a comment?

* Dropped the component name prefix in the class names to make the rendered classname a bit easier to read, like `.stage-selector_costume-canvas_2L_6h`. Is this a good pattern to continue, assuming CSS Modules prevents conflicts across components, or does that leave the CSS source classnames too genetic now, ie, `title`, `label` etc? 

* Related #100: Clicking the stage/backdrop image, then clicking + Import button, correctly opens the Stage/Backdrop Library. When a SelectorItem is selected though, how do we get React to open the backdrop library when the button is clicked?

* I also tried an alt way of showing Stage/Backdrop is selected. /cc @carljbowman thoughts?
screen shot 2017-02-26 at 1 15 07 pm

[Compare branches](https://github.com/lifeinchords/scratch-gui/compare/stage--gh100-gh65...lifeinchords:stage--gh100-gh65--alt)


lifeinchords:
Updated design. Note the space between border and canvas element when stage is selected.

screen shot 2017-02-27 at 3 40 58 pm

lifeinchords:
@rschamp Sounds good. I'll make a Todo to reuse those styles. Will file an issue for the stage thumbnail.

PR #152, opened 2017-03-02
Wire sprite info panel to VM

rschamp:
Sets up sprite info component to reflect data about the selected sprite in the panel.

In the interest of time:
* Every field is updated as you type, including x/y position. Typing `-` is interpreted as `-1`. Eventually this should only happen on blur or enter key pressed.  Good luck trying to update position while a sprite is moving 🙃 
* When you select the stage, the sprite info pane becomes disabled, with placeholder values set in the inputs
* Everything is handled with props passing. This is a good place for redux, but I needed it to be done for the play test tomorrow.
* I did minimal performance testing. I would not be surprised if some projects perform more slowly after this update.  I believe the redux refactor could help with this, since it would reduce the number of parent components that would need to re-render as sprite info changes.

/cc @lifeinchords @cwillisf @carljbowman

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "react-redux": "5.0.2",
     "react-style-proptype": "2.0.1",
     "redux": "3.6.0",
+    "redux-throttle": "0.1.1",
     "scratch-audio": "latest",
     "scratch-blocks": "latest",
     "scratch-render": "latest",
--- a/src/components/sprite-info/sprite-info.css
+++ b/src/components/sprite-info/sprite-info.css
@@ -36,29 +36,27 @@
 }
 
 /* TODO: refactor from descendant selector, to a regular class */
-.group img { 
+.group img {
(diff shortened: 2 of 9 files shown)


thisandagain:
### Validation of Sprite Names
One issue that may be ok to resolve later if we file a bug is that it is possible in the PR to have sprites with `''` as a name as well as multiple sprites with the same name. This will become important to resolve once we complete work on blocks like `point towards [x]` and `go to [x]`.

image

### Negative Integers in X / Y Position Fields
If you start typing a negative integer (e.g. `-50`) into the `x` or `y` position fields it will always prefix with `-1` which leads to unexpected results and can be very frustrating (`-50` turns into `-150`).

rschamp:
I fixed the performance issues altogether! Hahahaha!

rschamp:
And agreed on the other two points.

Issue #12, opened 2016-10-26
Design for sprite, costume, and backdrop libraries


tmickel:
The implementation should be at a point where some design can be applied, and this is needed for the prototype.

thisandagain:
Let's split this into two passes:
- [ ] Phase I: Basic library for February play test
- [ ] Phase II: Full library design

lifeinchords:
TODO: Safari overflow scroll in SpriteSelector pane

lifeinchords:
TODO: 6 sprites per row, vs 4

lifeinchords:
- [ ] Add a bit of space between rows in sprite library. On hover reveals them side overlapping

lifeinchords:
Related: #89

lifeinchords:
- [ ] Need space between items across rows on hover
- [ ] Width on items are not the same across rows

![12e8bd40-f1e1-11e6-882f-d0b89627eb8a](https://cloud.githubusercontent.com/assets/716168/23275561/f370e310-f9d4-11e6-88ce-2188fc33e154.png)

lifeinchords:
@carljbowman @thisandagain Should we close this? I think these are covered in  LLK/scratch-gui#355 and LLK/scratch-gui#391

carljbowman:
@lifeinchords - Yeah, let's go ahead and close this issue as we have completed an initial pass. We can create issues as needed as we continue to explore filtering and sprite tag/pack stuff.

PR #160, opened 2017-03-03
Libraries: Design pass 

lifeinchords:
Related #12 

- Design pass on libraries to make sprites look more like they do in the SpriteSelector. Fixed scrolling overflow when reaching the bottom

- Started a colors variables file

- Added a light transform on hover. Tried adding similar transform to SpriteSelectorItems, but antialiasing is too blurry for the image + text at that size

- Spacing between sprites is not uniform; deferring adding a proper tile grid with a gutter, until next iteration. Experiments with tile + waffle grids here: https://github.com/LLK/scratch-gui/compare/develop...lifeinchords:libraries-grid--gh12

#### Questions
- Is it possible to make the images all the same size? If we could, implementing a tile grid would make things lots simpler and fix some sizing quirks
- Are we considering paging via scroll? This could performance because not all the sprites need to render be inserted at once on load
- There's some weirdness in momentum scrolling in Chrome. Feels like it gets cut off short. Not sure if this is related to this PR or already existed

#### Todo
- The sprites on the last row don't flow well
- Should we make room explicitly in the design for scroll bar? it's doing weird things on hover


src/css/colors.css
```diff
@@ -0,0 +1,2 @@
+$ui-pane-gray: #f9f9f9;
+$blue: #4c97ff;
```
rschamp: Nit — newline at end of file.

lifeinchords: fixed

lifeinchords:
@rschamp @thisandagain Added some todo's + questions to description

PR #177, opened 2017-03-03
Buffer sprite info text fields

### LINKED ISSUE #154: Update sprite info on blur/submit
Currently all sprite info is updated onChange. That's a bit annoying for position, especially because `-` becomes `-1`.  

* Do validation/updating on blur
* Stop updating the inputs while the user has focus, so they can finish typing

rschamp:
Changes to the name, x, or y fields should only be sent to the VM when the field is blurred or when enter is pressed.

Still allow updates while the fields are focused, but unchanged.

Fixes #154

/cc @carljbowman

### DIFF
--- a/src/containers/sprite-info.jsx
+++ b/src/containers/sprite-info.jsx
@@ -25,7 +25,10 @@ class SpriteInfo extends React.Component {
         };
     }
     handleKeyPress (e) {
-        if (e.key === 'Enter') this.handleFlush();
+        if (e.key === 'Enter') {
+            this.handleFlush();
+            e.target.blur();
+        }
     }
     handleFlush () {
         if (this.state.name !== null) {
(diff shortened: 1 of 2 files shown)


paulkaplan:
@rschamp not sure, but it might be worth extracting the logic out into a reusable component that just handles buffered/validated inputs.

src/containers/sprite-info.jsx
```diff
@@ -25,7 +25,10 @@ class SpriteInfo extends React.Component {
         };
     }
     handleKeyPress (e) {
-        if (e.key === 'Enter') this.handleFlush();
+        if (e.key === 'Enter') {
+            this.handleFlush();
+            e.target.blur();
+        }
```
liam4: What's the purpose of blurring the input as soon as you press enter? Just curious.. :)

rschamp: That came from a design review of the feature. We thought it would be more expected for the field to lose focus when you hit enter, otherwise it is not as obvious that hitting enter did anything.

Now that you mention it, I think the blur is the only thing that we need to do, since the flushing will happen on blur anyway...

PR #212, opened 2017-03-06
Makes flags non-selectable + clickable

lifeinchords:
This fix still doesn't prevent ghost image dragging:

ghost-flag mov

 I think one solve would be to move the `src` from `img`, into the style as a `background-image`. If so, should we do this with other UI elements around the app?


src/components/green-flag/green-flag.css
```diff
@@ -3,6 +3,8 @@
     height: 1.1rem;
     opacity: 0.5;
     margin: 0.25rem 0.6rem;
+    user-drag: none;
```
rschamp: If the elements are still draggable, does the `user-drag` property give us anything?

lifeinchords: Nope, my mistake. I tried a few different strategies and had that leftover. Removed

Issue #105, opened 2017-02-16
[Safari] Sprite and stage selectors don't extend to bottom


towerofnix:
### Expected Behavior

The sprite and stage selectors (below the stage) extend to the bottom of the screen, as it looks in Chrome:

![Chrome screenshot](https://cloud.githubusercontent.com/assets/9948030/23023615/d105a19c-f42c-11e6-9d5d-09250973b472.png)

### Actual Behavior

The sprite and stage selectors don't extend to the bottom of the screen:

![Safari screenshot](https://cloud.githubusercontent.com/assets/9948030/23023650/f1d9cc04-f42c-11e6-82b9-cc124b6cc4b5.png)

### Steps to Reproduce

Open [scratch-gui](https://llk.github.io/scratch-gui/).

### Operating System and Browser

Mac OS 10.12, Safari 10.0

towerofnix:
I can't reproduce this anymore (still Safari, macOS 10.12).

PR #234, opened 2017-03-10
Made UI elements non-selectable

### LINKED ISSUE #162: Highlighted text appears draggable in editor
### Expected Behavior

Maybe the ghosted/ transparent text should not be draggable, and whatever you clicked on, e.g. the sprite, should still be draggable?

### Actual Behavior

When you highlight multiple bits of text, you can end up apparently dragging all of them around (they don't actually move, but the image is shown while dragging) - I don't know if we want that?

out

### Steps to Reproduce

- Highlight text from the sprite pane
- Click on the cat sprite and drag it

_Mac OS X El Capitan 10.11.6, Chrome Version 56.0.2924.87 (64-bit)_

lifeinchords:
Fixes #162 for Chrome + Firefox.
- Affects: SpriteSelector Pane, StageSelector Pane, Stage, MenuBar, Libraries Modal, Add/Import buttons
- This update does not fix issue on Edge.
- `SpriteSelector` pane is still broken in Safari, but this is related to #105. Once that is fixed, this might resolve without further updates needed.

### DIFF
--- a/src/components/sprite-info/sprite-info.css
+++ b/src/components/sprite-info/sprite-info.css
@@ -38,10 +38,7 @@ $form-radius: calc($space / 2);
     height: 1.25rem;
     padding: calc($space / 2);
     outline: none;
-
-    /* @todo: make this a mixin for all UI text labels */
     user-select: none;
-    cursor: default;
 }
 
 .icon {
--- a/src/components/sprite-info/sprite-info.css
+++ b/src/components/sprite-info/sprite-info.css
@@ -31,14 +31,12 @@ $form-radius: calc($space / 2);
     align-items: center;
 }
 
-/* @todo: refactor from descendant selector, to a regular class */
 .icon-wrapper {
     display: inline-block;
     box-sizing: content-box;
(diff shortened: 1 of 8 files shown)


src/components/sprite-info/sprite-info.css
```diff
@@ -31,14 +31,12 @@ $form-radius: calc($space / 2);
...
 }
 
-/* @todo: refactor from descendant selector, to a regular class */
 .icon-wrapper {
     display: inline-block;
     box-sizing: content-box;
     width: 1.25rem;
     height: 1.25rem;
     padding: calc($space / 2);
-    cursor: pointer;
```
paulkaplan: Does this make any visual changes or just refactor redundant code? It doesn't _seem_ to when trying it out, just wanted to check.

lifeinchords: I made a mistake, by defining `cursor` twice in the class. When reviewing it again now, I'm quite sure the `cursor: default` isn't needed either

PR #261, opened 2017-03-20
Use scratch-storage

cwillisf:
### Resolves

This is the GUI side of LLK/scratch-vm#427

### Proposed Changes

This change causes the GUI (specifically, the stage container) to:
1. create an instance of `ScratchStorage`, storing a reference locally,
2. add two standard web sources to the storage module, and
3. give the storage module instance to the VM.

There are several places in the GUI that could use the storage module for file access; those are not yet updated to use this module but they could be.

### Reason for Changes

This is the minimum necessary to make it safe to merge LLK/scratch-vm#498

### DIFF
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -11,8 +11,6 @@ const MenuBar = require('../menu-bar/menu-bar.jsx');
 const Box = require('../box/box.jsx');
 const styles = require('./gui.css');
 
-const Storage = require('../../lib/storage');
-
 const GUIComponent = props => {
     const {
         basePath,
@@ -72,12 +70,9 @@ const GUIComponent = props => {
 GUIComponent.propTypes = {
     basePath: React.PropTypes.string,
     children: React.PropTypes.node,
-    vm: React.PropTypes.instanceOf(VM)
+    vm: React.PropTypes.instanceOf(VM).isRequired
 };
 GUIComponent.defaultProps = {
-    basePath: '/',
-    vm: new VM(),
-    storage: new Storage()
+    basePath: '/'
 };
-GUIComponent.defaultProps.vm.attachStorage(GUIComponent.defaultProps.storage);
 module.exports = GUIComponent;
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -11,6 +11,8 @@ const MenuBar = require('../menu-bar/menu-bar.jsx');
 const Box = require('../box/box.jsx');
 const styles = require('./gui.css');
 
+const Storage = require('../../lib/storage');
+
 const GUIComponent = props => {
     const {
         basePath,
@@ -48,7 +50,7 @@ const GUIComponent = props => {
                             <GreenFlag vm={vm} />
                             <StopAll vm={vm} />
                         </Box>
-                        
+
                         <Box className={styles.stageWrapper} >
                             <Stage
                                 shrink={0}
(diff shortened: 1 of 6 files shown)


src/containers/stage.jsx
```diff
@@ -32,10 +33,12 @@ class Stage extends React.Component {
         this.attachRectEvents();
         this.attachMouseEvents(this.canvas);
         this.updateRect();
-        this.renderer = new Renderer(this.canvas);
-        this.props.vm.attachRenderer(this.renderer);
         this.audioEngine = new AudioEngine();
         this.props.vm.attachAudioEngine(this.audioEngine);
+        this.renderer = new Renderer(this.canvas);
```
paulkaplan: Was there a specific reason for changing the order of attaching audio / renderer?

src/containers/stage.jsx
```diff
@@ -32,10 +33,12 @@ class Stage extends React.Component {
...
+        this.props.vm.attachRenderer(this.renderer);
+        this.storage = new Storage();
+        this.props.vm.attachStorage(this.storage);
```
rschamp: The Stage seems like the right place to attach the renderer to the VM, pretty iffy on the audio engine, but seems wrong for Storage. It seems like something we might want to do in the GUI container, where we instantiate the VM?  Either that or in the app and pass the VM to the GUI with the storage already attached.

rschamp:
> There are several places in the GUI that could use the storage module for file access; those are not yet updated to use this module but they could be.

Whoops missed that. If you want to do that in a separate PR, that's fine with me.

src/containers/stage.jsx
cwillisf: The squirrel shows up if you're running scratch-render before LLK/scratch-render#122
The ordering of audio/renderer/storage is just alphabetical, no specific reason.

src/containers/stage.jsx
cwillisf: Cool - I just stuck it here because I saw that the audio and renderer modules were here. I'm happy to put it anywhere that makes sense. What's your preference?

rschamp: I think my preference would be in the GUI container — the default vm prop should be an instance of the VM with the storage attached.  That way if someone wanted to pass in a vm with a different storage attached, they could.  And we would maintain the base GUI component being "batteries included".

src/components/gui/gui.jsx
```diff
@@ -74,6 +76,8 @@ GUIComponent.propTypes = {
 };
 GUIComponent.defaultProps = {
     basePath: '/',
-    vm: new VM()
+    vm: new VM(),
+    storage: new Storage()
 };
+GUIComponent.defaultProps.vm.attachStorage(GUIComponent.defaultProps.storage);
```
cwillisf: @rschamp is this what you had in mind? I feel weird about this line of non-"declarative" code just hanging out here but I don't currently see a better place for it. Open to suggestion of course.

rschamp: I was thinking more like this, as to keep the interface to the GUI simple, since then if you want to use a different storage, you can just pass in your own VM with your storage (and whatever else) attached:
```javascript
const defaultVm = new VM();
defaultVm.attachStorage(new Storage());
GUIComponent.defaultProps = {
    basePath: '/',
    vm: defaultVm,
};
```

cwillisf: Ah, I see. That way you don't have a useless `defaultProps.storage` hanging around if you want a different one. Makes sense. I'd still like to officially* allow access to the storage module from outside the VM but thinking on it further I don't think this is the right place for it.

*Sure, you can do `vm.runtime.storage` but that doesn't exactly feel like a public API :)

PR #262, opened 2017-03-21
Load all Scratch 3.0 modules with Babel

cwillisf:
### Proposed Changes

This change adjusts the webpack configuration for compatibility with unpacked audio, render, storage, and VM modules. There are also a few other small tweaks to build settings. See line comments for details.

### Reason for Changes

This sets us up to build `scratch-gui` in a single step, allowing webpack to better optimize our files and reducing the duplication of code within our output bundles.

### Test Coverage

Test coverage has not changed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
 - '4.2'
-- 'stable'
+- 'node'
 cache:
   directories:
   - node_modules
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "GraphicaL User Interface for creating and running Scratch 3.0 projects",
   "main": "./src/index.js",
   "scripts": {
-    "build": "npm run clean && webpack --progress --colors --bail",
+    "build": "webpack --progress --colors --bail",
     "clean": "rm -rf ./build && mkdir -p build",
     "deploy": "touch build/.nojekyll && gh-pages -t -d build -m \"Build for $(git log --pretty=format:%H -n1)\"",
     "lint": "eslint . --ext .js,.jsx",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -16,6 +16,7 @@ module.exports = {
         host: '0.0.0.0',
         port: process.env.PORT || 8601
     },
+    devtool: 'cheap-module-source-map',
     entry: {
         lib: ['react', 'react-dom'],
         gui: './src/index.jsx'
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "GraphicaL User Interface for creating and running Scratch 3.0 projects",
   "main": "./src/index.js",
   "scripts": {
-    "build": "npm run clean && webpack --progress --colors --bail",
```
cwillisf: `npm run clean` runs `rm` which doesn't work on Windows. By removing the `npm run clean` step here, Windows users can at least `npm run build` even if `npm run clean` still won't work for them.

webpack.config.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
         host: '0.0.0.0',
         port: process.env.PORT || 8601
     },
+    devtool: 'cheap-module-source-map',
```
cwillisf: This option has treated me well over the past few days, though I'm happy to discuss and experiment with other options.
See also: https://webpack.js.org/configuration/devtool/#devtool

webpack.config.js
```diff
@@ -24,15 +25,18 @@ module.exports = {
         path: path.resolve(__dirname, 'build'),
         filename: '[name].js'
     },
+    resolve: {
+        symlinks: false
+    },
```
cwillisf: This affects the way webpack looks for things such as the `transform-object-rest-spread` plugin (see below) when dealing with `npm link`ed dependencies (say, `scratch-vm` for example). If `symlinks` is true (default) then webpack will look for `scratch-vm/node_modules/babel-plugin-transform-object-rest-spread` only and will fail if it doesn't exist. If `symlinks` is false it will find the one under `scratch-gui/node_modules/...`
See https://github.com/webpack/webpack/issues/985 for multiple pieces of conflicting advice. This setting works for me but may need further tinkering to make sure everything is correct.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
 - '4.2'
-- 'stable'
+- 'node'
```
cwillisf: Note: `stable` is a deprecated synonym for `node`.
See also https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Specifying-Node.js-versions

package.json
thisandagain: I think I'd rather fix the issue using something like [rimraf](https://github.com/isaacs/rimraf) than remove the clean step for Windows users.

webpack.config.js
```diff
@@ -24,15 +25,18 @@ module.exports = {
...
     externals: {
         React: 'react',
         ReactDOM: 'react-dom'
     },
     module: {
         rules: [{
-            test: /\.jsx?$/,
+            // allow ES2015 in any *.js or *.jsx file under .../scratch-*/src/...
+            test: /[\\/]+scratch-[^\\/]+[\\/]+src[\\/]+.+\.jsx?$/,
```
rschamp: Any way we can make this more readable using more than one regex? It seems like we should be able to use the other tools available to `Condition` https://webpack.js.org/configuration/module/#condition

package.json
cwillisf: Ah, rimraf seems cool. I feel kind of weird about `npm run build` doing a clean first, though. None of our other Scratch 3.0 repositories do that and it's not what I'm used to with other build systems, but maybe it's typical for npm / Node.js / webpack?

rschamp: FWIW, we do have it set up that way on www https://github.com/LLK/scratch-www/blob/develop/Makefile#L11. I would also rather have build do the clean step, since it ensures you don't have stray/stale files, which I think is important with all of the `npm link`ing we do, and switching branches when testing PRs etc.

I'll also note that `clean` also uses the Windows-unfriendly `mkdir -p`. But the `rimraf` README points out that can be resolved with another package, `mkdirp`😛

cwillisf:
Closing this in favor of #309 for now. There are a few other changes, like proper version numbers, which should happen before I try this kind of change again.

PR #265, opened 2017-03-22
Tabbed sound and costume panels

paulkaplan:
Builds on https://github.com/LLK/scratch-gui/pull/257 to add tabbed sound and costume panels

~The first commit of this is in #257, so ignore that, i'll rebase when it is merged~ Rebased

Adds sound and costume tabs, together with first iteration of a few actions like deleting and adding from the library.

There are a few places where there is some serious friction with the VM. @rschamp suggested leaving `todos` and creating issues on scratch-vm for it.

- [ ] Blockly doesn't like being resized while hidden. I had to trigger a browser resize event just to get blockly to resize, it would be best if there was a more direct way.
- [ ] The VM should provide a way to remove costumes and sounds from the editing target, handling triggering the relevant events, including what happens when you delete the active costume.


TheBrokenRail:
Here is some CSS that looks more like the concept image:
```
.tab {
  border-radius: 10px 10px 0 0;
  background-color: #F6F8FA;
  display: inline;
  float: left;
  color: #9AA1B5;
  width: 33.33%;
  width: calc(100% / 3);
  text-align: center;
}
.tab:hover {
  background-color: #FBFCFD;
  color: #6DADD5;
}
.tabSelected {
  border-radius: 10px 10px 0 0;
  background-color: #FFFFFF;
  display: inline;
  float: left;
  color: #40B9F5;
  width: 33.33%;
  width: calc(100% / 3);
  text-align: center;
}
.tabText {
  display: inline-block;
  font-weight: bold;
  margin: 10px;
  margin-top: 5px;
  margin-bottom: 5px;
  position: relative;
  font-size: 16px;
}
```

src/components/asset-panel/selector.jsx
```diff
@@ -0,0 +1,61 @@
...
+            >
+                {newText}
+            </Box>
+            <Box className={styles.listArea}>
+                {items.map((item, index) => {
+                    const _onItemClick = () => onItemClick(item);
+                    const _onDeleteClick = e => {
+                        e.stopPropagation();
+                        onDeleteClick(item);
+                    };
```
rschamp: A couple of things about these handlers:
* Since these are behavioral, they should be owned by a container
* Function definitions in `render` should be avoided for performance (which is why the linter complains about arrow functions as props). If you can't convert the component to a class and use methods instead, a wrapper component would be another way to avoid this (this is partially the reason for the `SpriteSelectorItem` component itself).

src/containers/costume-tab.jsx
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    handleSelectCostume (item) {
+        this.setState({selectedCostumeIndex: this.props.vm.editingTarget.getCostumeIndexByName(item.name)});
+    }
+
+    handleDeleteCostume (item) {
+        // @todo the VM should handle all of this logic
+        const {editingTarget} = this.props.vm;
+        const i = editingTarget.getCostumeIndexByName(item.name);
```
rschamp: Please avoid non-descriptive variable names, except for maybe `e` for events.

src/containers/sound-tab.jsx
```diff
@@ -0,0 +1,107 @@
...
+            'getSoundIndexByName',
+            'handleSelectSound',
+            'handleDeleteSound'
+        ]);
+        this.state = {selectedSoundIndex: 0};
+    }
+
+    getSoundIndexByName (name) {
+        // @todo should be in VM
+        let i = -1;
```
rschamp: Same note here about `i`.

paulkaplan:
@rschamp thanks for the feedback, I made the changes in e5e761d and b36d96e, ended up simplifying the code quite a bit as well!

paulkaplan:
@TheBrokenRail thanks! I integrated your recommendations, @carljbowman approves! 👍  Here is what it looks like now.

screen shot 2017-03-23 at 8 40 36 am

src/components/asset-panel/selector.jsx
```diff
@@ -0,0 +1,56 @@
...
+        <Box className={styles.wrapper}>
+            <Box
+                className={styles.newItem}
+                onClick={onNewClick}
+            >
+                {newText}
+            </Box>
+            <Box className={styles.listArea}>
+                {items.map((item, index) => (
+                    <SpriteSelectorItem
```
rschamp: Looks like it's time for this to get a new name! (Not in this PR)

src/components/gui/gui.css
```diff
@@ -21,24 +21,71 @@
...
+    color: #40B9F5;
+}
+
+.tabs {
+    position: relative;
     flex-grow: 1;
     flex-shrink: 0;
+    display: flex;
+    flex-direction: column;
+}
```
rschamp: Having these styles here is fine for now, but it would be nice to eventually pull these styles out and put them into our own tab components (which can just wrap the `react-tabs` library components).

src/containers/costume-tab.jsx
```diff
@@ -0,0 +1,97 @@
...
+const mapStateToProps = state => ({
+    editingTarget: state.targets.editingTarget,
+    sprites: state.targets.sprites,
+    costumeLibraryVisible: state.modals.costumeLibrary
+});
+
+const mapDispatchToProps = dispatch => ({
+    onNewCostumeClick: e => {
+        e.preventDefault();
+        dispatch(openCostumeLibrary());
```
rschamp: Now that this is triggered in more than one place, we should think about moving where the modals open from, maybe to `GUI`? Or some other kind of wrapper? I suppose it makes a kind of sense to have the modals located in the target pane, since you can't view the tabs without the pane, but it's a bit unintuitive.

carljbowman:
@rschamp - Yup, the 'Costumes' tab should read 'Backdrops' when the Stage is selected.

paulkaplan:
@rschamp I noticed that issue with the block toolbox. I could only get it to happen when i added a new sprite while the scripts tab wasn't active. It doesn't seem to respond to the `resize` hack...

TheBrokenRail:
Should I make a PR that adds a hover to the tabs?

rschamp:
@TheBrokenRail No thank you, we'll handle the design changes. [Check out issues with the "help wanted" label](https://github.com/LLK/scratch-gui/labels/help%20wanted) for things that we want PRs for, thanks!

Issue #256, opened 2017-03-16
`off` is not a function on vm


paulkaplan:
Submitting this so I don't forget about it. In the `blocks` container, there is a `vm.off(` in `detachVM` which gets triggered on componentDidMount. However, there is no such function... The correct function is (bizarrely) `.removeListener` :( 

This never comes up right now because that component never unmounts. But playing with a tabbed area (where it actually would unmount), I saw errors coming from this. 

https://github.com/LLK/scratch-gui/blob/develop/src/containers/blocks.jsx#L67-L72

PR #266, opened 2017-03-23
Change #off to #removeListener

paulkaplan:
Fixes https://github.com/LLK/scratch-gui/issues/256

`.on` is an alias for `.addListener`[1], but there is no alias for `.off` and it looks like there never will be [2] :( :( :( 

[1] https://github.com/nodejs/node/blob/master/lib/events.js#L307
[2] https://github.com/nodejs/node/pull/540


rschamp:
Alternatively we could alias `.off` in the VM, but for now, ignoring that `on` exists is probably more bugproof.

paulkaplan:
@rschamp I changed all `on` and `off` to `addListener` and `removeListener` (respectively). Ready for re-review.

Issue #87, opened 2017-02-13
Refactor Add From Library buttons into components. 


lifeinchords:
Currently they are individually styled.

paulkaplan:
Have been refactored!

PR #277, opened 2017-03-27
fix #108 add button is not centered in firefox

### LINKED ISSUE #108: Add Sprite + Add Backdrop buttons: icon not centered correctly
screen shot 2017-02-17 at 11 12 54 am

Safari: Version 10.0.1 (12602.2.14.0.7)
OS Sierra: 10.12.1 (16B2657)

bigeyex:
### Resolves

solved #108 add button not centered in Firefox

### Reason for Changes

There could be 2 ways in solving this:
1. adding "margin: auto" to center the img displayed in block;
2. turn the img to flexbox and center it with align-items.

I tried to do 2. But it somewhat does not work in Firefox (an issue in Firefox?). 
So I fallback to 1.

### Test Coverage

no link issue emerges

### DIFF
--- a/src/components/target-pane/target-pane.css
+++ b/src/components/target-pane/target-pane.css
@@ -44,6 +44,7 @@
 
 .add-button {
     height: 0.8rem;
+    margin: auto;
 }
 
 /* @todo: This is hacky. Better: move buttons in their corresponding panes, and set values relatively */


lifeinchords:
Sorry for the delay @bigeyex, thanks for submitting this. Option 1 with `margin: auto` works across browsers, though I think Option 2 is a better solution, since flexbox will help us when we make this button a component next, as part of #87. 

I think Option 2 should work by following [this answer](https://stackoverflow.com/questions/35464067/flexbox-not-working-on-button-element-in-some-browsers), and moving the flex properties from the `button` to a new `span` wrapper around the `img`

Would you like to give that a shot?

bigeyex:
Hi @lifeinchords , I updated Firefox to the newest version, this doesn't appear to be a problem. If this could be confirmed, then this PR and the corresponding issue could be closed.

lifeinchords:
Hi @bigeyex, thanks for the update. I confirmed this is no longer an issue in latest Firefox.
Latest Safari unfortunately still shows this issue. 

@paulkaplan We reviewed this earlier, I think this PR is good for now to bring in, and maybe consider Option 2 later if we need to. Good to merge?

Issue #140, opened 2017-02-28
StageSelectorItem: whitespace renders inside canvas


lifeinchords:
`canvas` element renders with whitespace inside of it. The blue border on the class when selected makes this more noticeable.

Each stage/backdrop renders with slightly different whitespace dimensions, making it's hard to programmatically fix/mask via the border.

@rschamp mentioned some potential solves, re: how the stage renders with a similar constraint.

![screen shot 2017-02-28 at 1 55 46 pm](https://cloud.githubusercontent.com/assets/716168/23438100/86e6f262-fdde-11e6-8bde-b0a157fa50a4.png)

PR #285, opened 2017-03-29
Fixes issue #140 whitespace renders inside canvas

bigeyex:
by adding a "vertical align" property inside the costume-canvas component

### Resolves

issue #140 
There's 2 ways to solve this issue: 
1. allow the canvas to align the image to the top, eliminating the space;
2. recompute the size of the canvas before/after rendering.

Approach 1 is taken for this pr. Because it is more elegant since the parent element
do not need to mingle with the dimensions of the image, which should be in the domain
of the canvas.

### Proposed Changes

1. In costume canvas, add a prop called "verticalAlign";
2. set "verticalAlign" to top when stage selector is using the canvas.

### DIFF
--- a/src/components/costume-canvas/costume-canvas.jsx
+++ b/src/components/costume-canvas/costume-canvas.jsx
@@ -48,7 +48,12 @@ class CostumeCanvas extends React.Component {
 
         // Rotation origin point will be center of the canvas.
         const contextTranslateX = this.canvas.width / 2;
-        const contextTranslateY = this.canvas.height / 2;
+        let contextTranslateY = this.canvas.height / 2;
+
+        // If vertical align is top, draw the image from the top of the canvas
+        if(this.props.verticalAlign === 'top') {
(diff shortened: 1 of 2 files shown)


paulkaplan:
@bigeyex thanks for this, sorry about the delay getting back to you on it. I think that we should take a different approach to https://github.com/LLK/scratch-gui/issues/140 though. I think that the thumbnail should render in exactly the way the stage renders. So if there is whitespace in the stage image, there should be whitespace in the thumbnail. There **isn't** whitespace on the actual stage though because the aspect ratio of the stage and the thumbnail are different. The simplest solution for that is to make the thumbnails the same aspect ratio as the stage. right now the stage is 56x44, but the stage is 4x3. I propose changing the stage thumbnail to be 4x3 aspect ratio (56x42). Then the whitespace problem will go away. What do you think about that? /cc @lifeinchords.

paulkaplan:
@bigeyex I made a PR for the fix I mentioned https://github.com/LLK/scratch-gui/pull/324

Issue #286, opened 2017-03-31
Sound Library does not play sounds on click


paulkaplan:
scratch-audio removed the ability to load sounds, delegating that to scratch-storage. The sound library which loads sounds both for library playback and adding them to the VM need to be updated to load sounds using the new APIs for storage and audio.

PR #293, opened 2017-04-03
Use decodeSound and scratch storage for loading sounds

paulkaplan:
Fixes https://github.com/LLK/scratch-gui/issues/286 by using the new storage system to load and play sounds. Also uses the new `addSound` API to correctly load sounds into the vm's audio engine (which is different from the sound libraries audio engine). 

Depends on https://github.com/LLK/scratch-vm/pull/516


paulkaplan:
Oh also /cc @ericrosenbaum, this should look very familiar, since it is basically what we do in the VM. But in this case I'm using that approach for the standalone audio engine that the sound library component owns.

PR #309, opened 2017-04-11
Standardize settings

cwillisf:
### Proposed Changes

This change ensures that line endings, Travis CI settings, and source map settings are similar to other Scratch 3.0 repositories. These were previously part of #262.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "./src/index.js",
   "scripts": {
     "build": "npm run clean && webpack --progress --colors --bail",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -67,5 +67,9 @@
     "webpack": "2.3.2",
     "webpack-dev-server": "^2.4.1",
     "xhr": "2.4.0"
+  },
+  "dependencies": {
+    "mkdirp": "^0.5.1",
+    "rimraf": "^2.6.1"
```
rschamp: Why not dev dependencies?

PR #321, opened 2017-04-21
Get `Asset` and `AssetType` from storage instance

### LINKED ISSUE #320: Get `Asset` and `AssetType` from storage instance
### Expected Behavior

Now that LLK/scratch-storage#9 has been merged, the GUI should access `scratch-storage`'s `Asset` and `AssetType` through a `scratch-storage` instance.

For example:
```js
this.storage.load(this.storage.AssetType.Project, '1234').then(project => {...});
```

### Actual Behavior

The `Asset` constructor and `AssetType` list are being accessed through the `ScratchStorage` constructor. This method of access is now deprecated.

For example:
```js
const ScratchStorage = require('scratch-storage');
this.storage.load(ScratchStorage.AssetType.Project, '1234').then(project => {...});
```

cwillisf:
### Resolves

Resolves #320 

### Proposed Changes

Access `Asset` and `AssetType` through an instance of ScratchStorage. Remove a now-unnecessary instance of `require('scratch-storage')`.

### Reason for Changes

Accessing `Asset` and `AssetType` as properties of `ScratchStorage` instead of an instance is deprecated as of LLK/scratch-storage#9. Accessing them through an instance ensures consistency between the storage runtime code and the values and objects being used with it, and allows for the possibility of a dynamic asset type list in the future.

### DIFF
--- a/src/lib/storage.js
+++ b/src/lib/storage.js
@@ -1,5 +1,4 @@
 const ScratchStorage = require('scratch-storage');
-const AssetType = ScratchStorage.AssetType;
 
 const PROJECT_SERVER = 'https://cdn.projects.scratch.mit.edu';
 const ASSET_SERVER = 'https://cdn.assets.scratch.mit.edu';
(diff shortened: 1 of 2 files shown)


src/lib/storage.js
```diff
@@ -1,5 +1,4 @@
 const ScratchStorage = require('scratch-storage');
```
cwillisf: I was originally going to try to get rid of this, maybe to move toward a string "glue the parts together in the playground" model, but we'd need to restructure non-storage things too before that's feasible.

PR #325, opened 2017-04-24
Stop clobbering loading projects with default data

rschamp:
If the app is loading with an initial project id, don't try to load the default project. Previously, this default data would load on top of the project being loaded.

Does not fix the issue of blocks not loading in the blocks pane, which appears to be separate.


paulkaplan:
LG

PR #330, opened 2017-04-25
Add prompt for variable creation

paulkaplan:
A simple first iteration of a new variable prompt. I wanted to get some feedback on the way this is implemented @rschamp. (The css definitely needs some work, and I avoided the header bar from the mockup because there isn't the data to fill it). 
_Update:_ Added the prompt header and refactored container/component divide.
screen shot 2017-04-26 at 10 17 26 am

Fixes https://github.com/LLK/scratch-blocks/issues/535

### DIFF
--- a/src/components/prompt/prompt.jsx
+++ b/src/components/prompt/prompt.jsx
@@ -1,70 +1,59 @@
 const React = require('react');
 const ReactModal = require('react-modal');
-const bindAll = require('lodash.bindall');
 const Box = require('../box/box.jsx');
 
 const styles = require('./prompt.css');
 
-class PromptComponent extends React.Component {
-    constructor (props) {
-        super(props);
-        bindAll(this, [
-            'handleOk',
-            'handleKeyPress'
-        ]);
-    }
-    handleKeyPress (event) {
-        if (event.key === 'Enter') this.handleOk();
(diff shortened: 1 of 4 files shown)


src/components/prompt/prompt.jsx
```diff
@@ -0,0 +1,74 @@
...
+    onCancel: React.PropTypes.func.isRequired,
+    onOk: React.PropTypes.func.isRequired,
+    placeholder: React.PropTypes.string
+};
+
+PromptComponent.defaultProps = {
+    placeholder: '...'
+};
+
+module.exports = PromptComponent;
```
rschamp: This should be split into a container and component.

paulkaplan: How do you mean? What part should be a container?

rschamp: The component should only call props for its handlers, and the container should implement them.

It doesn't matter too much at this point, other than for consistency, but as the components grow, it provides the split between the presentation and business logic.

paulkaplan:
@rschamp thanks for the feedback. I separated the container/component boundary and updated the binding with scratch-blocks. I also added a header per the design spec. Should be ready for a real review!

PR #331, opened 2017-04-27
Revert "Merge pull request #313 from paulkaplan/use-vm-data-for-sound…

paulkaplan:
This reverts commit d20238ae73e97c0aa6c12bd469a274b0aedd7130, reversing
changes made to a151f4bfb24ac7aee8591d97295743a2fba66ce2.

Reverting PR https://github.com/LLK/scratch-gui/pull/313

The code changed in the VM that made this possible is causing performance problems that make it basically unusable. As a quick example, try adding a sprite and then dragging a sprite around the stage. For this reason, I think we should revert this and https://github.com/llk/scratch-vm/pull/526 and rethink how we want to do it.


rschamp:
I'm going to take a crack at resolving this with the work I'm doing on serialization, so let's hold off on reverting til I can see if that helps.

PR #337, opened 2017-05-03
Add reducer, components and containers for monitors

paulkaplan:
Initial front-end for monitors including a reducer for updating monitors, an stage-overlayed component that displays monitors and draggable monitor components that emit position events at the end of their drags. 

monitor-updates

You can add monitors through the redux console by dispatching this event:
```
{
    type: 'scratch-gui/monitors/UPDATE_MONITORS',
    monitors: [{
        id: 'var1',
        category: 'data',
        label: 'Variable',
        value: '123',
        x: 20,
        y: 40
    }, {
        id: 'var2',
        category: 'motion',
        label: 'Variable2',
        value: '123123123',
        x: 200,
        y: 100
    }]
}
```

### DIFF
--- a/src/containers/monitor.jsx
+++ b/src/containers/monitor.jsx
@@ -1,0 +1,32 @@
+const bindAll = require('lodash.bindall');
+const React = require('react');
+
+const MonitorComponent = require('../components/monitor/monitor.jsx');
+
+class Monitor extends React.Component {
+    constructor (props) {
+        super(props);
+        bindAll(this, [
+            'handleDragEnd'
+        ]);
+    }
+    handleDragEnd (_, {x, y}) {
+        this.props.onDragEnd(
+            this.props.id,
+            x,
+            y
+        );
+    }
+    render () {
+        return (
+            <MonitorComponent
+                {...this.props}
+                onDragEnd={this.handleDragEnd}
+            />
+        );
+    }
+}
+
+Monitor.propTypes = MonitorComponent.propTypes;
+
+module.exports = Monitor;
--- a/src/containers/monitor.jsx
+++ b/src/containers/monitor.jsx
@@ -10,7 +10,7 @@ class Monitor extends React.Component {
             'handleDragEnd'
         ]);
     }
-    handleDragEnd (_, {x, y}) {
+    handleDragEnd (e, {x, y}) {
         this.props.onDragEnd(
             this.props.id,
             x,
--- a/src/reducers/monitors.js
+++ b/src/reducers/monitors.js
@@ -1,0 +1,27 @@
+const UPDATE_MONITORS = 'scratch-gui/monitors/UPDATE_MONITORS';
+
+const initialState = {
+    monitors: []
+};
+
+const reducer = function (state, action) {
+    if (typeof state === 'undefined') state = initialState;
+    switch (action.type) {
+    case UPDATE_MONITORS:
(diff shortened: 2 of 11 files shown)


src/reducers/monitors.js
```diff
@@ -0,0 +1,27 @@
...
+
+const initialState = {
+    monitors: []
+};
+
+const reducer = function (state, action) {
+    if (typeof state === 'undefined') state = initialState;
+    switch (action.type) {
+    case UPDATE_MONITORS:
+        return Object.assign({}, state, {monitors: action.monitors});
```
rschamp: If we want this to be less painful than the way targets work, then I think `UPDATE_MONITORS` should always supply the entire list of monitors with all the details. Then this can just return `action.monitors` rather than mucking around with merging (which can get out of sync when projects are loaded on top of each other).

Alternatively we should have `ADD_MONITOR`, `REMOVE_MONITOR`, `UPDATE_MONITOR` instead of dealing with the list as a whole.  But I think doing the whole list for now should be sufficient.

rschamp: Whoops misread this one.  The way you're doing it is how I was expecting. Except that there's the extra level of `monitors`. Since we don't have any other state to store, I think we can flatten this to a simple list.

paulkaplan: oh is it ok to have `state` be an array? I kind of assumed it had to be an object... not sure why i assumed that though...

src/containers/monitor.jsx
```diff
@@ -0,0 +1,32 @@
...
+const MonitorComponent = require('../components/monitor/monitor.jsx');
+
+class Monitor extends React.Component {
+    constructor (props) {
+        super(props);
+        bindAll(this, [
+            'handleDragEnd'
+        ]);
+    }
+    handleDragEnd (_, {x, y}) {
```
paulkaplan: yeah `_` is the actual mouse event, the second param is the drag event as provided by react-draggable (with properly offset coordinates from parent, for example). I was just avoiding the `unused-var` linter error, but yeah i can name it.

src/reducers/monitors.js
rschamp: You need the top-level state to be an object, but as far as I know, the slices can be anything you want.

paulkaplan:
@rschamp updated this with commits to address your questions.

One question I had was with the reducer using an array, is the way I did it correct, or should it return a copy of the array (by using `arr.slice(0)` or something). See https://github.com/LLK/scratch-gui/pull/337/commits/80127c5effa3afc029e5d44ef137d1593f822ab0

src/reducers/monitors.js
```diff
@@ -0,0 +1,25 @@
+const UPDATE_MONITORS = 'scratch-gui/monitors/UPDATE_MONITORS';
+
+const initialState = [];
+
+const reducer = function (state, action) {
+    if (typeof state === 'undefined') state = initialState;
+    switch (action.type) {
+    case UPDATE_MONITORS:
+        return action.monitors;
```
rschamp: I think the best practice is to always make a copy like `return [... action.monitors]`.  I was going to say in practice this won't be too important since we'll be making a copy in the VM. But we currently don't do that with targets, so I think we probably should copy here.

paulkaplan: roger that. updated in d01a5f5

PR #338, opened 2017-05-04
Remove shouldComponentUpdate from TargetPane

rschamp:
This "optimization" is actually a de-optimization. The deep compare happening on the sprites list will occur already via the components using the contents of the sprite list.

### DIFF
--- a/src/components/target-pane/target-pane.jsx
+++ b/src/components/target-pane/target-pane.jsx
@@ -1,5 +1,3 @@
-const isEqual = require('lodash.isequal');
-const omit = require('lodash.omit');
 const classNames = require('classnames');
 const PropTypes = require('prop-types');
 const React = require('react');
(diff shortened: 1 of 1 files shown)


src/components/target-pane/target-pane.jsx
```diff
@@ -23,119 +21,106 @@ const addIcon = require('./icon--add.svg');
...
-                    onChangeSpriteY={onChangeSpriteY}
-                    onDeleteSprite={onDeleteSprite}
-                    onSelectSprite={onSelectSprite}
-                />
-                <Box className={styles.stageSelectorWrapper}>
-                    {stage.id && <StageSelector
-                        backdropCount={stage.costumeCount}
-                        id={stage.id}
-                        selected={stage.id === editingTarget}
-                        url={
```
paulkaplan: I think this change is part of the other PR that has the vm storage changes. Probably don't want this here.

rschamp: Thanks! Will need to pull that back in, forgot that this file was part of that.

rschamp:
Need that change in the serialization PR, so I'll just include this in there.

PR #339, opened 2017-05-04
Integrate VM serialization changes

rschamp:
When the VM stops sending costume data over in sprite info reports and target updates, the sprite thumbnail image data need to be retrieved from storage.

To make this easier to do, I moved the vm into a reducer, and connected the relevant components to it, so that they convert asset ids into data.

Requires https://github.com/LLK/scratch-vm/pull/552

/cc @thisandagain @cwillisf

### DIFF
--- a/src/components/asset-panel/selector.jsx
+++ b/src/components/asset-panel/selector.jsx
@@ -29,6 +29,7 @@ const Selector = props => {
                     <SpriteSelectorItem
                         assetId={item.assetId}
                         className={styles.listItem}
+                        costumeURL={item.url}
                         id={index}
                         key={`asset-${index}`}
                         name={item.name}
--- a/src/components/asset-panel/selector.jsx
+++ b/src/components/asset-panel/selector.jsx
@@ -27,8 +27,8 @@ const Selector = props => {
             <Box className={styles.listArea}>
                 {items.map((item, index) => (
                     <SpriteSelectorItem
+                        assetId={item.assetId}
                         className={styles.listItem}
-                        costumeURL={item.url}
                         id={index}
                         key={`asset-${index}`}
                         name={item.name}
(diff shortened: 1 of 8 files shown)


rschamp:
Oh right, thanks. I think I didn't notice this before because I had inadvertently left in the `url` property on costumes for a while and didn't clean it up until the end.

rschamp:
@paulkaplan Pulled in that other PR to here, and fixed the sound tab.

paulkaplan:
My one concern with putting the whole vm in the reducer is abusing that for data. We could put only that one function in the reducer...?

paulkaplan:
@rschamp not sure if you want to deal with this here, but I noticed that some other components require the VM to be passed in as a prop. That could be moved into a `connect` at this point if you want.

paulkaplan:
@rschamp code looks good, but I had two questions I'm curious about your input on above. Neither are code issues though, so I'll turn this to green.

rschamp:
I agree that we can migrate everything to use Redux to get the VM  instance, but I think I might want to do that in a separate PR.  Still not sure how I feel about it because of how assuming Redux complicates reusing components (you need to provide a store to use the GUI now). But we'll have to solve that for the other connected components anyway, that issue is already there, so we may as well make things consistent.

As far as only adding the methods we need from the VM with the reducer, I think I would rather just be disciplined about that. Seems like unnecessary tediousness every time we want to use a different feature.

paulkaplan:
@rschamp sounds good. But just so the future knows, i'll be referring to this comment every time it comes up in the future 
> I would rather just be disciplined about that

rschamp:
@paulkaplan [Same here ;)](https://github.com/LLK/scratch-gui/blame/develop/src/containers/costume-tab.jsx#L38-L40)

PR #342, opened 2017-05-04
Add Save and Load buttons

rschamp:
Requires #339.  Isolated diff: https://github.com/rschamp/scratch-gui/compare/feature/serialization...feature/save-load

Use the de/serialization features of the VM to allow downloading and loading project JSON.

saveload

/cc @thisandagain

### DIFF
--- a/src/components/button/button.jsx
+++ b/src/components/button/button.jsx
@@ -1,0 +1,30 @@
+const classNames = require('classnames');
+const PropTypes = require('prop-types');
+const React = require('react');
+
+const styles = require('./button.css');
+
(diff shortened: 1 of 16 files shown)


src/components/button/button.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+const styles = require('./button.css');
+
+const ButtonComponent = ({
+    className,
+    onClick,
+    children,
+    ...props
+}) => (
+    <span
```
paulkaplan: Not sure if this is important now, but we should probably designate this as a button for ARIA using `role="button"`

PR #361, opened 2017-05-12
Only listen to targets updates

rschamp:
This reduces the complexity of the reducer, as well as how frequently it's run. The VM will emit a targets update after a step if any target info has changed.

### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/550. Brings the "medium complexity" project to comparable performance of the VM playground.

### Proposed Changes

Stop listening to `SPRITE_INFO_REPORT`, since as of https://github.com/LLK/scratch-vm/pull/563 the VM isn't emitting them anymore.

### Reason for Changes

A lot of time is spent in the reducer merging sprite info into the GUI's target list. Instead just use the current state of the list from `targetsUpdate`, which now is emitted each step which modifies target data.

Before (lots of time in reducer):
image

After (only a little time in reducer):
image

/cc @thisandagain @fsih

### DIFF
--- a/src/reducers/targets.js
+++ b/src/reducers/targets.js
@@ -15,8 +15,8 @@ const reducer = function (state, action) {
                 .filter(target => !target.isStage)
                 .reduce(
                     (targets, target, listId) => Object.assign(
-                        {[target.id]: {order: listId, ...target}},
-                        targets
+                        targets,
+                        {[target.id]: {order: listId, ...target}}
                     ),
                     {}
                 ),
--- a/src/reducers/targets.js
+++ b/src/reducers/targets.js
@@ -1,8 +1,5 @@
-const defaultsDeep = require('lodash.defaultsdeep');
-
 const UPDATE_EDITING_TARGET = 'scratch-gui/targets/UPDATE_EDITING_TARGET';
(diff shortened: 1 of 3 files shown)


src/reducers/targets.js
```diff
@@ -12,55 +9,26 @@ const initialState = {
...
-            )
-        });
     case UPDATE_TARGET_LIST:
         return Object.assign({}, state, {
             sprites: action.targets
                 .filter(target => !target.isStage)
                 .reduce(
-                    (targets, target, listId) => defaultsDeep(
+                    (targets, target, listId) => Object.assign(
                         {[target.id]: {order: listId, ...target}},
```
paulkaplan: Looks like `defaultsDeep` and `Object.assign` have different priority structures, where `defaultsDeep` has the objects listed in descending priority where `Object.assign` has them in increasing order of priority. e.g.

```
> a = {foo: 'bar'}
> b = {foo: 'baz'}
> require('lodash.defaultsdeep')(a, b)
{ foo: 'bar' }
> Object.assign({}, a, b)
{ foo: 'baz' }
```
So we might want to change the order of the args here?

Issue #312, opened 2017-04-19
Deleting costumes and sounds is broken


paulkaplan:
Seems to have been caused by the reporting of costumes/sounds from the VM.

paulkaplan:
Tracked this down to the way that `lodash.defaultsDeep` merges arrays. Here is a simple example:

```
oldState = {costumes: [1, 2, 3]}
newState = {costumes: [1, 2]}

defaultsDeep(newState, oldState) 
//=> {costumes: [1, 2, 3]} 
```
this is wrong in our case, we do not want to merge arrays, we just want the new array. 

@rschamp i see that instead of using arrays in many places we transform into objects keyed by ids, is that to prevent this issue? Should we have a costumes map of {id: costume, ...} instead of an array?

We could also use a modified defaults deep like https://www.npmjs.com/package/defaults-deep-preserve-arrays to preserve arrays.

rschamp:
@paulkaplan No, the reason for transforming to objects keyed by ID is just to be able to access targets without iterating over the list rather than any issue with arrays.  I'd be for using that modified defaults-deep. I'd also be for just not merging if we can avoid it.

If we can avoid transforming to an object, I'd be for that too. I think React may be able to help us there if we just key the components by the target ids.

PR #364, opened 2017-05-15
Use vm functions for deleting sounds and costumes

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/312

Depends on https://github.com/LLK/scratch-vm/pull/569

### Proposed Changes

Use the public VM functions for deleting costumes and sounds instead of reaching into the VM. Also update the logic for changing the selected index if it is out of range so that it doesn't rely on data directly in the VM. Instead move that into a conditional `setState` inside the `componentWillReceiveProps` lifecycle method. 

React docs say using `setState` inside of `componentWillReceiveProps` in this way is kosher for updating dependent state https://facebook.github.io/react/docs/react-component.html#componentwillreceiveprops

deleting-costumes-sounds


mrjacobbloom:
Regarding that functionality, I've had the issue from time to time where I leave a project running while I edit costumes (so I can see what a costume looks like with an effect on it, or other similar reasons) and when the project tries to change the sprite's costume I'm kicked out of the costume editor.

Issue #270, opened 2017-03-23
Block toolbox gets squashed when adding a sprite 


paulkaplan:
To reproduce:

1. Switch to the costume / sound tab (away from scripts).
2. Add a new sprite
3. Switch back to the scripts tab

![screen shot 2017-03-23 at 3 37 17 pm](https://cloud.githubusercontent.com/assets/654102/24266851/a128709e-0fde-11e7-99b8-d666c53fb626.png)

There is a current hack in the gui to trigger window resize when tabs change. I think that the toolbox has to be refreshed as well. See  https://github.com/LLK/scratch-gui/blob/develop/src/components/gui/gui.jsx#L31-L36

tmickel:
oh no!! one suspicion: rendering the blocks requires measuring the length of the texts, here: https://github.com/LLK/scratch-blocks/blob/develop/core/field.js#L436 - which requires that the blocks are visible. generally i think blockly expects that its visibility is controlled by https://github.com/LLK/scratch-blocks/blob/a21919e03b33563bf224ad0f1270847a8847fac3/core/workspace_svg.js#L755 - so that work can be done to rerender blocks/other UI components when appropriate. if a new blockly workspace is created when it's hidden by CSS or something else, maybe the above would happen...possibly you could fix it by calling `workspace.render` or something :)

p.s. the new library stuff is _totally amazing_

tmickel:
(another bug, without `workspace.setVisible` (and thus its hideChaff calls) for example, open a menu in a block and switch tabs - the menu stays open)

Issue #358, opened 2017-05-10
Focused block field appears after switching to sounds or costumes tab 


ericrosenbaum:
### Expected Behavior

Sounds and costumes tabs should not contain weird eyeball-like ghost-remnants of block input fields.

### Actual Behavior

They do.

### Steps to Reproduce

Drag a block with an input field, such as 'move 10 steps', to the workspace.
Click on the input field.
Click on the costumes or sounds tab.
(this also works for menus on blocks)

<img width="54" alt="screen shot 2017-05-10 at 4 58 22 pm" src="https://cloud.githubusercontent.com/assets/567844/25921091/2c8b6b94-35a2-11e7-883f-0da3d152e642.png">

### Operating System and Browser

Mac OS 10.11.6 Chrome 57.0.2987.133

ericrosenbaum:
oops should this be in scratch-blocks?

thisandagain:
Hmm. Not sure ... might have to do with tab switching. Needs investigation.

paulkaplan:
Yeah I think we should be hiding/showing blockly when switching tabs, something we will have to do in the GUI.

fsih:
#368 mostly fixes this but it hangs around for .25 seconds. Adding https://github.com/LLK/scratch-blocks/issues/893 in scratch blocks for the rest of the bug.

PR #368, opened 2017-05-17
Fix squashed blocks in flyout when adding a sprite while on another tab

fsih:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/270
Makes https://github.com/LLK/scratch-gui/issues/358 less bad

### Proposed Changes

When switching off the scripts tab, make it not visible. When switching onto it, make it visible, and reload the scripts. This means that UI from the scripts tab won't accidentally be visible in other tabs (like menus on blocks), and the scripts tab text length will be measured correctly when you re-open it.

Also does some clean-up in gui.jsx moving the handle tab change logic into the container.

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -52,7 +52,21 @@ class Blocks extends React.Component {
         this.attachVM();
     }
     shouldComponentUpdate (nextProps, nextState) {
-        return this.state.prompt !== nextState.prompt;
+        return this.state.prompt !== nextState.prompt || this.props.isVisible !== nextProps.isVisible;
+    }
+    componentDidUpdate (prevProps) {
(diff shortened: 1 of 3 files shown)


src/containers/blocks.jsx
```diff
@@ -169,11 +184,12 @@ class Blocks extends React.Component {
 }
 
 Blocks.propTypes = {
+    isVisible: PropTypes.bool.isRequired,
     options: PropTypes.shape({
         media: PropTypes.string,
         zoom: PropTypes.shape({
-            controls: PropTypes.boolean,
-            wheel: PropTypes.boolean,
+            controls: PropTypes.bool,
```
paulkaplan: 😮

Issue #210, opened 2017-03-06
"Go to x y" block does not update value when sprite is dragged


jwzimmer:
Reported in internal playtest 3/3/17.

### Actual vs. Expected Behavior

"Go to x y" block does not update value when sprite is dragged, but it should

### Operating System and Browser

_Mac & Chrome_

navignaw:
We probably want to update the "slide n secs to x y" block as well.

Is there any existing code for changing the block templates' arguments, or would that have to be implemented?

PR #373, opened 2017-05-19
Update shadow block values when sprite moves

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/210

### Proposed Changes

This PR adds a (debounced) event listener on `targetsUpdate` which updates the X and Y fields of the toolbox blocks (glide, set and move). It requires that the shadow blocks be addressable by ID. 

Depends on https://github.com/LLK/scratch-blocks/pull/896

### Reason for Changes

Fixes https://github.com/LLK/scratch-gui/issues/210. 

@rschamp we discussed a different way using a reducer instead of keeping it contained in the blocks container. Since I do not think anything else needs this data, it might be cleaner to keep it contained within here. At the very least this is a functional prototype to test the timing / interactions on. There is a 100ms debounce which _feels_ about right, i.e. it will not update when dragging or when a stack of blocks is executing without a wait. 

xy-updates


src/containers/blocks.jsx
```diff
@@ -93,8 +97,25 @@ class Blocks extends React.Component {
...
+            .getWorkspace()
+            .getBlockById(id)
+            .inputList[0]
+            .fieldRow[0]
+            .setValue(value);
+    }
+    onTargetsUpdate () {
+        if (this.props.vm.editingTarget) {
+            ['glide', 'move', 'set'].forEach(prefix => {
+                this.updateToolboxBlockValue(`${prefix}x`, this.props.vm.editingTarget.x.toFixed(0));
```
fsih: Is it sketchy to assume that the block ID will be a string in this format? Maybe the blocks that show x,y can somehow register as listeners?

paulkaplan: Yeah I wasn't too happy with this either. @rschamp do you have any thoughts on this?

src/containers/blocks.jsx
```diff
@@ -93,8 +97,25 @@ class Blocks extends React.Component {
...
+                this.updateToolboxBlockValue(`${prefix}y`, this.props.vm.editingTarget.y.toFixed(0));
+            });
+        }
     }
-
```
paulkaplan: Yup, the other methods do not have whitespace between them

src/containers/blocks.jsx
rschamp: I think this is ok for now. Presumably when we're building the toolbox programmatically these IDs will be defined close to here, so we can ensure they corellate with this code. I don't think there's any way to attach a listener to the actual fields, although maybe @rachel-fenichel has a better suggestion?

rachel-fenichel: Seems fine to me, since you're programmatically setting up the toolbox so you have full control over the IDs.

Issue #267, opened 2017-03-23
Backdrops tab is titled "Costumes" when Stage is selected


towerofnix:
### Expected Behavior

The tab between "Scripts" and "Sounds" is titled "Backdrops" when the Stage is selected, as in Scratch 2.0.

### Actual Behavior

The tab is titled "Costumes", irregardless of whether a sprite or the stage is selected.

### Steps to Reproduce

Select the stage in the sprite/stage browser.

### Operating System and Browser

Chrome 57.0.2987.110 on macOS 10.12.

thisandagain:
Good catch! Thanks for documenting @liam4

towerofnix:
This doesn't look to have a patch? PR #376 (which was the patch to this previously) is now closed.

PR #376, opened 2017-05-20
Costume tab title changes to "Backdrops" when stage is selected

neurosie:
### Resolves

#267 

### Proposed Changes

The "Costumes" tab title now changes to "Backdrops" when the stage is selected.

### Reason for Changes

Matching behavior from Scratch 2.0.

### DIFF
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -25,6 +25,7 @@ const GUIComponent = props => {
         children,
         vm,
         onTabSelect,
+        costumeTabText,
         tabIndex,
         ...componentProps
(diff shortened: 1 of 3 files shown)


paulkaplan:
Welcome and thanks for the contribution! 

Code looks good in terms of being in line with how we are doing this elsewhere. I wonder @rschamp if there is some easier way to do this. @rschamp Is it ok to compute this value in the `mapStateToProps` instead of passing it all into the component to do?

rschamp:
@paulkaplan Yes computing in `mapStateToProps` would be preferable.

PR #381, opened 2017-05-22
Use immutable maps for the monitor state in gui

fsih:
Part of https://github.com/LLK/scratch-vm/pull/574

/cc @paulkaplan


rschamp:
On https://github.com/LLK/scratch-vm/pull/574#discussion_r117853568 you said:
> It works, but it gives me a weird warning in the console:
> Warning: Failed prop type: Invalid prop `monitors` of type `Map` supplied to `MonitorList`, expected instance of `Map`.
>     in MonitorList (created by MonitorList)

The first issue is because we're building in `immutable` twice, once in the VM, and then again in the GUI. So the GUI can't tell that the object coming from the VM is an instance of `Map` (since it's a different `Map` than the GUI is using). There are two ways to fix this: one is to do what @cwillisf attempted a while ago, and to wrap the build process for the VM into the GUI — that way the GUI would use the same reference to `immutable` as the VM. Alternatively we could make `immutable` a peer dependency for VM, and an `external` in webpack. That way the GUI could provide the same `immutable` for both the VM and GUI.

Or we could just disable that proptype check. Or maybe use something like https://www.npmjs.com/package/react-immutable-proptypes and see if that fixes it?

You also said
> Also I had to convert to JS in the GUI since it gave me a warning that having Maps as UI elements was a temporary feature that might go away.

What did you mean by that here? I don't see the conversion happening?

fsih:
I thought .valueSeq() converted to array, but actually it doesn't. What was causing the error before was when I left out .valueSeq(), here's the error

Warning: Using Maps as children is not yet fully supported. It is an experimental feature that might be removed. Convert it to a sequence / iterable of keyed ReactElements instead. Check the render method of `Box`.

For your other comments, I'm not really sure where to start on these options. Can we pair on it sometime, or leave it for another project since it's only a warning?

fsih:
We decided to leave the warning be for now, since the build change is kind of a big project that we can pursue if more things like this come up.

src/components/monitor-list/monitor-list.jsx
```diff
@@ -2,31 +2,31 @@ const React = require('react');
...
-        {props.monitors.map(monitorData => (
+        {props.monitors.valueSeq().map(monitorData => (
             <Monitor
-                {...monitorData}
+                color={monitorData.color}
                 key={monitorData.id}
+                label={monitorData.label}
+                value={monitorData.value}
+                x={monitorData.x}
+                y={monitorData.y}
```
rschamp: Just curious — did you see if you can use the spread operator with `Record`s?

fsih: https://github.com/facebook/immutable-js/issues/936#issuecomment-236502840

Found this discussion about it. We decided offline that this is better than converting to an object and then destructuring.

paulkaplan:
LGTM!

fsih:
@paulkaplan please check the merge. In particular, in order to not convert to array, I lost the index and the layout code no longer works. Do you have ideas on that? Also, I don't see where the color on the monitor is being set, but it's always orange now.

paulkaplan:
Looks like `monitor-record` is not defined in the GUI. 

Additionally, I think we will lose any benefit of using immutable if we do that opcode -> label transformation in the reducer since we will never have shallow object equality. Maybe that should be moved into the monitor component?

fsih:
Adding @rschamp back, since after the merge, we can no longer pass the state straight through, it needs to go through the new monitor adapter.

@paulkaplan argues that since we have to call mapEntries() in the reducer, which essentially converts toJS(), we might as well call toJS in the VM before emit and not use immutable in the GUI. Immutable is also not providing much benefit since the VM is only sending updates when it knows there was a change.

paulkaplan:
LG once travis is passing and vm branch is merged

Issue #375, opened 2017-05-19
Variables incorrectly displayed on stage


ethanmriley:
### Expected Behavior

Like Scratch 2.0, clicking the check mark next to a variable displays the name and value of that variable on the stage.

### Actual Behavior

Only one variable, "data_variable," is ever displayed on the stage, regardless of the number of selected variables. 
The value of "data_variable" is the value of whichever variable was most recently selected. 

![variable_display](https://cloud.githubusercontent.com/assets/12638154/26269981/3ee3829c-3cc6-11e7-9f97-92b812f4f084.gif)

### Steps to Reproduce

1. Navigate to [the Scratch gui](https://llk.github.io/scratch-gui/).
2. In the "Data" panel, create two variables with different values. 
3. Select one or the other, and the bug is evident. 

### Operating System and Browser

Windows 10, Chrome 58.0.3029.110 (64-bit)

TheBrokenRail:
Also, if you uses a motion of any other category reporter, the monitor still uses the variable colors.

PR #382, opened 2017-05-23
Take responsibility for labeling, categorizing and positioning monitors

paulkaplan:
### Resolves

Together with https://github.com/LLK/scratch-vm/pull/576 this fixes https://github.com/LLK/scratch-gui/issues/375

### Proposed Changes

Create an opcode => label/category mapping in the GUI side. Consume opcodes from monitor reporting from the VM. Also allow undefined positions to be automatically (albeit naively) positioned by the GUI. 

The opcode => label map allows for either static `label`s or dynamic `labelFn` that consume param data from the block and map it to dynamic labels. This is used for variable monitors (the name of the variable) and for monitors with dropdowns. 

monitor-labelling


src/lib/monitor-adapter.js
```diff
@@ -0,0 +1,25 @@
...
+const isUndefined = a => typeof a === 'undefined';
+
+module.exports = function ({id, opcode, params, value, x, y}, monitorIndex) {
+    let {label, category, labelFn} = OpcodeLabels(opcode);
+
+    // Use labelFn if provided for dynamic labelling (e.g. variables)
+    if (!isUndefined(labelFn)) label = labelFn(params);
+
+    // Simple layout if x or y are undefined
+    // @todo scratch2 has a more complex layout behavior we may want to adopt
```
fsih: Maybe write the bugs with this layout behavior in an issue somewhere so we don't forget

paulkaplan: @fsih Added a comment about it in 843182c.

PR #386, opened 2017-05-24
Use library mouseenter and mouseleave to cycle sprite costumes

paulkaplan:
### Proposed Changes

Use the new mouseenter and mouseleave hooks to cycle through a sprites costume when it is hovered over. 

costume-cycling

Note that as shown in the gif, if a sprites costume does not have the correct svg positioning, it will not show or it may show off center. That is a separate issue that we will have to deal with.

### DIFF
--- a/src/containers/sprite-library.jsx
+++ b/src/containers/sprite-library.jsx
@@ -11,18 +11,55 @@ class SpriteLibrary extends React.Component {
     constructor (props) {
         super(props);
         bindAll(this, [
-            'handleItemSelect'
+            'handleItemSelect',
+            'handleMouseEnter',
+            'handleMouseLeave',
+            'rotateCostume'
         ]);
+        this.state = {
+            activeSprite: null,
+            costumeIndex: 0,
+            sprites: spriteLibraryContent
+        };
+    }
+    componentWillReceiveProps (newProps) {
+        if (!newProps.visible) clearInterval(this.intervalId);
     }
     handleItemSelect (item) {
         this.props.vm.addSprite2(JSON.stringify(item.json));
     }
+    handleMouseEnter (item) {
+        this.setState({activeSprite: item});
+        if (this.intervalId) clearInterval(this.intervalId);
+        this.intervalId = setInterval(this.rotateCostume, 300);
+    }
+    handleMouseLeave () {
+        this.intervalId = clearInterval(this.intervalId);
+    }
+    rotateCostume () {
+        const costumes = this.state.activeSprite.json.costumes;
+        const nextCostumeIndex = (this.state.costumeIndex + 1) % costumes.length;
+        this.setState({
+            costumeIndex: nextCostumeIndex,
+            sprites: this.state.sprites.map(sprite => {
+                if (sprite.name === this.state.activeSprite.name) {
+                    return {
+                        ...sprite,
+                        md5: sprite.json.costumes[nextCostumeIndex].baseLayerMD5
+                    };
+                }
+                return sprite;
+            })
+        });
+    }
     render () {
         return (
             <LibraryComponent
-                data={spriteLibraryContent}
+                data={this.state.sprites}
                 title="Sprite Library"
                 visible={this.props.visible}
+                onItemMouseEnter={this.handleMouseEnter}
+                onItemMouseLeave={this.handleMouseLeave}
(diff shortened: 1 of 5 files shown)


fsih:
Looks good to me

src/containers/sprite-library.jsx
```diff
@@ -11,18 +11,55 @@ class SpriteLibrary extends React.Component {
...
+    componentWillReceiveProps (newProps) {
+        if (!newProps.visible) clearInterval(this.intervalId);
     }
     handleItemSelect (item) {
         this.props.vm.addSprite2(JSON.stringify(item.json));
     }
+    handleMouseEnter (item) {
+        this.setState({activeSprite: item});
+        if (this.intervalId) clearInterval(this.intervalId);
+        this.intervalId = setInterval(this.rotateCostume, 300);
```
rschamp: Since `this.rotateCostume` needs `this.state.activeSprite`, I think it would be more correct to set the interval in the callback for `setState`, so that `activeSprite` is ensured to be there.

src/containers/sprite-library.jsx
```diff
@@ -11,18 +11,55 @@ class SpriteLibrary extends React.Component {
...
-            'handleItemSelect'
+            'handleItemSelect',
+            'handleMouseEnter',
+            'handleMouseLeave',
+            'rotateCostume'
         ]);
+        this.state = {
+            activeSprite: null,
+            costumeIndex: 0,
+            sprites: spriteLibraryContent
```
rschamp: Does this need to be state?

src/containers/sprite-library.jsx
```diff
@@ -11,18 +11,55 @@ class SpriteLibrary extends React.Component {
...
+    handleMouseLeave () {
+        this.intervalId = clearInterval(this.intervalId);
+    }
+    rotateCostume () {
+        const costumes = this.state.activeSprite.json.costumes;
+        const nextCostumeIndex = (this.state.costumeIndex + 1) % costumes.length;
+        this.setState({
+            costumeIndex: nextCostumeIndex,
+            sprites: this.state.sprites.map(sprite => {
+                if (sprite.name === this.state.activeSprite.name) {
```
rschamp: It would be more efficient to pre-process `this.state.sprites` (== `spriteLibraryContent`) to be indexed by name, so that this can be a simple lookup.

src/containers/sprite-library.jsx
paulkaplan: TIL setState can take a callback! Cool!

src/containers/sprite-library.jsx
paulkaplan: Yup, because it is not === spriteLibraryContent. rotateCostume modifies it by changing the active sprites md5 (i.e. image url).

src/containers/sprite-library.jsx
paulkaplan: Hmm I think it is kind of a "six of one, half dozen of the other" situation. The library component is expecting an array, so either we convert it to an array from an object every time we need it, or we treat it as an array and do this array lookup/map when we change it. Same operations either way I believe.

src/containers/sprite-library.jsx
paulkaplan: Fixed in e70daab

src/containers/sprite-library.jsx
rschamp: Whoops I misread this last night. You're right.

paulkaplan:
@rschamp Implemented those performance boosts (35392d8, c5b288a) and included @fsih suggestion of rotating instantly instead of waiting.

paulkaplan:
@lifeinchords FYI I had to remove the `transform: scale(1.02, 1.02);` in the CSS because it was causing a 90ms repaint of the entire grid. We might be able to find a different way to reimplement the subtle transform that is more performant. It is fine to have that transform on elements that aren't in a large grid. Ray and I spent an hour tracking down these performance problems involving repainting, maybe we could give a little demo on how we found the problems, it is /mildlyinteresting.

liam4:
Should the costume reset to default when the mouse exits, maybe? The library ends up being kind of chaotic when moving the mouse over the "letter" costumes:

"-Story" letters in the library; the costumes are all "shuffled" from moving the mouse over, so they look different and confusing

PR #397, opened 2017-05-30
Made selected tab more prominent

lifeinchords:
Updates border bottom color of selected tab to the same color as the background of what's under it, to better show what's selected/unselected

Depends on #396 to keep the width set

tabs-fixed-border-v2 mov


paulkaplan:
LG

Issue #425, opened 2017-06-06
Pressing escape in library browser should cancel dialog


liam4:
### Expected Behavior

Pressing the escape key while the library browser dialog is visible should close the dialog, like in the "create variable" dialog.

### Actual Behavior

The escape key does nothing in the library filter dialog. This is inconsistent with the "create variable" dialog and most system popup behaviour I know (e.g. pressing the escape key in the "open" system dialog cancels the open operation).

### Steps to Reproduce

Open the library browser in any way (e.g. clicking on the "add costume" button). Press the escape key.

### Operating System and Browser

macOS 10.12.5, Chrome 58.0.3029.110.

paulkaplan:
Definitely agree! Submitted a patch for this, https://github.com/LLK/scratch-gui/pull/427

PR #427, opened 2017-06-07
Make escape key exit library modals

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/425

### Proposed Changes

Pass `onRequestClose` to the react modal. It can handle escape-to-close automatically if you give it something to do when escape is hit. We already do this in the `Prompt` component.

### Reason for Changes

Escape-to-close is a common expectation. @lifeinchords we have talked about implementing this, turns out it is a one-liner :)


liam4:
What a simple and elegant change :)

PR #433, opened 2017-06-09
Deploy all branches but master to subdirectories

rschamp:
Check out the results of this PR build here! https://rschamp.github.io/scratch-gui/deploy/multi-branch/

I assigned everyone who might be interested in this change, since we may want something like this change in other repos (www?). Not everyone has to review it though.

### Proposed Changes

_Currently this works using an unmerged PR to tschaub/gh-pages. (Will update to an LLK fork if we decide to go this way until that PR is merged.)_

When deploying from Travis, push all branches except master to a subdirectory. So the build for master would be available at https://llk.github.io/scratch-gui/ but to view the build for e.g., develop, you would look at https://llk.github.io/scratch-gui/develop/.

As part of this change, stop building for multiple versions of Node, since everything is static. This allows us to get rid of travis-after-all, which makes it easier to have multiple deploy conditions.

I also removed the need for the `BASE_DIR` environment variable, so that a single build script was sufficient for viewing the build under any subdirectory. 

Also add a `watch` command, for testing out static builds.

### Reason for Changes

This will make us able to only release to https://llk.github.io/scratch-gui/ after testing it out on https://llk.github.io/scratch-gui/develop/. Any Greenkeeper PR will be automatically available on the branch published by Greenkeeper, making reviewing those PRs easier.

For PRs coming from forks, this makes it very easy to publish the results for the PR, so the reviewer doesn't have to build it themselves. And allows the contributor to make many PR builds available at once.

## To do
- [ ] Fork gh-pages to LLK for the deploy-by-branch fork
- [ ] Periodic cleanup builds to remove deleted branches from gh-pages

### DIFF
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -119,6 +119,6 @@ GUIComponent.propTypes = {
     vm: PropTypes.instanceOf(VM).isRequired
 };
 GUIComponent.defaultProps = {
-    basePath: '/'
+    basePath: './'
 };
 module.exports = GUIComponent;
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,33 +3,32 @@ sudo: false
 dist: trusty
 node_js:
 - 6
-- 'node'
 cache:
   directories:
   - node_modules
 env:
   global:
-  - BASE_PATH=/scratch-gui/
   - NODE_ENV=production
 install:
 - npm --production=false install
 - npm --production=false update
-after_script:
-- |
-  # RELEASE_BRANCHES and NPM_TOKEN defined in Travis settings panel
-  declare exitCode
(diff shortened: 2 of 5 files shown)


src/components/gui/gui.jsx
```diff
@@ -119,6 +119,6 @@ GUIComponent.propTypes = {
     vm: PropTypes.instanceOf(VM).isRequired
 };
 GUIComponent.defaultProps = {
-    basePath: '/'
+    basePath: './'
```
cwillisf: Is this still needed after removing `basePath` & `BASE_PATH` in all those other places?

rschamp: Not strictly, but useful since it allows updating the location of the scratch-blocks media if the directory structure changes.

PR #452, opened 2017-06-13
Feature/jest setup

### LINKED ISSUE #421: Setup testing for scratch-gui repo
My understanding is that the scratch-gui repo doesn't have any test setup currently. I'd like to help by drafting a design doc (unless you have one already?) and help do some research. If the scratch team reaches a decision on type of testing, testing framework, etc I can help get the implementation started.

If there has been any research/notes on this already, please let me know.

[Design doc draft](https://docs.google.com/document/d/1Q77jROqZyOaVjHicwNlmgyE_OIlE7rBlczzj6miok3w/edit)

From Ray:

> Unit would be most useful at the beginning. There are several frameworks for doing that for React, but we haven't investigated the ins and outs of each, which is something that would be really helpful to do on that issue.
> 
> And for questions about scratch-gui me and paul would be the best people to ask

bogusred:
### Resolves

 #421

### Proposed Changes

Configures unit test setup for testing React components for easier maintainability, reduce bugs, and ensure confidence when merging future changes.
- Uses React testing framework which is designed for React
- Uses Enzyme to isolate unit tests and remove child components
- Uses React test renderer package to aid in rendering components and creating snapshot tests.
- Unit tests are now run as part of existing `npm test`
- Unit tests can also be run alone and with a watcher. Documentation added to `README.md`

### Reason for Changes

See [Design doc](https://docs.google.com/document/d/1Q77jROqZyOaVjHicwNlmgyE_OIlE7rBlczzj6miok3w/edit#)

### Test Coverage

I've added a simple unit test for just the green flag component to start. I'm happy to add additional tests in follow up pull requests.


bogusred:
Thank you for taking the time to review.
Please let me know if you'd like me to rebase or squash this. Happy to do so but wasn't sure if you could do that easily on your end.

test/unit/green-flag.test.jsx
```diff
@@ -0,0 +1,41 @@
...
+        const component = renderer.create(
+            <GreenFlag active={false} vm={vm}/>
+        );
+        expect(component.toJSON()).toMatchSnapshot();
+    });
+
+    test('triggers onClick when active', () => {
+        const onClick = jest.fn();
+        const componentShallowWrapper = shallow(
+            <GreenFlag active={true} onClick={onClick} vm={vm}/>
```
paulkaplan: It looks like this test should be testing the green flag component, instead of the container. Green Flag container should not consume an onClick that is called, only the component should do that. The Green Flag container should call `vm.greenFlag` on click. Does this test actually pass? It shouldn't??

bogusred: Oh good catch! Not sure how that happened. I'll take a look and try to determine why this passed and revise it to use the component instead of the container.
I haven't worked with React before so I appreciate your guidance. I'll send an update in the next few days.

paulkaplan: Ah I think I understand why the test passes. It is a quirk with the way we are passing props in. Because you passed in a `onChange`, and the render code looks like this:
```
            <GreenFlagComponent
                active={this.state.projectRunning}
                onClick={this.handleClick}
                {...props}
            />
```
one of those props in `{...props}` is actually another onClick, which because of the way object spreads work, will override the other onClick.

@rschamp it may be a good idea to always pass in the object spread _before_ the other props, to make sure this kind of thing doesn't happen. i.e. 
```
            <GreenFlagComponent
                {...props}
                active={this.state.projectRunning}
                onClick={this.handleClick}
            />
```

To check this, you might try changing the green flag container to the above ^ and then see the test should fail.

rschamp: I think I set it up this way to allow overrides like this test uses. E.g., what if you want to force the green flag to appear active in more instances than the default, but you still want it to handle the click event to run the project? If you don't allow overriding it then you need to write your own container that duplicates all the other functionality except for the `active` prop.

We can decide we don't want to allow that kind of usage, but I think right now almost everything allows overriding from the outside with that in mind.

bogusred: Thanks for your comments. Can you provide clarification as to next steps here? Should the test be changed? If so, in what way? Seems clear to me that at the very least this test should be moved to a 'containers' folder so it's more easily distinguished as the container vs the component.

or if you'd like me to write a test for a different component/container that is less contentious just let me know which one you'd recommend.

paulkaplan: So first thanks again for all of this, it is wonderful! 

One question that I have is how you feel / how the world feels about snapshot testing. Pros: it is insanely easy to implement. Cons: it hides what the "correct" state is, so you don't get the opportunity for your tests to feed backwards into the way you are writing the code (something that happens to me often when writing tests, i.e. if the test writes/reads badly then the code is usually not good). Do you have any opinion about snapshot testing?

As a next step, maybe let's try adding a test for another component / container pairing. I'm particularly interested in how the redux model of separating component / container plays out in testing, which is where it is meant to shine.

rschamp: I think I'm happy with snapshot testing as scaffolding, since currently we have nothing. Once there's a base of tests, we should start writing them as we work on new features/bugs, and can remove snapshot tests later on (or not if they end up being useful).

bogusred:
Okie dokie. I'll send along an update within the next few days

test/unit/green-flag.test.jsx
bogusred: After working with this some more, I think where possible, we should use enzyme to do dom assertions because snapshot testing is:

- not declarative of what you're testing
- prone to human error because you have to look at a long dom snapshot

For example, with the sprite selector item component, the dom is different depending on whether it's in a selected state, unselected state and if a costumeURL property is passed. I could have tested this with the snapshot (which I did try) but I found using the enzyme tools much more declarative. e.g.

```
    test('does not have a close box when not selected', () => {
        selected = false;
        const componentShallowWrapper = shallow(getComponent());
        // this is very explicitly checking for a close button
        expect(componentShallowWrapper.find(CloseButton).exists()).toBe(false);
    });
```

vs this where i have to go look at the snapshot file to confirm i got what i expected (which is prone to human error)
```
    test('does not have a close box when not selected', () => {
        selected = false;
        const component = renderer.create(getComponent());
        // there's nothing here confirming that the close box is or isn't there
        expect(component.toJSON()).toMatchSnapshot();
    });
```

bogusred: Also i did want to say I don't think snapshot testing is bad or anything. It's better than nothing and easier to write. But where possible, it's easier to maintain, understand, and is safer to use enzyme.

bogusred:
pull request updated with these changes:
- make it clear the existing green-flag test is a container test
- add button component test
- add sprite-selector-item component test
- add sprite-selector-item container test

Please let me know if you'd like additional changes.

bogusred:
Hi I just wanted to follow up. It's been about two weeks. Do you think you'll have a chance to review soon? Thanks for your consideration.

paulkaplan:
Hi sorry about the delay! Thank you so much for putting in this work. @rschamp I think that we should merge and continue iterating on this w/r/t to whether we should use react-test-renderer vs. enzyme to json for the snapshotting. I think it is best to just start using it and see how it goes! I can resolve that conflict with the package json btw.

Kenny2github:
@paulkaplan it seems like you think that this pull is good to go. However, the actual status still says "changes requested" - so are there changes that still need to be made? If not, I'd suggest that you change your review status.

bogusred:
Huzzah! Thank you Paul for the review and for accepting my pull request! ❤️

paulkaplan:
No thank you for the contribution! I'm very excited to start using it and exploring it. There are some remaining questions I have about the usefulness of snapshot testing, whether we can combine the two test renderers we are using, and how to best test containers/components, but this is a terrific start. Thanks again!

bogusred:
Glad I could help! Good luck with it. I'm sure after trying different options you'll settle on whatever works best for your team and project. Let me know how it goes!

Issue #296, opened 2017-04-04
Record a new sound from the mic


paulkaplan:
Create a way to record your own sounds.

PR #469, opened 2017-06-22
Record a new sound from the mic

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/296

### Proposed Changes

Adds a new microphone recording flow:
mic-full-flow

/cc @carljbowman @thisandagain


src/lib/audio/audio-recorder.js
```diff
@@ -0,0 +1,165 @@
...
+        onUpdate(rms);
+    };
+
+    requestAnimationFrame(update);
+
+    // Wire everything together, ending in the destination
+    this.mediaStreamSource.connect(this.sourceNode);
+    this.sourceNode.connect(this.analyserNode);
+    this.analyserNode.connect(this.scriptProcessorNode);
+    this.scriptProcessorNode.connect(this.audioContext.destination);
```
ericrosenbaum: This line is unnecessary as far as I know!

paulkaplan: (addressed offline, for posterity) it looks like it is needed.

paulkaplan:
I'm wondering if the `lib/audio` code should really live in a separate repo (perhaps scratch-audio?) At first I thought the functionality was completely separate from scratch-audio, but since seeing @ericrosenbaums work recently, there is definitely a lot of overlap (like sampling RMS from an `analyserNode` and playing sound buffers). It would also keep the GUI as just front-end components. @ericrosenbaum what do you think about that? Is there a way we could use the `audioEngine` for the visualization / recording / playback I'm doing here?

mrjacobbloom:
I'd lump the audio editor in with the paint editor, which seems big enough to live in its own repo. Just my 2 cents

src/containers/record-modal.jsx
```diff
@@ -0,0 +1,123 @@
...
+                    md5: `${md5}.wav`,
+                    name: `recording ${this.props.vm.editingTarget.sprite.sounds.length}`
+                };
+
+                // Load the encoded .wav into the storage cache
+                const storage = this.props.vm.runtime.storage;
+                storage.builtinHelper.cache(
+                    storage.AssetType.Sound,
+                    storage.DataFormat.WAV,
+                    new Uint8Array(wavBuffer), // @todo ask cwf about the need for this?
```
rschamp: @cwillisf There's a question here for you ^

src/containers/record-modal.jsx
```diff
@@ -0,0 +1,123 @@
...
+    handleStopPlaying () {
+        this.setState({playing: false});
+    }
+    handleBack () {
+        this.setState({playing: false, samples: null});
+    }
+    handleSubmit () {
+        this.setState({encoding: true});
+
+        // Allow enough time for UI refresh before starting encoding
```
rschamp: What's the refresh we're waiting for? Can this be scheduled more accurately than with a timeout?

src/lib/audio/audio-buffer-player.js
```diff
@@ -0,0 +1,26 @@
+const SharedAudioContext = require('./shared-audio-context.js');
+
+const AudioBufferPlayer = function (samples) {
```
rschamp: Is there a reason this doesn't use the new class syntax?

src/lib/audio/audio-recorder.js
```diff
@@ -0,0 +1,128 @@
+const SharedAudioContext = require('./shared-audio-context.js');
+
+const AudioRecorder = function () {
```
rschamp: Same question here re: class syntax.

src/containers/record-modal.jsx
paulkaplan: So my hope was that if the wav encoding took a bit of time, we would have enough time to switch this to say `Loading...` before starting the encoding process. Does putting this in a callback of `setState` do that?

rschamp: Oh, I see. Yeah, as far as I know, the callback works for that purpose. I think it's called after the re-render. `componentDidUpdate` is the recommended way, but that doesn't seem right for a handler like this.

src/components/record-modal/record-modal.css
```diff
@@ -0,0 +1,183 @@
...
+
+.modal-overlay {
+    position: fixed;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    z-index: 1000;
+    background-color: rgba(94, 102, 125, .95);
+}
```
paulkaplan: Yeah we no longer have a generic small modal component (the large modal component is now full-screen and uses a text filter in the nav bar, so isn't general enough anymore). Should probably be refactored.

rschamp:
We should probably resolve the way to access the modal before merging.

paulkaplan:
Almost forgot about that ray. I added that and also refactored the way the modal worked so that it could be used by the three modal-y things (prompt, recorder and library). Paired with Carl on the add buttons. There are some unusual border-related edge cases, but we aren't concerned about them. 

image

PR #476, opened 2017-06-28
Mic recording with trimmers

paulkaplan:
Add trimmers and a playhead to recording. 

recording-with-trimmers


paulkaplan:
@rschamp @fsih can i get a review on this? Let me know if you have any questions!

src/lib/audio/audio-buffer-player.js
```diff
@@ -6,15 +6,40 @@ class AudioBufferPlayer {
...
+        const percentage = timeSinceStart / this.buffer.duration;
+        if (percentage + this.trimStart < this.trimEnd && this.source.onended) {
+            requestAnimationFrame(this.update.bind(this));
+            this.updateCallback(percentage + this.trimStart);
+        } else {
+            this.updateCallback = null;
+        }
     }
 
     stop () {
```
fsih: Should you remove the references to the functions like trimStart, trimEnd in stop()?

paulkaplan: Since those are just values, I was thinking they are ok to just leave to get normally garbage collected. The only reason I did some manual cleanup for certain things is because if you leave callback references to react components around, they can't be garbage collected and can cause you problems.

PR #479, opened 2017-06-28
Fix AudioContext issue in Safari

as-com:
### Resolves
Fixes white screen in Safari (and possibly old Chrome) caused by use of `AudioContext`

I didn't bother to open an issue

### Proposed Changes
Uses `webkitAudioContext` when `AudioContext` is not found

### Reason for Changes
Because people use Safari

### Test Coverage

???


paulkaplan:
So i tried this locally and it still fails on safari, I think you need to add `window.` to `AudioContext` as well to prevent undefined variable errors.

paulkaplan:
In particular, it looks like fixing that issue brings up more audio-related issues in Safari, if you are interested the repo for that is https://github.com/LLK/scratch-audio.

PR #482, opened 2017-06-29
"Add sprite/backdrop" buttons should have a discernible text

nterray:
In order to comply to WCAG 2.0[1], buttons must have discernible text that
clearly describe the purpose, so that it can be used by screen readers.

[1] https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/H65


liam4:
It might be good to keep an eye out for other buttons, too – off the top of my head I can think of the green flag and stop sign buttons ("start project" and "stop project", respectively), and the asset deletion buttons (in the costumes and sounds tabs, as well as in the sprite palette).

nterray:
Green flag and stop buttons already have a title ("Go" and "Stop", respectively). :)

Issue #477, opened 2017-06-28
Implement translation for Scratch-gui


chrisgarrity:
- [x] Include react-intl in scratch-gui and provide example of translating strings
- [ ] implement locale switching based on browser locale
- [x] implement locale switching in the ui
- [x] auto generate l10n source data from jsx source with babel plugin
- [x] process to update transifex from source l10n
- [x] process to pull translations from transifex - could be npm translations module?
- [x] build process including tx translations

paulkaplan:
@chrisgarrity is this issue referring to the strings in the GUI like "Record a sound", "Add costume", etc?

chrisgarrity:
@paulkaplan Yes, all the things that aren't the blocks.

paulkaplan:
@chrisgarrity can we split this issue into the units of work so that they can be prioritized and closed by individual PRs.

zyctinker:
hi I'm interested in this plan. I want to know when will you implement locale switching in the ui so that my students can use scratch 3.0 .   If you provided the API, I can take the task of Chinese translation. Thank you.

chrisgarrity:
@zyctinker this is just the start of the process of localizing 3.0 and doesn't include translation of the blocks themselves yet. When we're ready to have Scratch 3.0 translated, there will be a new project on our translation platform: https://www.transifex.com/llk/public/

chrisgarrity:
Tracking in 
- #507
- #509 
- #510 
- #511 
- #517 
- #518 
- #524

PR #487, opened 2017-06-30
Localize first elements

chrisgarrity:
First part of #477 

Bring in react-intl and demonstrate use of FormatMessage and defineMessages in costume-tab and sound-tab.

### DIFF
--- a/src/components/asset-panel/selector.jsx
+++ b/src/components/asset-panel/selector.jsx
@@ -1,5 +1,6 @@
 const PropTypes = require('prop-types');
 const React = require('react');
+const {FormattedMessage} = require('react-intl');
 
 const SpriteSelectorItem = require('../../containers/sprite-selector-item.jsx');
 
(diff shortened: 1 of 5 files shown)


paulkaplan:
@chrisgarrity woops looks like the linter caught a variable we didn't remove 'messageDescriptorPropTypes'

src/components/asset-panel/selector.jsx
```diff
@@ -55,7 +56,11 @@ const Selector = props => {
 
 Selector.propTypes = {
     buttons: PropTypes.arrayOf(PropTypes.shape({
-        text: PropTypes.string.isRequired,
+        message: PropTypes.shape({
+            id: PropTypes.string.isRequired,
+            defaultMessage: PropTypes.string,
+            description: PropTypes.string
+        }),
```
rschamp: This API seems unintuitive to me, since it makes localization information a part of the spec of the component. For components in the future, could we keep the `text` prop, but make a new PropType that can be either a string or a `FormattedMessage` component (e.g. `text: CustomPropTypes.localizedString`)?  I think that would help organizationally, since it would keep the localization info contained to the place it's used — here we have to update the component tree to account for it.

paulkaplan: After talking with @rschamp and @chrisgarrity, we should 
- [ ] change this prop to take a `FormattedMessage` (how to do proptypes for that? PropTypes.node? PropTypes.shape({type: FormattedMessage})?)
- [ ] change `defineMessage` in sound/costume tab to just be `FormattedMessage` directly

PR #489, opened 2017-06-30
Create base text wrapper & SVG implementation

cwillisf:
### Proposed Changes

This change adds a new family of classes designed to wrap text to a maximum width. The base `TextWrapper` class can theoretically be used to wrap text in any context, respecting Unicode grapheme cluster boundaries and (when possible) Unicode line break opportunities. It takes a "measurement provider" as an argument, and will use that provider to measure text widths. The `SVGTextWrapper` is some light glue to connect an SVG-based measurement provider to the base `TextWrapper`, though it will likely have a few additional features in the future. For example, I anticipate that it will have a method to return a single SVG node containing several lines of wrapped text.

### Reason for Changes

This is a step toward implementing say & think bubbles.

### Test Coverage

None, currently. We could potentially test the wrapping algorithm by mocking a measurement provider, but we'd need a DOM (or a simulated DOM) to test the SVG version.

### Known issues

Due to a limitation in the `grapheme-breaker` library I'm using, it's currently possible for a skin tone modifier to become separated from its associated emoji due to a line break. For more information, see:
- devongovett/grapheme-breaker#3
- devongovett/grapheme-breaker#4


package.json
```diff
@@ -38,8 +38,11 @@
     "eslint-config-scratch": "^3.0.0",
     "eslint-plugin-react": "^7.0.1",
     "gh-pages": "github:rschamp/gh-pages#publish-branch-to-subfolder",
+    "grapheme-breaker": "0.3.2",
     "html-webpack-plugin": "2.28.0",
+    "ify-loader": "1.0.4",
```
cwillisf: The `grapheme-breaker` and `linebreak` modules use `fs` in such a way that Webpack can't handle them directly. I found the suggestion to use `ify-loader` here: devongovett/linebreak#1

mrjacobbloom:
Does that mean the say/think blocks will probably support emojis? :o :D

liam4:
>:o :D

I believe you mean 😮 😁 ? 😀

cwillisf:
🤔😁😉

paulkaplan:
@cwillisf we now have a test system for utilities (you can check out `test/unit/util` for an example). If you still want to add any tests to this, jest is pretty easy to get into, I can give you a little intro if you want.

cwillisf:
@rschamp and I agreed to close this for now and bring it back once it's being actively used. The code will live on in my fork until then...

Issue #503, opened 2017-07-06
Shouldn't be able to start recording before mic access is given


paulkaplan:
### Expected Behavior

_Please describe what should happen_
You should not be able to start a recording without first giving access to your microphone.

### Actual Behavior

_Describe what actually happens_
You can hit the record button before giving mic access, which starts a "blank" recording, and probably breaks a bunch of other things. 

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_
Open in private browsing, click "record sound", click record before giving mic access.

---

This is a bug proper, but there are related design questions from the playtest, i.e. this from Andrew
> Is there a way we can make giving microphone access even more overt / obvious?"

paulkaplan:
I'm going to pull the design tag off of this and make a new issue for that part. I'll leave this as just the bug that you can start recording without mic access, which is solved by that PR #504

PR #504, opened 2017-07-06
Disable record button before mic access is given

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/503

### Proposed Changes

Disable the record button (and gray it out) before mic access is given.

### Reason for Changes

It should not be possible to start the recording before mic access is given. This doesn't really address the design problem mentioned in that issue, let's make a separate issue for that. 

@carljbowman thoughts? this is what it looks like

screen shot 2017-07-06 at 10 55 48 am
screen shot 2017-07-06 at 10 55 54 am

Note: the button is actually disabled, it cannot be clicked.

### DIFF
--- a/src/components/record-modal/record-modal.css
+++ b/src/components/record-modal/record-modal.css
@@ -64,6 +64,10 @@
     border: none;
 }
 
+.main-button-row button:disabled {
+    opacity: 0.25;
+}
+
 .main-button-row button:active, .main-button-row button:focus {
     outline: none;
 }
--- a/src/components/record-modal/recording-step.jsx
+++ b/src/components/record-modal/recording-step.jsx
@@ -27,7 +27,8 @@ const RecordingStep = props => (
                     />
                 ) : (
                     <span className={styles.helpText}>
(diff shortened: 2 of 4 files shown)


CosmicWebServices:
You should also change the text "Begin recording by clicking the button bellow" to something like "We need permission to use your mic"

paulkaplan:
@carljbowman @CosmicWebServices what about something like this?

screen shot 2017-07-06 at 1 18 34 pm

paulkaplan:
@ericrosenbaum thanks. Fixed by making sure that no callbacks are called after the audio recorder is disposed of.

CosmicWebServices:
@paulkaplan Nice but also I do not think all browsers have a dialog in that area...

paulkaplan:
@CosmicWebServices yeah, i think we'll have to deal with this more completely later in that linked issue. There is also the case where you deny mic access accidentally and then you need to click on some other button in the top right to give it (which we do not handle correctly right now).

Issue #497, opened 2017-07-06
Change the word "scripts" to "blocks"


ericrosenbaum:
We currently use the word "scripts" on the tab that shows you the blocks palette and blocks workspace. Let's consider changing this to the word "blocks."

carljbowman:
YAAAAAS!

CosmicWebServices:
Is there confusion related to this wording?

angryboy2016:
here i rename scripts to blocks on first tab

![scratch 3 0](https://user-images.githubusercontent.com/24998969/27932518-dff881c6-6252-11e7-90d3-ad24ab2e9013.PNG)

thisandagain:
@CosmicWebServices Yes. We use the word "blocks" throughout our help materials, but very rarely use the word "script" or "scripts". While testing with people who have never used Scratch before we have found that the use of the word "Scripts" can be confusing.

PR #522, opened 2017-07-10
Change 'Scripts' tab to 'Blocks'

thisandagain:
### Resolves

GH-497

### Proposed Changes

- Changes the "Scripts" tab to "Blocks"

### Reason for Changes

- Resolve inconsistency and confusion observed during play testing

### Test Coverage

- None


paulkaplan:
@thisandagain since https://github.com/LLK/scratch-gui/pull/487 we have `FormattedMessage` available for internationalizing strings. Wanna use that now, or just update it later 😜

PR #541, opened 2017-07-12
Initial sound editor area with playback and renaming

paulkaplan:
Depends on https://github.com/LLK/scratch-vm/pull/634 for the public API on the VM to rename sounds.

This is just the first few features of the sound editor, as well as some minor refactoring that was required to DRY things up. 

Included: 
- Sound renaming
- Sound playback

Recommend looking through commit-by-commit, as it may seem like a somewhat random set of refactors when taken as a whole.

image


src/css/forms.css
```diff
@@ -0,0 +1,54 @@
...
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    min-width: 0;
+}
+
+.input-form:focus {
+    border-color: #4c97ff;
+    box-shadow: inset 0 0 0 -2px rgba(0, 0, 0, 0.1);
+}
```
rschamp: Does something about the way we use inputs prevent us from having form components instead of this separate stylesheet? Could we refactor these into reusable components with their own styles instead? I would like to avoid having a library of classes not tied to specific components if possible, since this is how we get to unwieldy stylesheets where we don't know which classes are used where, or what will happen when we change them.

src/lib/audio/audio-util.js
```diff
@@ -0,0 +1,18 @@
...
+        const rms = Math.sqrt(sum / samples.length);
+        // Scale it
+        const unity = 0.55;
+        const val = rms / unity;
+        // Scale the output curve
+        return Math.sqrt(val);
+    }
+}
+
+module.exports = AudioUtil;
```
rschamp: Are tests possible/useful for this?

src/lib/audio/audio-util.js
```diff
@@ -0,0 +1,18 @@
+class AudioUtil {
+    static computeRMS (samples) {
```
rschamp: Now that we're using `import/export` syntax, we could just export this by itself.

src/containers/sound-editor.jsx
```diff
@@ -0,0 +1,110 @@
...
+            playhead: null, // null is not playing, [0 -> 1] is playing percent
+            chunkLevels: getChunkLevels(this.props.samples)
+        };
+    }
+    componentDidMount () {
+        this.audioBufferPlayer = new AudioBufferPlayer(this.props.samples, this.props.sampleRate);
+    }
+    componentWillReceiveProps (newProps) {
+        if (newProps.soundIndex !== this.props.soundIndex) {
+            this.audioBufferPlayer.stop();
```
rschamp: Does this dispose of the old samples?

src/css/forms.css
paulkaplan: Yeah we could do that. Do you think I should put the styles directly on `BufferedInput`, or create a HOC for buffered input that applies these inline-label-roundy-input styles?

src/lib/audio/audio-util.js
paulkaplan: Noob question about the export syntax style. I see that we use an eslint rule that that "requires the use of function expressions instead of function declarations" http://eslint.org/docs/rules/func-style. 

That prevents doing this, which I see in some codebases
```javascript
export function computeRMS (...) { ... }
```

 Is the proper way in "util" type files with multiple named exports to export at the bottom like this?

```javascript
const computeRMS = function (...) { ... }

// possibly other util functions

export {
    computeRMS,
    // the other named utils
};
```

rschamp: I think you can do `export computeRMS = function (...) { ... }` instead?

src/css/forms.css
rschamp: I think it makes sense to put the buffering behavior in a HOC, but perhaps the styles should be applied to an `Input` component that gets wrapped in that HOC?

paulkaplan: Ah gotchya. 54a9115 refactors the input stuff similar to the way www does it (styled dumb components + HOCs). There is definitely more that could be refactored there (which I think jest testing will make obvious), but seems like a good start.

src/containers/sound-editor.jsx
paulkaplan: It doesn't, but the next line reassigns so the whole thing will get garbage collected.

src/lib/audio/audio-util.js
```diff
@@ -0,0 +1,27 @@
+const computeRMS = function (samples, scaling = 0.55) {
```
rschamp: Also still curious about tests for these — is that practical?

src/components/sprite-info/sprite-info.jsx
```diff
@@ -3,14 +3,18 @@ import PropTypes from 'prop-types';
...
+import BufferedInputHOC from '../forms/buffered-input-hoc.jsx';
+
 import styles from './sprite-info.css';
 
 import xIcon from './icon--x.svg';
 import yIcon from './icon--y.svg';
 import showIcon from './icon--show.svg';
 import hideIcon from './icon--hide.svg';
 
+const BufferedInput = BufferedInputHOC(Input);
```
paulkaplan: We could but it would make the import component dependent on the hoc, which seems wrong. I think a different file might be warranted for common combinations of stuff from forms. For example, all the inputs are used with `buffered` but also used together with `label`, so maybe one component that has that pattern.

paulkaplan: We could but it would make the import component dependent on the hoc, which seems wrong. I think a different file might be warranted for common combinations of stuff from forms.

src/lib/audio/audio-util.js
paulkaplan: It is now that we have a test system! Check 8c2ced0. 

Thanks for pushing on that, writing those tests helped find an off-by-one error in the chunking code.

Issue #511, opened 2017-07-06
Add locale switching to the UI


chrisgarrity:
Will need a way for the user to switch to the language of their preference regardless of browser setting. Should be a saved preference for signed in users.

Part of #477

CosmicWebServices:
I am assuming that this could use the local storage api (or a cookie) unless you want this stored on server?

chrisgarrity:
@CosmicWebServices this is a place holder issue for designing the widget for switching languages (where should it be, icon? dropdown?, etc). Most of the other parts of #477 can get done without this, so splitting it out.

paulkaplan:
Implemented!

PR #551, opened 2017-07-19
Add language selection widget to menubar

chrisgarrity:
### Resolves

https://github.com/LLK/scratch-gui/issues/511 (add locale switching to UI)

### Proposed Changes

Adds a language selection widget to the menubar. Design is temporary (especially icon etc), this PR is just to get functionality working.

Ended up moving languages into src from the root directory, and replacing intl.jsx in lib with reducers/intl.js.


paulkaplan:
Pulling this branch and running the dev server throws an error:
```
Module not found: Error: Can't resolve '../../locale/messages.json'
```

src/reducers/intl.js
```diff
@@ -1,5 +1,26 @@
...
+// start with English
+export const intlInitialState = {
+    intl: {
+        defaultLocale: 'en',
+        locale: 'en',
+        messages: messages.en
+    }
+};
+
+export const updateIntl = locale => superUpdateIntl({
```
paulkaplan: This may be a style preference, but I would vote for us keeping all of our exports on a single line at the end. This file exports 4 things including a default, but you have to look all over the file to see that. I would vote for something more like

```javascript
...
export {
    IntlProvider,
    intlInitialState,
    updateIntl,
    intlReducer as default
}
```

or similar. @rschamp?

chrisgarrity:
@paulkaplan messages.json is generated by `npm run i18n:msgs`. It's run automatically for build, but you need to run it manually otherwise. Would it be better if `start` also generated messages? Also, folks won't have translations to use when generating messages unless they have transifex command line installed. The messages.json that you need for testing is: 
```{
  "en": {
    "action.addBackdrop": "Add Backdrop",
    "action.addCostume": "Add Costume",
    "action.recordSound": "Record Sound",
    "action.addSound": "Add Sound"
  },
  "es": {
    "action.addSound": "Add Sound",
    "action.addCostume": "Add Costume",
    "action.addBackdrop": "Add Backdrop",
    "action.recordSound": "Record Sound"
  },
  "fr": {
    "action.addSound": "Ajouter Son",
    "action.addCostume": "Ajouter Costume",
    "action.addBackdrop": "Ajouter Arrière-plan",
    "action.recordSound": "Enregistrement du Son"
  }
}
```
In the long term this should be an npm package.

paulkaplan:
Hmmm yeah not sure what the best fix is, but it definitely needs to be able to run from a clean clone of the repo using the dev server.

src/reducers/intl.js
chrisgarrity: Yes, that came from combining two files. Putting them all together at the end makes sense. It becomes a little less clear that IntlProvider and intlReducer are coming from `react-intl-redux`, but that's probably less important now.

paulkaplan: FWIW the airbnb javascript style guide (a generally sane and reliable source) explictly recommends against the export-from-import one liner. https://github.com/airbnb/javascript section 10.3: 
> Why? Although the one-liner is concise, having one clear way to import and one clear way to export makes things consistent.

src/components/language-selector/language-selector.jsx
```diff
@@ -0,0 +1,44 @@
...
+            <img
+                className={styles.languageIcon}
+                src={languageIcon}
+            />
+            <select
+                className={styles.languageSelect}
+                value={currentLocale}
+                onChange={onChange}
+            >
+                {Object.keys(languages).map(loc => (
```
rschamp: Could we change this name to `locale` for obviousness?

src/reducers/intl.js
rschamp: +1 to having all exports in one place.

src/components/language-selector/language-icon.svg
```diff
@@ -0,0 +1,72 @@
...
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
```
rschamp: I know this is just an svg but what the heck Illustrator?

src/components/language-selector/language-selector.jsx
chrisgarrity: Good catch, that was left over because originally currentLocale was just called locale and lint complained.

src/components/language-selector/language-icon.svg
chrisgarrity: I'm sure that Carl will design something better :)

rschamp:
@paulkaplan We decided to add the translation step ahead of both `start` and `build` until the translations can be installed from a separate npm module.  When that happens, they can just be required directly rather than needing this build step.

PR #558, opened 2017-07-21
localeData cleanup

### LINKED ISSUE #557: Clean up intl-related build artifacts
Due to this [dynamic import line](https://github.com/LLK/scratch-gui/blob/bc569e4fc6c4cbd13e3d0d4714d8c45e1764998f/src/reducers/intl.js#L10), data from all 200+ languages in `react-intl` are bundled by webpack into separate files, causing a small explosion in our build files (and build time).

To avoid this, update `languages.json` to an object like:

```javascript
import localeDataEn from 'react-intl/locale-data/en';
// other languages imported explicitly here...

import messages from '/path/to/messages.json';

export default {
  en: {
    name: 'English',
    localeData:  localeDataEn,
    messages: messages.en
  },
  // ...
};
```

This will simplify our localization data into one object, and should give webpack more hints for how to bundle the languages we're using into the main `gui.js` bundle.

chrisgarrity:
### Resolves

#557 

### Proposed Changes

move all locale data (react-intl/locale-data and gui messages) into locale.js. This also makes switching to an external node package easier because the package just needs to export the same thing as locale.js.

### DIFF
--- a/scripts/generate-locale-messages.js
+++ b/scripts/generate-locale-messages.js
@@ -42,12 +42,11 @@ Missing locales are ignored, react-intl will use the default messages for them.
 const fs = require('fs');
 const path = require('path');
 const mkdirp = require('mkdirp');
-const languages = require('../src/languages.json');
 
+const locales = ['en', 'es', 'fr'];
 const LANG_DIR = './translations/';
 const MSGS_DIR = './locale/';
 
-const locales = Object.keys(languages);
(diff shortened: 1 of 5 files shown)


chrisgarrity:
Ok, I rebased to fix merge conflicts. These changes will suffice until the new translations package is ready, and should be ready to review now @rschamp.

Issue #466, opened 2017-06-22
Can't add "Text" sprite from library


Kenny2github:
### Expected Behavior

The "Text" sprite in the library should be added to the project when clicked

### Actual Behavior

When clicked, nothing happened

### Screenshot

![scratch-gui-text-sprite](https://user-images.githubusercontent.com/28599280/27424533-f39ad904-5767-11e7-83c3-70760bf3002d.gif)

### Steps to Reproduce

Press the + to add a new sprite, search "text", click the lone sprite that appears, and nothing will happen

### Operating System and Browser

Windows 10 Chrome 58

paulkaplan:
@rschamp looks like one of the costume assets on this sprite does not exist, specifically 

```
                {
                    "costumeName": "text keep scratching",
                    "baseLayerID": -1,
                    "baseLayerMD5": "22a916248a294eadc07a7a065f5154e8.svg",
                    "bitmapResolution": 1,
                    "rotationCenterX": 135,
                    "rotationCenterY": 24
                },
```
which 404s on the URL https://cdn.assets.scratch.mit.edu/internalapi/asset/22a916248a294eadc07a7a065f5154e8.svg/get/

christan1029:
Only the "keep scratching" costume URL 404's; the other costume URLs are fine. Any way to find the actual URL? I tried looking for it in the scratch-flash repo, but I wasn't able to find it.

This is the costume:
![image](https://user-images.githubusercontent.com/29719007/28345067-8dc02a4c-6bf5-11e7-876f-e424e8cf9b7f.png)

liam4:
I've got the Scratch 2.0 link:

<pre>
<a href="https://cdn.assets.scratch.mit.edu/internalapi/asset/b3de9a77e2c1bb2f3e29f29a03d869ea.svg/get/">/asset/b3de9a77e2c1bb2f3e29f29a03d869ea.svg/get/</a> # Scratch 2.0
<a href="https://cdn.assets.scratch.mit.edu/internalapi/asset/22a916248a294eadc07a7a065f5154e8.svg/get/">/asset/22a916248a294eadc07a7a065f5154e8.svg/get/</a> # Scratch 3.0
</pre>

liam4:
Interestingly, the sprite library's JSON is stored [in the scratch-gui codebase itself](https://github.com/LLK/scratch-gui/blob/4352783b18dd26e8ef0bb6ad71e6a62fd62714d3/src/lib/libraries/sprites.json).

[...]

PR #561, opened 2017-07-24
Fix 404-ing costume in "Text" sprite

Kenny2github:
### Resolves

Solves #466

### Proposed Changes

Replaced 404-ing link to 3.0 asset with 2.0 asset

### Reason for Changes

The entire "Text" sprite was un-addable because of one 404-ing costume.


paulkaplan:
Hmmm it seems that the SVG has some quirk that isn't allowing the thumbnail to be rendered correctly. @rschamp has dealt with this issue several times before, he can probably help with this. We can't use this as-is because this is what it looks like as a thumbnail (note that it is fine on the stage, just the thumbnail is messed up)
image

Kenny2github:
@paulkaplan Okie dokie, once @rschamp looks into it I'll make the relevant changes - then this problem should be completely fixed 😄

Kenny2github:
@paulkaplan any updates?

src/lib/libraries/sprites.json
```diff
@@ -14328,7 +14328,7 @@
                 {
                     "costumeName": "text keep scratching",
                     "baseLayerID": -1,
-                    "baseLayerMD5": "22a916248a294eadc07a7a065f5154e8.svg",
+                    "baseLayerMD5": "b3de9a77e2c1bb2f3e29f29a03d869ea.svg",
```
rschamp: Please update to `97c9e6ef78d4f1987fd8c6e5042963cb.svg` and it will start working 🌈

Kenny2github: Will do!

Kenny2github: Done!

PR #567, opened 2017-07-26
Fix color conflict

### LINKED ISSUE #564: Language selector options are invisible until hovered over
### Expected Behavior

You should be able to see all the options for languages before choosing

### Actual Behavior

Instead, the background color AND the font color are both white - causing the options to be invisible until hovered over, when the background changes to blue.

### Steps to Reproduce

1. Open [scratch-gui](https://llk.github.io/scratch-gui/develop)
2. Click the language selector.
3. You can't see any options:
image

### Probable cause
In one of the `style`s, we have:
```css
.language-selector_language-select_8Vfnm {
    width: 100%;
    height: 1.85rem;
    border: 1px solid #4c97ff;
    -webkit-user-select: none;
    -moz-user-select: none;
    -ms-user-select: none;
    user-select: none;
    outline: none;
    background: rgba(255, 255, 255, 0.2);
    color: white;
    font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
}
```
As you can see, `background` is white (well, RGBA white) and `color` is also white.

Kenny2github:
### Resolves

#564 - Language selector options were invisible until hovered over

### Proposed Changes

Changes the text color to black, while keeping the background color as white

### Reason for changes

So that language menu options can be instantly seen, instead of having to hover over each one to find your language.

### DIFF
--- a/src/components/language-selector/language-selector.css
+++ b/src/components/language-selector/language-selector.css
@@ -18,6 +18,6 @@
     user-select: none;
     outline: none;
     background: rgba(255, 255, 255, 0.2);
-    color: white;
+    color: black;
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
 }
--- a/src/components/language-selector/language-selector.css
+++ b/src/components/language-selector/language-selector.css
@@ -9,6 +9,7 @@
 
 .language-icon {
     height:  2rem;
+    display:none;
 }
 
 .language-select {
@@ -17,7 +18,7 @@
     border: 1px solid #4c97ff;
     user-select: none;
     outline: none;
-    background: rgba(255, 255, 255, 0.2);
-    color: black;
+    background: rgba(255, 255, 255, 0.5);
+    color: #3373cc;
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
 }


paulkaplan:
Hmm that looks kinda strange 
image

@chrisgarrity could you take a look at this?

chrisgarrity:
The issue is that in some browsers/OS the select dropdown is styled from the css, and in others it's completely ignored. For example, the dropdown in Chrome on OSX is always black on gray even if the text is styled on white:
screen shot 2017-07-26 at 9 21 36 am

src/components/language-selector/language-selector.css
```diff
@@ -18,6 +18,6 @@
     user-select: none;
     outline: none;
     background: rgba(255, 255, 255, 0.2);
-    color: white;
+    color: black;
```
chrisgarrity: Black doesn't look great when everything else on the menu bar is white. 

Could you switch the color to #3373cc (our dark blue), increase the contrast on the background (try .5 instead of .2)

The grey icon probably won't look good next to the blue, and we really don't need it right now, so let's just remove it. 

Eventually we'll be doing a real design pass for the language selection, and this will be completely different.

Kenny2github: No problem! Will get to it now :)

PR #569, opened 2017-07-26
Change default block zoom to 0.675 from 0.75

paulkaplan:
### Proposed Changes

Change default zoom (`startScale` for blockly) to 0.675 from 0.75. This is what that looks like (new on the left, old on the right, at 1280x800 window size, non-retina screen):
image

### Reason for Changes

cc @carljbowman @thisandagain

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -237,7 +237,7 @@ Blocks.defaultOptions = {
     zoom: {
         controls: true,
         wheel: true,
-        startScale: 0.75
+        startScale: 0.675
     },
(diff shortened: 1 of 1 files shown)


thisandagain:
This feels like a step in the right direction to me. Thanks @paulkaplan

liam4:
I'm confused. The "Reason for Changes" section is literally empty..

> It only contains "*Explain why these changes should be made*"

..and the only other explanation is "This feels like a step in the right direction to me."?

Honestly, I'm curious why! That's all.

PR #578, opened 2017-08-01
Add trimming to the sound editor

paulkaplan:
### Resolves

_What Github issue does this resolve (please include link)?_

First of a few sound editor functionality chunks. This one adds the trim button and functionality.

Also fixes https://github.com/LLK/scratch-gui/issues/573

### Proposed Changes

_Describe what this Pull Request does_

Adds trim functionality to the sound editor.

### Test Coverage

_Please show how you have added tests to cover your changes_

Added a sound editor container test the tests several parts of the editor:
- Renaming sounds getting sent to the proper VM function
- Playing / stopping correctly triggering the audio buffer player
- Playing / stopping correctly modifying the state
- Trim button activation/submission setting the proper state
- Trim button submission getting sent to the proper VM function

Updated split tests between the component and the container.

PR #581, opened 2017-08-02
Sound editor effects UI 

paulkaplan:
__depends on https://github.com/LLK/scratch-gui/pull/578, review that first. Only the last commit here is new__

### Reason for Changes

This is the initial UI part of the sound effects part of the sound editor. This is includes the component interface and the component tests, and a stub of the actual implementation on the container side. I wanted to add this by itself because it involved refactoring some things to support internationalizing text.

### Test Coverage

Added new tests for effect buttons to the component, and updated the other tests to wrap with `<IntlProvider>` 

---
Note:
- The icons are completely temporary, not intended to be the finals
- This depends on https://github.com/LLK/scratch-gui/pull/578, review that first


test/unit/containers/sound-editor.test.jsx
```diff
@@ -3,7 +3,7 @@ import React from 'react'; // eslint-disable-line no-unused-vars
 import {mountWithIntl} from '../../helpers/intl-helpers';
 import configureStore from 'redux-mock-store';
 import mockAudioBufferPlayer from '../../__mocks__/audio-buffer-player.js';
-
+import {IntlProvider as Intl} from 'react-intl'; // eslint-disable-line no-unused-vars
```
rschamp: How come?

paulkaplan: Woops that was left over (and lint didn't catch it because it was force disabled!) good catch

PR #588, opened 2017-08-04
Sound editor  effects backend

paulkaplan:
### Proposed Changes

Adds the effect backend for the sound editor. Includes tests for the container doing the right thing, and some general tests on the effects.


src/lib/audio/audio-effects.js
```diff
@@ -0,0 +1,81 @@
...
+        // Some effects will modify the playback rate and/or number of samples.
+        // Need to precompute those values to create the offline audio context.
+        let sampleCount = buffer.length;
+        let playbackRate = 1;
+        switch (name) {
+        case effectTypes.ECHO:
+            sampleCount = buffer.length + 0.25 * 3 * buffer.sampleRate;
+            break;
+        case effectTypes.FASTER:
+            playbackRate = 1.5;
```
paulkaplan: The values for playback rate and volume change are not the same as the ones we playtested with. They should be updated to match.

test/__mocks__/audio-effects.js
```diff
@@ -0,0 +1,24 @@
+/* eslint-env jest */
+export default class MockAudioEffects {
+    static get effectTypes () { // @todo can this be imported from the real file?
```
rschamp: Probably? Importing and then returning `AudioEffects.effectTypes` seems like it should work?

src/lib/audio/effects/echo-effect.js
```diff
@@ -0,0 +1,33 @@
...
+
+        this.input.connect(this.effectInput);
+        this.effectInput.connect(this.delay);
+        this.delay.connect(this.compressor);
+        this.input.connect(this.compressor);
+        this.delay.connect(this.decay);
+        this.decay.connect(this.delay);
+        this.compressor.connect(this.output);
+    }
+}
```
rschamp: This stuff is pretty specialized, so should probably have some comments to explain what's happening?  The volume of this specialized code is also getting to the point that we should at least note which parts should move to a different repo. `scratch-audio`?

test/__mocks__/audio-effects.js
paulkaplan: Ack it seems that the `jest.mock` does something globally that stops this file from being allowed to import the actual AudioEffects... I would need to dig deeper into jest to figure out the right way for this.

src/lib/audio/effects/echo-effect.js
paulkaplan: Yeah, I basically figured that everything inside `lib/audio` would be extracted along with the sound editor.

PR #589, opened 2017-08-05
Add undo/redo functionality and tests to sound editor

paulkaplan:
### Proposed Changes

Adds the undo/redo functionality from the playtest version to the sound editor. Since undo/redo will probably be implemented at a different level in the future, I didn't put too much work into how it works. It uses a simple undo-stack and redo-stack approach. But I added tests to make sure everything is working correctly.

image

This is totally independent of the sound effect backend PR, and should be totally compatible with it functionally. 

### Test Coverage

Added component tests and container tests.


src/components/sound-editor/sound-editor.jsx
```diff
@@ -52,6 +54,16 @@ const messages = defineMessages({
         description: 'Title of the button to save trimmed sound',
         defaultMessage: 'Save'
     },
+    undo: {
```
fsih: Kind of tangential, should we put strings like this in a shared location, so translators don't have to translate the string for undo in several places?

src/containers/sound-editor.jsx
```diff
@@ -51,7 +59,11 @@ class SoundEditor extends React.Component {
             trimEnd: null
         });
     }
-    submitNewSamples (samples, sampleRate) {
+    submitNewSamples (samples, sampleRate, skipUndo) {
+        if (!skipUndo) {
```
fsih: Maybe it would be worthwhile to break out undo/redo to be independent of audio, so it could be tested by itself and maybe used in other places (then again if sound editor is going to be its own repo it wouldn't really be sharable)

src/containers/sound-editor.jsx
```diff
@@ -51,7 +59,11 @@ class SoundEditor extends React.Component {
...
+            this.redoStack = [];
+            this.undoStack.push(this.props.samples.slice(0));
```
fsih: Are these very large? Should we limit the size of the stacks?

src/components/sound-editor/sound-editor.jsx
paulkaplan: We may want to share certain strings. However it also can be useful to keep the separate because even strings that are the same in english might be slightly different in other languages depending on context. ¯\_(ツ)_/¯

src/containers/sound-editor.jsx
paulkaplan: They can get large yes. I'd like to save that for a performance pass, when we can look into similar questions about max lengths of recordings.

src/containers/sound-editor.jsx
paulkaplan: It's a good question, I think that because the editor may be extracted into a repo, the internal logic for undo/redo needs to be kept internal.

Issue #86, opened 2017-02-13
Confine Add from Library buttons to sprite and stage panes


lifeinchords:
![screen shot 2017-02-13 at 12 21 33 am](https://cloud.githubusercontent.com/assets/716168/22892260/e9cf3534-f1df-11e6-97dc-4e0ecdf12263.png)

lifeinchords:
Could you take care of the React related stuff @rschamp? and then I'll do CSS + positioning?

PR #597, opened 2017-08-09
Confine add buttons to stage and sprite selectors

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/86

### Proposed Changes

Restructures the layout to force the add sprite and add backdrop buttons to be within their respective panels. 

### Test Coverage

This is exactly the type of change that end-to-end testing will be able to automatically make sure doesn't break anything! until then, manually.


paulkaplan:
@rschamp what should we do with the vertical constraint?

rschamp:
Confine them to these areas? @carljbowman?

image

paulkaplan:
I'm just not sure that is an issue. I think our minimum resolution is basically ipad (1024x768), anything smaller than that I'm not sure it is worth considering

paulkaplan:
@rschamp In order to properly address that, we would need to make a min-height for the whole UI and make it possible to scroll. I'd rather deal with that part separately.

Issue #224, opened 2017-03-08
Context Menu


carljbowman:
Breaking out from meta issue: https://github.com/LLK/scratch-gui/issues/53

The context menu system is pervasive though-out the Scratch 2.0 GUI. Different options within the menu appear for different objects. 

One example of many — Context Menu on Sprite:
![screen shot 2017-03-08 at 4 43 58 pm](https://cloud.githubusercontent.com/assets/3409578/23725425/83b89f0a-041e-11e7-8050-3ec419269e17.png)

The context menu can be initiated by a right-click or shift click

carljbowman:
## 3.0 Context Menu
![sprite-tile-context-menu](https://user-images.githubusercontent.com/3409578/28621095-cf148766-71dd-11e7-9803-7b9ef7b19afc.png)

The context menu should generalized for use cases throughout the GUI. Eventually this style should be ported over to the Scratch-Blocks context menus.

------

## Here are some sample 2.0 use cases:

### Sprite _tile_ in Sprite Menu
<img width="488" alt="screen shot 2017-07-26 at 8 41 39 am" src="https://user-images.githubusercontent.com/3409578/28621459-018391e6-71df-11e7-8ff9-2fe273ffc5ab.png">

### Sprite on the Stage
<img width="488" alt="screen shot 2017-07-26 at 8 41 49 am" src="https://user-images.githubusercontent.com/3409578/28621469-0eb15aa6-71df-11e7-8263-4018cb5f9df4.png">

### Stage (we currently do not have this feature, but including for reference)
<img width="490" alt="screen shot 2017-07-26 at 8 41 56 am" src="https://user-images.githubusercontent.com/3409578/28621472-15467ce8-71df-11e7-9647-c7001ac9769c.png">

### Costume in Paint Editor
<img width="441" alt="screen shot 2017-07-26 at 8 42 03 am" src="https://user-images.githubusercontent.com/3409578/28621518-3b1b46d8-71df-11e7-9ee1-962d51d26376.png">

### Sound in Sound Editor
<img width="664" alt="screen shot 2017-07-26 at 8 42 13 am" src="https://user-images.githubusercontent.com/3409578/28621509-3363de50-71df-11e7-87e1-2a83c411c4d1.png">

liam4:
>Eventually this style should be ported over to the Scratch-Blocks context menus.

[...]

PR #605, opened 2017-08-10
Context menu

paulkaplan:
### Resolves

Part of implementing https://github.com/LLK/scratch-gui/issues/224. Specifically, the context menu with the delete menu item. Duplicate menu item will take VM work, so this is just the structure of the context menu.

### Proposed Changes

Using the `react-contextmenu` package, wrap the sprite selector in a `ContextMenuTrigger` and include a `ContextMenu` with the delete menu item. 

The `react-contextmenu` package includes no styling, so I created a custom `ContextMenu` and `MenuItem` that just inject custom classnames/styles into the components provided by `react-contextmenu`. 

### Reason for Changes

This will be the pattern we use for context menus throughout the app. 

### Test Coverage

Updated the sprite selector item tests because of the inclusion of an `intl` string, and added a test to make sure the right callbacks are attached to the context menu. 

image

image

---

@carljbowman I tried to match the style in the mocks, but let me know if you think anything should change. Also, the package I chose for context menu triggering also includes a "long-press to activate" feature, which makes it possible to bring up the menu on touch by long-touching.

### DIFF
--- a/test/unit/components/__snapshots__/sprite-selector-item.test.jsx.snap
+++ b/test/unit/components/__snapshots__/sprite-selector-item.test.jsx.snap
@@ -2,23 +2,14 @@
 
 exports[`SpriteSelectorItemComponent matches snapshot when selected 1`] = `
 <div
-  className="ponies undefined"
+  className="react-contextmenu-wrapper ponies undefined"
   onClick={[Function]}
-  style={
-    Object {
-      "alignContent": undefined,
-      "alignItems": undefined,
-      "alignSelf": undefined,
-      "flexBasis": undefined,
-      "flexDirection": undefined,
-      "flexGrow": undefined,
-      "flexShrink": undefined,
-      "flexWrap": undefined,
-      "height": undefined,
-      "justifyContent": undefined,
-      "width": undefined,
-    }
-  }
+  onContextMenu={[Function]}
+  onMouseDown={[Function]}
+  onMouseOut={[Function]}
+  onMouseUp={[Function]}
+  onTouchEnd={[Function]}
+  onTouchStart={[Function]}
 >
   <div
     className=""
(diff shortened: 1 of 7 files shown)


test/unit/components/__snapshots__/sprite-selector-item.test.jsx.snap
```diff
@@ -2,23 +2,14 @@
 
 exports[`SpriteSelectorItemComponent matches snapshot when selected 1`] = `
 <div
-  className="ponies undefined"
+  className="react-contextmenu-wrapper ponies undefined"
```
paulkaplan: Yeah, because the component is expecting a className, although it is optional.

test/unit/components/sprite-selector-item.test.jsx
```diff
@@ -1,11 +1,11 @@
 /* eslint-env jest */
 import React from 'react'; // eslint-disable-line no-unused-vars
-import {shallow} from 'enzyme';
+import {mountWithIntl, shallowWithIntl, componentWithIntl} from '../../helpers/intl-helpers';
 // eslint-disable-next-line no-unused-vars
 import SpriteSelectorItemComponent from '../../../src/components/sprite-selector-item/sprite-selector-item';
 import CostumeCanvas from '../../../src/components/costume-canvas/costume-canvas';
 import CloseButton from '../../../src/components/close-button/close-button'; // eslint-disable-line no-unused-vars
-import renderer from 'react-test-renderer';
+import {MenuItem} from '../../../src/components/context-menu/context-menu'; // eslint-disable-line no-unused-vars
```
paulkaplan: Woops, that is the danger of disabling eslint. I used in a previous iteration of the test and didn't remove it. Fixed.

Issue #579, opened 2017-08-02
Fit the layout into 1024px wide without scrolling


paulkaplan:
### Expected Behavior

_Please describe what should happen_

We want the full scratch-gui layout to fit into a screen 1024px wide (standard tablet width). 

### Actual Behavior

_Describe what actually happens_

The block editor is fixed at 600 and the stage is fixed at 480, plus 2 8px gutters = 1098px to fit the entire layout. 

@carljbowman we need to decide what to make smaller and by how much. Taking down the stage to 85% of it's size (to 408px) should allow the blocks+stage+gutters to fit. If we do that, we need to do two things:

- [ ] Make the stage component resilient to changing size on the fly, and making sure the drag events are properly scaled (non-obvious, need to pair with @cwillisf or @rschamp on this).
- [ ] Figure out how to get the sprite info panel to fit by either shrinking or hiding some of the items (name, x, y, rotation, direction, visibility, ...).  Implementing that part will be easy, but I'll need design guidance on it.

rschamp:
When we scale down the stage, we should take care to check out pixel art projects, and choose a size that scales nicely for them, since many may depend on the exact proportions to look good.

Kenny2github:
I noticed this earlier - because the GUI is a bit too wide it allows scrolling on an iPad - and since the “add sprite” and “add backdrop” buttons seem to be `position:fixed` they stay in place while the GUI scrolls, meaning that they’re out of place occasionally. I took a small screen recording, but I currently don’t have access to a pc so I can’t put it here yet - I will post it when I can.

PR #606, opened 2017-08-10
Allow GUI to fit into 1024px screen widths

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/579 by shrinking the stage and some minor changes to the sprite info area (hides icons, makes inputs slightly smaller). 

Depends on https://github.com/LLK/scratch-render/pull/154

### Proposed Changes

Still have to review these changes with @carljbowman, but the in-progress idea is to shrink the stage and to do a bit of responsive work on the sprite info area to fit everything into 1024 layout. 

Note in the code the width of the break is actually 1096, because that is where it was previously breaking. If we put the break at 1024, it would work for [1024, 1096+], but this way it works for 1024+. 

### Reason for Changes

Explained in https://github.com/LLK/scratch-gui/issues/579

### Test Coverage

I think the best way to test this might be to use the selenium tests with a browser window set to 1024 and make sure you can reach all the buttons? Not exactly sure on this though.

### DIFF
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -2,6 +2,7 @@ import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import React from 'react';
 import {Tab, Tabs, TabList, TabPanel} from 'react-tabs';
+import MediaQuery from 'react-responsive';
 import tabStyles from 'react-tabs/style/react-tabs.css';
 import VM from 'scratch-vm';
 
@@ -18,6 +19,13 @@ import MenuBar from '../menu-bar/menu-bar.jsx';
 
 import styles from './gui.css';
 
+const layout = {
+    fullStageWidth: 480,
+    fullStageHeight: 360,
+    smallerStageWidth: 480 * 0.85,
+    smallerStageHeight: 360 * 0.85,
+    fullSizeMinWidth: 1096
+};
 
 const GUIComponent = props => {
     const {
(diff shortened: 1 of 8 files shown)


src/components/gui/gui.jsx
```diff
@@ -18,6 +19,13 @@ import MenuBar from '../menu-bar/menu-bar.jsx';
 
 import styles from './gui.css';
 
+const layout = {
+    fullStageWidth: 480,
+    fullStageHeight: 360,
+    smallerStageWidth: 480 * 0.85,
+    smallerStageHeight: 360 * 0.85,
+    fullSizeMinWidth: 1096
+};
```
rschamp: From offline: Can we put these somewhere that can be imported from elsewhere? Or maybe make a wrapped `MediaQuery` component that has them embedded?

src/components/gui/gui.jsx
```diff
@@ -87,25 +95,31 @@ const GUIComponent = props => {
                         </Tabs>
                     </Box>
 
-                    <Box className={styles.stageAndTargetWrapper} >
-                        <Box className={styles.stageMenuWrapper} >
-                            <GreenFlag vm={vm} />
-                            <StopAll vm={vm} />
-                        </Box>
+                    <MediaQuery minWidth={layout.fullSizeMinWidth}>
+                        {isFullSize => (
```
rschamp: From offline: This is surprising, let's move it to the line above so it's clear it's related to `MediaQuery`.

src/components/gui/gui.jsx
```diff
@@ -87,25 +95,31 @@ const GUIComponent = props => {
...
+                                </Box>
 
-                        <Box className={styles.stageWrapper} >
-                            <Stage
-                                shrink={0}
-                                vm={vm}
-                            />
-                        </Box>
+                                <Box className={styles.stageWrapper}>
+                                    <Stage
```
rschamp: From offline, the `MediaQuery` can probably be moved to wrap just `Stage`.

src/components/sprite-info/sprite-info.css
```diff
@@ -79,3 +79,9 @@
     user-select: none;
     outline: none;
 }
+
+@media only screen and (max-width: 1096px) {
+    .non-essential {
+        display: none;
+    }
+}
```
rschamp: From offline: Let's consolidate to one method, and I think using `MediaQuery` will be the most flexible/compatible.

paulkaplan:
@rschamp thanks for the help thinking through how to do this. Made the changes you wanted in 40e3a86. And here is a gif for completeness: 

resizing-gif

PR #612, opened 2017-08-13
Fix all lint warnings except todo comments (including a bug!)

paulkaplan:
### Resolves

Fixes all the lint warnings except the `@todo` comments. Including a bug!

### Reason for Changes

We are going to get stricter about lint warnings, so wanted to fix these easy ones in advance. 

### Test Coverage

Check out the lint log!

### DIFF
--- a/src/containers/audio-trimmer.jsx
+++ b/src/containers/audio-trimmer.jsx
@@ -52,7 +52,7 @@ class AudioTrimmer extends React.Component {
     render () {
         return (
             <AudioTrimmerComponent
-                containerRef={el => (this.containerElement = el)}
+                containerRef={el => (this.containerElement = el)} // eslint-disable-line react/jsx-no-bind
                 playhead={this.props.playhead}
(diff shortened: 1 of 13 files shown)


src/containers/audio-trimmer.jsx
```diff
@@ -52,7 +52,7 @@ class AudioTrimmer extends React.Component {
     render () {
         return (
             <AudioTrimmerComponent
-                containerRef={el => (this.containerElement = el)}
+                containerRef={el => (this.containerElement = el)} // eslint-disable-line react/jsx-no-bind
```
paulkaplan: Well my thinking was basically that this is identical to the way you would do it with a regular ref (which eslint wouldn't complain about, because the rule doesn't apply for refs).

paulkaplan:
@rschamp the bug was in the for loop in the error message parsing in selenium-helpers. It had the return true inside the for loop, when it should have gone outside.

PR #618, opened 2017-08-15
Safari audio

paulkaplan:
### Resolves

Fixes audio context issues on safari (including mobile safari) encountered when using the sound editor. 

### Proposed Changes

1. The audio context must be "started" through user interaction. Use the `startaudiocontext` package (same one we use in `scratch-audio` to ensure that it is started properly.
2. Make sure to use `webkit` prefixed audio context for the offline audio context as well as the regular context. Note that the offline context (a) does not need to be "started" using `startaudiocontext` and (b) doesn't need to be shared the same way the online audio context does.
3. Changes to use the more universally supported `oncomplete` callback instead of the promise-based API for `startRendering`. 

### Reason for Changes

Sound editor isn't working in safari/mobile safari!

### Test Coverage

Existing tests!


ericrosenbaum:
Tested on iOS safari, and the sound editor works, and its effects (aside from robot) work. Cool!

PR #620, opened 2017-08-15
Update eslint packages and fix indentations

paulkaplan:
Update to eslint 4 and scratch-config 4, and update associated packages


src/containers/sound-library.jsx
```diff
@@ -40,9 +40,9 @@ class SoundLibrary extends React.PureComponent {
             };
             return this.audioEngine.decodeSound(sound);
         })
-        .then(soundId => {
-            this.player.playSound(soundId);
-        });
+            .then(soundId => {
+                this.player.playSound(soundId);
+            });
```
rschamp: This looks a bit weird without the previous `.then()` being indented similarly.

paulkaplan: Fixed in 0b71d4f

Issue #591, opened 2017-08-08
iPad touchscreen tap doesn’t count as “mouse down”


Kenny2github:
### Expected Behavior

When you tap on the stage with a touchscreen the “mouse down?” block should report true.

### Actual Behavior

Nothing when stage is tapped with touchscreen

### Steps to Reproduce

1. Drag into scripts area and run:
```scratchblocks
forever {
    point in direction (<mouse down?> * (90))
}
```
2. Tap on stage with iPad touchscreen
3. No direction change

### Operating System and Browser

iOS 11 Developer Beta, Chrome 60

rschamp:
Thanks @Kenny2github. Curious about the steps to reproduce — the `say` block is not currently implemented. Which build are these steps for?

Kenny2github:
I was using `say` as a random example - didn’t know it wasn’t implemented yet 😅 
The actual example was
```scratchblocks
forever {
    point in direction (<mouse down?> * (90))
}
```
And there was no direction change on tap.

rschamp:
Ok, no problem! Thanks for your report.

gengshenghong:
This seems like not a "mouse down" issue only, many mouse events does not work as well? Example: the "long press" and then "mouse move" does not work for stage and trimmer in sound editor on iPad. The Blocks part worked well (you can "long press" on a block and drag it to workspace). It seems like events on touch screens are not handled correctly by Scratch-GUI and Scratch-Blocks handled it quite well.

Thanks,
Shenghong

thisandagain:
We have not yet addressed issues with touch in the GUI. We will resolve this later in the year.

PR #627, opened 2017-08-16
Enable touch events for the stage and audio trimmers

paulkaplan:
Make the audio trimmers and stage work with touch events!

In addition to getting the trimmers and sprite dragging on the stage to work, this also makes all of the related mouse blocks work, so also fixes https://github.com/LLK/scratch-gui/issues/591. 

touch-events

### DIFF
--- a/src/containers/stage.jsx
+++ b/src/containers/stage.jsx
@@ -4,6 +4,7 @@ import React from 'react';
 import Renderer from 'scratch-render';
 import AudioEngine from 'scratch-audio';
 import VM from 'scratch-vm';
+import {getEventXY} from '../lib/touch-utils';
 
 import StageComponent from '../components/stage/stage.jsx';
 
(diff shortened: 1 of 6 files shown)


Kenny2github:
This seems great! I am currently on an iPad so this is good for that too.

It’s a bit concerning that Travis failed though...

paulkaplan:
@Kenny2github woops that is just a failing snapshot test, fixed in 0332606.

src/containers/stage.jsx
```diff
@@ -50,12 +51,18 @@ class Stage extends React.Component {
...
+        document.addEventListener('touchmove', this.onMouseMove);
+        document.addEventListener('touchend', this.onMouseUp);
         canvas.addEventListener('mousedown', this.onMouseDown);
+        canvas.addEventListener('touchstart', this.onMouseDown);
     }
     detachMouseEvents (canvas) {
         document.removeEventListener('mousemove', this.onMouseMove);
         document.removeEventListener('mouseup', this.onMouseUp);
+        document.removeEventListener('touchmove', this.onMouseMove);
+        document.removeEventListener('touchend', this.onMouseEnd);
```
rschamp: Looks like a typo here, should be `this.onMouseUp`?

paulkaplan: Thanks, fixed 	05f550a

Issue #640, opened 2017-08-23
Sprite selector layout is messed up on large screen sizes


paulkaplan:
### Expected Behavior

_Please describe what should happen_

The sprite selector + stage selector should fit snuggly under the stage, like this
![screen shot 2017-08-23 at 8 23 39 am](https://user-images.githubusercontent.com/654102/29615757-21885c88-87dd-11e7-9dcd-ad80f517513f.png)

### Actual Behavior

_Describe what actually happens_

The sprite selector grows! ![image](https://user-images.githubusercontent.com/654102/29615742-10fb1220-87dd-11e7-9939-994d1094276c.png)

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_

Load a project with many sprites on a large screen size.

### Operating System and Browser

_e.g. Mac OS 10.11.6 Safari 10.0_

Mac + Chrome 60

PR #641, opened 2017-08-23
Add max width to fix flex-grow errors

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/640

### Proposed Changes

Adds a max width (and helpful comment to future us) about sizing the sprite selector. 

### Reason for Changes

See screenshots in linked issue. 

### Test Coverage

No tests, but here is a screenshot before and after
screen shot 2017-08-23 at 8 33 10 am
screen shot 2017-08-23 at 8 32 42 am

### DIFF
--- a/src/components/gui/gui.css
+++ b/src/components/gui/gui.css
@@ -112,6 +112,9 @@
     */
     display: flex;
     flex-direction: column;
+
+    /* Fix the max width to max stage size (defined in layout_constants.js) + gutter size */
+    max-width: calc(480px + calc($space * 2));
 }
 
 .stage-wrapper {


src/components/sprite-selector/sprite-selector.css
```diff
@@ -2,7 +2,13 @@
...
 .sprite-selector {
     flex-grow: 1;
-    position: relative; 
+    /*
+      Because the stage doesn't grow beyond 480px, the max
+      of the sprite selector + stage selector must be 480px.
+      The stage selector is 72px + 8px gutter between.
+    */
+    max-width: 400px;
+    position: relative;
```
rschamp: The reason we need this comment is that the `max-width` rule is mixing layout logic into the component — the component (and the reader) shouldn't have to be aware of this context.

Is it not possible to take care of this in the component that controls the width of the stage? The sprite selector should fill the box it's given, and that box should be as wide as it needs to be to match the stage.

paulkaplan:
@rschamp good point, updated. I can't quite see how to get it in exactly the same place as the javascript, but it is now in the css corresponding to the layouting file.

PR #651, opened 2017-08-28
Do not render invisible components

paulkaplan:
This is a performance improvement. Reduces tick timing from ~30ms to ~18ms on my machine (in development). Also a generally good idea to not render components that aren't visible. 

Before
image

After
image

### Proposed Changes

Instead of passing `isVisible` to the library / tab components, just do not render them if they are not visible. 

The change from `Box` to `div` is because box wasn't being used for anything in particular, and it makes debugging harder.

### DIFF
--- a/src/containers/backdrop-library.jsx
+++ b/src/containers/backdrop-library.jsx
@@ -27,9 +27,9 @@ class BackdropLibrary extends React.Component {
     render () {
         return (
             <LibraryComponent
+                visible
                 data={backdropLibraryContent}
                 title="Backdrop Library"
-                visible={this.props.visible}
                 onItemSelected={this.handleItemSelect}
                 onRequestClose={this.props.onRequestClose}
             />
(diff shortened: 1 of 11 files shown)


src/containers/backdrop-library.jsx
```diff
@@ -27,9 +27,9 @@ class BackdropLibrary extends React.Component {
     render () {
         return (
             <LibraryComponent
+                visible
```
rschamp: I think we should go one step further and just remove the prop altogether. The right way to do it is conditionally rendering the component.

paulkaplan:
@rschamp should i remove it all the way down (it exists in the modal component as well, which the library renders) or just from library?

paulkaplan:
@rschamp I removed the visible property from library and modal (it was almost always just being set to true anyway) and also had to update a few places that were using the `visible` property. Those instances could always be changed to `componentWillUnmount`, which seems like a plus!

PR #655, opened 2017-08-29
Feature/extensions

cwillisf:
### Resolves

This is progress toward LLK/scratch-vm#643. It isn't useful without the VM portion of the implementation (PR coming soon).

### Proposed Changes

Add an "Add Extension" button to the GUI, which calls the VM's `loadExtensionURL` method.
Also, handle the VM's `EXTENSION_WAS_ADDED` event, registering the new blocks' JSON with `ScratchBlocks` and updating the toolbox to include the new blocks' XML.

### Reason for Changes

Extensions are cool, like bow ties.


src/containers/blocks.jsx
```diff
@@ -55,9 +60,18 @@ class Blocks extends React.Component {
...
+            this.state.prompt !== nextState.prompt ||
+            this.props.isVisible !== nextProps.isVisible ||
+            this.props.toolboxXML !== nextProps.toolboxXML
+        );
     }
     componentDidUpdate (prevProps) {
+        if (prevProps.toolboxXML !== this.props.toolboxXML) {
+            const selectedCategoryName = this.workspace.toolbox_.getSelectedItem().name_;
+            this.workspace.updateToolbox(this.props.toolboxXML);
+            this.setToolboxSelectedItemByName(selectedCategoryName);
```
paulkaplan: Can you add a comment around why this is needed? Is it a bug with scratch-blocks? If so, could you file an issue and leave a `@todo` here. Even if it isn't a bug, we should still leave a todo about modifying the scratch-blocks workspace/toolbox API so that we don't have to use properties that are marked as "private" (the trailing underscore).

src/components/gui/gui.jsx
```diff
@@ -36,6 +47,9 @@ const GUIComponent = props => {
         );
     }
 
+    const enableExtensions = window.location.search.includes('extensions');
+    const extensionButtonClasses = `${styles.extensionButton} ${enableExtensions ? '' : styles.hidden}`;
```
paulkaplan: I'm cool if you leave it this way for now, since it is obviously pretty temporary, but just an FYI we usually use the `classNames` helper for doing this. The syntax can seem funky at first, because it can take both class names and objects with {name: true/false} and usually uses the dynamic property definition thing from ES6. 

```javascript
import classNames from 'classnames';

<IconButton
    className={classNames(styles.extensionButton, {
        [styles.hidden]: window.location.search.includes('extensions')
    })}
/>
```

You can see a full example here, for instance https://github.com/LLK/scratch-gui/blob/develop/src/components/close-button/close-button.jsx#L8-L19

src/components/gui/gui.css
```diff
@@ -99,6 +99,7 @@
 .blocks-wrapper {
     flex-grow: 1;
     position: relative;
+    z-index: 0;
```
paulkaplan: What does this do?

src/containers/blocks.jsx
```diff
@@ -77,6 +91,14 @@ class Blocks extends React.Component {
         this.detachVM();
         this.workspace.dispose();
     }
+    setToolboxSelectedItemByName (name) {
+        const categories = this.workspace.toolbox_.categoryMenu_.categories_;
```
paulkaplan: Could you add a `@todo` here too for enhancing scratch-blocks to allow this kind of thing without digging around? I know something very similar was being added to scratch-blocks for the "all-in-one" toolbox, so it would be nice to indicate that this should use that when it is available

src/containers/blocks.jsx
```diff
@@ -91,6 +113,7 @@ class Blocks extends React.Component {
         this.props.vm.addListener('VISUAL_REPORT', this.onVisualReport);
         this.props.vm.addListener('workspaceUpdate', this.onWorkspaceUpdate);
         this.props.vm.addListener('targetsUpdate', this.onTargetsUpdate);
+        this.props.vm.addListener('EXTENSION_WAS_ADDED', this.onExtensionWasAdded);
```
paulkaplan: This might just be me, but the word "was" in this event and related handlers seems unneeded. Could it just be `EXTENSION_ADDED` and `onExtensionAdded`?

src/containers/blocks.jsx
```diff
@@ -264,4 +298,17 @@ Blocks.defaultProps = {
     options: Blocks.defaultOptions
 };
 
-export default Blocks;
+const mapStateToProps = state => ({
+    toolboxXML: state.toolbox.toolboxXML
+});
+
+const mapDispatchToProps = dispatch => ({
+    sendToolboxUpdate: toolboxXML => {
```
paulkaplan: I think the convention for dispatch props is to name them `onSomethingSomething`. The naming helps to keep track of where things come from. In the same vein, I think `onExtensionWasAdded` should be renamed to `handleExtensionAdded` to be consistent with our other naming (`handleSomething` vs. `onSomething`)

src/containers/blocks.jsx
```diff
@@ -167,6 +191,12 @@ class Blocks extends React.Component {
             this.workspace.resize();
         }
     }
+    onExtensionWasAdded (blocksInfo) {
+        this.ScratchBlocks.defineBlocksWithJsonArray(blocksInfo.map(blockInfo => blockInfo.json));
+        const dynamicBlocksXML = this.props.vm.runtime.getBlocksXML();
+        const toolboxXML = makeToolboxXML(dynamicBlocksXML);
+        this.props.sendToolboxUpdate(toolboxXML);
```
paulkaplan: Just a question, not a change request: @cwillisf is it possible to bind the vm directly to scratch-blocks to handle toolbox updates? Does the GUI have to be involved with that part beyond hooking the vm up with scratch-blocks, similar to the way the blocks events are connected directly with the vm? @rschamp do you think there are any consequences of doing one way vs. the other?

src/components/gui/gui.css
```diff
@@ -148,3 +149,17 @@
     height: $stage-menu-height;
     padding: $space;
 }
+
+.extensionButton {
```
paulkaplan: If you make that change I mentioned below about the `classNames` helper, it'll also automatically convert camel case to skewer case used in css, so this would become `.extension-button`

src/components/gui/gui.jsx
```diff
@@ -45,6 +59,11 @@ const GUIComponent = props => {
         tabSelected: classNames(tabStyles.reactTabsTabSelected, styles.isSelected)
     };
 
+    const onAddExtensionClick = () => {
```
paulkaplan: In general we would define this function in the container and pass it in, rather than define it in the component. Obviously it is very temporary so no need to fix it, but just FYI.

rschamp: I think I would lean to moving it in the container now for hygiene — so the container/component split remains clear.

src/containers/blocks.jsx
rschamp: I think we should keep scratch-blocks specific logic to a minimum in the VM, so that the VM isn't concerned with UI.  I also think it's a good design to not bind the blocks to the VM given how much we mutate scratch-blocks currently.

src/components/gui/gui.css
cwillisf: My button was ending up behind the flyout, so it was faded out and unclickable. This probably isn't the right solution... thoughts?

rschamp: I think this makes sense for that. It might be clearer if you put `z-index: 1` here and `z-index: 2` on the button though (plus a comment).

src/lib/make-toolbox-xml.js
```diff
@@ -0,0 +1,639 @@
+const separator = '<sep gap="45"/>';
+
+const motion = [
+    '<category name="Motion" colour="#4C97FF" secondaryColour="#3373CC">',
+    '    <block type="motion_movesteps">',
```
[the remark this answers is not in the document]
cwillisf: I tried a few variations, looking for something that both ESlint and my editor's syntax highlighting would accept. My editor didn't like what Eric had before so I kind of dismissed the use of ``` `` ```... but what you have works and I like it better. Thanks!

src/containers/blocks.jsx
cwillisf: TBH, I just copied this from @ericrosenbaum's work. It's possible that it's no longer necessary. I mean, the `updateToolbox` is necessary but I can try without the category switching logic.

src/components/gui/gui.css
cwillisf: I was wondering about that... thanks!

src/components/gui/gui.css
paulkaplan: I'm not sure how nested z-indices work, but the vertical category menu will put a 100% height div at z-index of 40 or so, it might make more sense to just z-index the button above that?

paulkaplan:
@cwillisf  those changes look good! As i mentioned offline, I think this will probably fail on travis until the VM stuff is merged.

PR #687, opened 2017-09-18
Fix eslint error from prop types

paulkaplan:
### Resolves

Fixes the broken build on develop that came from a minor version update in eslint. 

### Proposed Changes

Be explicit with react proptypes, even if it involves copy/pasting :( 

### Reason for Changes

Because the build was failing.

### Test Coverage

They pass now

---

Maybe worth noting that because the eslint version is not pinned to a specific version, develop was passing when the last merge was done, but something caused subsequent builds to use an in-range update to eslint, which then failed (probably because another PR updated the node_module cache?)

### DIFF
--- a/src/containers/recording-step.jsx
+++ b/src/containers/recording-step.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import bindAll from 'lodash.bindall';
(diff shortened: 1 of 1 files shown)


paulkaplan:
@rschamp lg'd offline. Merging

PR #691, opened 2017-09-19
Fix tests on Windows

cwillisf:
### Resolves

Jest-based integration tests currently do not run on Windows hosts.

Note that the integration tests don't run from Ubuntu-on-Windows without significant extra configuration. This change makes the tests work from a CMD prompt and maybe Cygwin.

### Proposed Changes

This PR includes a few different changes which, combined, allow the tests to run on Windows. LLK/scratch-storage#22 should be treated as a prerequisite.

### Reason for Changes

See the description of each commit below for a more thorough explanation. The short version:
- Using `[\\\\/]` in the  Jest test path arguments seems to be the only way to get the path to behave the same on both Windows and non-Windows platforms.
- Without `import 'chromedriver';`, Jest / Selenium can't find `chromedriver.exe` on Windows. This might also help other platforms...?
- Building as part of the integration test script makes iterating on the integration tests take a long time. However, it's important to build before running the integration tests. I've moved `npm run build` out of the integration test script and into the overall test script. In addition, I've changed the names of the test scripts as suggested by @rschamp.

/cc @fsih


thisandagain:
Related to:
https://github.com/LLK/scratch-gui/pull/726
https://github.com/LLK/scratch-gui/issues/727

PR #700, opened 2017-09-22
updating string ids to follow convention

### LINKED ISSUE #669: Update localization string ids to follow convention.
We decided that the naming convention for localization strings should be `repo.component.name`. scratch-paint is using the convention `paint.component.name`, but gui is currently inconsistent. 

Gui should use `gui.component.name`

chrisgarrity:
### Resolves
#669 

### Proposed Changes

revised string ids for localization according to our new convention. See specific comments/questions below.

### DIFF
--- a/scripts/generate-locale-messages.js
+++ b/scripts/generate-locale-messages.js
@@ -43,7 +43,7 @@ const fs = require('fs');
 const path = require('path');
 const mkdirp = require('mkdirp');
 
-const locales = ['en', 'es', 'fr'];
+const locales = ['en', 'es', 'de', 'ar', 'he'];
 const LANG_DIR = './translations/';
 const MSGS_DIR = './locale/';
 
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -26,7 +26,7 @@ const addExtensionMessage = (
     <FormattedMessage
         defaultMessage="Extensions"
         description="Button to add an extension in the target pane"
-        id="gui.targetPane.addExtension"
+        id="gui.gui.addExtension"
     />
 );
 
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -26,7 +26,7 @@ const addExtensionMessage = (
     <FormattedMessage
         defaultMessage="Extensions"
         description="Button to add an extension in the target pane"
-        id="targetPane.addExtension"
+        id="gui.targetPane.addExtension"
     />
 );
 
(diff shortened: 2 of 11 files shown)


src/components/gui/gui.jsx
```diff
@@ -26,7 +26,7 @@ const addExtensionMessage = (
     <FormattedMessage
         defaultMessage="Extensions"
         description="Button to add an extension in the target pane"
-        id="targetPane.addExtension"
+        id="gui.targetPane.addExtension"
```
chrisgarrity: The convention says `repo.component.name`. Does that make sense when the component is gui? Should this be `gui.gui.addExtension`?

src/components/sprite-selector/sprite-selector.jsx
```diff
@@ -14,7 +14,7 @@ const addSpriteMessage = (
     <FormattedMessage
         defaultMessage="Add Sprite"
         description="Button to add a sprite in the target pane"
-        id="targetPane.addSprite"
+        id="giu.SpriteSelectorComponent.targetPaneAddSprite"
```
chrisgarrity: is `targetPane` useful? Should this just be `gui.SpriteSelectorComponent.addSprite`

src/components/turbo-mode/turbo-mode.jsx
```diff
@@ -15,7 +15,7 @@ const TurboMode = () => (
             <FormattedMessage
                 defaultMessage="Turbo Mode"
                 description="Label indicating turbo mode is active"
-                id="controls.turboMode"
+                id="gui.turboMode"
```
chrisgarrity: What about when the string and the component name are the same? Should this be `gui.turboMode.turboMode` for consistency, or is that just overkill? Especially when there's only one string in the component.

scripts/generate-locale-messages.js
```diff
@@ -43,7 +43,7 @@ const fs = require('fs');
 const path = require('path');
 const mkdirp = require('mkdirp');
 
-const locales = ['en', 'es', 'fr'];
+const locales = ['en', 'es', 'de', 'ar', 'he'];
```
chrisgarrity: For the purposes of review, ignore the changes having to do with the list of locales, it's changed so we can test RTL and generally longer translations. And this will all go away once it gui imports scratch-l10n.

src/components/sprite-selector/sprite-selector.jsx
chrisgarrity: note gui typo

chrisgarrity: gui.spriteSelector.addSprite

paulkaplan:
@chrisgarrity Changes look good, but just a question about that new file, whether that was intended or note

chrisgarrity:
@paulkaplan good catch. I do need to add the transifex config at some point, but it should probably be in it's own PR (or at least one that is related to syncing new strings) - I'll remove that file and then merge

Issue #549, opened 2017-07-18
Thumbnails for all-white sprites should be visible


ericrosenbaum:
### Expected Behavior

It should be possible to see the shape of a sprite's costume in the thumbnail somehow, even if it is all white.

In Scratch 2.0, thumbnail images for sprites that have all-white costumes are given gray outlines so you can see their shape. Also, the selected state of the thumbnail turns the background the gray:

<img width="160" alt="screen shot 2017-07-18 at 10 15 44 am" src="https://user-images.githubusercontent.com/567844/28324528-a07db3bc-6ba9-11e7-951f-e1b7ff831220.png">
<img width="161" alt="screen shot 2017-07-18 at 10 15 25 am" src="https://user-images.githubusercontent.com/567844/28324536-a537a3ae-6ba9-11e7-924e-018886d9d011.png">

### Actual Behavior

Sprite thumbnails have white backgrounds:

<img width="155" alt="screen shot 2017-07-18 at 10 15 36 am" src="https://user-images.githubusercontent.com/567844/28324685-f8f09712-6ba9-11e7-9027-5bc147f739b9.png">

### Steps to reproduce

Visit https://llk.github.io/scratch-gui/#163165382

paulkaplan:
Here is the place where this is done in scratch-flash: https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/ui/SpriteThumbnail.as#L187-L193

paulkaplan:
Fixed!
![image](https://user-images.githubusercontent.com/654102/50114921-e3250880-0213-11e9-9b73-52ec2f8a8439.png)

PR #720, opened 2017-09-30
Fixing bug #549

javaone4kids:
Hi,

We are here in a workshop called javaone4kids and we just fixed a bug of `scratch-gui` .

### Resolves

https://github.com/LLK/scratch-gui/issues/549

### Proposed Changes

We added a css property for the sprite-selector-item when it is selected.

### DIFF
--- a/src/components/sprite-selector-item/sprite-selector-item.css
+++ b/src/components/sprite-selector-item/sprite-selector-item.css
@@ -22,6 +22,7 @@
 }
 
 .sprite-selector-item.is-selected {
+    background: #dbdbdb;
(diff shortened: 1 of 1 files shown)


thisandagain:
Thanks for the PR @javaone4kids ! Unfortunately this doesn't resolve the issue in a satisfactory way. Changing the background color to `#dbdbdb` creates a new issue with any costume that is that color.

javaone4kids:
Hi @thisandagain  ! Thanks for the quick review. This is indded an use case we didn't think of withi the kids (but we had only a 2 hours workshop).

PR #774, opened 2017-10-13
Update to react 16!

paulkaplan:
Updates all the required packages for react 16!

Fixes the tests that broke when updating to enzyme 3. They had to do with updating component props.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ cache:
   directories:
   - node_modules
 install:
-- npm --production=false install
-- npm --production=false update
+- npm install
+- npm update
 before_deploy:
 - npm --no-git-tag-version version 0.1.0-prerelease.$(date +%Y%m%d%H%M%S)
 - git config --global user.email $(git log --pretty=format:"%ae" -n1)
@@ -36,4 +36,3 @@ deploy:
   skip_cleanup: true
   email: $NPM_EMAIL
   api_key: $NPM_TOKEN
-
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
     "i18n:msgs": "node ./scripts/generate-locale-messages.js",
     "i18n:src": "babel src > tmp.js && rimraf tmp.js && ./scripts/build-i18n-source.js ./translations/messages/ ./translations/",
     "start": "npm run i18n:msgs && webpack-dev-server",
-    "test": "npm run test:lint && npm run build && npm run test:unit && npm run test:integration",
+    "test": "npm run test:lint && npm run test:unit && NODE_ENV=production npm run build && npm run test:integration",
```
rschamp: We should check if this works on Windows.

.travis.yml
```diff
@@ -11,9 +11,6 @@ node_js:
 cache:
   directories:
   - node_modules
-env:
-  global:
-  - NODE_ENV=production
 install:
 - npm --production=false install
 - npm --production=false update
```
rschamp: We can simplify this and remove `--production=false` if we're not setting `NODE_ENV=production` globally.

package.json
paulkaplan: The same change in scratch-paint worked for @fsih on windows

.travis.yml
paulkaplan: Fixed in 8ded00f

PR #788, opened 2017-10-17
importing l10n that includes paint strings

chrisgarrity:
Uses new version of scratch-l10n that supports localization of gui and paint. Part of https://github.com/LLK/scratch-paint/issues/16. 

Goes along with https://github.com/LLK/scratch-l10n/pull/6
This will be ready as soon as there's a  published release of l10n version 2.x


chrisgarrity:
Build is expected to fail until there is a 2.x version of scratch-l10n

Issue #799, opened 2017-10-20
Clicking on blocks doesn't run them


rschamp:
This may need to move to either scratch-blocks or scratch-vm, but logging here until we figure out the source of the issue.

### Expected Behavior

_Please describe what should happen_
When I click on a block, either in the palette or on the workspace, it should run. When I click on a reporter, it should report its value

### Actual Behavior

_Describe what actually happens_
When clicking on a block, the cursor appears to momentarily drag the block, and it doesn't run. Only when clicking stacks with hats, do the blocks in the stack run.

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_
1. Visit llk.github.io/scratch-gui/develop
2. Click <kbd>move 10 steps</kbd>
3. Observe that the cat does not move 10 steps, it just sits there, smugly

### Operating System and Browser

Chrome 61 and Safari 11, Mac OS 10.12 (probably all browsers)

towerofnix:
Note also that the block doesn't need to be in the palette; it won't work even if it's in the script area. Clicking on the block doesn't make it glow, like expected, either.

Also, of this script..

![when this sprite clicked / move 10 steps / repeat 10 { start sound meow / wait 0.1 secs }](https://user-images.githubusercontent.com/9948030/31826841-83cfaf4a-b58c-11e7-97f3-3c5181e22f21.png)

..clicking on any block, *except* the hat block, ran the script (and the whole script glowed, like expected). Clicking on the hat block did nothing.

paulkaplan:
It seems to be related to some of the recent changes to how scratch-blocks updates the toolbox / deals with workspace updates. We disable events on workspace refresh (because we clear / repopulate the workspace, we don't want to hear about delete/create events during that). However workspace refresh was recently changed to also cause a toolbox refresh, and the VM was relying on those events. That change will soon be reverted from scratch-blocks, but there will definitely be more cleanup involved.

PR #803, opened 2017-10-23
Revert broken performance improvement and add tests

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/799, which was caused by a scratch blocks change the [has been reverted](https://github.com/LLK/scratch-blocks/pull/1168). The revert commit is to take out the corresponding "performance improvement" that went along with the now-reverted scratch-blocks PR. 

In order to make sure these problems do not happen again, I've added tests for the following things which were broken:

- Clicking on blocks to run them from the toolbox, seeing reported values.
- (same but specifically for variables, something which was also fixed recently)
- Switching sprites updating the block menus. This would have been broken if we had reverted the scratch-blocks PR but not reverted this GUI commit. 

It's all a bit complicated :( 

But the good news is we now have tests that will stop it from happening in the future. 

The bad news is we had to take a step back from the performance improvements, but we now can approach them with a much better idea of the pitfalls.

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -82,10 +82,12 @@ class Blocks extends React.Component {
             return;
         }
         // @todo hack to resize blockly manually in case resize happened while hidden
+        // @todo hack to reload the workspace due to gui bug #413
         if (this.props.isVisible) { // Scripts tab
             this.workspace.setVisible(true);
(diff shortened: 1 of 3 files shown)


paulkaplan:
@rschamp LG'd offline.

Issue #768, opened 2017-10-13
Monitor visibility gets out of sync on switching sprites


ericrosenbaum:
### Expected Behavior

Monitor visibility set by the checkbox should stay in sync when switching sprites.

### Actual Behavior

Checkbox state and monitor visibility can get out of sync.

Seems related to https://github.com/LLK/scratch-gui/issues/749

### Steps to Reproduce

- Go to https://llk.github.io/scratch-gui/develop/
- Set the checkbox for the "x position" monitor to show it
- Switch to stage
- Switch back to Sprite1
- Monitor is still showing, but box is unchecked

griffpatch:
(and if you click the monitor again, a duplicate monitor is generated...
nice that they both work though *grin*).

This effect is also true for variables monitors, except that local variable
monitors only show the correct value when you are viewing the sprite the
variable belongs to (if you switch to another sprite while the local
variable is visible, then the monitor changes value to show '0' - see
project: https://llk.github.io/scratch-gui/develop/#26861174 and make 'aaa'
visible, then switch sprites back and forward to see the value change)

griffpatch:
Actually the "x position" is monitor also only shows the active sprite's 'x
position'. Maybe this is a know issue, I remember there being mention of
something similar before - sorry if I'm repeating things.

PR #811, opened 2017-10-25
Give fixed IDs to all monitored blocks

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/768, making all monitors appear in sync and correctly reflect the monitored blocks in the VM. 

Which incidentally means that the currently un-implemented "sprite specific" monitors do not work and show that they don't work, meaning when you switch sprites the checkbox stays checked. 

This is an improvement, because it accurately reflects the monitored block container in the VM.

It does this by assigning fixed IDs to all blocks in the toolbox that are monitored. We previously fixed the variable blocks so that they have deterministic IDs as well. 

Lacking other ideas, i just used the back half of the opcode as the fixed ID. Doesn't much matter since it is never referred to manually.

### DIFF
--- a/src/lib/make-toolbox-xml.js
+++ b/src/lib/make-toolbox-xml.js
@@ -371,7 +371,7 @@ const sound = `
                 </shadow>
             </value>
         </block>
-        <block type="sound_volume"/>
+        <block id="volume" type="sound_volume"/>
         <block type="sound_changetempoby">
             <value name="TEMPO">
                 <shadow type="math_number">
--- a/src/lib/make-toolbox-xml.js
+++ b/src/lib/make-toolbox-xml.js
@@ -186,9 +186,9 @@ const motion = `
         </block>
         <block type="motion_ifonedgebounce"/>
         <block type="motion_setrotationstyle"/>
-        <block type="motion_xposition"/>
-        <block type="motion_yposition"/>
-        <block type="motion_direction"/>
+        <block id="xposition" type="motion_xposition"/>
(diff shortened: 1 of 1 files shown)


src/lib/make-toolbox-xml.js
```diff
@@ -485,18 +485,18 @@ const sensing = `
         <block type="sensing_mousedown"/>
         <block type="sensing_mousex"/>
         <block type="sensing_mousey"/>
-        <block type="sensing_loudness"/>
-        <block type="sensing_timer"/>
+        <block id="loudness" type="sensing_loudness"/>
+        <block id="timer" type="sensing_timer"/>
         <block type="sensing_resettimer"/>
-        <block type="sensing_of">
+        <block id="of" type="sensing_of">
```
fsih: Is this one supposed to be monitorable?

src/lib/make-toolbox-xml.js
```diff
@@ -386,7 +386,7 @@ const sound = `
                 </shadow>
             </value>
         </block>
```
fsih: Should 'volume' be monitored?

src/lib/make-toolbox-xml.js
paulkaplan: Seems to be, it is a change from scratch 2

src/lib/make-toolbox-xml.js
paulkaplan: Yes! Thanks

Issue #813, opened 2017-10-27
Remove Top Blocks Category


carljbowman:
"Top" blocks was used extremely infrequently during workshops. Based on this observations and discussions in yesterday's product meeting, the top category should be removed. 

![desktop - 1024 x 768](https://user-images.githubusercontent.com/3409578/32109071-9e9bd48c-bb01-11e7-80d2-f90e8d51d752.png)

Kenny2github:
ah, well, it was a good idea while it lasted.

PR #820, opened 2017-10-30
Remove “top” category

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/813

### Proposed Changes

Removes the "top" category of blocks containing commonly used blocks at the top of the category menu.

### Reason for Changes

We tried it out with kids and found that it was used very infrequently.


DeleteThisAcount:
No! :( I'm sad

Issue #821, opened 2017-10-30
Prevent loading two copies of the same extension


ericrosenbaum:
### Expected Behavior

The user should be prevented from loading a duplicate of a particular extension.

Note that this issue is itself a duplicate (lol) of an issue in the VM (https://github.com/LLK/scratch-vm/issues/694). We have decided to implement two different solutions, one in the VM (tracked by the other issue, to come later), and one here. For the solution here in the gui, we should keep a list of loaded extension URLs. Clicking on an extension in the library that matches the list should close the modal and scroll to the extension category. 
  
### Actual Behavior

Loading a second copy of an extension causes duplicate blocks and other problems.

### Steps to Reproduce

Load an extension using the extension library
Load the same extension again
Note the duplicate blocks

PR #834, opened 2017-11-01
Prevent loading duplicate extensions

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/821

Depends on https://github.com/LLK/scratch-vm/pull/724

### Proposed Changes

- Prevent loading duplicate extensions by checking with the VM extension manager before loading
- If the extension has already been loaded, scroll the blocks menu to that extension category
- Render the extension library in the blocks container, so that it can scroll the blocks.

### Reason for Changes

This prevents the loading of duplicate extensions, at least for internal extensions.

### Test Coverage

[test coming]


paulkaplan:
@ericrosenbaum I added a commit to this branch with a test that makes sure clicking the same extension twice scrolls back down to the original extension. I'm satisfied (test-wise) with this integration test.

PR #837, opened 2017-11-02
Allow loading the default project while offline

### LINKED ISSUE #617: The GUI + default project should be able to run offline
### Expected Behavior

While offline, starting the GUI should work.

### Actual Behavior

The GUI attempts to load the assets for the default project JSON, but requires a network connection for these.

### Steps to Reproduce

Open the GUI without a network connection

rschamp:
Requires https://github.com/LLK/scratch-storage/pull/25. The tests will fail until that's published and the `scratch-storage` version is bumped here. I should note that this is a hack until we have `.sb3` files.

### Resolves

Resolves #617

### Proposed Changes

Cache the default project and its assets. Also begin loading projects using storage.

### Reason for Changes

The GUI should not require a network connection to run locally. This brings in the assets required to load the default project. This does not bring in all the assets from the libraries, but the editor should at least be able to load once without a network connection.

### Test Coverage

I tried to keep the existing tests equivalent and functional, but wasn't able to get enzyme/jest to wait for the promise in `componentDidUpdate` to resolve. So for now I'm just testing that the `storage.load` method is called with the appropriate arguments, and that the state is what I would expect.


test/unit/util/project-loader-hoc.test.jsx
```diff
@@ -1,51 +1,57 @@
...
 describe('ProjectLoaderHOC', () => {
     test('when there is no project data, it renders null', () => {
         const Component = ({projectData}) => <div>{projectData}</div>;
         const WrappedComponent = ProjectLoaderHOC(Component);
         window.location.hash = '#winning';
-        ProjectLoader.load = jest.fn((id, cb) => cb(null, null));
+        storage.load = jest.fn(() => Promise.resolve(null));
         const mounted = mount(<WrappedComponent />);
-        ProjectLoader.load.mockRestore();
+        storage.load.mockRestore();
```
paulkaplan: I'm not sure if `mockRestore` works this way. The documentation says:
> Beware that `mockFn.mockRestore` only works when mock was created with `jest.spyOn`. Thus you have to take care of restoration yourself when manually assigning `jest.fn()`.
https://facebook.github.io/jest/docs/en/mock-function-api.html#mockfnmockrestore

paulkaplan:
For posterity, do you mind giving a quick explanation about all the webpack loader changes? Particularly because several pieces of our infrastructure use very similar patterns?

src/lib/default-project/index.js
```diff
@@ -0,0 +1,49 @@
+import {TextEncoder} from 'text-encoding';
+import projectJson from './project.json';
+
+/* eslint-disable import/no-unresolved */
+import popWav from '!buffer-loader!./83a9787d4cb6f3b7632b4ddfebf74367.wav';
+import meowWav from '!buffer-loader!./83c36d806dc92327b9e7049a565c6bff.wav';
+import backdrop from '!buffer-loader!./739b5e2a2435f6e1ec2993791b423146.png';
+import penLayer from '!buffer-loader!./5c81a336fab8be57adc039a8a2b33ca9.png';
```
rschamp: These use buffer-loader so that the contents of these files are returned as a buffer, the same format as what storage returns.

src/lib/default-project/index.js
```diff
@@ -0,0 +1,49 @@
...
+import costume1 from '!raw-loader!./09dc888b0b7df19f70d81588ae73420e.svg';
+import costume2 from '!raw-loader!./3696356a03a8d938318876a593572843.svg';
```
rschamp: These use `raw-loader` for the same reason as above. Buffer-loader didn't seem to do what the storage engine expects, so I had to get the contents as a string and encode it for the storage engine.

test/unit/util/project-loader-hoc.test.jsx
rschamp: Thanks!

webpack.config.js
```diff
@@ -66,8 +66,12 @@ module.exports = {
             }]
         },
         {
+            test: /\.(png|wav)$/,
+            loader: 'raw-loader'
+        },
+        {
             test: /\.svg$/,
-            loader: 'svg-url-loader?noquotes'
```
rschamp: This change is just to get the svg loader in line with what's standard. Using svg-url-loader in the first place was a mistake, since we want a loader that can handle more than just svgs.

paulkaplan:
Can you rebase on develop to get the storage changes?

rschamp:
@paulkaplan Done, and now tests are passing!

Issue #836, opened 2017-11-02
<[] contains []> got removed?


Kenny2github:
I recall thisandagain opening an issue for the inclusion of a `<[] contains []>` Operators block - #601 
However, I see that it is no longer in the blocks palette...? When was it removed, and was it supposed to be?

paulkaplan:
Woops, looks like it was lost when we moved the toolbox xml from blocks to GUI. Thanks for spotting this. Hard to catch that kind of thing in code reviews because there is no previous diff. I'll fix it quick.

PR #838, opened 2017-11-02
Bring back the 'contains' operator block

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/836

### Proposed Changes

Brings back the new `contains` block which looks like it was lost in the shuffle between blocks xml and gui xml. 

I also noticed that the default values were wrong when this was originally implemented (`hello contains world`, while the spec says `hello world contains world`). I updated it to match the spec. @thisandagain this block is pretty wide, which makes it somewhat harder to notice that it is a boolean. Is the below screenshot ok?

image

h/t to @Kenny2github for noticing the regression.

### DIFF
--- a/src/lib/make-toolbox-xml.js
+++ b/src/lib/make-toolbox-xml.js
@@ -603,6 +603,18 @@ const operators = `
                 </shadow>
             </value>
         </block>
+        <block type="operator_contains" id="operator_contains">
+          <value name="STRING1">
+            <shadow type="text">
+              <field name="TEXT">hello</field>
+            </shadow>
+          </value>
+          <value name="STRING2">
+            <shadow type="text">
+              <field name="TEXT">world</field>
+            </shadow>
+          </value>
+        </block>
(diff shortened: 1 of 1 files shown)


paulkaplan:
@thisandagain also noticed this was also originally implemented with a trailing question mark, which doesn't match the spec either. That fix would go in scratch-blocks. Should I fix that too?

Kenny2github:
@paulkaplan when I implemented the block, @thisandagain originally said the default values should be "hello world" contains "world"; however, midway through, he noticed that it was too wide, so he asked me to change it to "hello" contains "world", which I did. The original issue didn't say anything about the question mark (I think?), but right now the lack of a question mark is the only thing that differentiates it from the Lists block `<[list v] contains []>`, and Operators blocks don't have question marks anyway, so I'd say don't include a question mark.

thisandagain:
Thanks @paulkaplan. Also good catch on the missing question mark.

PR #841, opened 2017-11-02
Load extension library data from a js file

ericrosenbaum:
### Proposed Changes

- Load extension library images from files. In order to do that:
- Switch to using webpack file-loader for loading svg and png files as data URIs
- Change library content file from json to a js file

### Reason for Changes

Previously we were loading extension library images from scratch assets using md5s. This strategy allows us to store the images in a repo (GUI for now).


towerofnix:
>This strategy allows us to store the images in a repo (GUI for now).

This makes me wonder if there'll be a dedicated repository for Scratch extensions in the future..!

PR #845, opened 2017-11-02
Show focus on language selector.

quachtina96:
### Resolves
There is no visual indicator of focus on the language selector when tabbing through the GUI. 

### Proposed Changes
Add white ring around the selector when it has focus (e.g. you have tabbed to the language selector).
Before:
image
After:
image

### Reason for Changes
Upon discussion with Carl, the highlight is consistent with other ways in which the focus on elements is shown.

### DIFF
--- a/src/components/language-selector/language-selector.css
+++ b/src/components/language-selector/language-selector.css
@@ -24,6 +24,5 @@
 }
 
 .language-select:focus {
-    border-color: white;
-    box-shadow: 0 0 0 -2px white;
+    border: 1px solid white;
 }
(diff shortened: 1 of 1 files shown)


src/components/language-selector/language-selector.css
```diff
@@ -22,3 +22,8 @@
     color: #3373cc;
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
 }
+
+.language-select:focus {
+    border-color: white;
+    box-shadow: 0 0 0 -2px white;
```
paulkaplan: Did you mean to have this negative 2 to actually be positive 2? A negative value wouldn't make much sense as a spread radius on a box shadow, it would be the same as having no box shadow at all?

Kenny2github: I think if all that is needed is a border, there wouldn't need to be box-shadow at all, right?

PR #847, opened 2017-11-02
Enable selection of sprite visibility by keyboard.

quachtina96:
### Resolves

- Cannot tab to individual show/hide sprite buttons (they both have same tab index)
image

- Cannot use keypresses (press space or enter) to activate show/hide sprite buttons
- Show and hide buttons do not visually show focus

### Proposed Changes
- You can tab to each button individually 
- User can press space or enter to activate either of the buttons when the button has focus on it
- Buttons have a border when they currently have focus

Focused on active Show Sprite Button:
image

Focused on inactive Show Sprite Button:
image

Focused on active Hide Sprite Button:
image

Focused on inactive Hide Sprite Button:
image

### Reason for Changes

Someone with limited motor skills may not be able to use the mouse to click on the show and hide sprite buttons. Keyboard controls may be more usable. Adding visual indicators to show that focus is on the buttons provides valuable feedback to users of keyboard controls.

### DIFF
--- a/src/components/sprite-info/sprite-info.css
+++ b/src/components/sprite-info/sprite-info.css
@@ -62,12 +62,22 @@
     border-bottom-left-radius: $form-radius;
 }
 
+.radio-left:focus {
+    border-color: #4c97ff;
+    box-shadow: inset 0 0 0 -2px rgba(0, 0, 0, 0.1);
+}
+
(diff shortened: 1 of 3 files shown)


VutonDesign:
Nice! 👍

PR #849, opened 2017-11-02
Hide icons from screenreader

quachtina96:
### Resolves
When using a screenreader on the Scratch GUI, images are encountered but are not meaningful (they are referred to as "/svg%3E") nor necessary.

### Proposed Changes
Have the screen reader skip over these images upon encountering them by using the aria-hidden attribute on these images.

### Reason for Changes
These images are not actionable nor necessary for the user to understand and use the Scratch GUI

### Test Coverage
I tested VoiceOver on Mac, checking that it skips over these images.

### DIFF
--- a/src/components/close-button/close-button.jsx
+++ b/src/components/close-button/close-button.jsx
@@ -20,6 +20,7 @@ const CloseButton = props => (
         <img
             className={styles.closeIcon}
             src={closeIcon}
+            aria-hidden="true"
         />
     </div>
 );
--- a/src/components/close-button/close-button.jsx
+++ b/src/components/close-button/close-button.jsx
@@ -7,6 +7,7 @@ import closeIcon from './icon--close.svg';
 
 const CloseButton = props => (
     <div
+        aria-label="Close"
         className={classNames(
             styles.closeButton,
             props.className,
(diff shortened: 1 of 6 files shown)


paulkaplan:
@quachtina96 thanks for these! 

I wonder if the problem is really that the "buttons" that these images are contained within are do not have the "role" attribute marking them as buttons? Would it be treated differently if the containing divs were correctly identified as role=button? I know @mewtaylor has been making some similar changes in scratch-paint.

The reason I ask is that trying to learn a bit about aria I encountered a few (possibly out-of-date recommendations about using ["extreme caution"](https://www.w3.org/TR/wai-aria-1.1/#aria-hidden) with `aria-hidden`. 

I'd love to have a consistent way of approaching icon buttons, since it is such a common feature of the scratch interface, and is used throughout our repos.

paulkaplan:
Oh and those tests are failing because they are checking against an old "snapshot" of the html. You can update those snapshots by running `npm run test:unit -- --updateSnapshot` (n.b. the two sets of dashes are intentional).

quachtina96:
@paulkaplan I see what you're saying! I agree that setting the aria role of the parent divs is a better solution so I've made the change. I've added an aria-label for the close button; however, it seems like the close button is used not just in the context of closing (dialogs) but also for deleting (sprites, sounds, backdrops). I'm not sure "close" is the best way to label the button given its multiple usages.

PR #853, opened 2017-11-02
Add alternate text for Scratch logo.

quachtina96:
### Resolves
When a screen reader is focused on the Scratch logo, the user hears "/svg%3E" since the logo is unlabeled.

### Proposed Changes
User of screen reader on Scratch GUI hears "Scratch logo image" voiced when navigation is on the Scratch logo image.

### Reason for Changes
Provides meaningful text describing the image for those who can't see.

### Test Coverage
Tested manually using VoiceOver on Mac

### DIFF
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -20,6 +20,7 @@ const MenuBar = function MenuBar () {
                 <img
                     className={styles.scratchLogo}
                     src={scratchLogo}
+                    alt="Scratch logo"
                 />
             </div>
             <SaveButton className={styles.menuItem} />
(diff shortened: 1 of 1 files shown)


src/components/menu-bar/menu-bar.jsx
```diff
@@ -18,6 +18,7 @@ const MenuBar = function MenuBar () {
         >
             <div className={classNames(styles.logoWrapper, styles.menuItem)}>
                 <img
+                    alt="Scratch logo"
```
thisandagain: Should this be "Scratch logo" or simply "Scratch"?

fsih: These tips might be helpful?
https://material.io/guidelines/usability/accessibility.html#accessibility-writing
It's not totally clear, but 'Label what an element does, not what it is' might lean toward 'Scratch' instead of 'Scratch logo'

quachtina96: Thanks for pointing this out! I changed to "Scratch" since "logo" isn't necessary

PR #854, opened 2017-11-03
Fix project json loading through storage

paulkaplan:
### Resolves

Hotfix bug loading projects by ID introduced by https://github.com/LLK/scratch-gui/pull/837

### Proposed Changes

Data comes through storage now, meaning it isn't parsed as JSON when it arrives, it is just an array buffer that needs to be stringified. 

### Test Coverage

Previously we were catching errors here and just logging them. Doing this breaks the error checking in the integration tests.


rschamp:
Thank you!

PR #856, opened 2017-11-03
Add compatibility testing example page

paulkaplan:
### Resolves

Add a simple example page `/compatibility-testing.html` that loads a project (from hash) in both scratch 3 player and an embedded iframe with the scratch 2 player. 

One note when using this locally is that I had to go into my chrome settings and whitelist `http://localhost:8601` to allow the flash player. It doesn't pop up an "allow" thing from the iframe for some reason. Will probably have to do the same thing with the deployed github url. 

side-by-side

### DIFF
--- a/src/examples/compatibility-testing.jsx
+++ b/src/examples/compatibility-testing.jsx
@@ -1,0 +1,74 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import {connect} from 'react-redux';
+
+import AppStateHOC from '../lib/app-state-hoc.jsx';
+import Controls from '../containers/controls.jsx';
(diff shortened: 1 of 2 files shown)


paulkaplan:
Future improvements might be things like splitting keyboard input between the two, adding UI to load projects more easily, having a list of projects that you can click through.

rschamp:
Right now when I load compatibility-testing.html, the default project displayed in the 3.0 editor is the default project with the Scratch cat, but it looks like you wanted it to default to project id 10015059?  Is that expected?

paulkaplan:
@rschamp is there a way to show the default project in the embed?

paulkaplan:
@rschamp made a change to sync the two.

PR #895, opened 2017-11-14
Append sprite name to monitor label for sprite-specific monitors.

fsih:
Working towards https://github.com/LLK/scratch-vm/issues/564
Depends on https://github.com/LLK/scratch-vm/pull/784

perspritemonitors

### DIFF
--- a/src/lib/monitor-adapter.js
+++ b/src/lib/monitor-adapter.js
@@ -11,6 +11,8 @@ const isUndefined = a => typeof a === 'undefined';
  * - Add missing XY position data if needed
  * @param {object} block - The monitor block
  * @param {string} block.id - The id of the monitor block
+ * @param {string} block.spriteName - Present only if the monitor applies only to the sprite
+ *     with given target ID. The name of the target sprite when the monitor was created
  * @param {number} block.index - The index of the monitor
  * @param {string} block.opcode - The opcode of the monitor
  * @param {object} block.params - Extra params to the monitor block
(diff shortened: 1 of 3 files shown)


src/lib/monitor-adapter.js
```diff
@@ -19,12 +21,15 @@ const isUndefined = a => typeof a === 'undefined';
  * @param {number} y - The monitor y position
  * @return {object} The adapted monitor with label and category
  */
-export default function ({id, index, opcode, params, value, x, y}) {
+export default function ({id, spriteName, index, opcode, params, value, x, y}) {
```
paulkaplan: Can we separate the monitor sprite name stuff into a separate PR? It depends on some VM changes, while the other parts of this do not.

src/lib/make-toolbox-xml.js
```diff
@@ -636,7 +637,9 @@ const makeToolboxXML = function (categoriesXML) {
     }
 
     everything.push(xmlClose);
-    return everything.join('\n');
+    return everything.join('\n').split('<TARGET_ID>')
+        .join(targetId); // targetIds are designed to not break XML.
+    // @todo consider something less hacky?
```
paulkaplan: Talked about this offline: instead of the categories being constants, make them functions that take in the target id and use `${...}` string interpolation instead.

src/lib/monitor-adapter.js
fsih: Broken out at https://github.com/LLK/scratch-gui/pull/899

PR #899, opened 2017-11-16
Refresh workspace with sprite-specific blocks when sprite changes

fsih:
Working towards LLK/scratch-vm#564

Give per-sprite monitor blocks unique IDs for each target. This way monitor blocks for different targets will be able to exist concurrently.

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -184,6 +184,8 @@ class Blocks extends React.Component {
         // Remove and reattach the workspace listener (but allow flyout events)
         this.workspace.removeChangeListener(this.props.vm.blockListener);
         const dom = this.ScratchBlocks.Xml.textToDom(data.xml);
+        // @todo This line rerenders toolbox, and the change in the toolbox XML also rerenders the toolbox.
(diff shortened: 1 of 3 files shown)


src/containers/blocks.jsx
```diff
@@ -172,6 +172,11 @@ class Blocks extends React.Component {
         this.workspace.reportValue(data.id, data.value);
     }
     onWorkspaceUpdate (data) {
+        // When we change sprites, update the toolbox to have the new sprite's blocks
+        if (this.props.vm.editingTarget) {
+            this.props.updateToolboxState(makeToolboxXML(this.props.vm.editingTarget.id));
+        }
```
paulkaplan: I'm ok with doing this until we figure out a better solution, but this is going to cause `updateToolbox` to be called twice: once as a result of this, and once as a result of the workspace update below. We should be able to keep that down to once, although it might involve some blocks change.

fsih: I've added a todo and filed issue https://github.com/LLK/scratch-gui/issues/901

PR #911, opened 2017-11-17
Warn user before navigating away/unloading

rschamp:
### Resolves

When you're working on your project, sometimes you accidentally swipe back or refresh when you don't mean to and you lose all your work!

### Proposed Changes

When the GUI is built for production, the page asks you (in modern browsers) if you really want to unload the page.  This isn't active while developing, only when the GUI is built with `NODE_ENV=production`.

### Reason for Changes

This way you'll be able to confirm before navigating away.

### Test Coverage

All the tests still work now. Should I update to add a test for this alert?

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -7,6 +7,11 @@ import ProjectLoaderHOC from './lib/project-loader-hoc.jsx';
 
 import styles from './index.css';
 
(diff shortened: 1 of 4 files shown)


thisandagain:
Resolved merge conflicts.

PR #931, opened 2017-11-20
Target specific palette

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/682

### Proposed Changes

Adds a `isStage` flag to the temporary "make toolbox XML" functionality that the GUI has. Show different blocks depending on whether or not a target is a stage. This level of specificity I think is acceptable for now, since this functionality will move out to the more general filtering of the VM with extensions. 

### Test Coverage

Added an integration test to make sure the label "Stage selected: no motion blocks" is visible when you have the stage selected.

### DIFF
--- a/src/lib/make-toolbox-xml.js
+++ b/src/lib/make-toolbox-xml.js
@@ -659,7 +659,7 @@ const xmlClose = '</xml>';
 
 /**
  * @param {!boolean} isStage - Whether the toolbox is for a stage-type target.
- * @param {!string} targetId - The current editing target
+ * @param {?string} targetId - The current editing target
  * @param {string?} categoriesXML - null for default toolbox, or an XML string with <category> elements.
  * @returns {string} - a ScratchBlocks-style XML document for the contents of the toolbox.
  */
--- a/src/lib/make-toolbox-xml.js
+++ b/src/lib/make-toolbox-xml.js
@@ -252,7 +252,7 @@ const looks = function (isStage, targetId) {
             <block id="backdropname" type="looks_backdropname"/>
         ` : `
             <block id="${targetId}_costumeorder" type="looks_costumeorder"/>
-            <block id="backdroporder" type="looks_backdroporder"/>
+            <block id="backdropname" type="looks_backdropname"/>
             <block id="${targetId}_size" type="looks_size"/>
         `}
         ${categorySeparator}
--- a/src/lib/make-toolbox-xml.js
+++ b/src/lib/make-toolbox-xml.js
@@ -2,9 +2,12 @@ const categorySeparator = '<sep gap="36"/>';
 
 const blockSeparator = '<sep gap="36"/>'; // At default scale, about 28px
 
-const motion = function (targetId) {
+const motion = function (targetId, isStage) {
     return `
     <category name="Motion" colour="#4C97FF" secondaryColour="#3373CC">
+        ${isStage ? `
+        <label text="Stage selected: no motion blocks"></label>
+        ` : `
         <block type="motion_movesteps">
             <value name="STEPS">
                 <shadow type="math_number">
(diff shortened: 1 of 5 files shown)


src/lib/make-toolbox-xml.js
```diff
@@ -208,34 +221,40 @@ const looks = function (targetId) {
...
+                </value>
+            </block>
+            ${blockSeparator}
+        `}
+        ${isStage ? `
+            <block id="backdroporder" type="looks_backdroporder"/>
+            <block id="backdropname" type="looks_backdropname"/>
+        ` : `
+            <block id="${targetId}_costumeorder" type="looks_costumeorder"/>
+            <block id="backdroporder" type="looks_backdroporder"/>
```
fsih: Should this be backdrop name instead of order?

src/reducers/toolbox.js
```diff
@@ -2,8 +2,7 @@ const UPDATE_TOOLBOX = 'scratch-gui/toolbox/UPDATE_TOOLBOX';
 import makeToolboxXML from '../lib/make-toolbox-xml';
 
 const initialState = {
-    // @todo switch this to make the stage's XML
-    toolboxXML: makeToolboxXML('')
+    toolboxXML: makeToolboxXML('', true)
```
fsih: What if we switch the argument order to isStage first, and make targetId optional? makeToolboxXML can assert that if isStage is false, there must be a targetId, but when the toolbox is for the stage, targetId should never be used

src/lib/make-toolbox-xml.js
```diff
@@ -354,45 +373,55 @@ const control = function () {
...
-        <block type="control_create_clone_of">
-            <value name="CLONE_OPTION">
-                <shadow type="control_create_clone_of_menu"/>
-            </value>
-        </block>
-        <block type="control_delete_this_clone"/>
+        ${isStage ? `
+            <block type="control_create_clone_of">
+                <value name="CLONE_OPTION">
+                    <shadow type="control_create_clone_of_menu"/>
```
fsih: Maybe for a different PR but this still has the 'myself' option

src/lib/make-toolbox-xml.js
paulkaplan: Yes, sorry

src/lib/make-toolbox-xml.js
paulkaplan: Hm, well fixed it since it is kinda related: 8aa1d6e

src/lib/make-toolbox-xml.js
paulkaplan: Fixed in c68cb09

src/reducers/toolbox.js
paulkaplan: Sounds good, fixed in 7b9e22c

src/lib/make-toolbox-xml.js
```diff
@@ -658,24 +658,24 @@ const xmlOpen = '<xml style="display: none">';
 const xmlClose = '</xml>';
 
 /**
- * @param {!string} targetId - The current editing target
  * @param {!boolean} isStage - Whether the toolbox is for a stage-type target.
+ * @param {!string} targetId - The current editing target
```
fsih: ?string

fsih: How do I un-out-date

src/lib/make-toolbox-xml.js
```diff
@@ -629,23 +658,24 @@ const xmlOpen = '<xml style="display: none">';
 const xmlClose = '</xml>';
 
 /**
+ * @param {!boolean} isStage - Whether the toolbox is for a stage-type target.
  * @param {!string} targetId - The current editing target
```
fsih: ?string

paulkaplan: Fixed 334d714

PR #950, opened 2017-11-27
Revert breaking change version update

chrisgarrity:
react-intl-redux is introducing breaking changes in minor version updates since they’re pre version 1.0

For now I just want to revert the change. Later I'l figure out what we need to do to handle the change (or maybe decide we don't want to depend on this package)

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "react-dom": "16.1.1",
     "react-draggable": "3.0.3",
     "react-intl": "2.4.0",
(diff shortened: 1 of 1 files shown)


paulkaplan:
I'm fine with this if you think it is the best way to go. Since it is pre-release, should we try to follow their updates to the API or is it a bunch of work to do so?

chrisgarrity:
I'll probably follow their updates, but it'd be nice to do it when it makes sense for us rather than reacting to the GUI being broken.

Issue #913, opened 2017-11-17
Remove all bitmap sprites, costumes and backdrops


ericrosenbaum:
Because our current paint editor does not support bitmaps, we should temporarily remove all bitmap assets: sprites, costumes and backdrops.

ericrosenbaum:
Also we'll need to remove any vector assets that include bitmaps inside of them (e.g. the backdrop "neon tunnel")

towerofnix:
I wrote a script to remove all non-SVGs and SVGs which contain `<image>` elements from the library files. The output is available in [this gist](https://gist.github.com/towerofnix/7ec4fbcb687a6e35472f8ed298a07ba3/revisions). I've published the script [here](https://github.com/towerofnix/scratch-library-filter); output files are also available in that repository for comparison (or, view the [revisions](https://gist.github.com/towerofnix/7ec4fbcb687a6e35472f8ed298a07ba3/revisions) tab of the gist).

fsih:
That's awesome @towerofnix, thanks!

thisandagain:
Amazing. Thanks @towerofnix !

PR #951, opened 2017-11-27
Remove all bitmap assets from libraries

thisandagain:
### Resolves

Resolves GH-913

### Proposed Changes

- Removes all bitmap assets from the sprite, costume, and backdrop libraries. This uses the new `libgen` utility which extracts all assets and builds libraries from a list of Scratch project IDs. In the case of this PR, the project that is being pulled from is `188453716`.

### Reason for Changes

- We would like to remove all bitmap assets until the Bitmap Paint Editor is complete.


thisandagain:
@ericrosenbaum Yeah. The "legacy" vector sprites all need to be cleaned up. I'm thinking about ideas for how to do this programmatically as manually editing 200+ costumes seems like not much fun.

paulkaplan:
@ericrosenbaum @thisandagain I think the backdrops being off by 100/100 is a new problem that was introduced in this PR? Was that the expected behavior?

Master
image

This
image

thisandagain:
Hmm! Not sure why that would have happened but I'll take a look. I'm guessing we might not be calculating the backdrop rotation center properly in `libgen`.

chexbox:
It looks like re-implementing bitmap sprites may be feasible, soon.

Issue #555, opened 2017-07-21
Changing languages resets the project state


paulkaplan:
### Expected Behavior

_Please describe what should happen_

Changing languages should leave the project state alone. For example, the blocks shouldn't disappear, the position of the sprites shouldn't change and newly added sprites shouldn't go away.

### Actual Behavior

_Describe what actually happens_

The project returns to its original default project state.

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_

Add a few blocks, reposition the sprite, etc. Then change languages and see it all reset.

paulkaplan:
Fixed

PR #968, opened 2017-11-29
Localize extension blocks

chrisgarrity:
Initial version of extension localization.

- pen translations are imported by default, we may want to consider whether the translations should be imported dynamically when the extension is added.
- blocks.jsx container is calling setLocale on the VM when the locale changes. Would this actually be better in the intl reducer?
- handleBlocksInfoUpdate is currently a duplicate of adding an extension, but it gets lots of warnings about overwriting the definition of blocks from blockly.js. Do we want to change the handler to remove the blocks before redefining them, or just remove the warning in scratch-blocks? [Update, left a todo comment to deal with scratch-blocks warnings, and simply call handleExtensionAdded]

The setLocale in componentDidUpdate never actually gets called because blocks always get remounted when the locale changes (see #555 )

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -59,6 +59,7 @@ class Blocks extends React.Component {
     componentDidMount () {
         this.ScratchBlocks.FieldColourSlider.activateEyedropper_ = this.props.onActivateColorPicker;
         this.ScratchBlocks.Procedures.externalProcedureDefCallback = this.props.onActivateCustomProcedures;
+        this.props.vm.setLocale(this.props.locale, this.props.messages);
 
         const workspaceConfig = defaultsDeep({},
             Blocks.defaultOptions,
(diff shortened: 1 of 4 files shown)


chrisgarrity:
By the way, I think the build is failing because this depends on the VM changes see https://github.com/LLK/scratch-vm/pull/777

chrisgarrity:
Currently blocked on the VM PR (in progress), and now it has merge conflicts. I'll re-open this when unblocked.

chrisgarrity:
@paulkaplan it's ready for review now - build is passing with the new VM :)

src/containers/blocks.jsx
```diff
@@ -72,17 +74,23 @@ class Blocks extends React.Component {
         addFunctionListener(this.workspace, 'zoom', this.onWorkspaceMetricsChange);
 
         this.attachVM();
+        this.props.vm.setLocale(this.props.locale, this.props.messages);
```
paulkaplan: This is a duplicate of line 63? I think you only want one of them?

chrisgarrity: @paulkaplan earlier line is gone now :)

paulkaplan:
LG

Issue #458, opened 2017-06-15
Monitors overlap each other on stage


jwzimmer:
### Expected Behavior

The monitors should never (automatically) appear in the same place.

### Actual Behavior

You can get one monitor to appear over another.

<img width="1280" alt="screen shot 2017-06-15 at 11 35 07 am" src="https://user-images.githubusercontent.com/1320456/27189437-593eba9a-51bf-11e7-9e36-f65fba837b1d.png">
<img width="1277" alt="screen shot 2017-06-15 at 11 34 51 am" src="https://user-images.githubusercontent.com/1320456/27189436-593ab4c2-51bf-11e7-84b6-3fab316354a4.png">

### Steps to Reproduce

- Create two variables with different-length names
- Uncheck the first one
- Recheck the first one

### Operating System and Browser

_Mac, Chrome 59_

Issue #472, opened 2017-06-26
Variable Monitor Cannot Be Moved


BGMead:
### Expected Behavior

_When you move a variable around, then hide it, then show it again, it should stay in the same place._

### Actual Behavior

_Instead, it goes somewhere completely random._

### Steps to Reproduce

_??? Not an advanced web developer, sorry._

### Operating System and Browser

_My browser / operating system: Windows 8, Chrome 58.0.3029.110, Flash 26.0 (release 0)_

paulkaplan:
Hi there! Thanks for the bug report. For the "Steps to Reproduce" could you just go through, step-by-step, the actions you took. For example you might say

1. Create a new variable
2. <what do you mean by "hide it"? There is no block for hiding variables yet>
3. ...

BGMead:
1. Create a new variable
2. Move it around
3. Hide the variable by toggling the checkmark box next to the variable name
4. Show it by clicking the checkmark box again
5. Boom. It moved.

paulkaplan:
Aha! Thanks for explaining.

Issue #463, opened 2017-06-19
Monitor list extends outside of stage


jwzimmer:
### Expected Behavior

The monitors get automatically placed under each other on the left side of the stage. When they run out of space on the left side of the stage, they should start a new column, still within the stage area.

### Actual Behavior

They can extend off the bottom of the stage into the sprite pane area.

<img width="1094" alt="screen shot 2017-06-19 at 11 03 09 am" src="https://user-images.githubusercontent.com/1320456/27291879-8d818648-54df-11e7-945c-218db7c1d9ac.png">

### Steps to Reproduce

This was on https://ericrosenbaum.github.io/scratch-gui/. I added the speech & wedo 2.0 extensions, then went back and forth between them checking the reporters. This added the same monitor to the stage multiple times (e.g. speech) which is in itself a bug (the checked/ un-checked state of a reporter that appears in the block pane for multiple extensions doesn't carry over across extensions, so you can add the variable multiple times), but @fsih mentioned that the monitors are placed without any reference to whether they are in the stage or not, so this sounds like it's a problem in the main gui prototype as well.

### Operating System and Browser

_Mac OS El Capitan, Chrome 59_

paulkaplan:
So the good news is that they no longer overflow the stage, because they now get hidden under the stage. However, the other bug with creating many repeats of of the same monitor is gone, so it is much harder to get this problem :)

fsih:
The Scratch 2.0 behavior seems to be to line them up nicely against the left edge at first, and after that fills up follow some seemingly non-random but indiscernible pattern
![image](https://user-images.githubusercontent.com/2855464/32577863-44070c24-c4a9-11e7-9148-7f4fea847c9e.png)

fsih:
(Current monitor positioning code is in scratch-gui/src/lib/monitor-adapter.js and simply uses the index)

PR #976, opened 2017-11-30
Monitor layout

fsih:
### Resolves
https://github.com/LLK/scratch-gui/issues/458 Monitors overlap each other on stage
https://github.com/LLK/scratch-gui/issues/472 Variable Monitor Cannot Be Moved
https://github.com/LLK/scratch-gui/issues/463 Monitor list extends outside of stage

### Proposed Changes

Algorithm for layout out monitors
Requirements:
- Monitors line up on the left edge of the screen first
- Monitors try to appear in a place where they won't overlap already showing monitors
- If a monitor is moved, then turned off and on again, it stays in the place it was put

Beyond that, I didn't try to match very closely the behavior of 2.0, which can be pretty strange. 
I tried to document some of it in https://github.com/LLK/scratch-gui/issues/463. You're pretty unlikely to want your entire project covered in monitors anyway.

Here is how I chose to implement this:
1. Monitors that were moved by the user always appear wherever they were last seen.
2. New monitors appear in the first leftmost, then topmost place they can fit without overlapping another monitor
    - If you make a variable, then hide its monitor, then make a new variable, the monitor for the new variable will appear in the very top left instead of below the space left open for the first monitor. This is different from how Scratch 2.0 does it and might take some iterating.
3. When the first column fills up, monitors start to line up one column to the right (avoiding rows where they would overlap the first column). A column is somewhat arbitrarily 50 pixels. (I tried having them go to the leftmost available spot, which was the best packing, but made them appear in a really weird order when you got to the second column.)
4. If the monitor doesn't fit anywhere, put it in the leftmost topmost place where it fits except that it goes off the screen to the left or bottom. Also, only do this if it's still at least 40 pixels on-screen. This is so that if you name your variable `asdfasdfasdfadsfafdafdsadfasdfasfsadfsafsafsdfsafdasfdsafdsafaasdfsafadsfaasfdasfdasfdsadfasdfsafdsadfasdfasfasdf`, it still tries to line it up somewhere.
5. If the entire screen is covered in monitors, place the monitor randomly on the left half of the screen.
image
(4 and 5 are overlapping because the monitor was moved on top of them after they spawned. Notice how 11 avoided the moved monitor)

### Reason for Changes

Fix bugs

### DIFF
--- a/src/components/monitor-list/monitor-list.jsx
+++ b/src/components/monitor-list/monitor-list.jsx
@@ -11,10 +11,9 @@ const MonitorList = props => (
     <Box
         className={styles.monitorList}
     >
-        {props.monitors.valueSeq().map((monitorData, index) => (
+        {props.monitors.valueSeq().map(monitorData => (
             <Monitor
                 id={monitorData.id}
-                index={index}
                 key={monitorData.id}
                 opcode={monitorData.opcode}
                 params={monitorData.params}
--- a/src/components/monitor/monitor.jsx
+++ b/src/components/monitor/monitor.jsx
@@ -16,13 +16,12 @@ const MonitorComponent = props => (
     <Draggable
         bounds="parent"
         defaultClassNameDragging={styles.dragging}
-        defaultPosition={{
-            x: props.x,
-            y: props.y
-        }}
         onStop={props.onDragEnd}
     >
-        <Box className={styles.monitor}>
+        <Box
+            className={styles.monitor}
+            componentRef={props.componentRef}
+        >
             <Box className={styles.label}>
                 {props.label}
             </Box>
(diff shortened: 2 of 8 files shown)


fsih:
poke

paulkaplan:
@fsih I'm concerned about how much this accesses the dom directly to write and read style properties, instead of using the APIs provided by react/react-draggable. There can be considerable performance penalties for doing things like reading and writing css properties from the dom during a drag. Want to just talk through that part when you are in? Specifically parts like this https://github.com/LLK/scratch-gui/pull/976/files#diff-352d73d98a71b1989c638208d9888b84R57

fsih:
I've moved the contents of setElement into componentWillUnmount and componentDidMount like we discussed. I also fixed passing in x, y as props from the VM to position new monitors. It goes through the same method of positioning as all other monitors now, so I removed default position from the draggable.

paulkaplan:
Great! Looks like there are some conflicts yet to resolve also?

fsih:
Merged!

PR #983, opened 2017-12-01
Try using sentry to make bug finding more effective

paulkaplan:
This is a small experiment to use Sentry to record uncaught errors from using scratch-gui. I don't expect that we will keep this in for a long time, but I've been trying it and it gives a lot of useful context.


paulkaplan:
Here is an example of the kind of context sentry gives us. Here is what happens if you import a JSON project from pre-typed variables time:

image

PR #986, opened 2017-12-01
Make libraries screen reader accessible.

quachtina96:
### Resolves
When a user opens a modal, the screen reader is moved out of the web content. When navigating into the  GUI, it reads "In Scratch 3.0 GUI, web content web content is empty". Modal and GUI becomes completely unusable.

### Proposed Changes

- Rather than hide the entire Scratch GUI from the screen reader when the modal is open, hide everything but the modal.  
- Include library items and close button in the tab order
- assign library items the button aria role
- Add an onBlur and onFocus listener to the library items so that one may preview the sound before selecting it, as is done with mouseEnter and mouseLeave.

### Reason for Changes
Someone who is blind or low-vision can now use the screen reader to identify various assets and add them to their projects! It also enables them to navigate out of the modal and back to the rest of the GUI.

### Test Coverage
Tested with VoiceOver screen reader on MacOS
1. Turn on VoiceOver on Mac (command + F5)
For each library... 

Item Selection
1. Navigate to 'Add _____' button to open the library
2. [control + option + rightArrow] or [tab] to move to next item
3. [control + option + rightArrow] or [tab] to move to previous item
4. [control + option + space] to select an item (and you are brought back to the GUI)

Library Exit
1. Navigate to 'Add _____' button to open the library
2. navigate to and select the close button (by tabbing or by moving right/left

Library Search
1. Navigate to 'Add _____' button to open the library
2. type to search
3. tab past close button to hear search results

### DIFF
--- a/src/components/close-button/close-button.jsx
+++ b/src/components/close-button/close-button.jsx
@@ -17,6 +17,7 @@ const CloseButton = props => (
             }
         )}
         role="button"
+        tabIndex="0"
         onClick={props.onClick}
     >
         <img
--- a/src/components/library-item/library-item.jsx
+++ b/src/components/library-item/library-item.jsx
@@ -46,7 +46,7 @@ class LibraryItem extends React.PureComponent {
                 </div>
             </div>
         ) : (
-            <div
+            <Box
                 className={styles.libraryItem}
                 role="button"
                 tabIndex="0"
@@ -64,7 +64,7 @@ class LibraryItem extends React.PureComponent {
                     </Box>
                 </Box>
                 <span className={styles.libraryItemName}>{this.props.name}</span>
-            </div>
+            </Box>
         );
     }
 }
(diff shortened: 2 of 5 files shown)


paulkaplan:
@quachtina96 testing this out, i was surprised that i could tab into the grid, but hitting enter doesn't submit the new sprite. Was that supposed to work? 

Love the way it triggers the mouse over/out when tabbing, allows the costumes to cycle and sounds to play!

paulkaplan:
Just found the answer to my question 

```
For native HTML <button> elements, the button's onclick event will fire both for mouse clicks and when pressing Space or Enter, while the button has focus. But if another tag is used  to create a custom button, the onclick event will only fire when clicked by the mouse cursor, even if role="button" is used. Because of this, the developer will have to add a separate key event handler to the element so that the button can be triggered when the Space or Enter key is pressed.
```

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role

I think given this, when we are using the `role="button"`, we should also add a keypress handler for "space" and "enter". I think it would be best to include that in this PR, since we are adding tab navigation here, I do not want to leave the submission part hanging. 

@quachtina96 I know you are probably quite busy, let me know if you want me to add that behavior, I'm more than willing to help.

PR #1068, opened 2017-12-19
Lock scratch-* dependencies to their current versions

paulkaplan:
Lock all the scratch-* dependencies to their current versions, allow greenkeeper to make PRs for the new versions and keep everything up-to-date. This will allow us to make sure that updates to other repos have the integration tests run before the GUI publishes an update. Also makes it easy to manage master since it will be strictly behind develop.


package.json
```diff
@@ -83,13 +83,13 @@
...
-    "scratch-audio": "latest",
-    "scratch-blocks": "latest",
-    "scratch-l10n": "latest",
-    "scratch-paint": "latest",
-    "scratch-render": "latest",
-    "scratch-storage": "^0.3.0",
-    "scratch-vm": "latest",
+    "scratch-audio": "0.1.0-prerelease.1511362862",
+    "scratch-blocks": "0.1.0-prerelease.1513170598",
+    "scratch-l10n": "2.0.20171211145754",
```
rschamp: Since eventually this version will get bumped periodically, every time there are new translations, we might want this to be patch or minor-pinned. Let's leave it for now though.

paulkaplan: Yeah we'll have to decide what exactly the versioning scheme is for language updates. are all language updates patch changes? Are new languages major/minor/patch? I assume major versions would still only be for things like package structure changes? Semver may or may not apply quite right to those ideas.

Kenny2github:
@paulkaplan Another case of the only commit coming after the review ;)

PR #1069, opened 2017-12-19
Partially Implement GH-830: Add coming soon tooltips

### LINKED ISSUE #830: Scratch 3.0 Preview "Coming Soon..." Tooltip
## Overview
Not all 2.0 features will be present in the 3.0 Preview. We will need a component (possibly a tooltip type mechanism) to message that these features will be added in the future, but are still in progress.

## UI
In all cases we will need to have a inactive appearance to the feature. In some cases we may need to add an inactive element to convey that the feature will exists (e.g. "Convert to Bitmap"). See sketches (below) for inactive state reference. 
Icons & Assets: [tooltips-assets.zip](https://github.com/LLK/scratch-gui/files/1519314/tooltips-assets.zip)

## Langauge
The hope was to have a collection of general phrases that would be randomly assigned to a tooltip upon each interaction. The phrases should have a fun tone and communicate the state of the editor is "Work in Progress". 
#### Current phrases (we may add more)
- Don't worry, we're on it (scratch cat emoji)
- Coming Soon...
- We're working on it (scratch cat emoji)

## Initial List of Features
- [x] Save
- [x] Load
- [x] Language Selector 
- [x] Large Stage / Small Stage / Full Screen Toggles (currently always Large Stage)
- [ ] Text Tool in Vector Editor
- [ ] Convert to Bitmap (Bitmap Paint Editor in general)
- [x] Variable More Options (e.g. Local scoping)

### Save
save

### Load
load

### Language Selector
language selector

### Stage Sizing Toggles
stage size

### Vector Editor Text Tool
text tool

### Convert to Bitmap
convert to bitmap

### Variable More Options (e.g. Local scoping)  
local and cloud vars

mewtaylor:
Lastly, this does a small refactor on the full screen implementation to include multiple small screen sizes too.

### Resolves

This partially resolves #830, implementing the component that can be used in react for "coming soon", and also applying it to:
* the small stage size
* save, load and language

### Proposed Changes

* Adds a new dependency, `react-tooltip` for the "Coming Soon" hover tips
* Adds a new component to generate "Coming Soon" messages
* There are a few slight changes to the full screen mode handling in order to accommodate both large/small screen states when not in full screen mode (and changes to the icon used /cc @carljbowman )

### Reason for Changes

Needed as part of #830.

@paulkaplan and @carljbowman – I still have to figure out how to do these tooltips for blocks. What I still need to do that is:
* The list of blocks that will be still "coming soon"
* A mock up, if it's any different than what we already have here
* Some guidance on how to do this in `scratch-blocks` 😸 

/cc @chrisgarrity since this introduces new translations.


paulkaplan:
@mewtaylor I think we should probably disable the "save" option, right now you can still click it and download json.

mewtaylor:
@paulkaplan ah good call sorry about that – I just pushed a change to disable it, but also to allow our `Button` components to be disabled in general (since they're `span`'s, they can't just received the `disabled` property and work). Let me know what you think!

src/containers/stage-header.jsx
```diff
@@ -22,19 +22,20 @@ class StageHeader extends React.Component {
...
+    isFullScreen: state.stageSize.isFullScreen
 });
 
 const mapDispatchToProps = dispatch => ({
-    onZoom: () => dispatch(setZoomed(true)),
-    onUnzoom: () => dispatch(setZoomed(false))
+    onSetStageLarge: () => dispatch(setStageSize(STAGE_SIZES.large)),
+    onSetStageSmall: () => dispatch(setStageSize(STAGE_SIZES.small)),
+    onSetStageFull: () => dispatch(setFullScreen(true)),
+    onSetStageUnFull: () => dispatch(setFullScreen(false))
```
paulkaplan: 🥛 🍷 🥃 🥂  half full? 😸

src/components/coming-soon/coming-soon.css
```diff
@@ -0,0 +1,62 @@
+@import "../../css/colors.css";
+
+.coming-soon,
+.coming-soon:before,
+.coming-soon:after {
+    opacity: 1 !important;
+}
+
+.coming-soon {
+    background-color: $data-primary !important;
```
paulkaplan: I've always wondered if the exclamation point is meant to mean "seriously!", in a code context it kind of looks like "NOT important". 

But actually, what is with all the importants? Could we solve it with some different scoping? Maybe not, just figured I'd ask because this file made my CSS senses ~~tingle~~ light on fire.

src/components/menu-bar/menu-bar.css
```diff
@@ -56,11 +57,11 @@
     line-height: $menu-bar-height; 
     cursor: pointer;
     text-decoration: none;
-    color: white;
+    color: $motion-tertiary;
     user-select: none;
 }
 
-.menu-item:hover {
+/*.menu-item:hover {
```
paulkaplan: Did we want to remove this completely instead of commenting out?

src/components/coming-soon/coming-soon.css
mewtaylor: @paulkaplan I also find the `important`s not ideal. it’s unfortunately the way the third party library is [set up](https://github.com/wwayne/react-tooltip#options), requiring them to override its default styling.

src/components/menu-bar/menu-bar.css
mewtaylor: @paulkaplan I'm fine with that – I thought I'd leave it in for when we re-enable, but you're right, we can always just re-implement rather than uncomment!

vutondesign:
@mewtaylor I'm guessing the left navbar text was meant to be changed to blue?

capture

Issue #973, opened 2017-11-30
Sound Editor Design Pass


paulkaplan:
Tracker issue for the design updates for the sound editor coming this iteration. Full design spec is in invision, but here is the overview: 
![image](https://user-images.githubusercontent.com/654102/33436508-3e383c6c-d5b3-11e7-8ecb-4d426fc41216.png)

PR #1084, opened 2017-12-20
Iteration 2 of sound editor design

paulkaplan:
- New icons
- Play button moved to prominent location
- Visually normalizes styling with costume editor

@mewtaylor I didn't bring in the new elements/styles directly from scratch-paint yet. I think that might as well wait until we extract them into components, if you are cool with that. 

### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/973

### Proposed Changes

image


thisandagain:
@paulkaplan Looks like this is failing on `Sound Editor Component › matches snapshot`

fsih:
Something we've noticed in the paint editor is that kids sometimes can't find the undo/redo buttons. They bring their mouse over the buttons at the top expecting it to be there, apparently don't find it, and end up using the eraser. I worry that if we make them smaller and in the center where they're less obvious we'll have the same issue in sound.

Issue #982, opened 2017-12-01
Scratch 3.0 Preview Browser Not Supported Modal


carljbowman:
## Overview
Some browsers will not be supported in the Scratch 3.0 Preview. This modal will communicate that this browser is not supported and point to the Preview FAQ for more information.

## Which Browsers
All Internet Explorer versions

## Mock
This follows the same pattern of [Scratch 3.0 Preview - Intro Modal](https://github.com/LLK/scratch-gui/issues/828)
![unsupported browser modal](https://user-images.githubusercontent.com/3409578/34525144-f33b239a-f06b-11e7-9290-8f14cb7e148a.png)

#### Illustration
![unsupported](https://user-images.githubusercontent.com/3409578/34525182-12cfc9d6-f06c-11e7-997e-d539200da062.png)

#### Copy
Header: `Internet Explorer is not Supported`
Copy: `We're very sorry, but Scratch 3.0 does not support Internet Explorer. We recommend trying a newer browser such as Google Chrome, Mozilla Firefox, or Microsoft Edge.`

#### Button
`Back` - Should call a browser Back command

#### Secondary Copy
Footer: `To learn more, go to the Preview FAQ.`
- FAQ URL`scratch.mit.edu/preview-faq`

carljbowman:
@chrisgarrity - There is still a lot to be worked out on this page. Gotta resolve the content and figure out which browsers will cause this alert. 

If it's okay with you, I am co-assigning you. My feeling is the lack of content should not block development.

carljbowman:
@chrisgarrity - Updated my previous content with all the content except the Illustration, which I will provide once I have completed it.

For sizing purposes here is a **temporary** asset:
This asset is at 2x, it should be displayed at 494px x 208px (we can adjust that ratio as we see fit, but a good starting point).
![unsupported-illustration](https://user-images.githubusercontent.com/3409578/33901607-553b9e2c-df40-11e7-9d0c-7f05b6b30b00.png)

carljbowman:
@chrisgarrity - Updated my original comment with an illustration asset.

PR #1087, opened 2017-12-20
Preview modals

### LINKED ISSUE #828: Scratch 3.0 Preview Introduction Modal
When we launch the Scratch 3.0 Preview we will need a modal to help contextualize the current state of 3.0, e.g. Calling out Features that are still in progress or that things might be a little buggy.

Info Considerations:
- Features in progress
- Things might be Buggy
- Link to Preview FAQ 
- Link to Feedback form ( to be put in the toolbar instead)

Initial UX Considerations / Questions:
- Does this open every time someone visits the Editor? Dismissible - "Do not show me again" check box type thing?
- We should also include links to the Preview FAQ and Feedback form in the Editor Toolbar
- It might be nice to include a Project ID load feature: simple a field to enter an ID or Project URL

chrisgarrity:
@mewtaylor it would be good to have your review early to see if this makes sense to use/refactor for the feedback form.

* current assets are placeholders
* text is localized
* modal shows every time the page is reloaded (no cookie)

### Resolves
fixes #828, #982 

### Proposed Changes

screen shot 2017-12-19 at 6 05 24 pm

screen shot 2017-12-20 at 11 14 02 am

### Test Coverage

- Automated tests run by clicking the tryit button.
- Will need to manually test on windows/ie to check the unsupported browser (I checked locally by changing 'MSIE' to 'Chrome' in the code and getting the other dialog.)


src/containers/preview-modal.jsx
```diff
@@ -0,0 +1,68 @@
...
+    handleTryIt () {
+        this.setState({previewing: true});
+        this.props.onTryIt();
+    }
+    handleCancel () {
+        window.history.back();
+    }
+    supportedBrowser () {
+        const userAgent = (typeof navigator === 'undefined') ? '' : navigator.userAgent;
+        if (userAgent === '' || userAgent.indexOf('Chrome') > 0 || userAgent.indexOf('Trident') > 0) {
```
mewtaylor: Probably not necessary, but wondering if we should use a library for this? Just in the case that the library has a maintainer that is keeping more up to date with any `userAgent` changes that might occur.

src/css/typography.css
```diff
@@ -0,0 +1,13 @@
...
+}
+
+h2 {
+    font-size: 1.5rem;
+    font-weight: bold;
+}
+
+p {
+    font-size: 1rem;
+    line-height: 1.5em;
```
mewtaylor: Should this be `rem` to match the other sizes?

src/containers/preview-modal.jsx
thisandagain: Agreed. This can often be notoriously tricky. I can recommend both of these two options:
https://www.npmjs.com/package/bowser
https://www.npmjs.com/package/platform

src/css/typography.css
chrisgarrity: Carl wanted the line height relative to the font-size, so it's em. If we later change font-size line-height with auto adjust.

chrisgarrity:
Note to self - remember to squash before merging

package.json
```diff
@@ -61,6 +61,7 @@
     "lodash.pick": "4.4.0",
     "minilog": "3.1.0",
     "mkdirp": "^0.5.1",
+    "platform": "^1.3.4",
```
thisandagain: Can you make this a strict version number?

Issue #1064, opened 2017-12-18
Replace "rotation style" with "size" in sprite info area


thisandagain:
We would like to remove the "Rotation" option in the sprite info area with a new selector for "Size" that fits between the "Show" and "Direction" options.

![image](https://user-images.githubusercontent.com/747641/34130680-c5c5ef32-e416-11e7-8e2c-c0b0829b167f.png)

carljbowman:
### Overview
Rotation-style is rather abstract and has hidden state. It does not feel like the right type of thing to include at the surface level within the Sprite Info area. The rotation style can still be set with blocks.

The new size input should adjust the size of the sprite, similar to other inputs in the space, e.g. X position dynamically adjusts the X position of the sprite.

### Mock
I believe we are using flex-box to position elements within the Sprite Info area, hence all the extra space around the Size input.
![sprite-info-update](https://user-images.githubusercontent.com/3409578/34211788-2a70162c-e568-11e7-840c-e05873f75f8f.png)

griffpatch:
Could the rotation mode be bundled into the direction box... a direction of
90 or -90 hardly seems right without the option to switch between
horizontal flipping or rotation. Could the rotation, drop down also have
the three options for static, flip & rotation modes as well as the
direction wheel?

thisandagain:
@griffpatch Would you mind filing that as a separate issue?

carljbowman:
@griffpatch - I have thought about this a little for the angle picker in the blocks (specifically point in direction), but I have some concerns about collapsing two properties that are programable into a single field. 

Color is an example of where we collapse may properties (H+S+B in 3.0 and Hue+Shade in 2.0) into a larger construct and that has always been rather confusing. I'd argue that this type of collapsing is a pattern we would want to avoid.

That said, I do think we should think more about how to highlight the relationships/impacts of compounding properties.

PR #1089, opened 2017-12-20
Add size and remove rotation style from sprite info area.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/1064

size-setter

### DIFF
--- a/src/components/forms/input.css
+++ b/src/components/forms/input.css
@@ -37,6 +37,8 @@
 }
 
 .input-small {
-    width: 3rem; 
+    width: 3rem;
+    padding: 0 0.5rem;
+    text-overflow: clip;
     text-align: center;
 }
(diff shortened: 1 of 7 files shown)


fsih:
Should the field be a bit bigger? It starts showing '...' around 200% which is pretty small

paulkaplan:
@fsih I agree, @carljbowman should just that field be larger, or each of the bottom inputs (direction & size)?

thisandagain:
Agreed. My inclination would be to slightly increase the width of both the direction and size fields.

paulkaplan:
Simpler solution was to take down the padding (it is center aligned anyway) for the smaller inputs, and remove the ellipses (since they take up 50% of the input). 

image

Issue #1110, opened 2017-12-21
Add default variable to all new projects


thisandagain:
We would like to add a default variable called "count" (set to `0`) to the toolbox for all new projects. This is to help newcomers to Scratch discover what a variable is and can do without having to go through the "Make a Variable" flow the first time.

thisandagain:
This is likely going to cause some localization problems. /cc @chrisgarrity

PR #1111, opened 2017-12-21
Add default "count" variable

paulkaplan:
### Resolves

Closes https://github.com/LLK/scratch-gui/issues/1110

### Proposed Changes

Add "count" as a (global) variable to the default project.

### Test Coverage

Added a check for this to the integration test dealing with variables.


chrisgarrity:
This isn't going to work for localizing as we don't ever change the contents of a project. One solution would be to load different 'default' projects based on the current language. That might be overkill, but there might be other things that we could localize in the default json.

PR #1116, opened 2017-12-21
WebGL unavailable modal

### LINKED ISSUE #888: Detect and show friendly error message if WebGL is not supported
### Expected Behavior

- GUI should present a friendly error message if WebGL is not supported

### Actual Behavior

- Fatal exception and stage fails to load

### Steps to Reproduce

- Use computer or browser without WebGL support
- Load https://llk.github.io/scratch-gui

chrisgarrity:
Minimal WebGL modal. Image asset will need to be replaced. 

Refactoring for later if there's time:
- pull together the 'unsupported' modals (webgl, browser)
- have welcome modal separate from unsupported-browser
- pull common localization strings into a separate module to reduce duplication.

### Resolves

fixes #888 

### Proposed Changes

screen shot 2017-12-21 at 5 55 37 pm

### Test Coverage

Find a machine that does not support WebGL (e.g. Jamaica, or old Samsung Chromebook)
- [ ] WebGL not supported modal appears as above
- [ ] WebGL link goes to https://en.wikipedia.org/wiki/WebGL#Support
- [ ] 'Back' button goes to the previous page
- [ ] Preview FAQ link goes to https://scratch.mit.edu/preview-faq


chrisgarrity:
I was able to test this at home using our old PC - Chrome and Edge both gave the WebGL not available modal.

chrisgarrity:
Unfortunately, running on IE doesn't show anything because there's an error in `registerBlockPackages()`:
ieerror

So I'm wondering if we should skip loading the GUI component if webGL is missing, or it's IE, and just show the modal with a static mock up of the GUI in the background.

chrisgarrity:
@paulkaplan @mewtaylor I started looking into fixing the IE issue as part of this PR, but the more I look into it, the less it has to do with WebGL. So I think this should get reviewed ignoring the IE issues, and then start a new PR for the IE problems/refactoring.

paulkaplan:
@chrisgarrity sounds reasonable to me. Should I review this then?

src/components/gui/gui.jsx
```diff
@@ -73,6 +74,7 @@ const GUIComponent = props => {
             {feedbackFormVisible ? (
                 <FeedbackForm />
             ) : null}
+            {(window.WebGLRenderingContext) ? null : (<WebGlModal />)}
```
chrisgarrity: This doesn't get called as WebGL errors cause the error-boundary to get used.

If the renderer starts handling the errors and failing by raising its own error, then this is unnecessary as it'll still get handled by the error-boundary. But if the renderer handles the errors, skips the stage, but doesn't otherwise fail we'll still want to show the unsupported modal.

src/containers/error-boundary.jsx
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    handleBack () {
+        window.history.back();
+    }
+
+    render () {
+        if (this.state.hasError) {
+            if (platform.name === 'IE') {
+                return <h1>Sorry Internet Explorer is not supported.</h1>;
```
chrisgarrity: This is a place holder. IE doesn't load so it never gets here.

src/containers/error-boundary.jsx
```diff
@@ -0,0 +1,48 @@
...
+            }
+            if (window.WebGLRenderingContext) {
+                const canvas = document.createElement('canvas');
+                if (!canvas.getContext('webgl')) {
+                    return <WebGlModalComponent onBack={this.handleBack} />;
+                }
+            } else {
+                return <WebGlModalComponent onBack={this.handleBack} />;
+            }
+            return <h1>Something went wrong.</h1>;
```
paulkaplan: If this is just a placeholder, can you file an issue for discussion about what the generic "something terrible happened" action should be?

chrisgarrity: 👍  https://github.com/LLK/scratch-gui/issues/1205

src/components/webgl-modal/webgl-modal.jsx
```diff
@@ -0,0 +1,98 @@
...
+import Box from '../box/box.jsx';
+import {defineMessages, injectIntl, intlShape, FormattedMessage} from 'react-intl';
+
+import styles from './webgl-modal.css';
+
+const messages = defineMessages({
+    label: {
+        id: 'gui.webglModal.label',
+        defaultMessage: 'Your Browser Does Not Support WebGL',
+        description: ''
```
paulkaplan: Needs a description

paulkaplan:
Other than the missing description for one of the formatted messages, this LG. I think the "Something went wrong" screen issue should be considered a critical one though. I know it replaces the blank screen, which doesn't happen much, but we definitely should figure out what should be shown.

Issue #829, opened 2017-11-01
Scratch 3.0 Preview Feedback form


carljbowman:
Create a mechanism to gather feedback during the Preview. Links to this feedback form will be sprinkle throughout the Preview experience so at any point a Scratcher can give feedback or report issues.

Grinthy:
Stop right there for a sec:

_for reporting stuff_

Github is for reporting stuff, right? Or is it anonymous?

thisandagain:
@grinthy Yes, but it is not reasonable or desireable to ask millions of kids to create Github accounts to tell us about issues they may encounter. For this we need a system that can handle a high volume of feedback that can be analyzed more easily and getting everything into a more structured data format (like a spreadsheet) will help with that.

carljbowman:
@mewtaylor - We would like the Google Form to appear in a modal iFrame over the GUI. Maybe we could use the modal feature that @chrisgarrity is working on for the [Intro Modal](https://github.com/LLK/scratch-gui/issues/828) and [Unsupported Browser Modal](https://github.com/LLK/scratch-gui/issues/982). 

## URL 
https://goo.gl/forms/KHVvX1nYY4IWFBPn1

## Sketches
### Give Feedback Button 
Add a button in the toolbar to open the Google Form in a modal
[feedback-icon.svg.zip](https://github.com/LLK/scratch-gui/files/1552605/feedback-icon.svg.zip)

![give feedback button](https://user-images.githubusercontent.com/3409578/33898182-ada9feec-df35-11e7-8d4d-a461a703b2d5.png)

### Quick sketch of Google Form in modal
Once we have this in a rough state we should dial in some of the interactions. In either case of a complete form or an incomplete form we will need a way to close the modal.
![feedback form modal](https://user-images.githubusercontent.com/3409578/33898202-b9717ec6-df35-11e7-9772-422018a40bdd.png)

redmangospros:
Cool, Like The Ideas. KIU

mewtaylor:
Implemented in #1120

PR #1120, opened 2017-12-22
GH-829: Add feedback form to menu bar

mewtaylor:
### Resolves

For #829

### Proposed Changes

Implements the feedback button and modal. It also includes a slight style change to the stage size buttons.

### Reason for Changes

This is needed for the deploy of alpha. Additionally, the small changes in styling to the stage buttons is needed due to a misunderstanding of the original implementation I did last week.


paulkaplan:
(marking needs work because it can't be merged until tests pass, but because it is just an unused variable lint error, the rest LG once that is fixed, so I approved)

mewtaylor:
Marking good to merge, tests now passing, thanks @paulkaplan !

PR #1123, opened 2017-12-23
Add next/previous backdrop options in backdrop block

sjhuang26:
### Resolves

[Issue #849 in the VM](https://github.com/LLK/scratch-vm/issues/849)

### Proposed Changes

Adds "next backdrop" and "previous backdrop" menu options to the "switch backdrop to [  ]" block.

### Reason for Changes

Consistency with Scratch 2.

### Test Coverage

None.

### DIFF
--- a/src/lib/blocks.js
+++ b/src/lib/blocks.js
@@ -44,9 +44,8 @@ export default function (vm) {
 
     const backdropsMenu = function () {
         if (vm.runtime.targets[0] && vm.runtime.targets[0].sprite.costumes.length > 0) {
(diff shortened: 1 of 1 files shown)


sjhuang26:
Note that because the string keys have been changed, the method in the VM that uses these string keys also has to be changed, which I did in [VM PR#882](https://github.com/LLK/scratch-vm/pull/882).

paulkaplan:
Oh ack I'm sorry, I didn't realize those strings already existed in the VM. And actually looking at the other places in the code, we are already using some regular strings. Maybe we should audit that usage all together instead of piece by piece. That is my bad. Let's just go back to the way you first had it with the regular strings. Sorry about that.

PR #1137, opened 2017-12-27
Implement paint-75: Add blank costume button

mewtaylor:
### Resolves

Implements https://github.com/LLK/scratch-paint/issues/75

### Proposed Changes

Adds a new button to the costume tab that creates a blank costume or backdrop.
* The naming convention of `costume[number]` or `backdrop[number]` currently follows 2.0
* Aligns the icon text in the costume tab to center to account for the larger text here.

### Reason for Changes

This is needed for alpha

### Unresolved Questions
* The spec calls for the text of these buttons to be hidden (and, I assume, in a title). Is that what we want as part of this PR too? /cc @carljbowman @paulkaplan


src/containers/costume-tab.jsx
```diff
@@ -41,27 +75,41 @@ class CostumeTab extends React.Component {
...
         const costumes = this.props.vm.editingTarget.sprite.costumes || [];
         this.setState({selectedCostumeIndex: Math.max(costumes.length - 1, 0)});
     }
-
+    handleNewBlankCostume () {
+        const emptyItem = costumeLibraryContent.find(item => (
+            item.name === 'Empty'
+        ));
+        const costumes = this.props.vm.editingTarget.sprite.costumes || [];
+        const name = this.props.vm.editingTarget.isStage ? `backdrop${costumes.length}` : `costume${costumes.length}`;
```
fsih: I think this might be off by 1 from what you want (if you have only backdrop1, it tries to add another backdrop called backdrop1) but it's fixed because vm's addCostume automatically runs StringUtil.unusedName to increment the number.

I think it's safe to always name new backdrops 'backdrop1' and new costumes 'costume1' here, since the string util will automatically handle the number.

Also we probably want internationalization for the word 'backdrop'

src/containers/costume-tab.jsx
```diff
@@ -82,31 +130,24 @@ class CostumeTab extends React.Component {
...
-                }]}
+                buttons={[
+                    {
+                        message: addBlankMessage,
+                        img: addBlankCostumeIcon,
+                        onClick: this.handleNewBlankCostume
+                    },
+                    {
+                        message: addLibraryMessage,
+                        img: addLibraryCostumeIcon,
```
fsih: Should we switch this from the cat to the add backdrop icon when you're on the stage? (../components/stage-selector/icon--backdrop.svg)

src/containers/costume-tab.jsx
mewtaylor: It was actually incrementing well it seemed to me at first, but good to know that was the vm! I will change it to `costume1` and `backdrop1`

src/containers/costume-tab.jsx
mewtaylor: Ah I didn't realize that was there! Will do

mewtaylor:
made those updates – thanks @fsih !

src/containers/costume-tab.jsx
```diff
@@ -18,6 +18,7 @@ import {
 } from '../reducers/modals';
 
 import addBlankCostumeIcon from '../components/asset-panel/icon--add-blank-costume.svg';
+import addLibraryBackdropIcon from '../components/asset-panel/icon--add-backdrop-lib.svg';
```
fsih: Maybe we should put shared icons in a shared place so we don't have to duplicate them?

PR #1158, opened 2018-01-03
Update preview illustration

chrisgarrity:
Replace the placeholder illustration with the new one from Carl

### Resolves

This is the last part of #828


thisandagain:
Looks great!

Issue #1162, opened 2018-01-03
Change default variable name to "my variable"


thisandagain:
Change the default variable name from "count" to "my variable" to more clearly connect it to the button above and remove some of the unhelpful ambiguity that came from "count".

kyleplo:
Why does the editor even need to start with a variable? I think is better if users create their own variables.

PR #1163, opened 2018-01-03
Change default variable name to 'my variable'. 

thisandagain:
### Resolves

Resolves GH-1162

### Proposed Changes

Changes the default variable name from "count" to "my variable"

### Reason for Changes

See GH-1162

### DIFF
--- a/src/lib/default-project/project.json
+++ b/src/lib/default-project/project.json
@@ -22,7 +22,7 @@
     "tempoBPM": 60,
     "videoAlpha": 0.5,
     "variables": [{
-        "name": "count",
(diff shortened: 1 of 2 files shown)


paulkaplan:
Mentioned this offline, but for posterity:

scratch-blocks replaces spaces with `&nbsp;` (aka `\u00A0`) in block text. For some reason selenium is not correctly normalizing whitespace the way it says it does, so you need to use `\u00A0` instead of `' '` when clicking block text for now.

thisandagain:
Note to self: squash during merge.

PR #513, opened 2017-07-07
Make Sprite/Sound/Costume deletion use the new prompt style

TheBrokenRail:
### Resolves

_Fixes #340 and #385_

### Proposed Changes

_Make Sprite/Sound/Costume deletion use the new prompt style_

### Reason for Changes

_So the prompt styles are consistent_

### Screenshot

asset-delete-new

PR #1170, opened 2018-01-04
Remove sprite-specific text from sprite/sound/etc deletion warning

### LINKED ISSUE #385: ALL asset-delete buttons show "delete this sprite?" dialog
### Expected Behavior

When clicking on an "X" button, the message displayed is based on what is being deleted. For example:

* "Are you sure you want to delete this sound?" when deleting a sound
* "Are you sure you want to delete this sprite?" when deleting a sprite
* "Are you sure you want to delete this costume?" when deleting a costume
* (etc for backdrops and such)

### Actual Behavior

The displayed prompt's message is always "Are you sure you want to delete this sprite?".

Though these dialogs are almost certainly temporary (i.e. they will be replaced by custom popup dialogs), the cause of the issue is still worth noting.

This problem stems from all asset [`Selector`s](https://github.com/LLK/scratch-gui/blob/64891259b1f58696208b20a46df917e58005a6b0/src/components/asset-panel/selector.jsx) using the [`SpriteSelectorItem`](https://github.com/LLK/scratch-gui/blob/64891259b1f58696208b20a46df917e58005a6b0/src/containers/sprite-selector-item.jsx#L23) class. It doesn't really make sense to have costume items using the sprite selector item; there should probably be a more generic parent class (ex. `SelectorItem`) that `SpriteSelectorItem`, and potential sound/costume-backdrop classes, inherit from.

### Steps to Reproduce

1. View the "sounds" or "costumes" tab.
2. Click on the "X" button inside a sound or costume's icon.
3. Read the dialog text that pops up.

macOS 10.12.5, Chrome 58.0.3029.110

towerofnix:
### Resolves

Fixes #385.

### Proposed Changes

Removes the sprite-specific message from the "delete" prompt that appears when you click on the X of any sprite, sound, costume, or backdrop card in a project.

### Reason for Changes

If anything, this is a hack :) - But it's appropriate. This fixes #385, which is scheduled for the February milestone; it's a very old bug (May of last year) and has been put into and brought out of backlog over the course of its lifespan. The *correct* fix is to make non-sprite elements use an element other than `SpriteSelectorItem`, but this is a *hard* fix. It means creating a general component for "selector items", then making subclasses for each selector item, and changing the GUI to appropriately use these; no quick task.

This is a simple fix and gets rid of the bug quickly. It should also be noted that putting a lot of time into fixing this *correctly* now probably isn't worth it - see @TheBrokenRail's pull request #513, which contained a fair amount of work but was closed to this comment:

> Sorry for the delay. This totally fell off my radar, that's my fault. We appreciate you porting this over to the new modal style, but the delete alerts are temporary for prototyping. **Eventually we will remove the alerts entirely.**

(Emphasis mine.) Since these alerts are eventually going to be removed altogether, there's no real reason to split this into multiple components right now. (Since they'll all behave the same, I suppose a better solution would just be to rename SpriteSelectorItem to SelectorItem; but this is also a fair amount of work to deal with! And not necessary to fix the bug; in fact, it would just contain the change in this PR anyways.)

### Test Coverage

Untested. I didn't try to run this locally, either (I made the change online through GitHub), but this is just a string change, so I can't imagine I broke anything :)

### DIFF
--- a/src/containers/sprite-selector-item.jsx
+++ b/src/containers/sprite-selector-item.jsx
@@ -21,7 +21,7 @@ class SpriteSelectorItem extends React.Component {
     }
     handleDelete () {
         // eslint-disable-next-line no-alert
-        if (window.confirm('Are you sure you want to delete this sprite?')) {
+        if (window.confirm('Are you sure you want to delete this?')) {
             this.props.onDeleteButtonClick(this.props.id);
         }
     }


thisandagain:
This seems fine to me for now and should help reduce some bug report volume when we release the preview. @paulkaplan @rschamp @carljbowman any objections?

Issue #1171, opened 2018-01-04
Pressing escape in fullscreen should exit fullscreen


towerofnix:
### Expected Behavior

Pressing the escape key while viewing a project in fullscreen should switch back to the editor.

### Actual Behavior

Nothing happens; the escape key is not handled.

### Steps to Reproduce

Enter fullscreen (click the right-most button above the corner of the stage) and press escape.

### Operating System and Browser

Debian (testing), Firefox 58.

kyleplo:
Also, can there be a way to do full-fullscreen, like in https://davidwalsh.name/demo/fullscreen.php

josiahneuberger:
@kyleplo #1195 implements requested behavior with 'escape' key; however, your request in comments for a 'F11' type full-screen mode should probably be made in a separate issue. Thanks for pointing out the missing functionality.

thisandagain:
Resolved in #1195

PR #1195, opened 2018-01-07
Exit full-screen mode by hitting escape key

josiahneuberger:
### Resolves

LLK/scratch-gui#1171

### Proposed Changes

The stage-header now listens for the 'escape' key press. If in full-screen mode, it exits this mode.

### Reason for Changes

Requested by user  @kyleplo.

### Test Coverage

No additional testing added.


thisandagain:
Thanks @josiahneuberger! We'll take a look this week.

Issue #1179, opened 2018-01-05
Green flag / stop should be brought in to edge of stage in full screen mode


paulkaplan:
Reported during workshop: 

When go to full-screen mode (on large screen), the green flag and stop sign are at the top-left of the window, but I would have preferred them to be above the top-left corner of the stage.

PR #1196, opened 2018-01-07
center/size stage-header controls in the same way as stage in fullscreen

josiahneuberger:
### Resolves

LLK/scratch-gui#1179

### Proposed Changes

Brings the menu controls in stage header component to be centered and of the same width as the stage in full-screen mode.

### Reason for Changes

In non-full-screen mode, the menu controls for the stage are aligned with the width of the stage. Full screen mode should match this behavior. 

### Test Coverage

No test added.

### DIFF
--- a/src/components/stage-header/stage-header.css
+++ b/src/components/stage-header/stage-header.css
@@ -16,6 +16,7 @@
 
 .stage-menu-wrapper {
     display: flex;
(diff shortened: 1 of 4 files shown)


thisandagain:
👍 Thanks @josiahneuberger

PR #1213, opened 2018-01-09
Support IE just enough to show unsupported modal

### LINKED ISSUE #982: Scratch 3.0 Preview Browser Not Supported Modal

chrisgarrity:
- Add object-assign polyfill to handle missing features in IE.
- Prevent audio library from trying to create an Audio Context is the browser is IE, as that fails during load. (This prevents the `SCRIPT445: Object doesn't support this action` error) 

Catch any other errors during render, and render Browser unsupported modal instead.

Fixes #982 

/cc @ericrosenbaum @rschamp

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,5 +115,8 @@
       "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/test/__mocks__/fileMock.js",
       "\\.(css|less)$": "<rootDir>/test/__mocks__/styleMock.js"
     }
+  },
+  "dependencies": {
+    "babel-polyfill": "6.26.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "css-loader": "^0.28.7",
     "enzyme": "^3.1.0",
     "enzyme-adapter-react-16": "1.1.0",
+    "es6-object-assign": "1.1.0",
     "eslint": "^4.7.1",
     "eslint-config-scratch": "^5.0.0",
     "eslint-plugin-import": "^2.8.0",
@@ -115,8 +116,5 @@
       "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/test/__mocks__/fileMock.js",
       "\\.(css|less)$": "<rootDir>/test/__mocks__/styleMock.js"
     }
-  },
-  "dependencies": {
-    "babel-polyfill": "6.26.0"
   }
 }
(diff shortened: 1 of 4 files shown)


paulkaplan:
@chrisgarrity nice! Works great, good catch about the audio context stuff!

My only question is whether we want to include only the polyfills we need instead of all of babel-polyfill, which is a very large library. I was able to get the "not supported" modal to show in ie11 with just a polyfill for object.assign. I think it would be good to be as light as possible with polyfills because they can make it difficult to track down other bugs sometimes.

package.json
```diff
@@ -115,5 +115,8 @@
       "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/test/__mocks__/fileMock.js",
       "\\.(css|less)$": "<rootDir>/test/__mocks__/styleMock.js"
     }
+  },
+  "dependencies": {
+    "babel-polyfill": "6.26.0"
```
paulkaplan: Also, should this be in dev dependencies with everything else?

chrisgarrity: babel-polyfill instructions explicity say to add to dependencies (https://babeljs.io/docs/usage/polyfill#installation). However, with the object.assign polyfill I tried adding it as a dev-dependency and it seems ok.

paulkaplan:
Code LG. Just for completeness, what version of IE does this work down to? I tested in 11 and it worked, but do you know how far back it works? @thisandagain how far back are we expecting to go?

chrisgarrity:
@paulkaplan looks like the polyfill needs to be a dependency after all - it doesn't work on IE on develop.

heyitsmeuralex:
Can't we just detect IE and display something with regular html/css rather than load a polyfil just to display an error?

chrisgarrity:
@heyitsmeuralex unfortunately the gui depends on a lot of different libraries, and it fails when **loading** in IE without the polyfill. The polyfill allows the GUI to load so that we can detect IE and show the browser unsupported message.

chrisgarrity:
Nevermind perhaps I checked before the merged code had deployed. Develop is working fine - yay!!

Issue #1207, opened 2018-01-08
() of () reporter is always missing the last sprite


fsih:
<img width="983" alt="screen shot 2018-01-08 at 12 47 42" src="https://user-images.githubusercontent.com/2855464/34684199-50b973fa-f472-11e7-87a7-7191d0aa6d11.png">

When you first have only the cat and stage, the dropdown only has the stage.
When you add another sprite, say Apple, now the dropdown only shows the stage and Sprite1 (the cat)

griffpatch:
I don't think two sprites should ever 'have' the same name?

hedgehog125:
It could be to do with the sprite names, the names might be used as ids and if they have the same id there's going to be problems.
In Scratch 2.0 if you create a sprite with an name that's already being used then it will add a 2 to the end (or increase the number at the end), this mechanic doesn't seem to exist in the current version so that could be the problem.

sjhuang26:
Yes, and duplicate sprite names are not allowed in Scratch 2 (if you try to change the sprite name to another sprite's name, it won't let you). This behavior should be added to Scratch 3.

fsih:
I discovered that this bug actually doesn't have to do with the duplicate names. I broke out the duplicate names bug to https://github.com/LLK/scratch-gui/issues/1209.

amazinigmech2418:
Maybe the JS array is manipulated incorrectly. Maybe it doesn't show the full array because of a miscalculated loop count. (Should be array.length in for loops)

fsih:
It was actually a mistake that the "() of ()" block had a checkmark next to it. It's working as intended that the sprite that's currently selected is missing from the dropdown for that sprite. This will be resolved by removing the checkbox. Sorry for the confusion everyone.

amazinigmech2418:
OK.

PR #1215, opened 2018-01-09
Include editing target in x of x menu

fsih:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/1207

### Proposed Changes

Include the current target in the list of sprites that can be selected in the () of () reporter.
This appends the current target to the start of the list to match Scratch 2.0.


fsih:
We decided that () of () shouldn't be monitorable at all, so this PR isn't relevant anymore

Issue #1114, opened 2017-12-21
x and y are squeezed in the target info pane


fsih:
### Expected Behavior

x and y have padding on both sides

### Actual Behavior

x and y have no padding on the left
### Steps to Reproduce

When your window width is small enough that the stage width gets smaller, the x and y fields get squeezed.
_Explain what someone needs to do in order to see what's described in *Actual behavior* above_
![image](https://user-images.githubusercontent.com/2855464/34275676-5dab3828-e66c-11e7-889e-cf46de3c8477.png)

### Operating System and Browser

Windows Chrome (seems to be Windows specific)

paulkaplan:
:( here is what it looks like on mac chrome

![image](https://user-images.githubusercontent.com/654102/34275901-30405eda-e66d-11e7-8b17-0b181859adff.png)

fsih:
The cause of this seems to be that we're not setting a width on those text inputs, so they are some default width set by the browser. On Macs that default width is smaller than on Windows. So we should probably set a default width for text inputs that corresponds to the width on Macs on Chrome, so that it will always look about the same as it does on Macs on Chrome.

Here's what the info box looks like when I set the width to 8rem, which seems about right:
![image](https://user-images.githubusercontent.com/2855464/34307879-5986d2e6-e718-11e7-85e7-2d08e0d8a876.png)

PR #1217, opened 2018-01-09
Set default input width

fsih:
### Resolves

https://github.com/LLK/scratch-gui/issues/1114
Before
image
After
image

### Proposed Changes

Set a default width on all input boxes. 8rem is close to and slightly less wide than it was by default on Mac Chrome, so now input boxes without widths defined on all platforms should look pretty close to how it used to look on Mac Chrome.

### Reason for Changes

Windows had a wider default input box, which was taking up too much space.


fsih:
I'm waiting on https://github.com/LLK/scratch-paint/pull/249 to put this in, since this exacerbates that bug.

PR #1223, opened 2018-01-10
Remove some blocks’ menu shadows to make them non-droppable

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/1199

Depends on https://github.com/LLK/scratch-blocks/pull/1339

### Proposed Changes

Remove the menu shadows for some blocks to make the input non-droppable (i.e. so that you cannot drag and drop a reporter block onto the input).

### Reason for Changes

See discussion in the issue above. Our focus is preventing confusion for beginners.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "redux-throttle": "0.1.1",
     "rimraf": "^2.6.1",
     "scratch-audio": "0.1.0-prerelease.1515596313",
-    "scratch-blocks": "0.1.0-prerelease.1515707006",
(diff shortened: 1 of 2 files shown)


paulkaplan:
@ericrosenbaum because this has a hard dependency on a scratch-blocks update, i'm going to mark it as "blocked" until https://github.com/LLK/scratch-blocks/pull/1339 is merged. Once it is merged, the update to the scratch-blocks version in the package json needed to pull it in should be included in this PR.

PR #1231, opened 2018-01-11
Use style sizing to fix issues with Edge + flex + canvas.

paulkaplan:
Additionally render the canvas at a multiplier of the device pixel
ratio, so we get hidpi images for supported screens. The fixed CSS
sizing allows us to do that.

The problem was that on Edge, flexbox apparently can grow canvas
elements, whereas on Chrome/FF the canvas width/height is respected. No
problem, we can set the width/height in CSS to fix it, which
conveniently allows us to also fix the low-res-on-retina-screens problem
at the same time.

Before on Edge
image

After on Edge
image

@cwillisf this may interfere with some of the work you are doing with the SVG rendering stuff. If so, feel free to defer this until that is done. But it is just a couple line change, so it should be easy to reconcile with whatever you are working on. But let me know if it causes any issues.

### DIFF
--- a/src/components/costume-canvas/costume-canvas.jsx
+++ b/src/components/costume-canvas/costume-canvas.jsx
@@ -105,12 +105,12 @@ class CostumeCanvas extends React.Component {
         return (
             <canvas
                 className={this.props.className}
-                height={this.props.height * window.devicePixelRatio}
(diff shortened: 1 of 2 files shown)


Kenny2github:
NANI?????
image
*ghost* approved the changes???

Issue #1227, opened 2018-01-11
Use scratch-render API to check for webGL support


paulkaplan:
Wanted to leave an issue to remember, once https://github.com/LLK/scratch-render/pull/226 is merged we should update GUI to

1. Use the static method on scratch-render `Renderer.isSupported()` to replace [line 77](https://github.com/LLK/scratch-gui/blob/5d07525b9d1e0d7e662eab53d99c8f878b17110b/src/components/gui/gui.jsx#L77) of `components/gui/gui.jsx`.
2. Replace the webGL check in the error boundary because it shouldn't be needed anymore.

/cc @chrisgarrity if you have thoughts about those 2?

PR #1238, opened 2018-01-12
Use new renderer methods to check for WebGL.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/1227

### Proposed Changes

0. Update renderer in package json to get new checking methods.
1. Use new renderer methods to check for WebGL.
2. Remove the WebGL checks from the error boundary. Instead, just do not
render the stage if it is not supported.

### Reason for Changes

The renderer can now tell us if it can be supported. Because we put up a fullscreen block if it isn't supported, do not bother trying to render the stage if the renderer is not supported (preventing errors which would lead to the error boundary. ). 

### Test Coverage

Tested locally using an extension to disable WebGL.

### DIFF
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -6,6 +6,7 @@ import MediaQuery from 'react-responsive';
 import {FormattedMessage} from 'react-intl';
 import tabStyles from 'react-tabs/style/react-tabs.css';
 import VM from 'scratch-vm';
+import Renderer from 'scratch-render';
 
 import Blocks from '../../containers/blocks.jsx';
 import CostumeTab from '../../containers/costume-tab.jsx';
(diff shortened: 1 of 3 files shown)


src/components/gui/gui.jsx
```diff
@@ -127,14 +132,18 @@ const GUIComponent = props => {
...
-                                <Stage
-                                    height={isFullSize ? layout.fullStageHeight : layout.smallerStageHeight}
-                                    shrink={0}
-                                    vm={vm}
-                                    width={isFullSize ? layout.fullStageWidth : layout.smallerStageWidth}
-                                />
-                            )}</MediaQuery>
+                            {/* eslint-disable arrow-body-style */}
+                            <MediaQuery minWidth={layout.fullSizeMinWidth}>{isFullSize => {
+                                return isRendererSupported ? (
```
chrisgarrity: Nice! It feels better to skip the Stage component if we don't have the renderer.

PR #1274, opened 2018-01-17
Manually update the scratch-* deps, greenkeeper seems behind.

paulkaplan:
@rschamp these aren't being left behind for any particular reason right?


rschamp:
Nope, I guess Greenkeeper was just more behind than we noticed. Let's hold off until https://github.com/LLK/scratch-vm/pull/897 is merged/built

paulkaplan:
OK. I can update this PR with the new VM and the new render when they build.

PR #1297, opened 2018-01-17
Sprite library now centered

### LINKED ISSUE #1277: Center each row in libraries
### Expected Behavior

- Each row in the library should be centered within the available space

### Actual Behavior

- Rows are left aligned

### Steps to Reproduce

- Open sprite library
- Resize browser window

dgnball:
### Resolves
#1277

### Proposed Changes
Changed the library style sheet to center each row on the sprite library.

### Reason for Changes
Center the grid on the sprite library.

### Test Coverage
Ran on Firefox 57.0.1 on Mac OS 10.13.1 and checked sprites were centered.

### DIFF
--- a/src/components/library/library.css
+++ b/src/components/library/library.css
@@ -20,7 +20,7 @@
 
 .library-scroll-grid {
     display: flex;
-    justify-content: flex-start;
+    justify-content: center;
     align-content: flex-start;
     background: $ui-pane-gray;
     flex-grow: 1;
--- a/src/components/library/library.css
+++ b/src/components/library/library.css
@@ -5,6 +5,7 @@
     position: absolute;
 
     display: flex;
+    flex-direction: column;
     height: 100%;
     width: 100%;
 
@@ -19,13 +20,8 @@
 }
 
 .library-scroll-grid {
-    display: flex;
+    display: grid;
     justify-content: center;
-    align-content: flex-start;
+    grid-template-columns: repeat(auto-fill, 160px);
     background: $ui-pane-gray;
-    flex-grow: 1;
-    flex-wrap: wrap;
-    overflow-y: auto;
-    height: calc(100% - $library-header-height);
-    padding: 0.5rem;
 }


src/components/library/library.css
```diff
@@ -20,7 +20,7 @@
 
 .library-scroll-grid {
     display: flex;
-    justify-content: flex-start;
+    justify-content: center;
```
thisandagain: This gets us part of the way there. This center aligns the items in the container element, but also aligns any items at the bottom of the library to the center as well.

### Before
image

### After
image

We want to have the items be centered in the space, but any "non-full" rows should be left aligned:
https://stackoverflow.com/questions/32802202/how-to-center-a-flex-container-but-left-align-flex-items

dgnball:
So I changed this part of the code to use a grid layout (and why not, seeing as we don't support IE anymore 😃) rather than flex as that made it a lot easier to get the effect your were looking for and it seemed to make the CSS easier to read. The boxes on each of the sprite libraries is slightly narrower but match the size specified in the underlying library-item size and it seems like you get more useful stuff to fit on the screen. I tested this on the latest FF, Chrome and Safari and it looks ok to my eye.

rschamp:
Thanks @dgnball! According to https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns, the grid-template-columns property is unsupported by Edge, so I would like to test this out there before merging. Have you looked at it on Edge yet?

dgnball:
Tried this on the EdgeHTML 16.16299. Seems fine to me (see below). Maybe the compatibility table is out of date.

image

thisandagain:
Looks good @dgnball. Tested on Edge, Firefox, and Chrome latest.

Ryuno-Ki:
@dgnball wrote
> Maybe the compatibility table is out of date.

FYI: If I read [this article](https://hacks.mozilla.org/2018/01/introducing-the-mdn-product-advisory-board/) correctly, the MDN product advisory board will change that soon~ish

PR #1309, opened 2018-01-19
Update deps

paulkaplan:
rolling all of the recent updates into one!

### DIFF
--- a/src/components/preview-modal/preview-modal.jsx
+++ b/src/components/preview-modal/preview-modal.jsx
@@ -54,6 +54,7 @@ const PreviewModal = ({intl, ...props}) => (
                 </button>
                 <button
                     className={styles.okButton}
+                    title="tryit"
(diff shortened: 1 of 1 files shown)


paulkaplan:
Sorry I'm stalling on this because it seems to break running the jest tests with locally linked scratch-vm, that is why the tests are passing on travis, but fail if you have scratch-vm linked locally. Because of that I can't merge this until figuring that out.

paulkaplan:
This PR now includes updates to all of our scratch-* deps and a few of the other react packages that were lagging behind. 

One important thing I left out of here was the update to jest 22. I was seeing an issue where running the tests with a locally linked scratch-vm would fail in the jest unit tests because of a strict-mode incompatibility with the node build of scratch-vm. I spent a day trying to fix it / track it down to no avail. @rschamp could we find a time to pair on it to figure that out? I could really use a hand on it.

PR #1310, opened 2018-01-19
Renderer dep

fsih:
Add the dependency on svg renderer to GUI. scratch-render will get the svg render code from GUI. Eventually, other parts of GUI will use the svg-renderer code.


fsih:
Closing for now while I figure out if this actually makes sense now

Issue #1320, opened 2018-01-22
Crash after recording sound and adding a new sprite


paulkaplan:
Crash can be reproduced by the following:

1. Go to the sound tab.
2. Create a recording, save it.
3. Click "Add Sprite", add the "Horse" sprite.
4. "Oops something went wrong"

PR #1321, opened 2018-01-22
Constrain selected sound index to the list of available sounds.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/1320

### Proposed Changes

Make sure that the "selected sound index" is within the range of available sounds.

### Reason for Changes

When switching sprites, the selected sound index is supposed to reset to zero. This is done in the `componentWillReceiveProps`. However the setState there is async. The sound tab can be rendered before that setState, so you can get to a place where the selected sound index is outside the range of available sprites. 

### Test Coverage

Added a regression integration test to make sure it works.

### DIFF
--- a/src/containers/sound-editor.jsx
+++ b/src/containers/sound-editor.jsx
@@ -190,8 +190,11 @@ SoundEditor.propTypes = {
 };
 
 const mapStateToProps = (state, {soundIndex}) => {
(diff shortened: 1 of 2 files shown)


paulkaplan:
@ericrosenbaum could you test/review this?

Issue #1328, opened 2018-01-23
When you click on a sprite, it switches to that sprite instead of triggering "this sprite clicked" hat


fsih:
### Expected Behavior

You used to be able to switch workspaces by dragging a sprite around, but just clicking a sprite shouldn't change workspaces.

Clicking should activate the "when this sprite clicked" stack

### Actual Behavior

You now switch to the workspace of the sprite you click on. "When this sprite clicked" hat doesn't work.

### Steps to Reproduce

1. Make 2 sprites. Put different code in each of them so you can tell the workspaces apart.
2. Be on sprite 1
3. Click on sprite 2
The workspace switches to sprite 2.
![spriteswitch](https://user-images.githubusercontent.com/2855464/35297657-ad819126-004d-11e8-9c19-71ba408b6ca6.gif)

### Operating System and Browser

Windows 10 Chrome. This is on both develop and preview.

paulkaplan:
Interesting. I can't reproduce on Mac/Chrome with and without a mouse, and an iPad on both preview/develop. Can anybody else repro this?

vutondesign:
I can repro this, but only when I double click on the sprite. (MacOS 10.13.2, Chrome 63)

Kenny2github:
I can’t repro this on iOS 11 (developer beta) - not even double tapping causes this problem.

fsih:
I'm only able to repro on Windows Chrome so far. Here's a project to use:
http://localhost:8601/#199623192
I'm pretty sure this only regressed recently

scratch-gui/src/containers/stage.onMouseDown gets called on the click. This sets mouseDownTimeoutId in the state. Then on Windows, immediately after, stage.onMouseMove is called. Since the timeout ID is indeed set, this behaves as if a double click was performed and causes a sprite change. e.movementX and e.movementY are both 0 in this event.

On Mac, stage.onMouseMove doesn't get called immediately after mouse down.

(But why doesn't the hat activate?)

fsih:
I found why the hat isn't called. In stage.onMouseMove, the state is set to isDragging. Then in stage.onMouseUp, 
```
if (this.state.isDragging) {
            this.onStopDrag();
            ....
} else {
            ...
            this.props.vm.postIOData('mouse', data);
            ...
}
```
So postIOData is never called by mouse up. postIOData is what eventually would call scratch-vm/src/io/mouse.js:Mouse._activateClickHats

fsih:
[...]

PR #1335, opened 2018-01-24
Ignore onMouseMove with no movement

fsih:
### Resolves

https://github.com/LLK/scratch-gui/issues/1328


paulkaplan:
Should we close this since it can't be repro'd? Or leave it open?

PR #1344, opened 2018-01-26
Fix integration tests for the recent chrome stable update

paulkaplan:
Fix the issues with the integration tests because of the issues introduced by chrome update (unclear if that is the root cause?)

The actual problem that we were seeing is that the promise for getting the project with the ID hash was not resolving, causing the tests to fail. I _believe_ this might be because the hash change is swallowed by the app and a reload doesn't actually happen. Something in the chrome/chromedriver update might have made this fail. 

In trying to fix this issue, I did a bit of refactoring that we had been meaning to do for a while, namely, splitting the one giant `test.js` into files for each part of the app ("Working with sounds", "Using the blocks", etc.)

That had a kind of "side-effect" of solving the problem because the project load tests now start from a fresh webdriver, meaning the page load is "real". As a side note, doing a project url load while on that page will cause a full page reload, because it isn't a hash "change"

The update to the chromedriver dependency is because the new stable (chrome 64) requires 2.35 chromedriver.


paulkaplan:
Ok so I confirmed that the root cause of this is that the promise from `driver.get` is not resolving if it is a hashchange that is caught.

Basically we had tests that did this:
```
get('/')
... do things
get('/#<PROJECT_ID') // <= this promise does not resolve on chrome 64/chromedriver 2.35
... do more things
```

So the "fix" for that in this PR is splitting those files up so that project loads happen on a fresh driver instead of after a different page load.

Issue #1349, opened 2018-01-26
"Run without screen refresh" label is not properly aligned with checkbox


thisandagain:
### Expected Behavior

- Label should be next to the checkbox

### Actual Behavior

- Label has extra padding in relation to the checkbox

![image](https://user-images.githubusercontent.com/747641/35454869-c4c53b7e-029d-11e8-9604-9dc9b34a525e.png)

### Steps to Reproduce

- Open "Make a Block" modal

### Operating System and Browser

Mac OS 10.13.2  |  Chrome Version 63.0.3239.132 (Official Build) (64-bit)

Narulacodes:
I can repro this from chromeOs, Chrome Version 63.0.3239.132 (Official Build) (64-bit) the same thing

amazinigmech2418:
This should be pretty easy to fix...

amazinigmech2418:
I think what needs to be done is probably update the CSS to change either the width of the checkbox or the padding/margins of the text.

amazinigmech2418:
I just submitted a PR for this! Hope it helps!

paulkaplan:
Thanks for helping us find the issue @amazinigmech2418. I mentioned on your PR, but I'll put it here too: this actually needs to be fixed as an update to https://github.com/LLK/scratch-gui/pull/1217, so I'm removing help-wanted and assigning to @fsih the original author of that PR.

PR #1360, opened 2018-01-29
Fix Checkbox Row Align

amazinigmech2418:
### Resolves

https://github.com/LLK/scratch-gui/issues/1349

### Proposed Changes

It will fix the align for the checkbox row

### Reason for Changes

It is needed to fix the align as mentioned in the issue.

### Test Coverage

I would do this, but don't know how to get GH pages to build for specifically the amazinigmech2418-develop branch since the other branches include a test Scratch 3 Scratch extension. Also, only the last commit is needed since the other ones were to delete the extension from the branch. @thisandagain I hope this can fix the bug!


amazinigmech2418:
@thisandagain I hope this helps!

amazinigmech2418:
I figured out .option-card had the text-align set to center and figured that was the problem with the alignment of the checkbox.

amazinigmech2418:
Yes! Travis passed!

Kenny2github:
0.0 that's a lot of commits for a single line change 🤔

paulkaplan:
Sorry this doesn't actually fix the problem. 

image

The issue is that the input has `width: 8rem`, which was introduced by https://github.com/LLK/scratch-gui/pull/1217. I'm going to close this and actually remove help-wanted because @fsih i think you should address this with a fix to that line in `typography.css`.

amazinigmech2418:
@Kenny2github Yeah... I had to delete some stuff from a previous branch to test Scratch 3 extensions... @paulkaplan OK. I didn't see that in the CSS at first.

Issue #1298, opened 2018-01-18
Variables By Id introduces warning related to variable/list monitors


kchadha:
Creating a new variable or list (with the [variables by id work in scratch-blocks](url)) issues the following warnings (respectively) in the console. These are non-fatal.

```raven.js:55 Warning: Failed prop type: Invalid prop `value` of type `number` supplied to `MonitorComponent`, expected `string`.```

```raven.js:55 Warning: Failed prop type: Invalid prop `monitors` of type `OrderedMap` supplied to `MonitorList`, expected instance of `OrderedMap`.```

**Comment by kchadha** [comment]
Updates: the second warning listed above is created regardless of whether a scalar variable or list is being created, and after talking to @fsih, it seems that these warnings (or at least the one about the OrderedMap) have been around for a while, and were not created because of https://github.com/LLK/scratch-blocks/pull/1350.

**Comment by Topppy** [comment]
the second warning。
Version of "immutable" package in scratch-gui is "3.8.2", while  version of "immutable" is "3.8.1"  in “scratch-vm” package. 
Monitors data is parsed by `vm.loadProject` which use "OrderedMap" of "3.8.1". 
while in monitor-list.jsx 
```
 monitors: PropTypes.instanceOf(OrderedMap), 
```
use "OrderedMap" of "3.8.2".  They are not the same "OrderedMap", ` monitors instanceof OrderedMap` will return` false`.

how to fix?
make sure that immutable in scratch-ui/package.json and scratch-vm/package.json are equal.
we can change scratch-gui/package.json:
```
"immutable": "3.8.1",
```

PR #1365, opened 2018-01-29
Fixing monitor prop type warning for scalar variable monitors.

kchadha:
### Resolves

Partial fix for #1298 

### Proposed Changes

Proptype validator for monitor prop type now asserts that the monitor value can be a string or a number.

#### Additional Notes

I couldn't seem to remove the second warning mentioned in #1298 about OrderedMap, but @fsih mentioned that that warning has been around for a while...

### Reason for Changes

#1298

### Test Coverage

Existing tests pass.


fsih:
I think this looks good, but Paul should review to be sure.

The second warning I think is caused by VM and GUI having different copies of the immutable library. Although the object is identical in code, it's not strictly equal.

PR #1366, opened 2018-01-29
Fix input css to apply only to text fields

fsih:
### Resolves

Fix https://github.com/LLK/scratch-gui/issues/1349
Previous PR broke non-text inputs

### DIFF
--- a/src/components/forms/input.jsx
+++ b/src/components/forms/input.jsx
@@ -9,14 +9,19 @@ const Input = props => {
     return (
         <input
             {...componentProps}
-            className={classNames(styles.inputForm, {
-                [styles.inputSmall]: small
-            })}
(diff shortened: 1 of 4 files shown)


paulkaplan:
image

fsih:
I pulled in the change to Input from scratch-paint which accepts className as a prop, and made the width apply only to the sprite name field.

amazinigmech2418:
You should just be able to change the original input selector to `input[type="text"]`...

fsih:
The `input[type="text"]` change caused all of the text inputs to be that wide, which made the sprite info pane too big, like in Paul's screenshot. This change was so it only applies to the sprite name field.

amazinigmech2418:
OK. @fsih

Issue #1408, opened 2018-02-05
Slightly moving a sprite switches to it, making it hard to activate "when sprite clicked"


morantsur:
(Noticed in microworlds playtesting)

clicking imprecisely on a sprite on the stage many times moves it a tiny bit, which makes this sprite be chosen, and does not activate "when sprite clicked" - I feel like it's much more sensitive than 2.0 in that.  

This happens to many kids, especially younger ones using a trackpad... 

Maybe consider adding a threshold for drag actions? 

(Similar to #1328, but with a different cause)

hedgehog125:
That would be useful, also you can drag sprites in full screen which is also annoying.

hedgehog125:
By a drag threshold do you mean as a delay or a minimum movement before it's a drag? Also, maybe adding the shadows that are in Scratch 2.0 might be useful as well because then you know when you're dragging.

paulkaplan:
There is a time threshold for drags, it is the same as the threshold for considering two separate clicks as a single double click. And yeah that timing threshold is longer on scratch 3, as in, it allows two clicks separated by a longer amount of time to act like a double click. In scratch 2, it is 400ms [src](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/util/GestureHandler.as#L64) and in scratch 3 it is 500ms [src](https://github.com/LLK/scratch-gui/blob/develop/src/containers/stage.jsx#L204)

That timing difference was originally introduced by @rschamp in https://github.com/LLK/scratch-gui/pull/159. Nothing about it was mentioned in the PR, so I think it can safely lowered to 400ms. 

In terms of a distance threshold for drag actions, I cannot find that in scratch-flash, but I'm not the best with that codebase. There could be an explicit threshold, or it could be a combination of other factors.

hedgehog125:
Ok.

PR #1434, opened 2018-02-09
Apply a drag threshold to make clicking sprites easier

paulkaplan:
Use the same drag threshold as from the blocks. It may need to be dialed
in further, but testing will be needed.

### Resolves

Resolvers https://github.com/LLK/scratch-gui/issues/1408, but may need further refinements to the drag threshold to dial it in.

### Proposed Changes

### Reason for Changes

Use a 3 px drag threshold before starting a drag on sprites. This gives a bit of breathing space for an unintentional movement while the mouse/touchpoint is down.

### Test Coverage

This will require a fair bit of manual testing on different devices / pointers. I'll mark it "needs-qa" so we can make sure to dial it in.

### DIFF
--- a/src/containers/stage.jsx
+++ b/src/containers/stage.jsx
@@ -15,6 +15,7 @@ import {
 } from '../reducers/color-picker';
 
 const colorPickerRadius = 20;
+const dragThreshold = 3; // Same as the block drag threshold
 
 class Stage extends React.Component {
     constructor (props) {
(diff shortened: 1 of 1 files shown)


hedgehog125:
What if you wanted to move a sprite less than 3 pixels though?

towerofnix:
@hedgehog125 You could start by moving the sprite more than 3 pixels, then move it backwards to the desired position.. or you could just use the set/change x/y blocks :)

hedgehog125:
@towerofnix Oh, ok. I didn't think of that.

towerofnix:
This is our new favorite commit ever.

We're implementing sprite drag and drop for Leopard: https://github.com/leopard-js/leopard/pull/201 - modeled right after Scratch's code/behavior, since we try to have equal runtime mechanics.

Scratch runs this "should we start dragging because we moved the mouse past a certain threshold?" operation *every frame* that the mouse is down and we're not currently in a drag. If the condition is met (your mouse is far from where you pressed it down), then it will attempt to start a drag... *every frame...* based on *where you pressed it down.*

This means you can press down where a draggable sprite is *going* to end up, move the mouse away, dance it around in place, and teleport the draggable sprite to the mouse once it comes by.

https://github.com/scratchfoundation/scratch-gui/assets/9948030/1952f431-7a2d-49d8-b209-58c2ba363c68

Pretty beautiful. LOL

Issue #1212, opened 2018-01-09
New "Make a Block" in palette appears off screen


griffpatch:
### Expected Behavior

When you "Make a Block", the new block is added to the left hand blocks palette and is easily visible.

### Actual Behavior

The new block is added to the left hand blocks palette, but is off the bottom of the screen.

![image](https://user-images.githubusercontent.com/6737342/34710903-0d912a8e-f515-11e7-989f-5259aa8858ac.png)

I wonder if a the blocks palette should have 2/3rds of a screen of blank scrollable area added to the bottom such that when a specific block category is selected, it's relevant blocks can be fully scrolled to the top of the blocks palette? Otherwise the highlighted blocks section is also never correct (see how the Operators category is currently highlighted in the screenshot? I wanted to be on the My Blocks one, but you can't select it - Adding dead space at the bottom of the blocks palette would allow me to scroll them such that you could select the category and keep it selected.)

### Steps to Reproduce

* Scroll to the 'My Blocks' category
* Click "Make a Block"
* Click "OK"

### Operating System and Browser

Windows PC running Chrome

ivanixgames:
Submitted PR https://github.com/LLK/scratch-gui/pull/1439

PR #1439, opened 2018-02-10
Proposed fix for Issue 1212

ivanixgames:
### Resolves
https://github.com/LLK/scratch-gui/issues/1212

### Proposed Changes

Scrolls to 'My Blocks' section after creating new block

Because potential conflict with localization in the future,
I select the last name in the toolbox category list.
Hopefully, 'My Blocks' will always remain the last item.
Otherwise, we could simply call scrollToCategoryByName('My Blocks')?

### Reason for Changes
Resolve issue posted.

### Test Coverage

Manually tested on local build and test server using 'npm start'
Two test cases:
1 - Create new block after clicking on 'My Block' category
2 - Create new block after clicking on higher category ('Operators') and then scrolling down and clicking 'Make a Block'

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -251,8 +251,15 @@ class Blocks extends React.Component {
         this.setState({prompt: null});
     }
     handleCustomProceduresClose (data) {
+
         this.props.onRequestCloseCustomProcedures(data);
-        this.workspace.refreshToolboxSelection_();
+        const ws = this.workspace;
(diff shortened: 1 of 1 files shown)


src/containers/blocks.jsx
```diff
@@ -251,8 +251,15 @@ class Blocks extends React.Component {
         this.setState({prompt: null});
     }
     handleCustomProceduresClose (data) {
+
```
towerofnix: Unnecessary newline, presumably?

PR #1448, opened 2018-02-13
Feature/untangle redux

### LINKED ISSUE #1336: Export for use on scratch-www
In order to use the GUI on scratch-www we need to make some things importable:
 - Combined GUI reducer
 - scratch-paint reducer, separately from the GUI reducer
   - Since scratch-paint expects its state to be under `state.scratchPaint`, it must be combined with the gui reducer separately. Otherwise, when scratch-gui is embedded into scratch-www, the scratchPaint state will be under `state.gui.scratchPaint`, so things will break
   - We want to export this as a part of scratch-gui so that scratch-www doesn't need scratch-paint as a separate dependency
   - We will need to re-combine the `gui` and `scratchPaint` reducers for the GUI "playground" in this repo, which can serve as an example of how to combine the reducers in www
 - The GUI component

chrisgarrity:
We should only merge one of the two export PRs. This one or https://github.com/LLK/scratch-gui/pull/1425

### Resolves

Alternate way to fix #1336. Exports GUI as a library. Users of GUI will need to provide the Intl prop if they want to change localization. All of the rest of GUI's state is self-contained.

### Proposed Changes

Moves standalone GUI build to `playground`. 

### Test Coverage

Current tests pass.


chrisgarrity:
@rschamp I tried moving the ErrorBoundary out of AppStateHOC and wrapping it around GUI in the playground after this commit, but it broke the IE modal...

src/playground/index.jsx
```diff
@@ -18,7 +18,13 @@ if (process.env.NODE_ENV === 'production' && typeof window === 'object') {
 // Register "base" page view
 analytics.pageview('/');
 
-const App = AppStateHOC(ProjectLoaderHOC(GUI));
+const ErrorWrappedGUI = props => (
+    <ErrorBoundary>
+        <GUI {...props} />
+    </ErrorBoundary>
```
rschamp: Maybe this should be a HOC as well?

test/integration/costumes.test.js
```diff
@@ -27,7 +27,7 @@ describe('Working with costumes', () => {
         await loadUri(uri);
         await clickXpath('//button[@title="tryit"]');
         await clickText('Costumes');
-        await clickXpath('//button[@title="Add Costume"]');
+        await clickXpath('//button[@title="Add Costume From Library"]');
```
rschamp: Interesting that this was necessary... 🤔

src/lib/app-state-hoc.jsx
```diff
@@ -12,25 +14,37 @@ const enhancer = composeEnhancers(
...
-                <ErrorBoundary>
-                    <WrappedComponent {...props} />
-                </ErrorBoundary>
-            </IntlProvider>
-        </Provider>
-    );
+    class AppStateWrapper extends React.Component {
+        constructor (props) {
+            super(props);
+            this.store = createStore(reducer, (props.intl || intlInitialState), enhancer);
```
rschamp: We should add some explanation about how this works.  Maybe in the doc comment? 
```javascript
 /*
  * Higher Order Component to provide redux state.  If an `intl` prop is provided, it will override the internal
  * `intl` redux state.
  * @param {React.Component} WrappedComponent - component to provide state for
  * @returns {React.Component} component with redux and intl state provided
  */
```

rschamp:
I don't think I understand the operation that broke the modal — let's look at it in person tomorrow.

rschamp:
@chrisgarrity Looks like there are still linting errors here.

chrisgarrity:
@rschamp ready to re-review (you can try it at https://llk.github.io/scratch-gui/revised-webpack/index.html)

package.json
```diff
@@ -2,14 +2,14 @@
...
-  "main": "./src/index.js",
+  "main": "./dist/scratch-gui.js",
   "scripts": {
     "build": "npm run clean && webpack --progress --colors --bail",
-    "clean": "rimraf ./build && mkdirp build",
+    "clean": "rimraf ./build && mkdirp build && rimraf ./dist && mkdirp dist",
     "deploy": "touch build/.nojekyll && gh-pages -t -d build -m \"Build for $(git log --pretty=format:%H -n1)\"",
     "i18n:src": "babel src > tmp.js && rimraf tmp.js && build-i18n-src ./translations/messages/ ./translations/",
     "start": "webpack-dev-server",
-    "test": "npm run test:lint && npm run test:unit && NODE_ENV=production npm run build && npm run test:integration",
```
rschamp: Why is this?

rschamp: Never mind, I forgot about how we set up the multiple targets for webpack

src/components/browser-modal/browser-modal.jsx
```diff
@@ -43,34 +43,18 @@ const BrowserModal = ({intl, ...props}) => (
                     className={styles.backButton}
                     onClick={props.onBack}
                 >
-                    <FormattedMessage
-                        defaultMessage="Back"
-                        description="Label for button go back when browser is unsupported"
-                        id="gui.unsupportedBrowser.back"
-                    />
+                    Back
```
chrisgarrity: This change is not making sense to me. In my version of `browser-modal` I have the intl and FormattedMessage parts....

Issue #1409, opened 2018-02-05
List monitor should have same color as list blocks


kchadha:
### Expected Behavior

Title says it all.

### Actual Behavior

List monitor background color is the same as variable blocks.

### Steps to Reproduce

1. Create a new project
2. Create a new list
3. Observe the monitor color.

The fix for this will involve 
1. updating the color list to include a list category here https://github.com/LLK/scratch-gui/blob/develop/src/components/monitor/monitor.jsx#L7-L13
2. using the list category in the opcode label code https://github.com/LLK/scratch-gui/blob/develop/src/lib/opcode-labels.js#L40

bartljak:
I submitted a pull request which has now been merged hence fixing this issue. Could someone close this issue?

kchadha:
Thanks for the fix!

PR #1449, opened 2018-02-13
Fixed issue #1409

bartljak:
### Resolves

Issue #1409 (https://github.com/LLK/scratch-gui/issues/1409)

### Proposed Changes

List monitor now is the same color as list blocks.

### Reason for Changes

Color for list monitor was the same as variable monitors before, this makes it easier to differentiate from lists and variables. 

### Test Coverage

All npm tests are passed.


paulkaplan:
@bartljak LG thanks!

PR #1455, opened 2018-02-16
Move active tab state into a reducer

paulkaplan:
This is a functionality-less refactor to promote the active tab state to a navigation reducer in order to allow other buttons to shift the tab focus (i.e. the "blank sprite" button needs to move to the costume editor).


src/components/gui/gui.jsx
```diff
@@ -35,16 +35,19 @@ const messages = defineMessages({
...
         enableExtensions,
         intl,
         feedbackFormVisible,
         vm,
         previewInfoVisible,
+        soundsTabVisible,
         onExtensionButtonClick,
-        onTabSelect,
-        tabIndex,
+        onActivateTab,
```
chrisgarrity: I know we removed the sort-prop-types lint rule, but I think this could still have a better order. How about moving all the `...Visible` props to just before the `on...` handler props, with the remaining props alphabetical? Alternatively let's fix the alphabetization (`intl` and `vm` are incorrectly sorted)

src/reducers/gui.js
```diff
@@ -9,6 +9,7 @@ import targetReducer from './targets';
 import toolboxReducer from './toolbox';
 import vmReducer from './vm';
 import stageSizeReducer from './stage-size';
+import navigationReducer from './navigation';
```
chrisgarrity: Wondering if this would ever be confused with overall navigation? i.e. usually the top bar on the website is the navigation bar (includes login etc). Perhaps it would be better named `tabsNavigationReducer`

PR #1468, opened 2018-02-20
Block drag

fsih:
### Resolves

Depends on https://github.com/LLK/scratch-vm/pull/939
Implement share the love UI.

### Proposed Changes
- Pass the hovered sprite in the target pane to the VM so that it knows what to do with end drag events. This is the mouse over and mouse out changes.
- Receive state about when blocks are being dragged outside of the blocks workspace, and animate the sprite when they are dragged over a sprite.

blockdragwiggle


mrjacobbloom:
I love the lil wiggle

hedgehog125:
Maybe the scripts should also shrink out and have the "whoosh" sound effect. Like when you delete blocks.

kyleplo:
Maybe there should be a little tooltip?
image

hedgehog125:
@kyleplo yes, that would be useful.

paulkaplan:
A few things:
1. there is a react prop error `Warning: Unknown event handler property `onBlockDragUpdate`. It will be ignored.`
2. I think we should have at least a minimal "confirm" animation/transition, even if it isn't fully spec'd. Right now it seems broken.
3. I am seeing an issue where grabbing and dragging on one side of a block doesn't get the wiggling to happen?

wiggle

carljbowman:
Wow! So awesome to have this feature so close. 

> I think we should have at least a minimal "confirm" animation/transition, even if it isn't fully spec'd. Right now it seems broken.

I agree with @paulkaplan. Are we holding off on the blocks animation due to perf?

If that's the case, we should have some feedback from the Sprite tile. Happy to pair on something.

mrjacobbloom:
I love the indication that other sprites can accept the love, that's a huge improvement over 2.0, but I'm not sure that it's obvious whether you're copying the blocks or moving them from one sprite to another... I think I already suggested this somewhere, but maybe when the blocks are dragged outside the workspace they could also appear semi-opaque in the workspace so you know they'll all be there when you let go?

carljbowman:
@mrjacobbloom - Totally agree. This is something that we are considering for the future, but would require some addtional work. Rather than blocking the Share the Love feature, we wanted to get something out. 

Def a great idea to iterate on in the future.

fsih:
@carljbowman I left off the ending animation just because we didn't finish deciding what it would be. I think you were going to experiment with it looking like generic blocks shrinking into the tile? Should I put in the blocks shrinking, which has the flashing issue?

fsih:
Fixed all but the drag sometimes not working.

paulkaplan:
For posterity, from a convo @fsih and I had offline, the flaky behavior of the hovering was due to using "mouseOut" and "mouseOver" instead of "mouseEnter" "mouseLeave", which is almost always what you really want (mouseOut/Over is triggered when going from parent to child as well as from outside to parent, which is... bad).

paulkaplan:
Once scratch-vm is updated to the most recent version in the package json this looks good!

Issue #923, opened 2017-11-20
Missing context menu option to duplicate a costume


ericrosenbaum:
### Expected Behavior

When you right-click a costume, the context menu should include an option to duplicate it.

### Actual Behavior

Can't duplicate it!

davidaylaian:
I added the context menu item in [this commit](https://github.com/davidaylaian/scratch-gui/commit/6c015d365cbf6f12d3bc7f7258260bdcfa560ee1), however it uses the duplicateCostume and duplicateSound scratchvm functions, which don't exist yet.

thisandagain:
Note to self: Needed for testing and workshops (CSF)

PR #1490, opened 2018-02-23
Duplicate costume

fsih:
Fixes https://github.com/LLK/scratch-gui/issues/923

Implement duplicate sound and costume. This required pulling in some sound updating code from Karishma's save/load branch.

### DIFF
--- a/src/components/asset-panel/selector.jsx
+++ b/src/components/asset-panel/selector.jsx
@@ -13,6 +13,7 @@ const Selector = props => {
         items,
         selectedItemIndex,
         onDeleteClick,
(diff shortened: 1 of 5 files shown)


paulkaplan:
@fsih this looks good! One thing is that with new GUI actions like this, it is good to add an integration test for them. I added one for duplicating sounds and one for duplicating costumes and pushed that latest commit to the branch (just b/c we want to merge/test this today). This will probably fail until the VM is updated to have the duplicate costume changes, so I'll push another commit with the package json update in a moment as well.

PR #1495, opened 2018-02-25
Update make-toolbox-xml.js

### LINKED ISSUE #1494: How come you can show and hide the backdrop?
### Expected Behavior

Show and hide blocks are sprite-only.

### Actual Behavior

Show and hide blocks show in the stage.

kyleplo:
### Resolves

#1494 

### Proposed Changes

Removes show and hide from stage toolbox.

### Reason for Changes

You can't show or hide the stage (and the blocks don't work anyway)

### DIFF
--- a/src/lib/make-toolbox-xml.js
+++ b/src/lib/make-toolbox-xml.js
@@ -234,9 +234,6 @@ const looks = function (isStage, targetId) {
         </block>
         <block type="looks_cleargraphiceffects"/>
         ${blockSeparator}
-        <block type="looks_show"/>
-        <block type="looks_hide"/>
-        ${blockSeparator}
         ${isStage ? '' : `
             <block type="looks_gotofrontback"/>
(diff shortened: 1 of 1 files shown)


src/lib/make-toolbox-xml.js
```diff
@@ -247,6 +244,8 @@ const looks = function (isStage, targetId) {
                 </value>
             </block>
             ${blockSeparator}
+        <block type="looks_show"/>
```
paulkaplan: These should go above the `looks_gotofrontback` block to maintain the current block ordering. Make sure to keep the a block separator between the show/hide and the layering blocks. 

Also, the indentation is wrong, it should be aligned with the `looks_gotofrontback` block since it is part of the same  indentation level.

kyleplo: Okay, I changed it.

Issue #1234, opened 2018-01-11
Implement "share the love"


kyleplo:
### Expected Behavior

When you drag a block over a sprite, it will add to that sprite

### Actual Behavior

Nothing happens
### Steps to Reproduce

Drag any block over the stage or other sprite

hedgehog125:
I don't think the alternative would work that well, as it would make moving assets between sprites a lot slower because you would have to put it into your backpack (which might already have a lot in) then take it out in the other sprite and then delete it from the backpack.

thisandagain:
Thanks for filing this @kyleplo. We call this feature "share the love". We currently have it scheduled for completion in February.

towerofnix:
Note that share the love also works with sounds and costumes (even dragging sprite costumes to the stage, and stage backdrops to sprites).

fsih:
Thanks @towerofnix! This issue will just be for the feature where you drag blocks to other sprites. We will implement sharing costumes and sound later. I've filed https://github.com/LLK/scratch-gui/issues/1470 for it.

Issue #1390, opened 2018-02-01
Show message for why 1.4 projects do not load


paulkaplan:
### Expected Behavior

_Please describe what should happen_

Trying to load a 1.4 project should not crash the editor, but it should explain that those projects are net yet supported. 

### Actual Behavior

_Describe what actually happens_

Editor crashes without explaining why.

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_

1. Go to a 1.4 project in the preview like https://llk.github.io/scratch-gui/develop/#1234

/cc @kchadha. @thisandagain @carljbowman how should we handle this before 1.4 support is available.

To provide some data: about 1/3rd of our hard-crash events reported are coming from trying to load 1.4 projects.

PR #1516, opened 2018-02-27
Nicer crash message format

### LINKED ISSUE #1205: What should GUI display if 'something terrible' happens
To handle the WebGL errors, I've added an [error boundary](https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html). If the failure is WebGL it'll show the missing WebGL modal, but what should it display if something unexpected fails.

chrisgarrity:
### Resolves

Fixes #1205.

### Proposed Changes

Adds crash-message component to render nicer message with css when unhandled errors occur. Text is not translated to avoid cascading errors.

### Test Coverage

Normal tests run.

Temporarily changed GUI component to throw an error to generate an unhandled error. Screenshot of new message:
screen shot 2018-02-26 at 6 37 31 pm

Try to load a 1.4 project e.g. `#1234`, to generate a crash.


kchadha:
@chrisgarrity This looks awesome!

Should we add additional messaging here around missing features? For example: #1390

chrisgarrity:
@kchadha I don't think we should add 1.4 messaging to this particular screen as it suggests they `reload` the page, and reloading a 1.4 project is not going to help. However, it looks pretty easy to detect the 1.4 unhandled error in the error boundary `Uncaught SyntaxError: Unexpected token S in JSON at position 0` and show a different message with a `back` button instead.

src/containers/error-boundary.jsx
```diff
@@ -28,21 +29,16 @@ class ErrorBoundary extends React.Component {
         window.history.back();
     }
 
+    handleReload () {
+        window.location.reload();
```
kchadha: Instead of calling this here, you can call `window.location.replace(window.location.origin)` to go back to the empty project url.

chrisgarrity:
@kchadha made that change

Issue #1403, opened 2018-02-04
Sprite selector - scroll bar not visible


morantsur:
On small screens, or when there are many sprites, no scrollbar is visible, but you can still scroll with the mouse - in our playtests, kids couldn't figure out how to get to the second row of sprites. 

![image](https://user-images.githubusercontent.com/15109618/35776704-1c666fd4-09aa-11e8-9407-82724972c5bd.png)

When you scroll with the mouse it shows up for a few seconds and disappears.
![image](https://user-images.githubusercontent.com/15109618/35776746-f2b48756-09aa-11e8-82da-70f885763f6d.png)

### Expected Behavior
Scroll bar visible whenever the sprite selector area is too small?

PR #1520, opened 2018-02-27
Add constraint on min height for large stage

morantsur:
### Proposed Changes

Take screen height into account when deciding on stage size (full/small), instead of just screen width. 

Not sure if that's the right solution - let me know if you have other ideas. Also not sure what the min size should be - in this case I took a little over 768px, which is the same as the width (1096, as a little over 1024) - would love your opinion. 

### Reason for Changes

Should help with #1403 - after discussing with @rschamp and @ericrosenbaum, we think that maybe the reason you can't see enough of the sprite area is because the screen width is bigger than the min supported size for the big stage, but their height isn't (For example, 11in chromebook with a screen resolution of 1366x768)

It doesn't solve the issue with the missing scroll bar, but should at least make the stage smaller which will make it less problematic. 

### Test Coverage

Can add a test if we decide it's a good solution :)

### DIFF
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -144,16 +144,21 @@ const GUIComponent = props => {
                         </Box>
                         <Box className={styles.stageWrapper}>
                             {/* eslint-disable arrow-body-style */}
-                            <MediaQuery minWidth={layout.fullSizeMinWidth}>{isFullSize => {
-                                return isRendererSupported ? (
-                                    <Stage
-                                        height={isFullSize ? layout.fullStageHeight : layout.smallerStageHeight}
-                                        shrink={0}
-                                        vm={vm}
-                                        width={isFullSize ? layout.fullStageWidth : layout.smallerStageWidth}
-                                    />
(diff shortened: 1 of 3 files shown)


paulkaplan:
@carljbowman you may want to take a look at this? I don't have the context for it, but @rschamp / @ericrosenbaum might be able to talk through it?

carljbowman:
@morantsur - Thanks for raising this issue. 

TLDR: Our current media queries are slightly broken and need some more consideration before adding this addtional of behavior. 

Overall I think this is heading in the right direction, but we are a little hesitant to pull this in right now. The current system of media queries can break and need some more tuning. Building on that system might cause more breaking in unexpected ways.

Another factor: the touching-color feature puts some constraints on how we can scale the stage. We need to intentionally pick some tiers that avoid creating sub-pixels which could have dramatic impacts on touching-color.

With those things in mind the proposed direction is as follows. We would like to try would be to put energy into the Stage Size Toggle UI (up next to fullscreen). Then on page load use media-queries to set these toggles.

PR #1532, opened 2018-02-28
Always report mouse-up to VM and send flag indicating end of drag

kchadha:
### Resolves

Resolves LLK/scratch-vm#940

### Proposed Changes

Reports mouse-up data to VM even in the case of ending a drag. Also sends VM flag indicating whether this mouse-up was the end of a drag (which the VM should then use so that the `when this sprite clicked` block doesn't get activated on a drag: LLK/scratch-vm#957).

### Reason for Changes

Compatibility bugfix.

### Test Coverage

Existing tests pass.

#### Note: This PR depends on LLK/scratch-vm#957

### DIFF
--- a/src/containers/stage.jsx
+++ b/src/containers/stage.jsx
@@ -188,16 +188,15 @@ class Stage extends React.Component {
         });
         if (this.state.isDragging) {
             this.onStopDrag();
-        } else {
-            const data = {
-                isDown: false,
-                x: x - this.rect.left,
-                y: y - this.rect.top,
-                canvasWidth: this.rect.width,
-                canvasHeight: this.rect.height
-            };
-            this.props.vm.postIOData('mouse', data);
         }
+        const data = {
+            isDown: false,
+            x: x - this.rect.left,
+            y: y - this.rect.top,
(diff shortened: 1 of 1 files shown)


src/containers/stage.jsx
```diff
@@ -186,18 +186,18 @@ class Stage extends React.Component {
...
-        } else {
-            const data = {
-                isDown: false,
-                x: x - this.rect.left,
-                y: y - this.rect.top,
-                canvasWidth: this.rect.width,
-                canvasHeight: this.rect.height
-            };
-            this.props.vm.postIOData('mouse', data);
+            data.wasDragged = true;
```
rschamp: This might be simpler to read (and more intuitive with people's understanding of `const`) if it was just part of the `data` object:
```javascript
        const data = {
            isDown: false,
            x: x - this.rect.left,
            y: y - this.rect.top,
            canvasWidth: this.rect.width,
            canvasHeight: this.rect.height,
            wasDragged: this.state.isDragging
        };
```

kchadha:
@rschamp Thanks for the review! I made the change you suggested. I'll merge after the build finishes

Issue #1475, opened 2018-02-21
Backdrop should automatically switch after adding new backdrop


fsih:
### Expected Behavior

Like sounds and sprites, the new backdrop should become the selected backdrop

### Actual Behavior

Backdrop doesn't switch

### Steps to Reproduce

Add new backdrop from library

### Operating System and Browser

_e.g. Mac OS 10.11.6 Safari 10.0_

towerofnix:
Hmm... I think this already works? When I add a new backdrop, the stage switches to that backdrop. I'm on Firefox 59.0b11, for what it's relevant.

kyleplo:
Also, do you think it would be useful to delete the blank default backdrop if the user doesn't paint on it and they add a new one?

fsih:
Fixed except for add surprise background and add paint background from the stage selector

PR #1547, opened 2018-03-02
Move backdrop library back to costumes tab

fsih:
Move the backdrop library back in to costume-tab. This fixes https://github.com/LLK/scratch-gui/issues/1475 for all but adding a surprise or paint backdrop from the info pane (Fixes adding a backdrop from library from the costume tab, adding a surprise/paint backdrop from the costume tab, and adding a backdrop from library from the info pane)

Fixing surprise backdrop from the info pane is more complex so I plan to make another PR.

### DIFF
--- a/src/index.ejs
+++ b/src/index.ejs
@@ -9,7 +9,7 @@
     <!-- Sentry error logging to help with finding bugs -->
     <script src="https://cdn.ravenjs.com/3.22.1/raven.min.js" crossorigin="anonymous"></script>
     <script>
-        Raven.config('https://42b7d13da8ad4d68b13e57c5e54f9a23@sentry.io/273218').install();
+        // Raven.config('https://42b7d13da8ad4d68b13e57c5e54f9a23@sentry.io/273218').install();
     </script>
     <!-- /Sentry -->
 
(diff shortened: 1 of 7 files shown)


thisandagain:
@fsih Can you please investigate the CI failure on this PR?

src/index.ejs
```diff
@@ -9,7 +9,7 @@
     <!-- Sentry error logging to help with finding bugs -->
     <script src="https://cdn.ravenjs.com/3.22.1/raven.min.js" crossorigin="anonymous"></script>
     <script>
-        Raven.config('https://42b7d13da8ad4d68b13e57c5e54f9a23@sentry.io/273218').install();
+        // Raven.config('https://42b7d13da8ad4d68b13e57c5e54f9a23@sentry.io/273218').install();
```
paulkaplan: Probably do not want to commit this.

fsih:
The tests seem to pass sometimes and fail other times for me locally.

paulkaplan:
@fsih this is not a flaky test, the test is failing for a good reason. Clicking the "add backdrop" button without being on the costumes tab does not actually work. The reason is that because the costume library is a child of the costumes tab, and the costumes tab is not rendered when it is not visible, adding a backdrop does not work without the costumes tab being selected first.

fsih:
That's so strange, the integration test has been working consistently for me locally. But yes, that sounds like the right thing to fix, thanks!

fsih:
I have changed the backdrop library button to switch you to the costumes tab. This better matches the behavior of Scratch 2.

fsih:
I realized some weird behavior caused by this change :/
If you open the backdrop library and then press 'X' to quit, you are still in the costumes tab

PR #1556, opened 2018-03-05
Fixes the add backdrop integration test

fsih:
Fixes the integration test that is flaky.


paulkaplan:
As i mentioned in https://github.com/LLK/scratch-gui/pull/1547, we need to do some more research into the cause, this doesn't fix the cause.

PR #1582, opened 2018-03-07
Costume tab detects added costume and switches to the current costume of the edit target

fsih:
Fixes https://github.com/LLK/scratch-gui/issues/1475

### DIFF
--- a/src/containers/costume-tab.jsx
+++ b/src/containers/costume-tab.jsx
@@ -67,7 +67,6 @@ class CostumeTab extends React.Component {
             'handleSelectCostume',
             'handleDeleteCostume',
             'handleDuplicateCostume',
-            'handleNewCostume',
             'handleNewBlankCostume',
             'handleSurpriseCostume',
             'handleSurpriseBackdrop'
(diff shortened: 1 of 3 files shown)


src/containers/costume-tab.jsx
```diff
@@ -96,10 +96,23 @@ class CostumeTab extends React.Component {
...
-        if (this.props.editingTarget !== editingTarget) {
+        if (this.props.editingTarget === editingTarget) {
+            // Switch to a newly added costume if there is one
+            const oldTarget = this.props.sprites[editingTarget] ?
+                this.props.sprites[editingTarget] : this.props.stage;
+            // @todo: Check that the costume is actually new by making sure it doesn't
+            // exist in the old costume list. This is blocked by
+            // https://github.com/LLK/scratch-vm/issues/967
+            if (oldTarget.costumeCount < target.costumeCount) {
+                this.setState({selectedCostumeIndex: target.currentCostume});
```
paulkaplan: Is it possible to combine the new costume case with the deleted costume case by just checking if oldTarget.costumeCount !== target.costumeCount, and if that is true just set selected to currentCostume? I think this would reduce the amount of "assuming" we are doing by just always using the target.currentCostume in any case where we need to manually reset the selected costume.

In fact you could switch around these "ifs" and just ask: "is the editing target different or has the costume count changed? If so, switch to the currentCostume.

fsih: Sure, but do you prefer this solution over https://github.com/LLK/scratch-gui/pull/1597 ? I didn't put reviewers on this PR because I wanted to have both to compare, and I think the other change is better overall, since the costume tab is guessing less.

fsih: I discovered an issue with both options ):

If you have the costume changing rapidly in the project (forever: next costume), then the target.currentCostume might not match the newly added costume by the time this code runs. This would also be true for delete costume if I implemented your suggestion; you might not see the index move 1 up like you would expect, but instead see the selected costume change to a random costume.

fsih: Ok, I've integrated your and Ray's suggestions!

thisandagain:
@paulkaplan Can you please re-review this?

PR #1585, opened 2018-03-07
Try using headless chrome for integration tests

paulkaplan:
Following the instructions https://developers.google.com/web/updates/2017/04/headless-chrome

I'm trying to see if this will allow for more stable tests and/or faster integration tests. I'm submitting this PR to test it on travis. It actually works with no additional changes locally, which is kind of remarkable!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ sudo: required
 dist: trusty
 addons:
     chrome: stable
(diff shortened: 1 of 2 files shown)


paulkaplan:
@chrisgarrity would you be able to test this locally? Actually same question to @fsih?

PR #1621, opened 2018-03-15
Load Button Loading/Error State

kchadha:
(a.k.a. Save/Load pt. 2)

### Proposed Changes

This PR is a continuation of the save/load work, and accommodates the changes made in LLK/scratch-vm#979 and LLK/scratch-parser#21. This PR depends on these two sets of changes, and should not be merged until these others have been merged.

Changes in this PR include:
- Hooking up load button to loading state animation
- Adding a filter to the file input (such that we can only load files with specific extensions). '.zip' will be removed when save/load is finalized.
- Add explicit error handling on calls to loadProject: The changes in LLK/scratch-vm#979 make it so that loadProject no longer throws errors, but instead rejects a promise. The changes in this PR catch these rejected promises (e.g. loading an invalid project -- example: project ID 1234) and sets an error state and throws an error directly in the renderer.
- Uses new version of scratch-vm's `loadProject` function to handle both JSON strings (returned by storage) as well as input buffers (from the load-button file input), instead of using `loadProject` for one and `loadProjectLocal` (introduced by #1564) for the other.

### Reason for Changes

Continuing Save/Load work.

### Test Coverage

Existing tests pass.
(Travis build will fail until LLK/scratch-parser#21 and LLK/scratch-vm#979 are merged)


src/components/load-button/load-button.jsx
```diff
@@ -26,6 +26,7 @@ const LoadButtonComponent = ({
             </ButtonComponent>
             <input
                 disabled
+                accept=".sb,.sb2,.sb3,.zip"
```
cwillisf: I'd suggest removing `.zip` and adding `.sbx` (ScratchX project)

src/containers/load-button.jsx
```diff
@@ -13,10 +18,23 @@ class LoadButton extends React.Component {
...
         ]);
+        this.state = {
+            loadingError: false,
+            errorMessage: ''
+        };
     }
     handleChange (e) {
+        this.props.openLoadingState();
+        // Remove the hash if any (without triggering a hash change event or a reload)
+        history.replaceState({}, document.title, '.');
```
rschamp: Things that modify the state of the page should not be buried in individual components, and should instead be handled by redux or a HOC that the "GUI component user" has control over.  I think our best bet atm would be to get this behavior into the `ProjectLoaderHOC` somehow, since that already is handling hash changes.

src/components/load-button/load-button.jsx
kchadha: I removed .zip and .sb (since we can't support it yet). Projects now should save as .sb3 as well. I'm leaving .sbx out since they probably can't be fully supported, but we can always reconsider this later.

kchadha:
@cwillisf, to answer your questions: 

> A few "have you thought about"s, since I can't quite tell the behavior just by reading this:

> If a load fails and then a subsequent load succeeds, does it clear out the error message display?
> Is it possible to start a second load while one is already in progress?

The way this code is set up, it should not be possible to start another load after a load fails. If a load fails, the 'Oops' page shows up, and users have to reload the gui via the reload button on the 'Oops' page..

It should also not be possible to start a second load (_via the load button_) while one is already in progress. Starting a load triggers the loading state animation, and it covers the whole gui, including the menu bar where the save/load buttons are located. It _is_ possible to start a second load while a load is already in progress via the hash in the url. With some minimal testing, this seems to work fine. The second load is the project that loads and there aren't any unexpected errors in the console.

thisandagain:
@kchadha Is this ready for re-review?

src/containers/load-button.jsx
kchadha: @rschamp, I reverted the changes we were working on (as a fix for this). Could you take another look?

PR #1631, opened 2018-03-19
Don't prevent default on keyboard events

fsih:
vm-listener-hoc's prevent default on keyboard events prevents the text tool in the paint editor from working.


fsih:
I traced it back to here:
https://github.com/LLK/scratch-gui/commit/e43653a389006f61cdcd620ed161e98c475b693b#diff-9d5f54ae39f220ba1ec30f2374d08cb3R56

rschamp:
I believe this is what prevents the project from responding when you're typing elsewhere?

fsih:
I tested typing a into various text fields, and the cat didn't turn unless I didn't have any text fields focused. I'm going to put it in and let's watch out for issues.
image

Issue #1552, opened 2018-03-05
Rotate through messages during load sequence


thisandagain:
While loading projects, displaying a single message (e.g. "Loading sprites ...", "Loading costumes ...", etc.) can lead to the perception that the editor has "hung" and gotten stuck. My instinct is that this is reinforced by the behavior of the current Scratch 2.0 loading screen which shows a progress bar.

My recommendation would be for us to rotate through messages every few seconds to better reinforce that the 3.0 editor is "doing something".

PR #1647, opened 2018-03-21
Make the loader message update every 5 seconds

paulkaplan:
### Resolves

Resolvers https://github.com/LLK/scratch-gui/issues/1552

### Proposed Changes

Rotate the loading messages every 5 seconds to make sure people know it is still going!

### Reason for Changes

From @thisandagain 
> While loading projects, displaying a single message (e.g. "Loading sprites ...", "Loading costumes ...", etc.) can lead to the perception that the editor has "hung" and gotten stuck. My instinct is that this is reinforced by the behavior of the current Scratch 2.0 loading screen which shows a progress bar.
> My recommendation would be for us to rotate through messages every few seconds to better reinforce that the 3.0 editor is "doing something".

loading-messages

### DIFF
--- a/src/components/loader/loader.jsx
+++ b/src/components/loader/loader.jsx
@@ -5,140 +5,171 @@ import styles from './loader.css';
 import topBlock from './top-block.svg';
 import middleBlock from './middle-block.svg';
 import bottomBlock from './bottom-block.svg';
-
-const LoaderComponent = () => {
-    const messages = [
-        {
-            message: (
-                <FormattedMessage
-                    defaultMessage="Creating blocks …"
-                    description="One of the loading messages"
-                    id="gui.loader.message1"
-                />
-            ),
-            weight: 50
-        },
-        {
-            message: (
-                <FormattedMessage
-                    defaultMessage="Loading sprites …"
-                    description="One of the loading messages"
-                    id="gui.loader.message2"
-                />
-            ),
-            weight: 50
-        },
-        {
-            message: (
-                <FormattedMessage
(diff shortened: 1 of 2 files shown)


src/components/loader/loader.jsx
```diff
@@ -5,140 +5,171 @@ import styles from './loader.css';
...
+                description="One of the loading messages"
+                id="gui.loader.message7"
+            />
+        ),
+        weight: 1
+    },
+    {
+        message: (
+            <FormattedMessage
+                defaultMessage="Preparing emojiis …"
```
kyleplo: Emojis is spelled wrong

src/components/loader/loader.jsx
```diff
@@ -5,140 +5,171 @@ import styles from './loader.css';
...
+                description="One of the loading messages"
+                id="gui.loader.message3"
+            />
+        ),
+        weight: 50
+    },
+    {
+        message: (
+            <FormattedMessage
+                defaultMessage="Loading extensions …"
```
kyleplo: What if the project doesn't have extensions?

thisandagain:
@paulkaplan Looks like the build failed. I'm attempting to re-run, but regardless you may want to resolve the type @kyleplo caught above.

paulkaplan:
Woops, linting error was breaking the build. Fixed. Also fixed the typo, thanks @kyleplo. In terms of the extensions line, again this loading is temporary since we do not have the ability to get fine grained loading info right now.

BryceLTaylor:
@paulkaplan Could you add some test cases to this issue?

PR #1670, opened 2018-03-26
Enable video sensing extension

mzgoddard:
### Proposed Changes

Enable videoSensing extension with its first incremental implementation from https://github.com/LLK/scratch-vm/pull/1001


thisandagain:
Looks good! We should probably shift the order to be grouped with the other "enabled" extensions.

PR #1692, opened 2018-03-29
Custom procedure enter

### LINKED ISSUE #1649: Pressing <enter> in Custom Procedure modal should create custom procedure
### Expected Behavior

Pressing <enter> on keyboard while in a modal should act as a submit for that modal.

### Actual Behavior

Pressing <enter> key on the keyboard after using the custom procedure modal seems to result in a no-op. This is not consistent with the behavior of the variable/list/broadcast message modals.

### Steps to Reproduce
1. Open a new 3.0 project (preview.scratch.mit.edu)
2. Click on "Make a block" in the toolbar
3. Edit the custom procedure using the modal
4. Press 'enter' on keyboard

kyleplo:
### Resolves

#1649

### Proposed Changes

Adds an onKeyPress event to custom procedure modal, and runs the onOK handler if the key pressed is Enter. Basically, it makes pressing enter in the modal submits the new procedure.

### Reason for Changes

> @kchadha  said "Pressing key on the keyboard after using the custom procedure modal seems to result in a no-op. This is not consistent with the behavior of the variable/list/broadcast message modals."

### DIFF
--- a/src/components/custom-procedures/custom-procedures.jsx
+++ b/src/components/custom-procedures/custom-procedures.jsx
@@ -14,8 +14,8 @@ const CustomProcedures = props => (
     <Modal
         className={styles.modalContent}
         contentLabel="Make a Block"
-        onRequestClose={props.onCancel}
         onKeyPress={props.onKeyPress}
+        onRequestClose={props.onCancel}
     >
         <Box
             className={styles.workspace}
--- a/src/components/custom-procedures/custom-procedures.jsx
+++ b/src/components/custom-procedures/custom-procedures.jsx
@@ -15,7 +15,7 @@ const CustomProcedures = props => (
         className={styles.modalContent}
         contentLabel="Make a Block"
         onRequestClose={props.onCancel}
-        onKeyPress={porps.onKeyPress}
+        onKeyPress={props.onKeyPress}
     >
         <Box
             className={styles.workspace}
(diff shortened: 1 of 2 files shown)


src/components/custom-procedures/custom-procedures.jsx
```diff
@@ -14,6 +14,7 @@ const CustomProcedures = props => (
     <Modal
         className={styles.modalContent}
         contentLabel="Make a Block"
+        onKeyPress={props.onKeyPress}
```
paulkaplan: You will need to pass this in from the custom procedure container, right now nothing is being passed for `onKeyPress`. Should be added around line 110 of src/containers/custom-procedures.jsx

kyleplo: Should work now

src/containers/custom-procedures.jsx
```diff
@@ -109,6 +113,7 @@ class CustomProcedures extends React.Component {
                 onAddLabel={this.handleAddLabel}
                 onAddTextNumber={this.handleAddTextNumber}
                 onCancel={this.handleCancel}
+                onKeyPress={this.handleOk}
```
paulkaplan: This will submit the modal any time a key is pressed, definitely not what you want. I think this line should be `onKeyPress={this.handleKeyPress}`

kyleplo:
I'm a bit confused. Do I still have changes to make?

paulkaplan:
@kyleplo have you tested this locally? It doesn't seem to work for me. I don't think there is anything handling `onKeyPress` for the `Modal` (which is a custom component, not a native DOM element, so it can't just have `onKeyPress` added to it).

kyleplo:
@paulkaplan I don't have a way to local test. (I'm not experienced with Node.js) If you want, just close this and let someone else fix this.

PR #1718, opened 2018-04-05
Store sample rate/count for recorded sounds

### LINKED ISSUE #1707: First recorded sound is named "recording 0" when there are no other sounds
### Expected Behavior

Generally we use 1-indexing instead of 0-indexing, so nothing that Scratch automatically numbers in the UI should start with 0. 

### Actual Behavior

If you record a new sound in a sprite that has no sounds, the new recording is named "recording 0".

### Steps to Reproduce

- Delete any sounds in the sprite
- Record a new sound

kchadha:
### Resolves

Resolves #1707 
Towards more easily allowing adding metadata to sounds in the sound editor tab.

### Proposed Changes

Changes include:

- Add sample rate and sample count to sound object being added to the vm when a sound is recorded. This data is already provided for library sounds and for sounds imported from sb2 projects, so this change standardizes the availability of these properties across all sounds.

- Update the name property of the sound object to just be 'recording 1'. The VM already assigns fresh names to sounds, costumes, and sprites to remove duplication, so it will handle renaming the sound so that it doesn't overlap with other sound names for the given sprite. This change resolves #1707.

Note: these changes need to go in before the save-load work, so that the sb3 validation can correctly check that all given sounds have the rate and sampleCount properties provided.

### Test Coverage

Existing tests pass.

### DIFF
--- a/src/containers/record-modal.jsx
+++ b/src/containers/record-modal.jsx
@@ -77,8 +77,7 @@ class RecordModal extends React.Component {
                     format: '',
                     dataFormat: 'wav',
                     rate: this.state.sampleRate,
-                    sampleCount: clippedSamples.length,
-                    name: `recording ${this.props.vm.editingTarget.sprite.sounds.length}`
+                    sampleCount: clippedSamples.length
                 };
 
                 // Load the encoded .wav into the storage cache and get resulting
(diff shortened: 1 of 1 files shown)


src/containers/record-modal.jsx
```diff
@@ -90,6 +91,9 @@ class RecordModal extends React.Component {
 
                 // update vmSound object with md5 property
                 vmSound.md5 = `${md5}.${vmSound.dataFormat}`;
+                // The VM will update the sound name to a fresh name
+                // if the following is already taken
+                vmSound.name = 'recording 1';
```
paulkaplan: It looks like we generally give names without spaces between word/number, and our automatic renamer doesn't handle whitespace (i.e. recording 1 and recording1 are distinct names)

kchadha:
@paulkaplan Removed the whitespace from the name.

Issue #1722, opened 2018-04-05
Trailing slash after project ID prevents at least one project from loading


ericrosenbaum:
### Expected Behavior

It should be possible to use a URL pattern like in Scratch 2, with a trailing slash.

### Actual Behavior

In some cases, a URL with a trailing slash prevents the project from loading with `No 'Access-Control-Allow-Origin'` error.

### Steps to Reproduce

compare:
https://llk.github.io/scratch-gui/develop/#214170924
https://llk.github.io/scratch-gui/develop/#214170924/

kchadha:
@cwillisf This seems like a really weird bug.
One solution would be to pattern match on the hash.

Below is just some observations I made about the second project listed above

In the second project, it seems to actually fetch the right project from the server (somehow), but as an .sb2 archive instead of the project json (thus the problem lies in [this line of code](https://github.com/LLK/scratch-gui/blob/develop/src/lib/project-loader-hoc.jsx#L34), where we're assuming we get a buffer representing a json file and we're hoping to turn it into a json string). The good news there is that we can now remove the toString call because the vm/scratch-parser now does the right thing with zip files!

My questions are, (1) why does fetching a projectId with `/` at the end behave differently on different projects (e.g. the two types of errors @ericrosenbaum mentioned above), (2) why does putting a `/` at the end for the second project mentioned above suddenly return the .sb2 file representing that project instead of the project json...

@colbygk, maybe you know something about this as well..?

PR #1727, opened 2018-04-05
Remove toString call in the project-loader.

kchadha:
### Proposed Changes

No longer need to convert the data from the loaded project to a string. This was also causing an error in the issue mentioned below.

### Reason for Changes

Resolves part of #1722 (namely the second project mentioned in that issue). Also for ergonomics; the toString call is no longer necessary now that vm/parser can load .sb2 files.

### Test Coverage

Existing tests pass.


ericrosenbaum:
Just a note that this doesn't fully fix the the second example in https://github.com/LLK/scratch-gui/issues/1722 (somehow it loads two different versions of the project??), but @rschamp suggested the trouble may be specific to the github.io environment, so not to worry about it.

PR #1728, opened 2018-04-05
Hook up save load buttons to file/edit options in menu bar

kchadha:
Everything stays disabled and still has tooltips.

### Proposed Changes

Hook up save/load functionality to the file/edit options in the menu bar as a temporary solution while we wait for dropdown menus.

### Reason for Changes

Unblocks save/load progress.

### Test Coverage

Existing tests pass.


rschamp:
We're going to do this with the real options in dropdowns — for the record this works though!

PR #1750, opened 2018-04-09
Fixes for load button issues.

### LINKED ISSUE #1747: Issues with using load button twice
### Expected Behavior

1) Load Button should be able to load the same file twice
2) Cancelling picking a file to load after successfully loading a project via load button should not trigger loading animation.

### Actual Behavior

The opposite of the above.

### Steps to Reproduce

For problem 1 above:
1) Use the load button (on the save-load branch since it is disabled on develop) to load a project.
2) After successfully loading a project, make changes to the project
2) Use the load button to load the same project again.
3) Observe that nothing happens. (The loading state animation should have come up and the project should have been re-loaded, e.g. your changes should be gone)

For problem 2 above:
1) Use the load button (on the save-load branch since it is disabled on develop) to load a project.
2) After successfully loading a project, use the load button to bring up the file browser again, but this time click cancel (or esc).
3) Observe that the loading animation triggers even though we cancelled the load.

kchadha:
### Resolves

Resolves #1747 

### Proposed Changes

Fixes to issues in #1747 that were caused due to the way we were handling changes to the input element.

### Reason for Changes

Issues that came up during smoke testing.

### Test Coverage

Existing tests pass.

### Notes on testing these changes
Since the load button doesn't exist anymore, these changes should be manually tested. In order to test these changes, pull in these changes as well as the changes from this PR #1728.
Since the save/load code being referenced on develop is not the same as that in the save-load branch, you should create your own .sb3 file (via creating your own project from scratch in 3.0 or via importing a 2.0 project) this is because .sb3 files saved from the save-load branch won't load on develop.

### DIFF
--- a/src/containers/load-button.jsx
+++ b/src/containers/load-button.jsx
@@ -24,18 +24,24 @@ class LoadButton extends React.Component {
         };
     }
     handleChange (e) {
-        this.props.openLoadingState();
         // Remove the hash if any (without triggering a hash change event or a reload)
         history.replaceState({}, document.title, '.');
         const reader = new FileReader();
+        const thisFileInput = e.target;
         reader.onload = () => this.props.vm.loadProject(reader.result)
             .then(() => {
                 this.props.closeLoadingState();
+                // Reset the file input after project is loaded
+                // This is necessary in case the user wants to reload a project
+                thisFileInput.value = null;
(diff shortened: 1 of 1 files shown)


paulkaplan:
LG!

Issue #1653, opened 2018-03-21
Can Drag in Fullscreen


amazinigmech2418:
### Expected Behavior

When in fullscreen, you shouldn't be able to drag a sprite.

### Actual Behavior

When in fullscreen, you can drag a sprite.

### Steps to Reproduce

Make a project
Go to fullscreen
Drag a sprite

### Operating System and Browser

Chrome on Windows 10

kyleplo:
It should drag if the user has used the 
![image](https://user-images.githubusercontent.com/31634240/37874191-a3f6c290-2ff8-11e8-8c53-00b8eae98e75.png) block.

unize4:
I also have this problem.
 I solved this problem temporarily by judging the state of the sprite in the 'onStartDrag' function of 'stage.jsx' when in fullscreen.
like this：
`if (!this.props.vm.editingTarget.draggable) return;`

Issue #1687, opened 2018-03-29
Player mode


rschamp:
The GUI should have the ability to change its appearance to "player" mode, which is similar but distinct from Fullscreen mode.

There should be no blocks palette, menubar, etc. The only visible portion should be the stage, green flag, stop all, and full screen buttons.

Only sprites with the "draggable" property enabled should be draggable.

This mode should be set-able with a prop (`mode=player|fullscreen|editor`? `playerMode={true}`?) on the exported GUI component.

PR #1764, opened 2018-04-12
Add editor drag mode for popping sprites of stage.

paulkaplan:
Also integrate the "ignore draggability, always drag" behavior of editor
mode.

Right now, editor drag mode is hooked up to the "isFullscreen" flag,
since it is the closest thing we have to player mode now.

### Resolves

Resolves https://github.com/LLK/scratch-render/issues/128, implemented in the GUI because (a) we needed the sprite drag out of the stage for the backpack and (b) the shadow is trivial in the GUI (`filter: drop-shadow`) but would be complex in the renderer.

Resolves https://github.com/LLK/scratch-gui/issues/1653

Doesn't create a "player" mode, but introduces a flag that can be used when making player mode, so related to https://github.com/LLK/scratch-gui/issues/1687

### Proposed Changes

Adds a flag to the stage `useEditorDragStyle`, that when enabled changes the drag behavior in the following ways:
- Make the sprite not visible in the renderer while dragging.
- Use the renderer `extractDrawable` function to render the sprite on its own canvas, managed by the GUI.
- Adds a drop shadow to the sprite canvas. 
- Changes the CSS to allow dragging outside the stage. 
- Hooks up the `isFullscreen` flag to also determine whether to use editor drag style (`useEditorDragStyle = !isFullScreen`)
- When not using editor drag style, check targets for draggability before starting a drag on them.

### Test Coverage

@BryceLTaylor when we test this, there are a few things we should make sure to do:
- Make sure not-draggable sprites can be dragged in the editor,  but cannot be dragged when you make it fullscreen (matching scratch2 behavior)
- Sprites can be dragged all around!
- Sprite XY position is still fenced even when dragging (i.e. the x/y inputs should not go way outside stage bounds, and the sprite should pop back to the fenced position when you drop the sprite).
- It should work on touch devices just as well. 

drag-outside

/cc @cwillisf you might want to take a look at the way I'm actually using the data from `extractDrawable`. One hiccup was that to draw image data I needed to create a `Uint8ClampedArray`, the imagedata constructor cannot be used with `Uint8Array` directly. But I used the constructor that reads from the underlying buffer, so it doesn't do a copy and isn't slow. But if you have a different idea about this let me know.

### DIFF
--- a/src/components/stage/stage.css
+++ b/src/components/stage/stage.css
@@ -14,6 +14,9 @@
 
     /* @todo: This is for overriding the value being set somewhere. Where is it being set? */
     background-color: transparent;
+
+    /* Allow custom touch handling to prevent scrolling on Edge */
+    touch-action: none;
 }
 
 .with-color-picker {
--- a/src/components/stage/stage.css
+++ b/src/components/stage/stage.css
@@ -8,6 +8,10 @@
     */
     display: block;
 
+    /* Attach border radius directly to canvas to prevent needing overflow:hidden; */
+    border-radius: $space;
+    border: 1px solid $ui-black-transparent;
+
     /* @todo: This is for overriding the value being set somewhere. Where is it being set? */
     background-color: transparent;
 }
@@ -30,10 +34,6 @@
 
 .stage-wrapper {
     position: relative;
-    border-radius: $space;
-    border: 1px solid $ui-black-transparent;
-    /* Keep the canvas inside the border radius */
-    overflow: hidden;
 }
 
 .stage-wrapper-overlay {
@@ -78,3 +78,11 @@
     pointer-events: none;
     overflow: hidden;
 }
+
+.dragging-sprite {
+    position: absolute;
+    top: 0;
+    left: 0;
+    z-index: 1000; /* Above everything so it is draggable into other panes */
+    filter: drop-shadow(5px 5px 5px $ui-black-transparent);
+ }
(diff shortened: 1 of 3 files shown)


paulkaplan:
oh wait this isn't done, in 2.0 the sprite snaps back to the pre-drag position if it is dropped outside.

cwillisf:
> One hiccup was that to draw image data I needed to create a Uint8ClampedArray

Yeah, unfortunately the WebGL spec demands that it's a `Uint8Array` exactly.

> I used the constructor that reads from the underlying buffer, so it doesn't do a copy

👍 I think that's the right solution. If we end up with several callers that all want `Uint8ClampedArray` then we could build this step into `extractDrawable`.

paulkaplan:
Did some browser compatibility checking around the `ImageData` constructor (h/t @rschamp for that), looks like it is well supported. But that check did reveal some issues with dragging on Edge touch devices (Surface), which I followed up with a fix for.

src/components/stage/stage.css
```diff
@@ -78,3 +81,11 @@
     pointer-events: none;
     overflow: hidden;
 }
+
+.dragging-sprite {
+    position: absolute;
+    top: 0;
+    left: 0;
+    z-index: 1000; /* Above everything so it is draggable into other panes */
```
rschamp: It's always "above everything" until that next thing that wants to be. It would be nice to keep it low enough to not be totally arbitrary.  In scratch-blocks the category menu is 40 so I made the dropdowns use 50... so could you make this 60? Or maybe 100?

PR #1766, opened 2018-04-12
Add File and Edit menu dropdowns

### LINKED ISSUE #1399: 3.0 Toolbar
## 2.0 Overview 
The Scratch 2.0 Toolbar contains lots of various information, of which I have tried to break up into smaller categories: Nav/Menus, Account, and Project. 

These sections of information change relative to the state of a project: Share/Unshared, My Project/Someone Else's Project, and Logged-in/Logged-out. Below are the outlined examples of where these states intersect and which information is shown related to that state.

### Logged-out > Project
logged-out - project

### Logged-out > Someone Else's Project
logged-out - scratcher project

### Logged-in > My Project > Unshared
logged-in - my project - unshared

### Logged-in > Someone Else's Project
logged-in - scratcher project

rschamp:
### Resolves

Resolves #1399

### Proposed Changes

* Add menu component and subcomponents for dropdowns in the menu bar
* Enable Download and Upload items within the File menu
* Cosmetic tweaks to the menu bar, leftover from #1683 

### Reason for Changes

Finish the framework necessary for implementing menubar-related features. Enable local save/load!

### Test Coverage

I'm sorry but I only had just enough time to make it work. I will file an issue.


rschamp:
Demo here: https://rschamp.github.io/scratch-gui/menubar-dropdowns/

src/components/menu-bar/menu-bar.jsx
```diff
@@ -20,91 +32,180 @@ import communityIcon from './icon--see-community.svg';
...
+                            </MenuItemTooltip>
+                        </MenuSection>
+                        <MenuSection>
+                            <LoadButton>{(renderFileInput, loadProject, loadProps) => (
+                                <MenuItem
+                                    onClick={loadProject}
+                                    {...loadProps}
+                                >
+                                    Upload from your computer
+                                    {renderFileInput()}
```
kchadha: This is cool! If I understand it correctly, you're passing in a function as a prop that renders the file input. This might be something I want to do for file and camera upload sounds/costumes/sprites...

rschamp: Yeah — it's an alternative to using a higher order component that eliminates the need for using an intermediary component. With a HOC this could have been passed in as a prop instead, but then you would need to create a component to wrap in the HOC. This way it can be inline.

src/containers/load-button.jsx
```diff
@@ -49,26 +48,34 @@ class LoadButton extends React.Component {
...
         } = this.props;
-        return (
-            <LoadButtonComponent
-                inputRef={this.setFileInput}
-                onChange={this.handleChange}
-                onClick={this.handleClick}
-                {...props}
-            />
-        );
+        return this.props.children(this.renderFileInput, this.handleClick, props);
```
kchadha: I'm not sure I fully understand what's happening here... is it just a way of passing arguments that the children of this component can use?

rschamp: Comments would have helped, sorry :/  This is the piece that makes the HOC-replacement above work. This component expects its "children" to be a function. Then it calls that function with arguments that would have been props had it been a HOC.  So while it is a component, you can use it to enhance any component you like directly within the render function. With a HOC you have to wrap a component class ahead of time to use it.

src/containers/project-saver.jsx
```diff
@@ -3,7 +3,21 @@ import PropTypes from 'prop-types';
...
+ * The component can then be used to attach project saving functionality
+ * to any other component:
+ *
+ * <ProjectSaver>{(saveProject, props) => (
+ *     <MyCoolComponent
+ *         onClick={saveProject}
+ *         {...props}
+ *     />
+ * )}</ProjectSaver>
+ */
```
kchadha: 👍

rschamp:
@apple502j Good question — I was tempted to enable this but ran into an issue. While we have the ability to toggle turbo mode, the VM does not emit events when turbo mode is toggled. This makes it difficult to make the UI reflect the turbo mode state. So I left that disabled until we have those events in the VM.

PR #1784, opened 2018-04-13
Limit toolbox updates to when we tell it to update

### LINKED ISSUE #901: Don't render the toolbox twice on sprite change
gui/src/containers/blocks.js's onWorkspaceUpdate calls updateToolboxState, and also clearWorkspaceAndLoadFromXml, which rerenders the toolbox. When updateToolboxState is called, blocks.js's componentDidUpdate method calls workspace.updateToolbox, which is unnecessary. We should avoid rerendering the toolbox more often than necessary.

gnarf:
### Resolves

#901

### Proposed Changes

* Force the `toolboxRefreshEnabled` flag on the workspace to always be false (regardless of what `Blockly.Xml.domToWorkspace` says)

* Defer the toolbox update a frame to allow the click handler to return earlier

### Reason for Changes

* Removes approximately 400ms from the click event handler.
* Defers 200ms of it to the "next frame" so it still feels a bit jumpy, but since nothing really changes much you don't notice it.

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -74,7 +74,9 @@ class Blocks extends React.Component {
         // entire toolbox every time we reset the workspace.  We call updateToolbox as a part of
         // componentDidUpdate so the toolbox will still correctly be updated
         this.setToolboxRefreshEnabled = this.workspace.setToolboxRefreshEnabled.bind(this.workspace);
-        this.workspace.setToolboxRefreshEnabled = () => this.setToolboxRefreshEnabled(false);
+        this.workspace.setToolboxRefreshEnabled = () => {
+            this.setToolboxRefreshEnabled(false);
+        };
 
         // @todo change this when blockly supports UI events
         addFunctionListener(this.workspace, 'translate', this.onWorkspaceMetricsChange);
@@ -107,6 +109,11 @@ class Blocks extends React.Component {
             clearTimeout(this.toolboxUpdateTimeout);
             this.toolboxUpdateTimeout = setTimeout(() => {
                 this.workspace.updateToolbox(this.props.toolboxXML);
+                // In order to catch any changes that mutate the toolbox during "normal runtime"
+                // (variable changes/etc), re-enable toolbox refresh.
+                // Using the setter function will rerender the entire toolbox which we just rendered.
+                this.workspace.toolboxRefreshEnabled_ = true;
+
                 const currentCategoryPos = this.workspace.toolbox_.getCategoryPositionByName(categoryName);
                 this.workspace.toolbox_.setFlyoutScrollPos(currentCategoryPos + offset);
             }, 0);
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -218,19 +218,15 @@ class Blocks extends React.Component {
         // Remove and reattach the workspace listener (but allow flyout events)
         this.workspace.removeChangeListener(this.props.vm.blockListener);
         const dom = this.ScratchBlocks.Xml.textToDom(data.xml);
-        // @todo This line rerenders toolbox, and the change in the toolbox XML also rerenders the toolbox.
-        // We should only rerender the toolbox once. See https://github.com/LLK/scratch-gui/issues/901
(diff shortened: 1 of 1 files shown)


gnarf:
```
$ cd llk/scratch-blocks
$ grep -rC8 setToolboxRefreshEnabled core
core/workspace_svg.js-
core/workspace_svg.js-/**
core/workspace_svg.js- * Update whether this workspace has toolbox refreshes enabled.
core/workspace_svg.js- * If enabled, the toolbox will refresh when appropriate.
core/workspace_svg.js- * If disabled, workspace will not refresh until re-enabled.
core/workspace_svg.js- * Use to avoid refreshing during a batch operation, for performance.
core/workspace_svg.js- * @param {boolean} enabled Whether refreshes should be enabled.
core/workspace_svg.js- */
core/workspace_svg.js:Blockly.WorkspaceSvg.prototype.setToolboxRefreshEnabled = function(enabled) {
core/workspace_svg.js-  var reenabled = (!this.toolboxRefreshEnabled_ && enabled);
core/workspace_svg.js-  this.toolboxRefreshEnabled_ = enabled;
core/workspace_svg.js-  if (reenabled) {
core/workspace_svg.js-    // Newly enabled.  Trigger a refresh.
core/workspace_svg.js-    this.refreshToolboxSelection_();
core/workspace_svg.js-  }
core/workspace_svg.js-};
core/workspace_svg.js-
--
--
core/xml.js- * Clear the given workspace then decode an XML DOM and
core/xml.js- * create blocks on the workspace.
core/xml.js- * @param {!Element} xml XML DOM.
core/xml.js- * @param {!Blockly.Workspace} workspace The workspace.
core/xml.js- * @return {Array.<string>} An array containing new block ids.
core/xml.js- */
core/xml.js-Blockly.Xml.clearWorkspaceAndLoadFromXml = function(xml, workspace) {
core/xml.js-  workspace.setResizesEnabled(false);
core/xml.js:  workspace.setToolboxRefreshEnabled(false);
core/xml.js-  workspace.clear();
core/xml.js-  var blockIds = Blockly.Xml.domToWorkspace(xml, workspace);
core/xml.js-  workspace.setResizesEnabled(true);
--
core/xml.js-  workspace.clear();
core/xml.js-  var blockIds = Blockly.Xml.domToWorkspace(xml, workspace);
core/xml.js-  workspace.setResizesEnabled(true);
core/xml.js:  workspace.setToolboxRefreshEnabled(true);
core/xml.js-  return blockIds;
core/xml.js-};
core/xml.js-
core/xml.js-/**
core/xml.js- * Decode an XML DOM and create blocks on the workspace.
core/xml.js- * @param {!Element} xml XML DOM.
core/xml.js- * @param {!Blockly.Workspace} workspace The workspace.
core/xml.js- * @return {Array.<string>} An array containing new block ids.

```

src/containers/blocks.jsx
```diff
@@ -97,9 +103,13 @@ class Blocks extends React.Component {
...
-            this.workspace.updateToolbox(this.props.toolboxXML);
-            const currentCategoryPos = this.workspace.toolbox_.getCategoryPositionByName(categoryName);
-            this.workspace.toolbox_.setFlyoutScrollPos(currentCategoryPos + offset);
+            // rather than update the toolbox "sync" -- update it in the next frame
+            clearTimeout(this.toolboxUpdate);
+            this.toolboxUpdate = setTimeout(() => {
+                this.workspace.updateToolbox(this.props.toolboxXML);
+                const currentCategoryPos = this.workspace.toolbox_.getCategoryPositionByName(categoryName);
+                this.workspace.toolbox_.setFlyoutScrollPos(currentCategoryPos + offset);
+            }, 0);
```
cwillisf: Super minor: I'd consider renaming the variable so its purpose is more obvious. Maybe `toolboxUpdateId` or `pendingToolboxUpdateId` or `theGreatAmericanNovelAboutUpdatingToolboxesShallBeTrackedWithThisId` or something. That last one might be a bit verbose, though.

gnarf: `toolboxUpdateTimeout` is where I'd lean I think...

gnarf:
just rebased this again to try to get rid of the pr test run failures

PR #1786, opened 2018-04-13
Remove feedback form

thisandagain:
### Proposed Changes

- Remove feedback form and associated button in the menu bar

### Reason for Changes

- Preparing to share preview with Scratch community (we will be managing feedback from within the community rather than a separate form)

/cc @carljbowman @christanbalch


paulkaplan:
Love PRs like this 
image

PR #1787, opened 2018-04-13
Use unique ids for the toolbox elements that change per sprite

gnarf:
### Proposed Changes
https://github.com/LLK/scratch-blocks/pull/1442#issuecomment-381216217
In order to use the "reuse toolbox blocks" scratch-blocks patch, we need to make sure blocks that have shadows that will change based on sprite id are updated.


gnarf:
Also I'd like to note, this would have no effect on current performance - but fixes a bug found during testing of the "recycle toolbox" scratch-blocks PR - so we might as well land this now?

Issue #1797, opened 2018-04-17
Change default values for string operators


thisandagain:
The default values for the string operators are "hello" and "world." Although tradition for programming language introductions, newcomers wouldn't know that, and they don't help at all to understand what the string operators do . 

To make the function of the string operators more understandable, recommendation is to change to:

join apple banana
letter 1 of apple
length of apple
apple contains a ?

![image](https://user-images.githubusercontent.com/4721032/38640854-73062820-3da3-11e8-90bd-5e957225703c.png)

---
Moved from LLK/scratch-vm#1049
/cc @ntlrsk

hedgehog125:
Would this be clearer? 
<img width="193" alt="screen shot 2018-04-17 at 16 16 01" src="https://user-images.githubusercontent.com/16543008/38879222-b7d6a2ac-425a-11e8-92be-e93405f0cc00.png">

PR #1798, opened 2018-04-17
Adjustments to string operator default arguments

thisandagain:
### Resolves

Resolves GH-1797

### Proposed Changes

image


paulkaplan:
just a heads up there is an integration test that needs to be changed here that checks the reported value of the default "join" block. `test/integration/blocks.test.js:36`

PR #1828, opened 2018-04-20
Add USE_HEADLESS=no npm run test:integration option

gnarf:
### Proposed Changes

* Add an environment variable `USE_HEADLESS` to the selenium helper which if set to `no` disables headless (any other value including blank is "yes")

### Reason for Changes

* Easy access to turn off headless mode


rschamp:
Thanks @gnarf!

PR #1841, opened 2018-04-21
Make the library filter input full height

rschamp:
This way you can click anywhere to gain focus.

Demo: https://rschamp.github.io/scratch-gui/bugfix/search-height/


rschamp:
Your modification looks great, thanks!

PR #1844, opened 2018-04-21
Center library content

### LINKED ISSUE #1277: Center each row in libraries

rschamp:
Demo: https://rschamp.github.io/scratch-gui/bugfix/center-library-content/

Center library content using CSS grid.

According to https://caniuse.com/#feat=css-grid `display: grid` has the same browser support as Scratch 3.0.

### Resolves

Resolves #1277

### Proposed Changes

This splits the libraries into "featured" and "unfeatured" sections. Theoretically one library could have both sections, but we currently only have one per library. This will need to be tweaked when that happens but is a starting point (unfeatured extensions will need to be the same width as featured extensions, and not be the same dimensions as sprites, costumes, etc.).

### Test Cases
* The sprite, costume, backdrop, sound and extension libraries should appear to be centered with the last row still left justified at any window width
* The above, in our lowest supported browser versions

### DIFF
--- a/src/components/library-item/library-item.css
+++ b/src/components/library-item/library-item.css
@@ -6,9 +6,9 @@
     flex-direction: column;
     align-items: center;
     justify-content: center;
-    flex-basis: 160px;
-    max-width: 160px;
-    height: 160px;
+    flex-basis: $library-item-width;
+    width: $library-item-width;
+    height: $library-item-height;
     margin: $space;
     padding: 1rem 1rem 0 1rem;
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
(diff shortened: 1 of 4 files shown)


paulkaplan:
🤷‍♂️  i'm a bit sad because I think support for webgl/webaudio actually predates css grid on most browsers. any chance this can at least do the old behavior if css grid isn't supported?

paulkaplan:
Sorry @rschamp this isn't mergeable it is causing errors when combined with the cards, looks like it is trying to lowercase a thing that isn't a string. Looking into it

thisandagain:
Closing until we have time to readdress.

PR #1858, opened 2018-04-24
Allow the back button to work with libraries

rschamp:
Demo: https://rschamp.github.io/scratch-gui/feature/modal-back-button/

Push to history when the library mounts. Whenever popstate occurs after mounting, request to be closed.

Behavior captured in the `Modal` container. If we want to re-use this functionality for anything else that uses `ModalComponent`, just use the `Modal` container rather than `ModalComponent`.

### Resolves

Follow-up from Friday's smoke test

### Proposed Changes

Enable the back button when opening modals. You can click back once, and clicking forward again doesn't re-open the modal. This is the most logical way of allowing to back out of modals but not keeping the rest of your navigation history logged while using the GUI (we could think about doing this for fullscreen modals, and tab changes, but @carljbowman discussed offline and thought this might be undesirable).

### Reason for Changes

The libraries seem like new pages so users will expect to be able to use back to close them.

### DIFF
--- a/src/components/library/library.jsx
+++ b/src/components/library/library.jsx
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 import LibraryItem from '../library-item/library-item.jsx';
-import ModalComponent from '../modal/modal.jsx';
+import Modal from '../../containers/modal.jsx';
 import Divider from '../divider/divider.jsx';
 import Filter from '../filter/filter.jsx';
 import TagButton from '../../containers/tag-button.jsx';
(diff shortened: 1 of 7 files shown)


rschamp:
@paulkaplan As far as I can tell, there's not really any way around this since we need to be able to tell the difference between the browser initial state and our pushed state, and there is no API for deleting history, even history you have pushed. So my compromise was to push this one piece of state and reuse it when possible, so we only ever have one thing in the history.  The alternative could be to trigger a "back" navigation whenever closing modals, but that has the opposite problem of having an unused "forward" state that we can't delete. Also just seems overcomplicated to use back for that purpose in a react app.

It's like this:
```
Loaded:
null <- current state

Open modal:
null
modal 1 <- current state

Hit back:
null <- current state
modal 1

Open modal 2 after hitting back on modal 1:
null
modal 2 <- current state (overwrites forward history)

Close modal:
null
modal 1 <- current state (unused)

Open modal 2 after closing modal 1:
null
modal 2 <- current state (manually replaced)
```

PR #1866, opened 2018-04-24
Bitmap import and export

fsih:
### Resolves
Along with https://github.com/LLK/scratch-render/pull/266, LLK/scratch-vm#1082 , and https://github.com/LLK/scratch-paint/pull/404 , supports import and export bitmaps

This PR should wait on all the others to deploy, then update the dependencies in package.json before going in.


paulkaplan:
Just FYI https://github.com/LLK/scratch-gui/pull/1871 brought in the render update, since it could come in without needing any changes from GUI

src/containers/paint-editor-wrapper.jsx
```diff
@@ -55,11 +59,12 @@ const mapStateToProps = (state, {selectedCostumeIndex}) => {
     } = state.targets;
     const target = editingTarget && sprites[editingTarget] ? sprites[editingTarget] : stage;
     const costume = target && target.costumes[selectedCostumeIndex];
+    const resolution = costume && costume.dataFormat === 'png' ? costume.bitmapResolution / 2 : 1;
```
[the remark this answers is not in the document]
cwillisf: Maybe `const resolution = costume && costume.bitmapResolution ? costume.bitmapResolution / 2 : 1`?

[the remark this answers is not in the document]
fsih: Discovered that GUI/VM rotation center units actually match paint's, so there's no need to look up resolution here 👍

fsih:
Huzzah, unblocked and ready for review!

src/containers/paint-editor-wrapper.jsx
```diff
@@ -59,7 +73,8 @@ const mapStateToProps = (state, {selectedCostumeIndex}) => {
         name: costume && costume.name,
         rotationCenterX: costume && costume.rotationCenterX,
         rotationCenterY: costume && costume.rotationCenterY,
-        svgId: editingTarget && `${editingTarget}${costume.skinId}`,
+        imageFormat: costume.dataFormat,
```
kchadha: Should this line also check that costume exists first e.g. `costume && costume.dataFormat`?

src/containers/paint-editor-wrapper.jsx
```diff
@@ -22,28 +22,42 @@ class PaintEditorWrapper extends React.Component {
...
+                image,
+                rotationCenterX,
+                rotationCenterY);
+        } else {
+            this.props.vm.updateBitmap(
+                this.props.selectedCostumeIndex,
+                image,
+                rotationCenterX,
+                rotationCenterY,
+                2 /* bitmapResolution */);
```
cwillisf: In the long term, do you imagine this `2` going away since that will be the only case at runtime?

src/containers/paint-editor-wrapper.jsx
fsih: Thanks!

src/containers/paint-editor-wrapper.jsx
fsih: Yes, I think it would be good if one day Scratch 2 projects still using bitmap resolution 1 were converted by the 2.0-coping-strategies repo right away, and then the project json, gui, vm, paint editor, storage and renderer could all share the same definitions of size and rotation center and not store a resolution. Right now VM is in charge of dividing values by bitmap resolution before sending them to the renderer.

PR #1867, opened 2018-04-25
Framework and demo content for how-tos inside the editor

paulkaplan:
how-to-demo

This is a big PR, but a lot of the "files changed" are just new static assets in the form of gifs and thumbnails. I assume we probably don't want all these static images in source control forever (about 15mb worth right now), but @rschamp @thisandagain we should talk about how we want to deal with that. 

This PR includes both "video" and "step-by-step" cards, and they can actually be interwoven (technically possible, not sure we want to stylistically). 

A couple of things to note:
- Naming: since this was built up initially in phases as a prototype, there are lots of different words used to describe the units of UI here: decks, cards, tips, how-tos, etc. We should probably just make a note in the code of any of these words we want to change. 
- This is the first "library" to use a `.jsx` file as the index, using `FormattedMessage` instead of fixed strings. If this is the way we want to go with the costume/sound/sprite libraries, awesome. If not, we might want to not introduce this as a convention.
- @thisandagain we may want to add a banner/notice to the top of the how-to library for resource developers about that content being subject to change, since I'm sure it will be changed somewhat before launch. Not sure if that is the best way to do it, or if we can communicate it directly instead. 
- @ericrosenbaum I made the "More things to try!" card an explicit step in the sequence for each set of cards, so that we could (for now) manually specify what other decks should appear at the end. This seemed better than just choosing randomly, or only ever showing the first 3. I just copy/pasted the same set for all of them, however, so it would be helpful if you have the chance to go in and make them a bit more meaningful.
- hat tip to Eric S. for the awesome videos.

TODO for myself: 
- [ ] Add integration test for the how-to library and the decks. 
- [ ] Talk to @BryceLTaylor about things to look for in the smoke test
- [ ] Verify looks/usability on mobile devices/more browsers
- [ ] Squash these commits before merging.


rschamp:
Try it out: [cards](https://llk.github.io/scratch-gui/cards)

thisandagain:
:tada: Thanks @paulkaplan

Re: resource developer notice. I think we can message that directly. I’ll follow-up on this.

rschamp:
I can't fully review before the smoke test today, but it looks good to merge for that! I'll follow up after with some suggestions.

src/components/cards/cards.jsx
```diff
@@ -0,0 +1,284 @@
...
+            />
+            <FormattedMessage
+                defaultMessage="All How-Tos"
+                description="Title for button to return to how-to library"
+                id="gui.cards.all-how-tos"
+            />
+        </div>
+        {totalSteps > 1 ? (
+            <div className={styles.stepsList}>
+                {Array(totalSteps).fill(0)
```
rschamp: We have a similarly awkward thing going on in www for a year selection. I think it'd be worth it to start using the `lodash.range` util for this kind of stuff for readability.

src/components/cards/cards.jsx
```diff
@@ -0,0 +1,284 @@
...
+                description="Title card with more things to try"
+                id="gui.cards.more-things-to-try"
+            />
+        </div>
+        <div className={styles.decks}>
+            {deckIds.map(id => (
+                <div
+                    className={styles.deck}
+                    key={`deck-preview-${id}`}
+                    onClick={onActivateDeckFactory(id)}
```
rschamp: Curious about this pattern — since you're defining a function either way, is there an advantage to passing in a factory like this vs doing it inline, like `onClick={() => onActivateDeck(id)}`?  I think both are fine, just curious about this choice.

src/components/cards/cards.jsx
```diff
@@ -0,0 +1,284 @@
...
+        id: PropTypes.shape({
+            name: PropTypes.node.isRequired,
+            img: PropTypes.string.isRequired,
+            steps: PropTypes.arrayOf(PropTypes.shape({
+                title: PropTypes.node,
+                image: PropTypes.string,
+                video: PropTypes.string,
+                deckIds: PropTypes.arrayOf(PropTypes.string)
+            }))
+        })
```
rschamp: It would be easy to make a mistake and modify content without React noticing, so since only `content[activeDeckId]` is used, it would be more error-proof to pass in `name`, `img` and `steps` directly (easy to do with a spread like `{...content[activeDeckId}`).

src/components/library-item/library-item.css
```diff
@@ -87,11 +87,12 @@
 .featured-text {
     font-weight: bold;
     padding: 10px;
-    height: 140px;
+    /* height: 140px; */
```
rschamp: Cruft

src/containers/tips-library.jsx
```diff
@@ -0,0 +1,75 @@
...
+            featured: true
+        }));
+
+        if (!this.props.visible) return null;
+        return (
+            <LibraryComponent
+                data={decksLibraryThumbnailData}
+                filterable={false}
+                title="How-Tos"
+                visible={this.props.visible}
```
rschamp: this visible prop is superfluous since library modals are either displayed or not rendered (also already taken into account above).

Issue #1778, opened 2018-04-13
Upload costume from camera


kchadha:
### Expected Behavior

Should be able to create a new costume from a camera image.

### Actual Behavior

![image](https://user-images.githubusercontent.com/1786240/38745254-06d3ff18-3f12-11e8-9760-ffe60d2dcba3.png)

PR #1880, opened 2018-04-27
Refactor Video IO

kchadha:
### Resolves

Towards resolving #1778.

### Proposed Changes

Refactor video IO so that there is only ever a single call to `getUserMedia` but stage video and camera modal video are controlled separately.

The functionality of the video IO device specific to the scratch-vm runtime and the renderer are left in the scratch-vm IO device.

### Reason for Changes

Towards #1778.

### Test Coverage

Changes can be tested on https://llk.github.io/scratch-gui/camera-costume/, but it should be noted that this PR doesn't contain the camera modal/enabled camera upload button in the costume tab, as that is still a work in progress.

#### Related PR
These changes are related to LLK/scratch-vm#1092, and both PRs should be reviewed together.
The scratch-vm PR should be merged first.


src/containers/stage.jsx
```diff
@@ -6,6 +6,7 @@ import VM from 'scratch-vm';
 import {connect} from 'react-redux';
 
 import {getEventXY} from '../lib/touch-utils';
+import {VideoProvider} from '../lib/camera';
```
gnarf: I'd rather `import VideoProvider from '../lib/video-provider' -- let it require whatever it wants from `lib/camera`

src/lib/camera.js
```diff
@@ -0,0 +1,377 @@
...
+const requestDisableCheck = () => {
+    requestStack.pop();
+    if (requestStack.length > 0) return false;
+    return true;
+};
+
+/**
+ * Video Manager for video extensions.
+ */
+class VideoProvider {
```
gnarf: can we make this it's own file that exports this class?  It would be easier to show as an example in a comment in -vm I think that way.

src/lib/camera.js
```diff
@@ -0,0 +1,377 @@
...
+            this._workspace.push(workspace);
+        }
+        return workspace;
+    }
+}
+
+/**
+ * Video Manager for Camera Modal
+ */
+class ModalVideoManager {
```
gnarf: I think this should get it's own file too

kchadha:
@gnarf Broke out VideoProvider and ModalVideoManager into separate classes, and did some code cleanup.

kchadha:
@rschamp, this PR doesn't cover that change. Those will be coming in a separate PR. The branch that's posted there is just to demonstrate the functionality of the video feed being controlled separately in the camera modal vs. the video motion extension. 
E.g. a running project that uses the video motion extension and turns the video off while the user has the camera modal open should not affect the camera modal, but should turn the stage video off.

kchadha:
@rschamp, @paulkaplan, @thisandagain, @gnarf, does this PR need any more changes?

PR #1896, opened 2018-04-27
Revert to pre-bitmap paint editor for deploying

paulkaplan:
Looks like bitmap paint editor is still having issues that are causing crashes, so excluding it from todays deploy.


rschamp:
Try it out: [revert-bitmap-paint](https://llk.github.io/scratch-gui/revert-bitmap-paint)

PR #1909, opened 2018-04-29
Player mode

### LINKED ISSUE #1687: Player mode

chrisgarrity:
Adds player-only mode to GUI.

### Resolves

Fixes #1687

### Proposed Changes
Add isPlayerOnly property on GUI. Move isFullScreen into mode reducer with isPlayerOnly.

Full screen needs to be a separate prop so that GUI knows what to show when exiting full-screen (either player-only or full editor).

Set `box-sizing: border-box` for all of gui-body. Required the update to width and height on some elements that were still content-box sized.

### Reason for Changes

Player mode for www

### Test Coverage

Current tests run, player playground example uses `player` mode, so player integration test checks player.

To check manually:
- [ ] sprites are not draggable in player mode
- [ ] stage is the default size when not in full screen mode.
- [ ] clicking the full screen button toggles full-screen.
- [ ] green flag, stop, stage-size, and full-screen buttons all look the right size in editor and player.

### DIFF
--- a/src/components/sprite-info/sprite-info.css
+++ b/src/components/sprite-info/sprite-info.css
@@ -31,9 +31,8 @@
 
 .icon-wrapper {
     display: inline-block;
-    box-sizing: content-box !important;
-    width: 1rem;
-    height: 1rem;
+    width: calc(2rem + 2px);
+    height: calc(2rem + 2px);
     padding: 0.5rem;
     outline: none;
     user-select: none;
--- a/src/components/sprite-info/sprite-info.css
+++ b/src/components/sprite-info/sprite-info.css
@@ -31,7 +31,7 @@
 
 .icon-wrapper {
     display: inline-block;
-    box-sizing: content-box;
+    box-sizing: content-box !important;
     width: 1rem;
     height: 1rem;
     padding: 0.5rem;
--- a/src/components/stage-header/stage-header.jsx
+++ b/src/components/stage-header/stage-header.jsx
@@ -9,7 +9,6 @@ import Button from '../button/button.jsx';
 import {ComingSoonTooltip} from '../coming-soon/coming-soon.jsx';
 import Controls from '../../containers/controls.jsx';
 import {getStageSize} from '../../lib/screen-utils.js';
-import {MODES} from '../../reducers/mode';
 
 import fullScreenIcon from './icon--fullscreen.svg';
 import largeStageIcon from './icon--large-stage.svg';
@@ -49,7 +48,7 @@ const messages = defineMessages({
 const StageHeaderComponent = function (props) {
     const {
         isFullScreen,
-        mode,
+        isPlayerOnly,
         onKeyPress,
         onSetStageLarge,
         onSetStageFull,
@@ -86,7 +85,9 @@ const StageHeaderComponent = function (props) {
         );
     } else {
         const stageControls =
-            mode === 'editor' ? (
+            isPlayerOnly ? (
+                []
+            ) : (
                 <div className={styles.stageSizeToggleGroup}>
                     <ComingSoonTooltip
                         place="left"
@@ -124,8 +125,6 @@ const StageHeaderComponent = function (props) {
                         </Button>
                     </div>
                 </div>
-            ) : (
-                []
             );
         header = (
             <Box className={styles.stageHeaderWrapper}>
@@ -159,7 +158,7 @@ const StageHeaderComponent = function (props) {
 StageHeaderComponent.propTypes = {
     intl: intlShape,
     isFullScreen: PropTypes.bool.isRequired,
-    mode: PropTypes.oneOf(Object.keys(MODES)),
+    isPlayerOnly: PropTypes.bool.isRequired,
     onKeyPress: PropTypes.func.isRequired,
     onSetStageFull: PropTypes.func.isRequired,
     onSetStageLarge: PropTypes.func.isRequired,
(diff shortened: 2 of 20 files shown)


src/components/sprite-info/sprite-info.css
```diff
@@ -31,7 +31,7 @@
 
 .icon-wrapper {
     display: inline-block;
-    box-sizing: content-box;
+    box-sizing: content-box !important;
```
rschamp: Same here about `!important`.  I think it might be easier to just adjust the widths and heights so that we don't actually need to use `box-sizing: content-box` at all.

src/components/stage-header/stage-header.jsx
```diff
@@ -78,51 +80,54 @@ const StageHeaderComponent = function (props) {
             </Box>
         );
     } else {
+        const stageControls =
+            mode === 'editor' ? (
```
rschamp: Here we should compare against one of the `MODES` in case this string `'editor'` changes for some reason. But I think better would be if we didn't import that enum at all (which I would consider internal to the reducer) and instead imported a helper function like `isPlayerMode(state.guiMode.mode)`.  And even better would be if we used that in `mapStateToProps` and just had `isPlayerMode` as a boolean prop wherever we needed to know about the mode.

src/components/stop-all/stop-all.css
```diff
@@ -14,7 +14,10 @@
 }
 
 .stop-all:hover {
-    transform: scale(1.2);
+    /* Scale stop image by 1.2, but keep background static */
+    width: 1.5rem;
+    height: 1.5rem;
+    padding: 0.25rem;
```
rschamp: It looks like we're fighting with something here? AFAIK transform should help with issues like that since it doesn't affect width/height calculation.

src/containers/stage.jsx
```diff
@@ -393,8 +393,8 @@ Stage.defaultProps = {
 const mapStateToProps = state => ({
     isColorPicking: state.colorPicker.active,
     isFullScreen: state.stageSize.isFullScreen,
-    // Do not use editor drag style in fullscreen mode.
-    useEditorDragStyle: !state.stageSize.isFullScreen
+    // Do not use editor drag style in fullscreen or player mode.
+    useEditorDragStyle: !(state.stageSize.isFullScreen || state.guiMode.mode === 'player')
```
rschamp: The `'player'` magic string should be imported (or use the `isPlayerMode` helper function I proposed).

src/lib/app-state-hoc.jsx
```diff
@@ -37,8 +40,21 @@ const AppStateHOC = function (WrappedComponent) {
...
+            } else {
+                mode = modeInitialState;
+            }
 
-            this.store = createStore(reducer, intl, enhancer);
+            this.store = createStore(
+                reducer,
+                {
+                    intl: intl,
+                    guiMode: mode
```
rschamp: Curious about `mode` vs `guiMode`. I think I agree that `mode` might be too ambiguous. Since it's distinct from full screen mode, maybe we should stick to that and just call it `isPlayerMode` similar to `isFullScreen`?

rschamp:
Looking good @chrisgarrity! I'm leaning to just having this state as a boolean altogether rather than an enumerable state since they're not exclusive of each other.  In fact maybe `isFullScreen` should get pulled into the `mode` reducer, and have the state look like `{fullScreen: true/false, playerMode: true/false}`?

src/components/stage-header/stage-header.jsx
chrisgarrity: +1 on the isPlayerMode boolean. I just realized that mode was when I thought there would be three modes (editor, player, fullscreen) and just kept it due to inertia when I pulled out fullscreen.

src/components/stop-all/stop-all.css
chrisgarrity: I was seeing an interesting rendering artifact due to layering when I moused over the stop button - a large white rectangle would briefly flash over the top left quarter of the stage. It didn't happen on hover for the green flag, so I switched the stop button to use the same css as the green flag. In any case, it seems like having consistent styling for the hover on the buttons makes sense.

src/lib/app-state-hoc.jsx
chrisgarrity: Yes ambiguity, and `state.mode.mode` just looks wrong.

src/lib/app-state-hoc.jsx
```diff
@@ -37,8 +40,22 @@ const AppStateHOC = function (WrappedComponent) {
...
+                mode = modeInitialState;
+            }
 
-            this.store = createStore(reducer, intl, enhancer);
+            this.store = createStore(
+                reducer,
+                {
+                    intl: intl,
+                    mode: mode
+                },
```
rschamp: I'm surprised that this works since previously the initial state was just the initial state for intl, but now it's an object containing keys for each state.  What you have now actually makes more sense to me — maybe it just wasn't working all the way before?

src/lib/app-state-hoc.jsx
```diff
@@ -37,8 +40,22 @@ const AppStateHOC = function (WrappedComponent) {
...
+                enhancer);
         }
         componentDidUpdate (prevProps) {
             if (prevProps.intl !== this.props.intl) updateIntl(this.props.intl);
```
rschamp: We need a similar check/update here for mode.

src/lib/app-state-hoc.jsx
chrisgarrity: It was working because the initial state includes the `intl` key.

Issue #1632, opened 2018-03-19
Oops! Something went wrong when adding costume


fsih:
### Expected Behavior

Never Oops

### Actual Behavior

Oops

### Steps to Reproduce
It's kind of hard to reproduce, but I've gotten it to happen a few times now.
1. Add 'forever, next costume' script to the cat and run it
2. Set the selected costume index of the stage to at least 1. That is, add at least 1 backdrop, and select a non-first backdrop in the list of backdrops in the costumes tab
3. Add a bunch of surprise sprites

Eventually you will get a crash. "TypeError: Cannot read property 'assetId' of undefined": this indicates that the renderer is trying to access a costume index that doesn't exist for the current editing target.

![newcostumecrash](https://user-images.githubusercontent.com/2855464/37614289-be1ecad6-2b80-11e8-8a12-79d7b781e89b.gif)

### Operating System and Browser

Windows 10 Chrome

Clark-E:
I can’t confirm this right now, but I’m pretty sure that the exception gets thrown [here](https://github.com/LLK/scratch-gui/blob/75f086e065cdb7fac4aa4e524e6fbd8e1d3f7393/src/components/asset-panel/selector.jsx), line 42. This seems to be the only place in the code where this could happen.

Also, concidering the code in question is related to adding in sprite selectors, there’s a good chance the problem is there.

fsih:
I think this is caused by these 2 separate dispatches in scratch-gui/src/lib/vm-listener-hoc:
```
        onTargetsUpdate: data => {
            dispatch(updateEditingTarget(data.editingTarget));
            dispatch(updateTargets(data.targetList));
        },
```

Because this is 2 separate dispatches instead of 1, there is an instant when you add a new sprite where the editing target isn't in the targets list.

In `scratch-gui/src/containers/costume-tab`'s `componentWillReceiveProps`:
```
        const target = editingTarget && sprites[editingTarget] ? sprites[editingTarget] : stage;
        if (!target || !target.costumes) {
            return;
        }

        // If switching editing targets, update the costume index
        if (this.props.editingTarget !== editingTarget) {
            this.setState({selectedCostumeIndex: target.currentCostume});
        }
```

When the editing target isn't found in the sprites list, it defaults to assuming that you are looking at the stage's costumes. It tries to switch the selected costume index to the selected costume in the stage, which doesn't exist in the new sprite, and crashes.

[...]

PR #1915, opened 2018-04-30
Protect against showing costume tab for target that does not exist.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/1632

### Proposed Changes

Make sure not to render the costume tab if the data indicates a sprite is selected that is not in the targets list. 

### Reason for Changes

I decided to make this defensive change instead of changing the reducer API because I want that to involve a more holistic refactor after discussing with @rschamp 

@fsih can you try this branch out, you should be able to find it (after travis builds) here:
paulkaplan.me/scratch-gui/fix-costume-crash or by pulling this locally.


fsih:
Tested and it looks like it's fixed 👍
LGTM if you just fix the lint error

PR #1953, opened 2018-05-03
Make error catching and reporting better

paulkaplan:
### Resolves

Fixes an issue where it is difficult for us to find out where crashing errors are coming from by looking at GA.

### Proposed Changes

2 major things:
1. Separate errors coming from unsupported browsers from those coming from supported browsers. This is because basically every visit from IE counts as a "Fatal Error" like the actual bugs. Track those as a separate type of error.
2. Add ErrorBoundary's around the sounds, costumes and blocks tabs to report where the errors are coming from. This has the added benefit of allowing certain kinds of errors to be "recovered" from, because it only shows the "oops" page within the tab, instead of over the whole page. So it should still be possible to download your project if that happens.

By forcing an error to occur in the blocks container, this is what happens
image

### Test Coverage

Manually tested that the correct analytics events are being emitted for errors on both supported and unsupported browsers, including IE11.

### DIFF
--- a/src/lib/app-state-hoc.jsx
+++ b/src/lib/app-state-hoc.jsx
@@ -71,7 +71,7 @@ const AppStateHOC = function (WrappedComponent) {
             return (
                 <Provider store={this.store}>
                     <IntlProvider>
-                        <ErrorBoundary>
+                        <ErrorBoundary action="Top Level App">
                             <WrappedComponent {...this.props} />
                         </ErrorBoundary>
                     </IntlProvider>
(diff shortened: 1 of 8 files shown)


src/lib/app-state-hoc.jsx
```diff
@@ -71,7 +71,7 @@ const AppStateHOC = function (WrappedComponent) {
             return (
                 <Provider store={this.store}>
                     <IntlProvider>
-                        <ErrorBoundary>
+                        <ErrorBoundary action="Top Level App">
```
chrisgarrity: I know Action can be anything, but I'm wondering if there are conventions that would make it more useful for filtering in analytics

paulkaplan: At this point only that they be different from each other and meaningful.

PR #1964, opened 2018-05-03
Check for old versions of currently supported browsers

### LINKED ISSUE #1391: Should we show unsupported modal for older browsers?
We are seeing hard crash reports from people on things like chrome 54 (10 major versions older than current), firefox 52 (4 major versions older than current) and safari 10 (only 1 major version older than current, but scratch doesn't work at all on it).

Should we show a warning to those people like we do for IE users?

chrisgarrity:
### Resolves

#1391

### Proposed Changes

Check for Chrome < 55, Firefox < 53, and Safari < 10

### Reason for Changes

We decided that we should prevent people from using old versions.

### Test Coverage

Tried on Mac and Windows - everything was fine, but I have updated versions of all the browsers. I changed the code to check for a higher version of Chrome and was able to get the unsupported browser message.

### Browser Coverage
Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/browser-modal/browser-modal.jsx
+++ b/src/components/browser-modal/browser-modal.jsx
@@ -31,7 +31,7 @@ const BrowserModal = ({intl, ...props}) => (
             <p>
                 { /* eslint-disable max-len */ }
                 <FormattedMessage
-                    defaultMessage="We're very sorry, but Scratch 3.0 does not support Internet Explorer, Opera or Silk. We recommend trying a newer browser such as Google Chrome, Mozilla Firefox, or Microsoft Edge."
+                    defaultMessage="We're very sorry, but Scratch 3.0 does not support this browser. We recommend installing the latest version of Google Chrome, Mozilla Firefox, or Microsoft Edge."
                     description="Unsupported browser description"
                     id="gui.unsupportedBrowser.description"
                 />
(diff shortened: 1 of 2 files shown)


chrisgarrity:
Tests pass if I comment out the Safari version check.
Tests pass if I don't run headless.

This probably isn't something we should do right now. Or we could not check version for Safari for now, and add it back when it's supported (there is a [PR](https://github.com/bestiejs/platform.js/pull/145) on platform to detect headless, but it's been open since January).
/cc @paulkaplan @rschamp @thisandagain

rschamp:
How set are we on using `platform`? On scratchr2, we use [bowser](https://www.npmjs.com/package/bowser) which looks more widely used and active.

chrisgarrity:
We're not particularly set on `platform`, but if headless Chrome still shows up as Safari it's not going to help. I didn't see anything in bowser that looks for 'Headless Chrome'

chrisgarrity:
I'm going to close this PR. I can resubmit a PR if we sort out the browser sniffing.

PR #1981, opened 2018-05-04
Add undo testing to paint editor integration test

paulkaplan:
We had an issue in todays in-person test which saw the undo buttons not working correctly in the paint editor, so I figured we should just add a step to one of our automated integration tests to try those buttons.

### DIFF
--- a/test/integration/costumes.test.js
+++ b/test/integration/costumes.test.js
@@ -88,4 +88,19 @@ describe('Working with costumes', () => {
         const logs = await getLogs();
         await expect(logs).toEqual([]);
     });
+
+    test('Undo/redo in the paint editor', async () => {
+        await loadUri(uri);
+        await clickXpath('//button[@title="tryit"]');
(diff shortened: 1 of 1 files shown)


test/integration/costumes.test.js
```diff
@@ -85,6 +85,17 @@ describe('Working with costumes', () => {
         await clickText('costume1', scope.costumesTab);
         await clickText('Convert to Vector', scope.costumesTab);
```
rschamp: It looks like something got screwed up here, the new test stole the bottom of this one?

paulkaplan:
@rschamp yeah sorry, looks like a bad rebase/merge. Fixed!

paulkaplan:
That is a good point @fsih, maybe something we can loop back to (we could select for a button that specifically is not disabled, for example). I'm going to close this because i included it in the other PR that has even more integration tests, but thanks for taking a look!

PR #1996, opened 2018-05-07
Add several new integration tests to the GUI

paulkaplan:
### Resolves

Personal fear that the surprise/empty buttons in thermometer menus were generally untested. Now they are!

Also this should make it easier to add tests for all the new file upload options as well /cc @kchadha.

Also /cc @BryceLTaylor since this is relevant to you!

### DIFF
--- a/test/integration/costumes.test.js
+++ b/test/integration/costumes.test.js
@@ -25,7 +25,7 @@ describe('Working with costumes', () => {
         await driver.quit();
     });
 
-    test('Adding a costume', async () => {
+    test('Adding a costume through the library', async () => {
         await loadUri(uri);
         await clickXpath('//button[@title="tryit"]');
         await clickText('Costumes');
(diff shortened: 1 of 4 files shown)


test/integration/costumes.test.js
```diff
@@ -38,6 +38,32 @@ describe('Working with costumes', () => {
         await expect(logs).toEqual([]);
     });
 
+    test('Adding a costume by surprise button', async () => {
+        await loadUri(uri);
+        await clickXpath('//button[@title="tryit"]');
+        await clickText('Costumes');
+        const el = await findByXpath('//button[@aria-label="Choose a Costume"]');
+        await driver.actions().mouseMove(el)
+            .perform();
```
rschamp: Cool!  If we're doing this a lot already, maybe a `moveMouseToElement` (or something) helper?  I can imagine needing more of this in the future with drag and drop, for example.

paulkaplan: Yeah, I'm not sure what the right level of abstraction is for it though, because if you need to do anything else you'll need to drop back down to this level. I think we should wait for some more uses.

PR #2011, opened 2018-05-09
Consume monitor mode and position from VM

paulkaplan:
### Resolves

Another step towards importing monitors from scratch 2 projects from the VM.

### Proposed Changes

Pass in monitor x/y data as well as mode data from the VM. This change will not do anything without the PR that imports monitors from SB2 files in the VM, but it does not break without it. With the VM update, monitors will be imported with the correct mode and positioning.

Also made a minor CSS update to make large monitors a bit closer in size to SB2, which I found was important when we got SB2 importing working.

Remaining work for monitors in the GUI includes:
- Sending mode updates to the VM, not storing them in state
- Adding slider and monitor modes to the GUI
- Hooking up the new slider/monitor modes to VM actions to update variables.

To be clear, this does not depend on the VM update, so it can be merged before it. But it was tested/developed using the new VM monitors import branch, so it will start working when that branch comes in.

### DIFF
--- a/src/components/monitor-list/monitor-list.jsx
+++ b/src/components/monitor-list/monitor-list.jsx
@@ -16,10 +16,13 @@ const MonitorList = props => (
             <Monitor
                 id={monitorData.id}
                 key={monitorData.id}
+                mode={monitorData.mode}
                 opcode={monitorData.opcode}
                 params={monitorData.params}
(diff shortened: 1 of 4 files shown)


paulkaplan:
Also forgot to mention, since not all the modes are available, I added a default case for now:

image

PR #2014, opened 2018-05-09
Add "read-only" slider and list monitors.

paulkaplan:
## Resolves

Adds first iteration UI for list and slider monitors. They are "read-only" in this PR, i.e. they cannot be edited, the sliders do not work. I wanted to do that in a separate PR, because it may require VM APIs that do not exist yet.

### Proposed Changes

You can now:
- Create list monitors from scratch 3 toolbox
- Import list monitors with the XY/width/height set from scratch 2
- Use double-click or context menu to change variable monitors to slider mode.

Manually tested along with monitor imports:

image

Depends on https://github.com/LLK/scratch-vm/pull/1128, which is merged, i'll update this PR when it is built.

### DIFF
--- a/src/components/monitor/monitor.jsx
+++ b/src/components/monitor/monitor.jsx
@@ -7,6 +7,8 @@ import {ContextMenu, MenuItem} from '../context-menu/context-menu.jsx';
 import Box from '../box/box.jsx';
 import DefaultMonitor from './default-monitor.jsx';
 import LargeMonitor from './large-monitor.jsx';
+import SliderMonitor from './slider-monitor.jsx';
+import ListMonitor from './list-monitor.jsx';
 
 import styles from './monitor.css';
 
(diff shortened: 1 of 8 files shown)


src/components/monitor/monitor.jsx
```diff
@@ -21,44 +23,62 @@ const categories = {
...
+    large: LargeMonitor,
+    slider: SliderMonitor,
+    list: ListMonitor
 };
 
 const MonitorComponent = props => (
     <ContextMenuTrigger id={`monitor-${props.label}`}>
         <Draggable
             bounds=".monitor-overlay" // Class for monitor container
+            cancel=".no-drag" // Class used for slider input to prevent drag
```
rschamp: Can't these be imported and scoped? Sorry I missed the bounds one.

src/components/monitor/slider-monitor.jsx
```diff
@@ -0,0 +1,49 @@
...
+            <div
+                className={styles.value}
+                style={{background: categoryColor}}
+            >
+                {value}
+            </div>
+        </div>
+        <div className={styles.row}>
+            <input
+                className="no-drag" // Class used on parent Draggable to prevent drags
```
rschamp: Same here

src/containers/monitor.jsx
```diff
@@ -68,39 +80,51 @@ class Monitor extends React.Component {
...
         this.setState({mode: 'large'});
     }
+    handleSetModeToSlider () {
+        this.setState({mode: 'slider'});
+    }
     setElement (monitorElt) {
         this.element = monitorElt;
     }
     render () {
+        // console.log(this.props);
```
rschamp: Cruft

src/containers/monitor.jsx
```diff
@@ -68,39 +80,51 @@ class Monitor extends React.Component {
...
+                height={this.props.height}
+                max={this.props.max}
+                min={this.props.min}
                 mode={this.state.mode}
+                width={this.props.width}
                 onDragEnd={this.handleDragEnd}
                 onNextMode={this.handleNextMode}
                 onSetModeToDefault={this.handleSetModeToDefault}
                 onSetModeToLarge={this.handleSetModeToLarge}
+                onSetModeToSlider={showSliderOption ? this.handleSetModeToSlider : null}
```
rschamp: Might be simpler to handle this in the handler and then always pass it in? Then you don't have to conditionally check the type of this prop in the component.

src/components/monitor/monitor.jsx
```diff
@@ -21,44 +23,62 @@ const categories = {
...
-            </MenuItem>
-            <MenuItem onClick={props.onSetModeToLarge}>
-                <FormattedMessage
-                    defaultMessage="large readout"
-                    description="Menu item to switch to the large monitor"
-                    id="gui.monitor.contextMenu.large"
-                />
-            </MenuItem>
-        </ContextMenu>
+        {props.mode === 'list' ? null : (
```
rschamp: This is kind of doing work we already have figured out in the container, could we just pass the filtered list of available modes for this instead?

src/lib/monitor-adapter.js
```diff
@@ -25,8 +25,8 @@ export default function ({id, spriteName, opcode, params, value}) {
     }
 
     // If value is a number, round it to six decimal places
-    if (typeof value === 'number' || (typeof value === 'string' && String(parseFloat(value)) === value)) {
-        value = Number(Number(value).toFixed(6));
```
rschamp: Just curious... wat?

paulkaplan:
Discussed offline. This code will go through a few more iterations in the next few days, so won't worry too much about the architecture right now.

Issue #942, opened 2017-11-23
Implement list monitors


Kenny2github:
### Expected Behavior

List monitors should be like 2.0

### Actual Behavior

List monitors attempt to act like variable monitors, with strange results:
![image](https://user-images.githubusercontent.com/28599280/33167403-60930ba6-d078-11e7-9881-32c2a63fc3bc.png)

### Steps to Reproduce

1. Create a list
2. Look at the Stage

### Operating System and Browser

Chrome 62 Windows 10

thisandagain:
Thanks @Kenny2github. This is a feature that hasn't been implemented yet.

towerofnix:
See also #268:

>**Dropping text files onto lists to insert the text content into the list.** I'm bolding this because I think it'll otherwise be missed, and would be a surprisingly useful feature – for example, as a means of loading save files into a project with a custom save/load mechanic. It's already a feature of 2.0 to import/export list data by using the context menu, but a more obvious and robust way would be nice.

Definitely not a critical thing to implement, but maybe worth considering while designing.

Several Scratch games use lists exporting/importing as a save/load mechanic, but unfortunately it currently requires the user to see inside the project and right-click on the list monitor. It certainly doesn't need to be that complicated. So, adding onto that idea - should there be a way for the user to import or export a displayed list as a text file, just by clicking a button on the list monitor?

thisandagain:
Yup! Agreed. We'll implement the CSV import/export in a separate issue / scope of work, but we are tracking it. 😄

towerofnix:
>Yup! Agreed. We'll implement the CSV import/export in a separate issue / scope of work, but we are tracking it. :smile:

:tada:

> CSV import/export

..Does this mean you'll be implementing 2-dimensional lists? :stuck_out_tongue_winking_eye:

thisandagain:
@towerofnix LOL. One step at a time. 😜

DeleteThisAcount:
when can we expect 4d lists?

picklesrus:
I know this isn't done yet, but here are couple of things to make sure to test when implementation is done:
- When a list is renamed, the monitor's name should be updated too
- Color of the monitor should match the color of the list category instead of the variable category.

kyleplo:
bump

rachel-fenichel:
Frequently reported in Scratch 3.0 feedback.

PR #2019, opened 2018-05-10
Make slider and list monitors update variables in the VM

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/942

Makes the slider and list monitors interactive:
- slider monitors update variables
- list monitors do... a lot:
  - clicking a list item turns it into an input, selecting all the text
  - when active, there is a button that removes the item from the list and activates the next item
  - when active, the <kbd>Tab</kbd> key activates the next item (<kbd>Shift</kbd> goes backwards) which wraps the around
  - when active, the <kbd>Enter</kbd> key adds a new blank item after the current item, and activates it (<kbd>Shift</kbd> inserts above, h/t @towerofnix for pointing that out).
  - when active, the <kbd>ArrowDown</kbd> and <kbd>ArrowUp</kbd> activate the next / prev items (h/t @towerofnix for pointing that out)
  - when active, you can type to change the item. updates get propagated to the VM on blur
  - you can add new items to the bottom by clicking "+"
  - you can resize the list monitor by dragging the "="

Also @kchadha I didn't implement the VM apis for updating variables, but I made an isolated utility that does it which we should be able to emulate in the VM (right now the utility digs into the VM to make the changes). Felt better to design the use-case first, then go back and design the APIs. 

I chose what seems to be the go-to library for virtualizing long lists, `react-virtualized`, which allows us to avoid both the react render and the dom elements of a 200k element list, making scrolling and editing quick. It gets a bit slower after 100k, probably more due to GC thrashing of the list objects, but we can follow up with that later (immutable for lists?)

list-slider

---

Note not implemented:
- Import from CSV in list monitor context menu
- Hide in the context menu for any monitors
- Min/Max slider setting in context menu (min/max is imported correctly from 2.0, but there is no UI for changing it in this PR)
- Blinking yellow in list monitors when values are changed/added/removed.

### DIFF
--- a/src/components/monitor/list-monitor-scroller.jsx
+++ b/src/components/monitor/list-monitor-scroller.jsx
@@ -1,0 +1,109 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import bindAll from 'lodash.bindall';
+
+import styles from './monitor.css';
+import {List} from 'react-virtualized';
+
+class ListMonitorScroller extends React.Component {
+    constructor (props) {
+        super(props);
+        bindAll(this, [
+            'rowRenderer',
+            'noRowsRenderer',
+            'handleEventFactory'
+        ]);
+    }
+    handleEventFactory (index) {
+        return () => this.props.onActivate(index);
+    }
+    noRowsRenderer () {
+        return (
+            <div className={styles.listEmpty}>
+                {'(empty)' /* TODO waiting for design before translation */}
+            </div>
+        );
+    }
+    rowRenderer ({index, key, style}) {
+        return (
+            <div
+                className={styles.listRow}
+                key={key}
+                style={style}
+            >
+                <div className={styles.listIndex}>{index + 1 /* one indexed */}</div>
+                <div
+                    className={styles.listValue}
+                    dataIndex={index}
+                    style={{background: this.props.categoryColor}}
+                    onClick={this.handleEventFactory(index)}
+                >
+                    {this.props.activeIndex === index ? (
+                        <div className={styles.inputWrapper}>
+                            <input
+                                autoFocus
+                                autoComplete={false}
+                                className={classNames(styles.listInput, 'no-drag')}
+                                spellCheck={false}
+                                type="text"
+                                value={this.props.activeValue} /* eslint-disable-line */
+                                onBlur={this.props.onDeactivate}
+                                onChange={this.props.onInput}
+                                onFocus={this.props.onFocus}
+                                onKeyDown={this.props.onKeyPress} // key down to get ahead of blur
+                            />
+                            <div
+                                className={styles.removeButton}
+                                onMouseDown={this.props.onRemove} // mousedown to get ahead of blur
+                            >
+                                {'✖︎'}
+                            </div>
+                        </div>
+
+                    ) : (
+                        <div className={styles.valueInner}>{this.props.values[index]}</div>
+                    )}
+                </div>
+            </div>
+        );
+    }
+    render () {
+        const {height, values, width, activeIndex, activeValue} = this.props;
+        // Keep the active index in view if defined, else must be undefined for List component
+        const scrollToIndex = activeIndex === null ? undefined : activeIndex; /* eslint-disable-line no-undefined */
+        return (
+            <List
+                activeIndex={activeIndex}
+                activeValue={activeValue}
+                height={(height) - 44 /* Header/footer size, approx */}
+                noRowsRenderer={this.noRowsRenderer}
+                rowCount={values.length}
+                rowHeight={24 /* Row size is same for all rows */}
+                rowRenderer={this.rowRenderer}
+                scrollToIndex={scrollToIndex} /* eslint-disable-line no-undefined */
+                values={values}
+                width={width}
+            />
+        );
+    }
+}
+
+ListMonitorScroller.propTypes = {
+    activeIndex: PropTypes.number,
+    activeValue: PropTypes.string,
+    categoryColor: PropTypes.string,
+    height: PropTypes.number,
+    onActivate: PropTypes.func,
+    onDeactivate: PropTypes.func,
+    onFocus: PropTypes.func,
+    onInput: PropTypes.func,
+    onKeyPress: PropTypes.func,
+    onRemove: PropTypes.func,
+    values: PropTypes.arrayOf(PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.number
+    ])),
+    width: PropTypes.number
+};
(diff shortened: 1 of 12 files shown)


towerofnix:
Note some extra list controls from 2.0 that might not be implemented in this (I haven't tested this PR myself):

* Hold <kbd>Shift</kbd> while pressing <kbd>Enter</kbd> to insert an empty list item *above* the currently selected item, and then select it.
* Use the <kbd>Up</kbd> and <kbd>Down</kbd> arrow keys to select items up and down in the list. (These are functionally equivalent to <kbd>Shift</kbd>-<kbd>Tab</kbd> and <kbd>Tab</kbd>, respectively.)
* <kbd>Tab</kbd>/<kbd>Shift</kbd>-<kbd>Tab</kbd> (and their arrow key equivalents) wrap if you're on the bounds of the list - that is, pressing down on the last item will focus the first item, and pressing up on the first item will select the last item.

towerofnix:
May be worth considering what happens to the list focus while the list is being changed *by a script*, by the way. For example, if an item is deleted or inserted above the currently focused item, it shouldn't suddenly be that the item *below* the focused item is selected. I think Scratch 2.0 was a little wonky with this, especially if the list was being modified by scripts very very frequently while focused.

paulkaplan:
Yeah, I'd like to get this in and then see what feels right.

src/components/monitor/monitor.jsx
```diff
@@ -29,7 +29,10 @@ const modes = {
 };
 
 const MonitorComponent = props => (
-    <ContextMenuTrigger id={`monitor-${props.label}`}>
+    <ContextMenuTrigger
+        holdToDisplay={props.mode === 'slider' ? -1 : 1000}
```
rschamp: Just so I understand — this is the prop that lets long-press open the menu? So 1s hold to open?

src/containers/list-monitor.jsx
```diff
@@ -0,0 +1,183 @@
...
+            'handleAdd',
+            'handleResizeMouseDown'
+        ]);
+
+        this.state = {
+            activeIndex: null,
+            activeValue: null,
+            // TODO These will need to be sent back to the VM for saving
+            width: props.width || 80,
+            height: props.height || 200
```
rschamp: I think you want `typeof props.width === 'undefined' ? 80 : props.width` so that you can set 0 for these if you need to.

rschamp: Also it is a smell to set state to the value of a prop so I think a comment is necessary to explain. Do you need a change handler from the parent component to update the width/height?  Or maybe the width/height props would be better named `initialWidth`/`initialHeight`?

src/containers/list-monitor.jsx
```diff
@@ -0,0 +1,183 @@
...
+            const newListValue = listValue.slice(0, previouslyActiveIndex + newValueOffset)
+                .concat([newListItemValue])
+                .concat(listValue.slice(previouslyActiveIndex + newValueOffset));
+            setVariableValue(vm, targetId, variableId, newListValue);
+            const newIndex = (previouslyActiveIndex + newValueOffset) % newListValue.length;
+            this.setState({
+                activeIndex: newIndex,
+                activeValue: newListItemValue
+            });
+        }
```
rschamp: This was simpler than I thought it was gonna be, nice work!

kchadha:
@paulkaplan Some notes from testing:

- Can't set slider min and max (is there a separate issue out for this)?
- With list monitors, you can click on an item slot to focus it and type in it, but clicking outside the list monitor on the stage does not un-focus the item slot. Clicking outside of the stage does.
- I can't tell if it's just my machine (could be) but it seems that rapid clicking the `+` icon to add a new item slot is kind of slow (especially in comparison to the rapid item add for the `add () to [list]` block
- should the `+` and `=` 'icons' have some padding around them (see image below)?

image

- You can use the down arrow to go through all the item slots in a list and it wraps back around to the top, however the up arrow does not wrap back around to the bottom (this is inconsistent with Scratch 2.0). Same with `tab` vs. `shift + tab`

Note: these don't feel like blockers to me, and can be tracked in separate issues

src/lib/variable-utils.js
```diff
@@ -0,0 +1,24 @@
+// Utility functions for updating variables in the VM
+// TODO these should be moved to top-level VM API
```
rschamp: Please make an issue to migrate these this iteration.

src/components/monitor/list-monitor-scroller.jsx
```diff
@@ -0,0 +1,109 @@
...
+                        <div className={styles.valueInner}>{this.props.values[index]}</div>
+                    )}
+                </div>
+            </div>
+        );
+    }
+    render () {
+        const {height, values, width, activeIndex, activeValue} = this.props;
+        // Keep the active index in view if defined, else must be undefined for List component
+        const scrollToIndex = activeIndex === null ? undefined : activeIndex; /* eslint-disable-line no-undefined */
```
kchadha: why set to undefined and not null?

kchadha:
Another note:
Having items in a list, and going through them changing each makes the old item flash briefly before changing to the new item during input switching. See gif below:

list_monitor_item_switching

src/components/monitor/monitor.jsx
paulkaplan: Yup, which is the default. The more important part is that it is disabled for sliders.

src/containers/list-monitor.jsx
paulkaplan: I think I explicitly do not want zero width/height, anything with zeros should get set back to the default.

I think setting the state to the props value is the standard way to deal with "transient" values (like a buffered input, etc.) The prop is "initial" i suppose, although once the w/h get sent back to the VM and propogated back here, it won't really be "initial". Once we do that (TODO comment above) we will just reset the state to the props in componentWillReceiveProps.

src/components/monitor/list-monitor-scroller.jsx
paulkaplan: Sorry I tried to make it clear in the comment above. The component I'm using (List) takes scrollToIndex as an optional prop (which means it'll be undefined if it isn't passed in). Apparently they are looking specifically for undefineds, and `null` is not an acceptable value.

paulkaplan:
@kchadha @rschamp thanks for the comments, updated or responded accordingly. 

@kchadha I'll make an issue for wrapping the other direction, darn `%` operator with negative numbers, i wish you worked the way i want you to! And w/r/t the flickering, I'll have to think more about how to prevent that. Basically as soon as it loses focus it switches back to the current value of that field, which hasn't yet been updated... I'll file it for completeness

PR #2020, opened 2018-05-10
Project loader refactor

chrisgarrity:
### Proposed Changes

Separate getting the project id from the `location.hash` (needed for playground) from loading the project by Id - needed by all uses of GUI.

### Reason for Changes

When GUI is embedded (e.g. in www), it will be passed the project id, and the main app will be managing the URL.

### Test Coverage

Changed the current project-loader test to check that hash-parser+project-loader continues to work as before. Added new project loader test.

### Browser Coverage
Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/stage-wrapper/stage-wrapper.css
+++ b/src/components/stage-wrapper/stage-wrapper.css
@@ -1,7 +1,7 @@
 @import "../../css/units.css";
 @import "../../css/colors.css";
 
-.stage-wrapper {
+.stage-wrapper * {
     box-sizing: border-box;
 }
 
--- a/src/lib/hash-parser-hoc.jsx
+++ b/src/lib/hash-parser-hoc.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import bindAll from 'lodash.bindall';
 
 /* Higher Order Component to get the project id from location.hash
  * @param {React.Component} WrappedComponent component to receive projectData prop
@@ -8,24 +9,22 @@ const HashParserHOC = function (WrappedComponent) {
     class HashParserComponent extends React.Component {
         constructor (props) {
             super(props);
-            this.fetchProjectId = this.fetchProjectId.bind(this);
-            this.updateProject = this.updateProject.bind(this);
+            bindAll(this, [
+                'handleHashChange'
+            ]);
             this.state = {
                 projectId: null
             };
         }
         componentDidMount () {
-            window.addEventListener('hashchange', this.updateProject);
-            this.updateProject();
+            window.addEventListener('hashchange', this.handleHashChange);
+            this.handleHashChange();
         }
         componentWillUnmount () {
-            window.removeEventListener('hashchange', this.updateProject);
+            window.removeEventListener('hashchange', this.handleHashChange);
         }
-        fetchProjectId () {
-            return window.location.hash.substring(1);
-        }
-        updateProject () {
-            let projectId = this.fetchProjectId();
+        handleHashChange () {
+            let projectId = window.location.hash.substring(1);
             if (projectId !== this.state.projectId) {
                 if (projectId.length < 1) projectId = 0;
                 this.setState({projectId: projectId});
--- a/src/lib/hash-parser-hoc.jsx
+++ b/src/lib/hash-parser-hoc.jsx
@@ -1,0 +1,49 @@
+import React from 'react';
+
+/* Higher Order Component to get the project id from location.hash
+ * @param {React.Component} WrappedComponent component to receive projectData prop
+ * @returns {React.Component} component with project loading behavior
+ */
+const HashParserHOC = function (WrappedComponent) {
+    class HashParserComponent extends React.Component {
+        constructor (props) {
+            super(props);
+            this.fetchProjectId = this.fetchProjectId.bind(this);
+            this.updateProject = this.updateProject.bind(this);
+            this.state = {
+                projectId: null
+            };
+        }
+        componentDidMount () {
+            window.addEventListener('hashchange', this.updateProject);
+            this.updateProject();
+        }
+        componentWillUnmount () {
+            window.removeEventListener('hashchange', this.updateProject);
+        }
+        fetchProjectId () {
+            return window.location.hash.substring(1);
+        }
+        updateProject () {
+            let projectId = this.fetchProjectId();
+            if (projectId !== this.state.projectId) {
+                if (projectId.length < 1) projectId = 0;
+                this.setState({projectId: projectId});
+            }
+        }
+        render () {
+            return (
+                <WrappedComponent
+                    projectId={this.state.projectId}
+                    {...this.props}
+                />
+            );
+        }
+    }
+
+    return HashParserComponent;
+};
+
+export {
+    HashParserHOC as default
+};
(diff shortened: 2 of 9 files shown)


src/components/stage-wrapper/stage-wrapper.css
```diff
@@ -1,7 +1,7 @@
 @import "../../css/units.css";
 @import "../../css/colors.css";
 
-.stage-wrapper {
+.stage-wrapper * {
     box-sizing: border-box;
```
rschamp: Should this be `.stage-wrapper, .stage-wrapper *` so the `.stage-wrapper` itself retains the previous `box-sizing`?  Also, is this in this PR intentionally?

src/lib/hash-parser-hoc.jsx
```diff
@@ -0,0 +1,49 @@
...
+/* Higher Order Component to get the project id from location.hash
+ * @param {React.Component} WrappedComponent component to receive projectData prop
+ * @returns {React.Component} component with project loading behavior
+ */
+const HashParserHOC = function (WrappedComponent) {
+    class HashParserComponent extends React.Component {
+        constructor (props) {
+            super(props);
+            this.fetchProjectId = this.fetchProjectId.bind(this);
+            this.updateProject = this.updateProject.bind(this);
```
rschamp: We usually use [`lodash.bindall`](https://github.com/LLK/scratch-gui/blob/342ac4a2448f0ec6bbf5468cdd942c63d735dc34/src/containers/tag-button.jsx#L10) for this.

src/lib/hash-parser-hoc.jsx
```diff
@@ -0,0 +1,49 @@
...
+            this.state = {
+                projectId: null
+            };
+        }
+        componentDidMount () {
+            window.addEventListener('hashchange', this.updateProject);
```
rschamp: Seeing it in this new context, I think a better name for `updateProject` here would be `handleHashChange`.

src/lib/hash-parser-hoc.jsx
```diff
@@ -0,0 +1,49 @@
...
+            this.updateProject();
+        }
+        componentWillUnmount () {
+            window.removeEventListener('hashchange', this.updateProject);
+        }
+        fetchProjectId () {
+            return window.location.hash.substring(1);
+        }
+        updateProject () {
+            let projectId = this.fetchProjectId();
```
rschamp: Since `fetchProjectId` is only used once, I think we can just inline `window.location.hash.substring(1)` here.

test/unit/util/hash-project-loader-hoc.test.jsx
```diff
@@ -0,0 +1,64 @@
...
+        const originalLoad = storage.load;
+        storage.load = jest.fn(() => Promise.resolve(null));
+        const mounted = mount(<WrappedComponent />);
+        storage.load = originalLoad;
+        window.location.hash = '';
+        const mountedDiv = mounted.find('div');
+        expect(mountedDiv.exists()).toEqual(false);
+    });
+
+    test('when there is no hash, it loads the default project', () => {
```
rschamp: I think we should split this in two:
1. When there is no hash, `HashParserHOC` passes `0` as the `projectId` prop
2. If `0` is passed to  `ProjectLoaderHOC` it loads the default project

test/unit/util/hash-project-loader-hoc.test.jsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import ProjectLoaderHOC from '../../../src/lib/project-loader-hoc.jsx';
+import HashParserHOC from '../../../src/lib/hash-parser-hoc.jsx';
+import storage from '../../../src/lib/storage';
+import {mount} from 'enzyme';
+
+jest.mock('react-ga');
+
+describe('Hash/ProjectLoaderHOC', () => {
```
rschamp: I think we should only test HashParserHOC in this file, since we already have a file dedicated to `ProjectLoaderHOC`.

test/unit/util/hash-project-loader-hoc.test.jsx
```diff
@@ -0,0 +1,64 @@
...
+    test('when there is no project data, it renders null', () => {
```
rschamp: This seems like a pure `ProjectLoaderHOC` test, so it should probably stay in the other file.

test/unit/util/hash-project-loader-hoc.test.jsx
```diff
@@ -0,0 +1,64 @@
...
+        const Component = ({projectData}) => <div>{projectData}</div>;
+        const WrappedComponent = HashParserHOC(ProjectLoaderHOC(Component));
+        window.location.hash = '#winning';
+        const originalLoad = storage.load;
+        storage.load = jest.fn(() => Promise.resolve(null));
+        const mounted = mount(<WrappedComponent />);
+        storage.load = originalLoad;
+        window.location.hash = '';
```
rschamp: This is how it was before, but it's kind of confusing how this works. The above two lines are just setting global state to how it was found, so the test is idempotent. Either we should do this at the end of all the tests, or we should leave a comment here that says that's what we're doing.

test/unit/util/hash-project-loader-hoc.test.jsx
```diff
@@ -0,0 +1,64 @@
...
+        storage.load = jest.fn((type, id) => Promise.resolve(id));
+        const mounted = mount(<WrappedComponent />);
+        expect(mounted.state().projectId).toEqual(0);
+        expect(storage.load).toHaveBeenCalledWith(
+            storage.AssetType.Project, 0, storage.DataFormat.JSON
+        );
+        storage.load = originalLoad;
+    });
+
+    test('when there is a hash, it tries to load that project', () => {
```
rschamp: This doesn't have to involve `HashParserHOC`, and can move to the other test file: "When a non-0 projectId is passed to ProjectLoaderHOC, it tries to load that project".

test/unit/util/hash-project-loader-hoc.test.jsx
```diff
@@ -0,0 +1,64 @@
...
+        storage.load = jest.fn((type, id) => Promise.resolve({data: id}));
+        const mounted = mount(<WrappedComponent />);
+        expect(mounted.state().projectId).toEqual('winning');
+        expect(storage.load).toHaveBeenLastCalledWith(
+            storage.AssetType.Project, 'winning', storage.DataFormat.JSON
+        );
+        storage.load = originalLoad;
+    });
+
+    test('when hash change happens, the project data state is changed', () => {
```
rschamp: This can be a pure `HashParserHOC` test: "When  hash change happens, the `projectId` state changes".

src/components/stage-wrapper/stage-wrapper.css
chrisgarrity: It's actually irrelevant whether stage-wrapper has box-sizing set. The important part was all the children - which was missing.

PR #2041, opened 2018-05-11
Interact with blocks categories by id instead of by name

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/1429
Depends on https://github.com/LLK/scratch-blocks/pull/1517

### Proposed Changes

- Add an id to each category in the toolbox XML (following https://github.com/LLK/scratch-vm/pull/1141, extension categories now have this id as well)
- Extension library provides extensionId instead of extensionURL
- When switching sprites, use category id to select category, scroll to it, and request position and length
- When selecting an extension in extension library, use the id so the menu can scroll to it

### Reason for Changes

Category names will change due to localization, so we need to use stable ids.

It also provides support for auto-scrolling to an extension category (on adding it) even if the extension name shown in the library does not match the name shown in the category menu.


paulkaplan:
Marking this as "blocked" because it needs to be updated with a new dependency before it can be merged.

paulkaplan:
LG

PR #2068, opened 2018-05-15
Add Dockerfile for docker to build this Scratch Web UI.

philipz:
### Resolves

Using docker tool for portable.

### Proposed Changes

Just add A Dockerfile.

### Reason for Changes

No more need install node.js environment, just only install docker.

### Test Coverage

No yet.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,1 @@
+FROM node:6.14.2-onbuild


colbygk:
Will discuss later on how we want to incorporate this into a longer term docker strategy

PR #2083, opened 2018-05-16
Feature/refactor gui store

chrisgarrity:
### Proposed Changes

Refactor the gui redux store to allow it to be a part of an enclosing app's store. 
Also provide a way for an enclosing app to set the mode, along with a function to handle switching to editor mode (aka 'see inside').

### Reason for Changes

Needed for new project page

### Test Coverage

Current tests run.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -445,12 +445,12 @@ Blocks.defaultProps = {
 };
 
 const mapStateToProps = state => ({
-    anyModalVisible: Object.keys(state.modals).some(key => state.modals[key]),
-    extensionLibraryVisible: state.modals.extensionLibrary,
+    anyModalVisible: Object.keys(state.scratchGui.modals).some(key => state.scratchGui.modals[key]),
+    extensionLibraryVisible: state.scratchGui.modals.extensionLibrary,
     locale: state.intl.locale,
     messages: state.intl.messages,
-    toolboxXML: state.toolbox.toolboxXML,
-    customProceduresVisible: state.customProcedures.active
+    toolboxXML: state.scratchGui.toolbox.toolboxXML,
+    customProceduresVisible: state.scratchGui.customProcedures.active
 });
 
 const mapDispatchToProps = dispatch => ({
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -445,7 +445,10 @@ Blocks.defaultProps = {
 };
 
 const mapStateToProps = state => ({
-    anyModalVisible: Object.keys(state.scratchGui.modals).some(key => state.scratchGui.modals[key]),
+    anyModalVisible: (
+        Object.keys(state.scratchGui.modals).some(key => state.scratchGui.modals[key]) ||
+        state.scratchGui.mode.isFullScreen
+    ),
     extensionLibraryVisible: state.scratchGui.modals.extensionLibrary,
     locale: state.intl.locale,
     messages: state.intl.messages,
--- a/src/containers/gui.jsx
+++ b/src/containers/gui.jsx
@@ -24,12 +24,13 @@ class GUI extends React.Component {
     constructor (props) {
         super(props);
         this.state = {
-            loading: true,
+            loading: !props.vm.initialized,
             loadingError: false,
             errorMessage: ''
         };
     }
     componentDidMount () {
+        if (this.props.vm.initialized) return;
         this.audioEngine = new AudioEngine();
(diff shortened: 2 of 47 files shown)


src/containers/gui.jsx
```diff
@@ -43,6 +44,7 @@ class GUI extends React.Component {
                 // error page gets rendered if project failed to load
                 this.setState({loadingError: true, errorMessage: e});
             });
+        this.props.vm.initialized = true;
```
rschamp: This is a hack... we should figure out what the best way to ask the vm if it's already been started.  What we really want to know is if there's a project loaded / if the vm is running.  It looks like the most accurate would be `if (typeof this.props.vm.runtime._steppingInterval !== 'undefined')`... but I think we'd want to just make an API on `scratch-vm` for this purpose.

src/index.js
```diff
@@ -0,0 +1,28 @@
...
+import {applyMiddleware, compose} from 'redux';
+import throttle from 'redux-throttle';
+
+const guiReducers = {
+    intl: IntlReducer,
+    scratchGui: GuiReducer,
+    scratchPaint: ScratchPaintReducer
+};
+
+const guiMiddleware = compose(applyMiddleware(throttle(300, {leading: true, trailing: true})));
```
rschamp: It seems like we would want this to be an array, rather than a pre-composed middleware, so its members can be composed together with other middleware in the parent app.

src/lib/app-state-hoc.jsx
```diff
@@ -1,15 +1,18 @@
...
+import intlReducer from '../reducers/intl.js';
+
+import guiReducer, {guiInitialState, initFullScreen, initPlayer} from '../reducers/gui';
+
+import {setPlayer, setFullScreen} from '../reducers/mode.js';
+
+import {ScratchPaintReducer} from 'scratch-paint';
 
 const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
 const enhancer = composeEnhancers(
```
rschamp: We should pull this from `../reducers/gui` so we don't duplicate it.

src/containers/gui.jsx
```diff
@@ -97,16 +101,18 @@ GUI.propTypes = {
...
+    activeTabIndex: state.scratchGui.editorTab.activeTabIndex,
+    blocksTabVisible: state.scratchGui.editorTab.activeTabIndex === BLOCKS_TAB_INDEX,
+    cardsVisible: state.scratchGui.cards.visible,
+    costumesTabVisible: state.scratchGui.editorTab.activeTabIndex === COSTUMES_TAB_INDEX,
+    importInfoVisible: state.scratchGui.modals.importInfo,
+    isPlayerOnly: state.scratchGui.mode.isPlayerOnly,
+    loadingStateVisible: state.scratchGui.modals.loadingProject,
+    previewInfoVisible: state.scratchGui.modals.previewInfo,
+    targetIsStage: state.scratchGui.targets.stage &&
+    state.scratchGui.targets.stage.id === state.scratchGui.targets.editingTarget,
```
rschamp: This formatting is a little confusing (`state.scratchGui.targets.stage.id` looks like the start of a prop), I would expect something like:
```js
targetIsStage: (
    state.scratchGui.targets.stage &&
    state.scratchGui.targets.stage.id === state.scratchGui.targets.editingTarget
),

// or

targetIsStage: state.scratchGui.targets.stage &&
    state.scratchGui.targets.stage.id === state.scratchGui.targets.editingTarget,
```

src/containers/blocks.jsx
```diff
@@ -445,13 +445,12 @@ Blocks.defaultProps = {
 };
 
 const mapStateToProps = state => ({
-    anyModalVisible: Object.keys(state.modals).some(key => state.modals[key]) ||
-            state.mode.isFullScreen,
-    extensionLibraryVisible: state.modals.extensionLibrary,
+    anyModalVisible: Object.keys(state.scratchGui.modals).some(key => state.scratchGui.modals[key]),
```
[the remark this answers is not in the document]
chrisgarrity: Nope - that needs fixing. It was an update to the conflict...

PR #2096, opened 2018-05-18
Deploy 5/18/18 iteration to Preview

BryceLTaylor:
Features
 * Font tool - You can change the font in the text tool
 * Block toolbox updates - Infrastructure changes, should not change behavior at all
 * Toolbox Padding - Can scroll down below the bottom of the blocks in the blocks toolbox
 * BIG MONITORS - Monitors in fullscreen/player mode should be consistent with Scratch 2.0

PR #2104, opened 2018-05-21
Smoke test branch -> Develop

rschamp:
See #2096 
Pull in updates to smoke test branch into develop, so it is even with master.


rschamp:
Try it out: [smoke](https://llk.github.io/scratch-gui/smoke)

PR #2133, opened 2018-05-23
Show a coming-soon backpack

paulkaplan:
Another small step toward the backpack! The bottom bar with coming soon.

image

@carljbowman I want to make sure the colors on this are right. 

@thisandagain @rschamp I want to make sure I understand what we want in the stand-alone preview build. This PR allows the backpack visibility to be controlled via a prop (to make it easier to remove for offline build, etc.) but I did enable it for the standalone preview build. Is that what we want to do? Or did we want to hide it completely from preview?

### DIFF
--- a/src/components/backpack/backpack.css
+++ b/src/components/backpack/backpack.css
@@ -1,0 +1,20 @@
+@import "../../css/units.css";
+@import "../../css/colors.css";
+
+.backpack-container {
+    flex-shrink: 1;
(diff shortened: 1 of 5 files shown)


carljbowman:
@paulkaplan - Everything looks good to me.

thisandagain:
@paulkaplan As we discussed offline I think this looks good.

PR #2136, opened 2018-05-23
Make new default backdrop included in build

paulkaplan:
### Resolves

Fixes an issue where the default project was still making network requests, something we are trying to avoid

### Proposed Changes

Download the new default backdrop and put it in the default-project folder where the old default backdrop used to be. Update the relevant manifest parts

### Reason for Changes

https://github.com/LLK/scratch-gui/pull/2075 updated the default backdrop to a new blank one, but did not put the file in the default-project folder so that it could be available without making network requests. This fixes that.


thisandagain:
Thanks for catching that @paulkaplan

Issue #2120, opened 2018-05-22
New variables and lists do not appear in the blocks palette if made after the paint editor is used.


paulkaplan:
### Expected Behavior

_Please describe what should happen_

Newly added variables and lists should show up in the toolbox and their monitors should be added to the stage. 

### Actual Behavior

_Describe what actually happens_

After switching back and forth between the costumes tab, newly created variables do not show up. 

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_
Quoting from the [original report from the forum](https://scratch.mit.edu/discuss/topic/299791/?page=128#p3114770)

1. Start a new project.
2. Create variable A and List B.  Notice both A and B show up in the blocks palette and both monitors are on the stage.
3. Click Costumes.
4. Click Code.
5. Create variable C.  Notice that it does not appear in the blocks palette nor on the stage.
6. Drag a show variable block to the script area and select C.  When this block is clicked, variable C still does not appear on the stage.

---

This seems to be related to the toolbox not refreshing, because the variable _is_ actually created: notice that in step 6, you can choose "C" from the dropdown menu. Additionally, if after step 5 you switch to the stage, you can see the variable "c" show up in the toolbox. 

@rschamp this feels like a regression from the work with recycled toolbox.

PR #2146, opened 2018-05-24
Fix toolbox refreshing for new variables after switching tabs

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/2120

### Proposed Changes

re-enable toolbox refreshing, the way we do after doing a real toolbox refresh, after the blocks tab is re-shown after being hidden. 

The full workspace refresh which is done the line before causes refreshing to be disabled. See the comments on line 148-150 for more info.

### Reason for Changes

Toolbox refreshing was not happening when it needed to after switching tabs. 

### Test Coverage

Repro the test case in the lined issue.

### Browser Coverage

not browser related.

----
/cc @gnarf, @rschamp

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -125,6 +125,8 @@ class Blocks extends React.Component {
         if (this.props.isVisible) { // Scripts tab
             this.workspace.setVisible(true);
             this.props.vm.refreshWorkspace();
+            // Re-enable toolbox refreshes without causing one. See #updateToolbox for more info.
+            this.workspace.toolboxRefreshEnabled_ = true;
             window.dispatchEvent(new Event('resize'));
         } else {
             this.workspace.setVisible(false);


src/containers/blocks.jsx
```diff
@@ -125,6 +125,8 @@ class Blocks extends React.Component {
         if (this.props.isVisible) { // Scripts tab
             this.workspace.setVisible(true);
             this.props.vm.refreshWorkspace();
+            // Re-enable toolbox refreshes without causing one. See #updateToolbox for more info.
+            this.workspace.toolboxRefreshEnabled_ = true;
```
gnarf: post merge +1

I had assumed that anytime we needed to `refreshWorkspace` it would also be causing us to change the `toolboxXML` prop which would cause an `updateToolbox` -- thanks for catching this bad assumption here!

Issue #1524, opened 2018-02-27
Share the love to the stage


fsih:
### Expected Behavior

Be able to share blocks from sprites to the stage

### Actual Behavior

Can't share to the stage right now

PR #2151, opened 2018-05-25
Share blocks to the stage

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/1524

sharing-to-stage

/cc @carljbowman I just borrowed the exact styling for the sprite tile for the droppability indication, the hover animation and the dropped indication. The wiggle is a a bit crazy for the long bar, but I kinda love it. Any other thoughts?

### Proposed Changes

Hook up the stage to work the same way as the sprite selector tiles. Stage has a target id the same way the sprite tiles do, so the block drop works without any additional work. 

Also refactors the animation stuff from sprite selector because the autoprefixer does a lot of that automatically.

### Test Coverage

Working on testing different browsers.

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/sprite-selector/sprite-selector.css
+++ b/src/components/sprite-selector/sprite-selector.css
@@ -79,31 +79,11 @@
 }
 
 .raised:hover {
-    -webkit-animation-name: wiggle;
-    -ms-animation-name: wiggle;
-    -ms-animation-duration: 500ms;
-    -webkit-animation-duration: 500ms;
-    -webkit-animation-iteration-count: 1;
-    -ms-animation-iteration-count: 1;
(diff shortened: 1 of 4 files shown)


paulkaplan:
@fsih I updated this PR removing the wiggle, using only the scaling. For your other question, I think that would be a good thing to bring up offline in IPM or something, I don't know the answer.

PR #2165, opened 2018-05-28
add package-lock.json .gitignore

t-kazu:


Kenny2github:
The package-lock.json file is specifically meant to block two pull requests with package modifications from being merged together. Pulls with a modified package-lock.json should remove the package-lock.json file from the commits. The package-lock.json file should not be git ignored.

Also, in the future, please open an [issue](../issues/new) and wait for it to be marked "help wanted" before submitting a pull request.

PR #2185, opened 2018-05-30
Prevent dragging the "+" button

### LINKED ISSUE #2112: Clicking the "+" button in list monitors makes the list look like it's being dragged
### Expected Behavior

The list monitor shouldn't look like it's being dragged when the "+" button is clicked. (That is, the list should not be draggable by the "+" button.)

### Actual Behavior

Clicking the "+" button picks up the list monitor (i.e. gives it a drop shadow), and the monitor can then be dragged.

### Steps to Reproduce

Show a list monitor, and click on the "+" button. The effect is especially noticeable if you hold down the mouse when you click the button.

Firefox 61.0b6, Debian (testing).

Kenny2github:
### Resolves

Fixes #2112 

### Proposed Changes

Adds the "no-drag" class to the "+" button so it doesn't 

### Reason for Changes

To prevent dragging by the "add an item" button 😛

### DIFF
--- a/src/components/monitor/list-monitor.jsx
+++ b/src/components/monitor/list-monitor.jsx
@@ -26,7 +26,7 @@ const ListMonitor = ({draggable, label, width, height, value, onResizeMouseDown,
         </div>
         <div className={styles.listFooter}>
             <div
-                className={styles.addButton}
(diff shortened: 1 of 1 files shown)


Kenny2github:
Woah, that was quick

PR #2187, opened 2018-05-30
Backpack load contents from server

paulkaplan:
### Resolves

Next step for the backpack is to get basic loading of the backpack contents into the list. This PR loads up the first page of contents.

The functionality here is very limited (just loads a single page into the list), but this PR is mostly about infrastructure in terms of backpack-api and session token/username information.

As a compromise to allow the backpack to be tested independently from the servers, I made the backpack container first try to get session information from the state tree matching `scratch-www`, then if that is not available try to get it from url params the same way the host is retrieved. 

I've tested this using the backpack server locally, using the included dummy scratchteam/sessionid user. 

It loads content when the backpack is toggled, and looks like this:
image

Depends on https://github.com/LLK/scratch-gui/pull/2175, the first two commits are shared, look at the last two commits for these changes

### DIFF
--- a/src/containers/backpack.jsx
+++ b/src/containers/backpack.jsx
@@ -2,26 +2,51 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import bindAll from 'lodash.bindall';
 import BackpackComponent from '../components/backpack/backpack.jsx';
+import {getBackpackContents} from '../lib/backpack-api';
+import {connect} from 'react-redux';
 
 class Backpack extends React.Component {
     constructor (props) {
         super(props);
         bindAll(this, [
-            'handleToggle'
+            'handleToggle',
+            'refreshContents'
         ]);
         this.state = {
+            offset: 0,
+            itemsPerPage: 20,
+            loading: false,
             expanded: false,
             contents: []
         };
     }
     handleToggle () {
-        this.setState({expanded: !this.state.expanded});
+        const newState = !this.state.expanded;
+        this.setState({expanded: newState, offset: 0});
+        if (newState) {
+            this.refreshContents();
+        }
+    }
+    refreshContents () {
+        if (this.props.token && this.props.username) {
+            this.setState({loading: true});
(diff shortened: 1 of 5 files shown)


src/lib/backpack-api.js
```diff
@@ -0,0 +1,19 @@
+const getBackpackContents = ({
+    host,
+    username,
+    token,
+    limit,
+    offset
+}) => fetch(`${host}${username}?limit=${limit}&offset=${offset}`, {
```
rschamp: We have been using `xhr` for requests: https://github.com/LLK/scratch-gui/blob/0baf6852fcc0ff42ff39a94e27e39b17378d757b/src/components/costume-canvas/costume-canvas.jsx#L79. Even though this is probably here temporarily, I think we should use the same method for requests everywhere.

src/containers/backpack.jsx
```diff
@@ -0,0 +1,80 @@
...
+    refreshContents () {
+        if (this.props.token && this.props.username) {
+            this.setState({loading: true});
+            getBackpackContents({
+                host: this.props.host,
+                token: this.props.token,
+                username: this.props.username,
+                offset: this.state.offset,
+                limit: this.state.itemsPerPage
+            }).then(contents => {
```
rschamp: There might be an error we need to handle.

src/lib/backpack-api.js
```diff
@@ -0,0 +1,19 @@
...
+    headers: {'x-token': token}
+})
+    .then(d => d.json())
```
rschamp: Not sure if/how we should handle the error here, or just raise it up to the component and let the component do it.

paulkaplan:
@rschamp updated per your comments:
- Using XHR instead of fetch, but wrapped it in promises so I can still use then/catch in the container
- Fixed proptypes
- Added an error state. I used statusCode !== 200 because I saw we do something similar in www, but I could make it more fine-grained if you think.

Issue #2189, opened 2018-05-30
Tutorial cards appear below monitors when brought over the stage.


kchadha:
### Expected Behavior

Tutorial cards should be on top of everything else in the editor.

### Actual Behavior

Tutorial cards appear below monitors when repositioned over the stage.

![image](https://user-images.githubusercontent.com/1786240/40743526-c605b4e4-641f-11e8-858f-a9dd00185371.png)

sjhuang26:
I will work on this.

paulkaplan:
Has been fixed

PR #2226, opened 2018-06-02
Move z-index values to separate file

### LINKED ISSUE #2205: Move all z-index values to separate SCSS file
Z-index values are currently scattered between many SCSS files. This leads to conflicts about which element is on top such as #2189, where two elements each have a z-index of 100. Perhaps moving all z-index values into a separate SCSS file would clarify situations like this.

sjhuang26:
### Resolves

#2205
#2189

### Proposed Changes

Move z-index values into a new file called "z-index.css".

### Reason for Changes

This makes it obvious which components overlap what.

### Test Coverage

No tests needed.

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/cards/card.css
+++ b/src/components/cards/card.css
@@ -1,9 +1,10 @@
 @import "../../css/units.css";
 @import "../../css/colors.css";
+@import "../../css/z-index.css";
 
 .card-container {
     position:absolute;
-    z-index: 100;
+    z-index: $z-index-card;
 }
 
 .left-card, .right-card {
@@ -13,7 +14,7 @@
     background: $ui-white;
     border: 1px solid $ui-tertiary;
     width: 10px;
-    z-index: 99;
+    z-index: 10;
     opacity: 0.9;
     overflow: hidden;
 }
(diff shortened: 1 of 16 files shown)


src/components/cards/card.css
```diff
@@ -13,7 +14,7 @@
     background: $ui-white;
     border: 1px solid $ui-tertiary;
     width: 10px;
-    z-index: 99;
+    z-index: 10;
```
paulkaplan: Shouldn't these be converted to variables as well?

sjhuang26: I don't think it's necessary. The card component creates a stacking context, so the z-index values of left-card and right-card specify only the relative order of elements within the card and are thus independent from the rest of the app.

Issue #2229, opened 2018-06-04
Webpack build changes break Safari 9, 10 and IE11


paulkaplan:
Those browsers see a blank page, it seems to be due to a compilation error on webpack, leaving `const`s in the built code.

mzgoddard:
I think this is us using the new uglifyjs-webpack-plugin. I believe its adding const's in. We need to configure uglifyjs-webpack-plugin to do back-compat for Safari 9, 10 and IE 11.

mzgoddard:
I think it's this change: https://github.com/LLK/scratch-gui/commit/8b7f1d636b04511abb3be758a37deca5505d887c

`node_modules/react-popover/index.js` uses const and the webpack configuration does not babel it.

mzgoddard:
Hmm so scratch-paint has been using react-popover for 8 months https://github.com/LLK/scratch-paint/commit/3eaf8047aa54462a39d6fcb85e910d3412678575 And react-popover has had this const use for the same amount of time.

I'm not sure why this hasn't come up before. It may have been incidentally babel'd before. But I'm not sure I see what step would have done that.

rschamp:
I think this was just missed... this branch from before the build updates also fails in Safari 9: https://llk.github.io/scratch-gui/greenkeeper/react-16.4.0/

paulkaplan:
Fixed by ^ in scratch-render-fonts.

Safari 9 is still broken, always was due to `Intl` polyfill missing.

PR #2233, opened 2018-06-04
Add IE11 and Safari 9/10 tests to browser smoke test using Sauce

paulkaplan:
@BryceLTaylor I wanted to make sure that we don't keep making mistakes w/r/t older browsers, so I gave the hosted VMs on Sauce a try (we had intended to use this in the past, but haven't had the time/need).

I wanted to get this smoke test into the repo so that we can continue building on it in the future. I propose we use this script in the following way:

Step 1 (this PR): Instead of having "test IE and old Safari" on the manual smoketest checklist, use this script instead.

This script can be run locally in the following way 
```
SAUCE_USERNAME=<sauce username> \
SAUCE_ACCESS_KEY=<sauce access key> \
SMOKE_URL=https://llk.github.io/scratch-gui/develop \
npm run test:smoke
```

The sauce username/access key can be a free trial account, or if this is useful, a team account. For reference, the free trial account should have enough build time to satisfy at least 30 weeks worth of running this script once a week.

Stage 2 (future): Run this stage on Travis in a special build stage that is run during PRs to master

We would definitely need a team account for this.

This is enough to catch problems like https://github.com/LLK/scratch-gui/issues/2229

---

Also worth noting, of course, that these tests _fail_ right now, due to #2229, but they pass if I change it to use other browsers.

@BryceLTaylor let me know what you think. We can chat IRL if there is a better way to integrate this into the smoke test process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "test:integration": "jest --runInBand test[\\\\/]integration",
     "test:lint": "eslint . --ext .js,.jsx",
     "test:unit": "jest test[\\\\/]unit",
+    "test:smoke": "jest --runInBand test[\\\\/]smoke",
     "watch": "webpack --progress --colors --watch"
   },
   "author": "Massachusetts Institute of Technology",
--- a/test/helpers/selenium-helper.js
+++ b/test/helpers/selenium-helper.js
@@ -17,6 +17,7 @@ class SeleniumHelper {
             'findByText',
             'findByXpath',
             'getDriver',
+            'getSauceDriver',
             'getLogs',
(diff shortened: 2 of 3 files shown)


paulkaplan:
@BryceLTaylor I updated this test so that the issues from #2229 are covered by this test, Safari 9 is still broken but investigation showed it was always broken due to lack of `Intl` polyfill.

Can this be merged, since it now passes?

paulkaplan:
Yeah, that would be a good step 3. I'm curious how much work it would be to get all of our existing tests to run on several platforms this way.

PR #2253, opened 2018-06-05
Run smoke test if pull request against master

paulkaplan:
Add a step to the travis yml to run the smoke tests automatically when making a PR against master. This was the "step 2" from https://github.com/LLK/scratch-gui/pull/2233

I looked at the documentation, it is kinda strange but my understanding is  TRAVIS_BRANCH is the branch you are PR'ing _against_ when TRAVIS_PULL_REQUEST is true, but it is the branch that is being built when TRAVIS_PULL_REQUEST is false... [src](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,9 @@ addons:
 node_js:
 - 8
 env:
+  global:
   - NODE_ENV=production
+  - SMOKE_URL=https://llk.github.io/scratch-gui/$TRAVIS_PULL_REQUEST_BRANCH
 cache:
   directories:
   - node_modules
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -17,6 +17,9 @@ before_deploy:
 - npm --no-git-tag-version version 0.1.0-prerelease.$(date +%Y%m%d%H%M%S)
 - git config --global user.email $(git log --pretty=format:"%ae" -n1)
 - git config --global user.name $(git log --pretty=format:"%an" -n1)
+script:
+- npm test
+- if [ "$TRAVIS_PULL_REQUEST" != "false" ] && [ "$TRAVIS_BRANCH" == "master" ]; then npm run test:smoke; fi
```
rschamp: `!= "false"` please don't not do this :)  I think `"$TRAVIS_EVENT_TYPE" == "pull_request"` is what you want.

PR #2260, opened 2018-06-05
Add Amazon Polly extension to the extensions library

thisandagain:
Adds the "Amazon Polly" (Text-to-Speech) extension to the library. Still lots of work to do on the extension itself, but this is far enough for testing / feedback.


DeleteThisAcount:
sounds cool!

PR #2261, opened 2018-06-06
Better project loading analytics and errors

paulkaplan:
Fixes an issue where we can't see which projects are giving us errors, and which modes projects are being brought in by (file vs. server)

Making the project load analytics use `label` instead of `value` makes it possible to see which projects are getting loaded, making it easier to see those projects for users with errors.

Including the window.location.hash in the project loader error message also creates a more direct way to see where load errors are coming from.

### DIFF
--- a/src/lib/project-loader-hoc.jsx
+++ b/src/lib/project-loader-hoc.jsx
@@ -44,7 +44,7 @@ const ProjectLoaderHOC = function (WrappedComponent) {
                         analytics.event({
                             category: 'project',
                             action: 'Load Project',
-                            value: projectId,
+                            label: projectId,
                             nonInteraction: true
                         });
                     }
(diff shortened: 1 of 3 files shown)


src/lib/project-loader-hoc.jsx
```diff
@@ -44,7 +44,7 @@ const ProjectLoaderHOC = function (WrappedComponent) {
                         analytics.event({
                             category: 'project',
                             action: 'Load Project',
-                            value: projectId,
+                            label: projectId,
```
kchadha: Is this name change significant? e.g. something (not in this PR) is referring to label and this was a bug fix? Or is this just a clarification change?

paulkaplan: Sorry it is in the PR description, but `value` is only for numeric data in GA and only used for summing, we want the `label` field which allows you to search/list by those properties

Issue #607, opened 2017-08-10
Blocks - add "When this sprite touches [sprite]" hat


ericrosenbaum:
This block would be a hat that uses the same logic as the existing boolean reporter, "touching [mouse-pointer]?", which has a menu to choose the mouse pointer, the edge, or a sprite. 

One reason to add this block is that it would greatly simplify a common case for beginners, where they want to detect that one sprite is touching another, often in a game. One example is in the "Catch Game" Scratch cards:

<img width="343" alt="screen shot 2017-08-10 at 4 26 05 pm" src="https://user-images.githubusercontent.com/567844/29192922-49b49b5e-7df1-11e7-853b-14676b268433.png">

This stack could be reduced to a much simpler one- the new hat would replace the forever block, the if block, and the boolean reporter:

When this sprite touches bowl
Play sound pop
Go to random position
Set y to 180

thisandagain:
I really like this idea.

I'm curious about what should be included in the drop-down here and what the default should be. In 36% of shared projects there is only one sprite. Should we think about matching what is in the `touching [drop down] ?` reporter (see: https://github.com/LLK/scratch-gui/issues/599)?

I'm also curious about the language on the block. A slight variation could be `when touching [drop down]` as to match the `touching [drop down] ?` and be a bit more terse. Thoughts?

/cc @ntlrsk @carljbowman

towerofnix:
>I'm also curious about the language on the block.

[So,](url) `

towerofnix:
>I'm also curious about the language on the block. A slight variation could be when touching [drop down] as to match the touching [drop down] ? reporter and be a bit more terse. Thoughts?

"When touching" might imply that the following blocks will be run *constantly while* touching [argument]. For example, a user might try to do this:

!["when touching mouse"; "turn 15 degrees"](https://user-images.githubusercontent.com/9948030/29199428-581c81e8-7e23-11e7-8630-46a02dc7baed.png)

..and then be surprised to find that the sprite only turns *once*, as soon as the mouse is touching it!

Whereas "when this sprite touches" makes more sense in that case, I think:

!["when this sprite touches mouse"; "turn 15 degrees"](https://user-images.githubusercontent.com/9948030/29199420-47cf98a2-7e23-11e7-81f2-d4c2fc2aa0c4.png)

Plus, I think the actual solution would jump out more clearly with "when this sprite touches"; compare:

[...]

Issue #599, opened 2017-08-10
Blocks - update "touching [drop down] ?" reporter


thisandagain:
Update `touching [drop down] ?` block as per discussions with the resources team. This change simplifies the creation of some "low floor" projects by bypassing the need for conditionals to detect when a sprite hits certain parts of the stage.

### Block Spec
`touching [drop down] ?`

### Defaults
`touching [any edge] ?`

### Drop-down
- `any edge`
- `left edge`
- `right edge`
- `top edge`
- `bottom edge`
- `mouse-position`
- `[-----]`
- `[dynamic sprite list]`

/cc @ntlrsk @carljbowman

Kenny2github:
Shouldn’t it be `mouse-pointer` instead of `mouse-position`?

thisandagain:
@Kenny2github Good catch! Fixed.

mrjacobbloom:
Potentially related to https://github.com/LLK/scratch-blocks/issues/789

amazinigmech2418:
Maybe since this feature will be added, there should be a bounce block for the other detectable edges.

thisandagain:
Decided to hold off on this for now. Can revisit in the future. :-)

ntlrsk:
Before closing - two more options, each adds a block but makes booleans easier to start using.

Option A: 
Make touching sprite more apparent
block 1: **touching** [pointer/ edge / right edge/ left edge /...] 
block 2: **touching** [__] **sprite**

OR Option B:
block 1: **touching** [pointer / sprite names]
block 2: **touching** [any/left/right..] **edge**

Each has downsides but also lowers the floor for booleans since constructions such as <(x position) > 180> take a lot of explaining and dexterity to construct.

Kenny2github:
@ntlrsk I'd go with A since it seems more logical to organize the options that way; I guess the touching () sprite block's default would simply be blank (and therefore always be false)?

Mr-Dave2:
For the sake of beginners you might want to add in the GUI a way to disable these new blocks since they are always 'hot'.  For example, if a program has {"when this sprite touches mouse-pointer", Hide } then trying to manually position that sprite on the stage will be confusing.  This should already be a problem for "when this sprite is clicked" except that event is now a trailing-edge event (which is a separate issue for some current projects).

PR #2265, opened 2018-06-06
Add when touching hat

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/607
Depends on https://github.com/LLK/scratch-vm/pull/1214
Depends on https://github.com/LLK/scratch-blocks/pull/1559

### Proposed Changes

Add a hat block for when a sprite is touching the mouse pointer, edge of the stage, or another sprite.

### Reason for Changes

This is a "low floor" feature, making it easier to make simple projects where you want a sprite to do something when it touches e.g. another sprite. 

### Test Coverage

I built a simple pong game to test when touching a sprite, played with making lots of clones that react when touching the mouse, and made a sprite change color when touching the edge as it bounced around. Generally the behavior should be the same as when using `forever [ if touching? [  ] ]`, except that the hat will run just once instead of continuously while touching. 

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


ericrosenbaum:
@Kenny2github that's right!

Kenny2github:
Should this block have the same new options as the "touching []" block? there was an issue: #599 to add more options to that reporter - should this block have the same options?

PR #2281, opened 2018-06-07
Stage size toggle

### LINKED ISSUE #56: Normal Stage Size/Small Stage Size Toggle
image

cwillisf:
### Resolves

Resolves #56 

### Proposed Changes

The small stage / large stage toggle buttons in the stage header are now functional. Compared to the full "large stage" size, the small stage mode is scaled by 50%. In combination with the existing media query-based scaling, there are now three stage sizes:
- `large`: 100%, when the toggle is in "large" mode and the browser window is wide,
- `largeConstrained`: 85%, when the toggle is in "large" mode but the browser window is more narrow, and
- `small`: 50%, when the toggle is in "small" mode, regardless of the browser window size.

### Reason for Changes

Implements feature described in #56, similar to the Scratch 2.0 stage size toggle.

### Browser Coverage
Mac
 * [x] Chrome 
 
Windows
 * [x] Chrome 
 * [x] Edge


cwillisf:
Note that the icons are currently incorrect in that the large-stage button is always blue and the small-stage button is always gray, but since the colors are in the SVG files themselves I figure that can be fixed in a follow-up change. The buttons currently fade to indicate enabled/disabled state instead of changing the icon color as designed.

src/components/gui/gui.jsx
```diff
@@ -257,6 +277,15 @@ GUIComponent.defaultProps = {
...
     },
-    basePath: './'
+    basePath: './',
+    stageSizeMode: STAGE_SIZES.large
 };
-export default injectIntl(GUIComponent);
+
+const mapStateToProps = state => ({
+    // This is the button's mode, as opposed to the actual current state
+    stageSizeMode: state.scratchGui.stageSize.stageSize
```
rschamp: I think we're conflating two distinct things here. There's the mode, large and small, and there's the display size, which is responsive to screen width.  The display size doesn't change in small mode, but it adjusts in large mode.  We'd never want the `stageSizeMode === STAGE_SIZES.largeConstrained` but that seems like it could be a possibility since it's choosing from the same enumeration.  I think we want to distinguish these things.

src/reducers/stage-size.js
```diff
@@ -32,6 +28,5 @@ const setStageSize = function (stageSize) {
 export {
     reducer as default,
     initialState as stageSizeInitialState,
-    setStageSize,
-    STAGE_SIZES
```
rschamp: I suggest exporting another enum called `STAGE_SIZE_MODES` which is only large and small, to distinguish from the `STAGE_SIZES` (or `STAGE_DISPLAY_SIZES`?).

src/containers/stage-header.jsx
```diff
@@ -45,7 +46,7 @@ StageHeader.propTypes = {
     isFullScreen: PropTypes.bool,
     isPlayerOnly: PropTypes.bool,
     onSetStageUnFull: PropTypes.func.isRequired,
-    stageSize: PropTypes.oneOf(Object.keys(STAGE_SIZES)),
+    stageSize: PropTypes.oneOf(Object.values(STAGE_SIZES)),
```
rschamp: Using `Object.keys` here was intentional, since we want to refer to the values by its key until the value is needed.  I don't think it's consequential in the way that we're using it, but it can have an effect under some optimization circumstances, where the keys of objects are renamed by the optimizer, so you want to use keys rather than values (which are untouched by the optimizer) as much as possible.

rschamp:
Oh looks like there's also a failing integration test, do you know what that's about?

src/components/sprite-info/sprite-info.jsx
```diff
@@ -68,80 +72,116 @@ class SpriteInfo extends React.Component {
...
+            </div>
+        );
+
+        if (stageSize === STAGE_SIZES.small) {
+            return (
+                <Box className={styles.spriteInfo}>
+                    <div className={classNames(styles.row, styles.rowPrimary)}>
+                        <div className={styles.group}>
+                            {
+                                stageSize === STAGE_SIZES.small ?
```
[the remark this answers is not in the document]
cwillisf: Ah, good catch. A previous version had all the checks embedded like this... looks like I forgot to remove some when I put in the big `if`.

src/containers/stage-selector.jsx
```diff
@@ -79,12 +79,11 @@ class StageSelector extends React.Component {
...
-            assetId,
-            id,
-            onActivateTab,
-            onSelect,
-            /* eslint-enable no-unused-vars */
+            assetId: _assetId,
+            dispatchSetHoveredSprite: _dispatchSetHoveredSprite,
+            id: _id,
+            onActivateTab: _onActivateTab,
+            onSelect: _onSelect,
```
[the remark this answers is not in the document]
cwillisf: I didn't know about `_.omit` before... interesting. It does say it's "considerably slower" than `_.pick` but I can't imagine it's slow enough to worry about if it's part of rendering.

Either way, I'd suggest doing something about this as a separate change... is that cool with you?

src/containers/stage-header.jsx
cwillisf: I switched it to `Object.values` because in an earlier version I had `largeConstrained: 'large-constrained'` in the `STAGE_SIZES` enum and that was causing errors. Isn't `Object.values` correct, though, even considering that the keys might be renamed by the optimizer? For example:
```js
const myEnum = {key1: 'value1', key2: 'value2'}
var myProp = myEnum.key1
console.log(myProp); // logs "value1"
```

cwillisf:
I'm not really sure why the integration tests are failing... any guesses? I probably won't be back on this until Monday but if you have thoughts I'd love to hear them :)

src/containers/stage-header.jsx
rschamp: Well either keys or values will work, just will take up different amounts of space depending on  the optimization.

You're right though. We want the prop to be the value, so, you know, we can read it...   `<Component prop={myEnum.KEY} />` not `<Component prop={"KEY"} />` 🤕✋ (face palm)

cwillisf: Good news! They added this emoji with Unicode 9.0: 🤦

cwillisf:
I believe I addressed all of your comments, @rschamp, except for the `Object.values` thing. I looked at the other situations in `scratch-gui` where we use `Object.keys` and I believe they're different from what I have here in that they're named objects instead of a simple enum. I think that's why `Object.values` makes sense here but not in those other places.

Also, I forgot that I said I was going to put off the `omit` thing, so I accidentally went ahead and did it. Oh well.

Also also, I rebased onto `develop` to fix a merge conflict.

paulkaplan:
@cwillisf a couple of things I noticed:
1. `Object.values` does not have full browser support on our supported version list (Safari 10) so you'd need to add a polyfill to use it. We encountered this with scratch-paint previously https://github.com/LLK/scratch-paint/issues/193
2. There is something wrong with the styling of those buttons, they should follow the convention of the visibility toggle (i.e. the active button should be blue, the inactive one should be gray, both should have cursor: pointer) but it currently isn't changing the blue and the small stage button doesnt have cursor: pointer  stage-toggle
3. @carljbowman noticed that the tooltips on the library buttons is z-indexed below the scrollbars, something that never came up before the small-stage layout: 
image

I think (1) should be addressed before merging because it crashes Safari 10 (tested on an iPad running 10.2). I'd be happy to contribute fixes for (2) and (3), since that is very UI-y stuff.

cwillisf:
@paulkaplan:
1. Ah. Phooey. Thanks! Will fix.
2. Yeah, the color is currently embedded in the SVGs and there aren't alternate-color versions of those SVGs, so I'm just using them as-is for the moment. See also [this comment](https://github.com/LLK/scratch-gui/pull/2281#issuecomment-395572079). I didn't notice the cursor thing, though... I guess that's a side effect of the style that I'm using on the "off" part of the toggle. I'll see if I have a reasonable way to fix that. Do we have any other toggles like this in the UI? I can't think of any off the top of my head...
3. I'm happy to fix the z-index thing but I don't know how best to do it. I'll look into the CSS to see if I can learn by example but: is there a (Scratch GUI) convention here? Like "tooltips should always go to z-index 1234" or something?

cwillisf:
OK, I think we're all set now except for the test failure. I'm out of time for that at the moment but I can take another look later/tomorrow.

Also, regarding z-index: what I did is better than it used to be, but there are still problems. I think we need to do a z-index pass over a bunch of the UI as a separate change. Here's an example of the sort of thing that still looks wrong:
image

paulkaplan:
@cwillisf oof, yeah it looks like the stage size add button needs the same z-index. I added a commit to this PR with that change, which fixes the issue from your screenshot. Good to merge now!

cwillisf:
@paulkaplan Oh... I was expecting we'd have to make a) the main buttons, b) the pop-out button strips, and c) the labels for the buttons all have different z-index values. But if setting the z-index of the sage's add button fixes it then that's awesome -- thanks!

cwillisf:
I filed LLK/scratch-gui#2329 to track the "Creating variables" integration test problem.

PR #2303, opened 2018-06-08
Fix "Opening the backdrop library causes the stage backdrop editor to become selected"

### LINKED ISSUE #2188: Opening the backdrop library causes the stage backdrop editor to become selected
### Expected Behavior

The backdrop editor should be selected when I choose a backdrop. If I don't do anything in the library, my place in the editor shouldn't change.

### Actual Behavior

Clicking the "Choose a backdrop" button and then hitting <kbd>Esc</kbd>, leaves the editor in the Stage's paint editor tab.
add-backdrop

### Steps to Reproduce

In a new project, with a sprite's code tab open, click the "Choose a backdrop" button, then hit <kbd>Esc</kbd> (or otherwise close the library without doing anything).

Not browser related

sjhuang26:
### Resolves

#2188 

### Proposed Changes

When the "choose a backdrop" button is pressed (the biggest button in the action menu) but the backdrop library is closed without selecting a backdrop, the Stage's paint editor tab is **not** opened. Instead, nothing happens.

Behind the scenes, there are several changes:
* The BackdropLibrary component is no longer nested in the CostumeTab component; it has been moved to the root GUI component, as discussed in the issue.
* Clicks on the action menu don't propagate up to the stage selector. This way, a click to the action menu won't select the stage as well.
* The third commit ("fix bug") explicitly shifts focus to the stage immediately after a backdrop is added, which is necessary.

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/action-menu/action-menu.jsx
+++ b/src/components/action-menu/action-menu.jsx
@@ -76,8 +76,8 @@ class ActionMenu extends React.Component {
         // (possibly slow) action is started.
         return event => {
             ReactTooltip.hide();
+            if (fn) fn(event);
             this.setState({forceHide: true, isOpen: false}, () => {
-                if (fn) fn(event);
                 setTimeout(() => this.setState({forceHide: false}));
             });
         };
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -15,6 +15,8 @@ import StageWrapper from '../../containers/stage-wrapper.jsx';
 import Loader from '../loader/loader.jsx';
 import Box from '../box/box.jsx';
 import MenuBar from '../menu-bar/menu-bar.jsx';
+import CostumeLibrary from '../../containers/costume-library.jsx';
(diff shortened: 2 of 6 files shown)


src/components/action-menu/action-menu.jsx
```diff
@@ -76,8 +76,8 @@ class ActionMenu extends React.Component {
         // (possibly slow) action is started.
         return event => {
             ReactTooltip.hide();
+            if (fn) fn(event);
```
paulkaplan: Why is this change made? The point of the forceHide:true => fn(event) => forceHide: false is to allow the menu to close before the (possibly long running) callback is called.

PR #2317, opened 2018-06-11
use new SoundPlayer in SoundLibrary

mzgoddard:
Depends on llk/scratch-audio#88

### Resolves

- Stopping sounds in SoundLibrary fades them out avoiding potential audio clipping

### Proposed Changes

Use new `AudioEngine.decodeSoundPlayer` to decode sounds in SoundLibrary and have a promise resolving to a SoundPlayer instance returned. This SoundPlayer can control playback on its own and fades out when told to stop playback.

### Reason for Changes

`scratch-audio` is being rewritten to help resolve vm bugs like `deleting a clone stops sprite sounds`. SoundLibrary depends on a small part of `scratch-audio` so that is being updated to use the slightly different new API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "redux-mock-store": "^1.2.3",
     "redux-throttle": "0.1.1",
     "rimraf": "^2.6.1",
-    "scratch-audio": "0.1.0-prerelease.1528394075",
+    "scratch-audio": "0.1.0-prerelease.1528996828",
     "scratch-blocks": "0.1.0-prerelease.1528907522",
     "scratch-l10n": "3.0.20180611175036",
     "scratch-paint": "0.2.0-prerelease.20180614161221",
--- a/src/containers/sound-library.jsx
+++ b/src/containers/sound-library.jsx
@@ -29,20 +29,63 @@ class SoundLibrary extends React.PureComponent {
             'handleItemMouseEnter',
             'handleItemMouseLeave'
         ]);
+
+        /**
+         * AudioEngine that will decode and play sounds for us.
+         * @type {AudioEngine}
+         */
+        this.audioEngine = null;
+        /**
+         * A promise for the sound queued to play as soon as it loads and
+         * decodes.
+         * @type {Promise<SoundPlayer>}
+         */
+        this.playingSoundPromise = null;
     }
     componentDidMount () {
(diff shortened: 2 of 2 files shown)


paulkaplan:
@ericrosenbaum I assigned this to you and added the `blocked` label, when the corresponding scratch-audio PR goes in, update the `package.json` with that update in this PR and then it can be review/merged

ericrosenbaum:
There are some situations in which the sounds fail to stop on mouse leave, which results in a sound (or several) that is playing that cannot be stopped, even after closing the library. One way to reproduce this is to click the "loops" filter in the sound library and move the mouse rapidly over several sounds.

paulkaplan:
@ericrosenbaum it looks like this is because the `handleMouseEnter` creates the sound player in an async way, but the corresponding `handleMouseLeave` tries to stop the sound right away, meaning it is possible to mouse in and mouse out in a way that causes the "stop" to be attempted before the sound player has loaded, leaving it running. Furthermore, the reference to the sound player gets overwritten the next time a new sound is played, causing it to keep playing even after closing the modal.

@mzgoddard one fix for this is just to add a check before replacing the sound player reference to make sure to stop any existing sound players. Something like adding this to line 60
```
           .then(soundPlayer => {
                if (this.playingSound !== null) {
                    this.playingSound.stop();
                }
                this.playingSound = soundPlayer;
```

---

A related existing bug, not necessary to fix in this PR: it is possible to close the modal in the time between triggering a load and when the sound player is loaded, which i can replicate by slowing down the network and mousing into a loop and hitting `esc` before the network loads the sound.

For that kind of case, react recommends using [cancelable promises](https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html). Fixing in that way could fix both problems at once in a more complete way, since the code suggestion above could leave a sound playing after you mouse out, but before you mouse into another one or exit.

mzgoddard:
@paulkaplan Good point. I think we can fix the duplicate playback without cancelable promises and instead store the promise and put the stop instructions as additional callbacks to the promise. That'll have it perform some extra work but I think that'll be easier to understand than cancelable promises?

ericrosenbaum:
This is amazing! Sounds so great just mousing over the sound library now.

mzgoddard:
This might fix https://github.com/LLK/scratch-audio/issues/82?

Issue #2325, opened 2018-06-13
Enable the language menu


chrisgarrity:
First part of implementing gui localization is to re-enable the language menu. [Note: this doesn't include actually changing the editor interface/blocks, but it's a prerequisite]

It should still default to disabled for most users, but have a way to enable it for testing.

PR #2326, opened 2018-06-13
Enable language selection menu

chrisgarrity:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/2325

### Proposed Changes

Allows the language selection menu to be active if `?enable=language` is in the URL, otherwise shows a ‘coming soon’ tooltip. This can also land on a branch if we don’t want that available on preview.

- language menu opens if the globe is clicked
- language menu closes if mouse is clicked anywhere outside the language selector
- language list is imported from scratch-l10n
- current locale is updated if another language is selected (i.e. the current language selected changes)

Not in this PR:
- styling of language selector
- actually changing the messages!

You can try it at https://chrisgarrity.github.io/scratch-gui/feature/enable-language-selector/?enable=language

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/language-selector/language-selector.css
+++ b/src/components/language-selector/language-selector.css
@@ -1,23 +1,6 @@
 @import "../../css/colors.css";
 @import "../../css/units.css";
 
-.group {
-    z-index: 50;
-    top: $menu-bar-height;
-    left: 0;
-    position: absolute;
-    border: 1px solid $ui-black-transparent;
-    border-radius: 0 0 8px 8px;
-    background-color: $motion-primary;
-    padding: 0;
-    margin: 0;
-    min-width: 186px;
-    max-width: 260px;
-    overflow: visible;
-    color: $ui-white;
-    box-shadow: 0 8px 8px 0 $ui-black-transparent;
-}
-
 .language-icon {
     height:  1.5rem;
 }
@@ -28,7 +11,6 @@
 
 .language-select {
     margin: .5rem;
-    width: calc(100% - 1rem);
     height: 1.85rem;
     border: 1px solid $motion-primary;
     user-select: none;
--- a/src/components/language-selector/language-selector.css
+++ b/src/components/language-selector/language-selector.css
@@ -2,10 +2,20 @@
 @import "../../css/units.css";
 
 .group {
-    display: inline-flex;
-    flex-direction: row; /* makes columns, for each label/form group */
-    align-items: center;
-    vertical-align: middle;
+    z-index: 50;
+    top: $menu-bar-height;
+    left: 0;
+    position: absolute;
+    border: 1px solid $ui-black-transparent;
+    border-radius: 0 0 8px 8px;
+    background-color: $motion-primary;
+    padding: 0;
+    margin: 0;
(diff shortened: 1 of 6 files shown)


src/components/language-selector/language-selector.css
```diff
@@ -2,10 +2,20 @@
...
+    top: $menu-bar-height;
+    left: 0;
+    position: absolute;
+    border: 1px solid $ui-black-transparent;
+    border-radius: 0 0 8px 8px;
+    background-color: $motion-primary;
+    padding: 0;
+    margin: 0;
+    min-width: 186px;
+    max-width: 260px;
```
paulkaplan: Such specific fixed values are pretty suspicious in CSS. It seems the same thing can be achieved by not setting a min/max width here and removing the width from `.language-select` below?

src/components/language-selector/language-selector.jsx
```diff
@@ -1,56 +1,49 @@
...
-                        src={dropdownCaret}
-                    />
-                </React.Fragment>
-            )}
-        </div>
-    </Box>
-);
-
+                    <select
+                        aria-label="language selector"
```
paulkaplan: I don't think `aria-label` is appropriate here, [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) says 

> The aria-label attribute is used to define a string that labels the current element. Use it in cases where a text label is not visible on the screen.

I think the right place for the aria-label would be on the "globe" icon button, on whatever element the click is bound to. Additionally, I think the aria-label should be in a `formatMessage(...)`

src/containers/language-selector.jsx
```diff
@@ -1,18 +1,85 @@
...
+            'handleClick',
+            'ref'
+        ]);
+        if (props.open) this.addListeners();
+    }
+    componentDidUpdate (prevProps) {
+        if (this.props.open && !prevProps.open) this.addListeners();
+        if (!this.props.open && prevProps.open) this.removeListeners();
+    }
+    addListeners () {
```
paulkaplan: This logic seems to be the same as what is used for the dropdown menus in general, do we need to duplicate this here? Doesn't the `Popover` or `Dropdown` component do this "close by clicking outside" thing?

src/containers/language-selector.jsx
```diff
@@ -1,18 +1,85 @@
...
+        document.addEventListener('mouseup', this.handleClick);
+    }
+    removeListeners () {
+        document.removeEventListener('mouseup', this.handleClick);
```
paulkaplan: The use of `mouseup` here will prevent this from working correctly on touch devices. In addition to sharing the logic with the `menu` component, could you add `touchstart` and `touchend` handlers to support closing the menus on touch?

PR #2333, opened 2018-06-13
Prune gh-pages to match our branch list after each build

rschamp:
### Proposed Changes

Trim the directories in gh-pages after each build to match our active branches. It also begins building all branches, including master into a subdirectory on gh-pages, and due to this change, redirects llk.github.io/scratch-gui/ to llk.github.io/scratch-gui/master/. We should land #2334 first so the gh-pages branch is already set up for this change.

### Reason for Changes

Right now there is no cleaning up of directories in gh-pages when branches are deleted. This means that our gh-pages branch is continually growing, and occasionally gets to the point where GitHub fails to build the gh-pages site after successful pushes to the gh-pages branch.  This in turn causes confusion by developers when changes from a successful build aren't seen.

### Test Coverage

I've tested this out on my own fork. It does what I expect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,10 +36,8 @@ deploy:
     branch: master
   skip_cleanup: true
   script: npm run --silent deploy -- -x -a -r https://${GH_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git
-# - provider: script
-#   on:
-#     branch: greenkeeper/*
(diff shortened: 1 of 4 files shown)


rschamp:
[:+1:](https://git-man-page-generator.lokaltog.net/#cda47e7778a9b295fa98a2e387f8eb65)

PR #2354, opened 2018-06-14
Fix target area max-width problem

paulkaplan:
I don't love this solution, but I remember not being able to figure out a way to not specify the max width because it is needed in order for the flowing to work. Without the max-width, the sprite tiles try to be in a row, pushing over everything else.

### Resolves

Fixes the small-stage (and constrained stage) sizes when you have multiple sprites.

### Proposed Changes

Adjust the max-width depending on the stage-size

image
image
image

There seems to be an unrelated issue, I think introduced by me in the reordering PR, that makes the sprite tiles sometimes take up too much space, you can see the empty space in the small stage pic above. That is definitely unrelated to this though.

I'd love to figure out a non-hard-coded way to do this, btw, but I haven't been able to figure it out yet.

### DIFF
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -240,7 +240,7 @@ const GUIComponent = props => {
                             ) : null}
                         </Box>
 
-                        <Box className={styles.stageAndTargetWrapper}>
+                        <Box className={classNames(styles.stageAndTargetWrapper, styles[stageSize])}>
                             <StageWrapper
                                 isRendererSupported={isRendererSupported}
                                 stageSize={stageSize}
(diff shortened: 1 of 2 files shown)


src/components/gui/gui.jsx
```diff
@@ -240,7 +240,7 @@ const GUIComponent = props => {
                             ) : null}
                         </Box>
 
-                        <Box className={styles.stageAndTargetWrapper}>
+                        <Box className={classNames(styles.stageAndTargetWrapper, styles[stageSize])}>
```
cwillisf: I'm not familiar with this syntax... this looks to me like it would pick either:
- `styles.stageAndTargetWrapper` and `'.large'`, or
- `styles.stageAndTargetWrapper.large` but not `styles.stageAndTargetWrapper`

My guess is that my interpretation is incorrect, but could you explain what's actually going on here?

Issue #2401, opened 2018-06-19
Sharing a costume between sprites shares the wrong costume


BryceLTaylor:
### Expected Behavior

Dragging a costume from one sprite to another should copy that costume to the other sprite

### Actual Behavior

The sprite you drag over becomes costume 1.  The remaining costumes move down.  Any costumes with a higher number than the one dragged stay in the same place.

Then the costume that is now in the position that the one you dragged WAS in gets copied to the other sprite.

### Steps to Reproduce

Create a new project
Add a new sprite from library
Choose the buildings sprite (it works with others, but this one has a lot of costumes.)
Select the costumes tab
Select one of the middle buildings (#4-7 ish)
Drag that costume from the assets list to the cat sprite in the sprite list.
Note: the costume you just dragged is now at the top of the list and the others have moved down
Select the Cat sprite
Note, it has the costume that, in the buildings sprite, NOW has the same number as the one you dragged used to have

![wrongcostume2](https://user-images.githubusercontent.com/1088291/41617989-e7f93732-73cf-11e8-9170-055cf065fef1.gif)

### Operating System and Browser

Chrome on Mac

PR #2391, opened 2018-06-19
Backpack 4: adding costumes and sounds from the backpack

paulkaplan:
backpack-adding-costume-sounds

Based on https://github.com/LLK/scratch-gui/pull/2187, only one commit ahead, so only need to review this commit: 
https://github.com/LLK/scratch-gui/commit/b1f1c3ac0b51405a43e5ea67d7766ab05c0d7f5d

This adds drop handlers to the costumes/sounds tabs to allow them to respond to drop events by either (a) adding the asset directly if it is of the same type (i.e. costumes tab + BACKPACK_COSTUME) or switch to the correct tab and then add it (i.e. costume tab + BACKPACK_SOUND switches to the sounds tab, then adds the sound, and vice versa)

PR #2404, opened 2018-06-19
Fix sharing assets ordering

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/2401

### Proposed Changes

Make sure not to call onDrop unless the newIndex is an actual value. This makes sure that the sorting onDrop from the costume tab does not get called before the onDrop from the sprite selector, which was leading to the appearance that the wrong costume was being shared in #2401 

Tested by going through the same process as listed in #2401


paulkaplan:
Rolled into https://github.com/LLK/scratch-gui/pull/2391, closing

PR #2405, opened 2018-06-19
Initial version of loading or switching language.

chrisgarrity:
Add `?locale=<localecode>` to load in a new langage.
Add `?enable=language` to enable the language selector
Or combine the two: `?enable=language&locale=<localecode>`
Still accepts a hash, but requires that the hash be a sequence of digits.

Added new locales reducer to gui:
- initializes default  react-intl/localeData
- initializes default locale messages for all locales
- default locale is ‘en’ (can be overridden by URL)

Removed dependency on react-intl-redux, added ConnectedIntlProvider (uses correct locale messages from redux based on current locale)

### Proposed Changes

This PR depends on https://github.com/LLK/scratch-blocks/pull/1596

* Language selector dispatches an update to the IntlProvider with new locale and locale messages
* `blocks.jsx` has a new method `setLocale` that runs on mount and when the locale changes.
* `setLocale` refreshes the workspace and toolbox
* GUI ui updates based on the change to the IntlProvider
* Locales is a new reducer that is initialized with messages for all locales.
* removed dependency on `react-intl-redux` it made things more complicated than needed.

### Reason for Changes

Support Scratch in multiple languages.

### Test Coverage

Hash parser now requires that project id be a sequence of digits -  modified unit test to work with numbers and use the default project id if hash is not a number.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


chrisgarrity:
@paulkaplan the build failure is due to the dependency on the outstanding scratch-blocks PR. Tests all pass when using a local scratch-blocks that includes the https://github.com/LLK/scratch-blocks/pull/1596 fix

src/containers/blocks.jsx
```diff
@@ -461,8 +472,8 @@ const mapStateToProps = state => ({
         state.scratchGui.mode.isFullScreen
     ),
     extensionLibraryVisible: state.scratchGui.modals.extensionLibrary,
-    locale: state.intl.locale,
-    messages: state.intl.messages,
+    locale: state.scratchGui.locales.locale,
+    messages: state.scratchGui.locales.messages[state.scratchGui.locales.locale],
```
rschamp: Can't the reducer do this work for everyone, so we can continue to refer to just `state.scratchGui.messages`?

src/containers/language-selector.jsx
```diff
@@ -36,16 +39,19 @@ class LanguageSelector extends React.Component {
...
     children: PropTypes.node,
+    currentLocale: PropTypes.string.isRequired,
+    locales: PropTypes.objectOf(PropTypes.objectOf(PropTypes.string)),
     onChangeLanguage: PropTypes.func.isRequired
 };
 
 const mapStateToProps = state => ({
-    currentLocale: state.intl.locale
+    currentLocale: state.scratchGui.locales.locale,
+    locales: state.scratchGui.locales.messages
```
rschamp: Looking at https://github.com/chrisgarrity/scratch-gui/blob/a04554229f9311595c5d1624839f01bf1c6a8dc7/src/components/language-selector/language-selector.jsx#L29, it doesn't look like we even need to pass this in, since it's taken directly from `scratch-l10n`?  It doesn't seem right that the value of `locales` is set to `messages`.

src/reducers/locales.js
```diff
@@ -0,0 +1,53 @@
...
+import editorMessages from 'scratch-l10n/locales/editor-msgs';
+
+addLocaleData(localeData);
+
+const UPDATE_LOCALES = 'scratch-gui/locales/UPDATE_LOCALES';
+const SELECT_LOCALE = 'scratch-gui/locales/SELECT_LOCALE';
+
+const initialState = {
+    locale: 'en',
+    messages: editorMessages
```
rschamp: To avoid the need to look up messages by locale, while still allowing something to update the `messages`, maybe we should hold two things in state, `messagesByLocale` and `messages`. Where every time `locale` or `messagesByLocale` is updated, we set `messages` to `messagesByLocale[state.locale]`. Then connected components can just map `messages: state.messages` rather than needing to do the lookup and have knowledge about the internal workings of intl?

rschamp: That's to say, here maybe we can have
```js
initialState = {
    locale: 'en',
    messagesByLocale: editorMessages,
    messages: editorMessages['en']
}
```

chrisgarrity:
@rschamp BTW, tests are still expected to fail - until scratch-blocks PR lands.

PR #2407, opened 2018-06-19
Backpack step 5: saving costumes and sounds

paulkaplan:
backpack-saving

Add some basic payload-ifiers for packaging costumes and sounds for the backpack, and add the drop detecting code for triggering the saves. @thisandagain Note how the indeterminate order issue causes the ordering to be... indeterminate.


src/containers/backpack.jsx
```diff
@@ -2,18 +2,24 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import bindAll from 'lodash.bindall';
 import BackpackComponent from '../components/backpack/backpack.jsx';
-import {getBackpackContents} from '../lib/backpack-api';
+import {getBackpackContents, saveBackpackObject, deleteBackpackObject, soundPayload, costumePayload} from '../lib/backpack-api';
```
amazinigmech2418: The length is 128 which is above 120 which is the maximum character count per line. You should maybe change the names to shorter versions like gBackPackContents, saveBackpackObj, delBackpackObject, sndPayload, and costumePayload can stay the same.

amazinigmech2418: It would reduce the length to 118.

src/components/asset-panel/selector.jsx
```diff
@@ -15,6 +15,7 @@ const Selector = props => {
         buttons,
         containerRef,
         dragType,
+        dragPayload,
```
amazinigmech2418: 18:9  error  'dragPayload' is assigned a value but never used. Allowed unused vars must match /^_/  no-unused-vars
  18:9  error  'dragPayload' is missing in props validation                                           react/prop-types

Maybe, it isn't defined in props...

amazinigmech2418:
@paulkaplan My comments may help you fix the Travis errors.

src/containers/backpack.jsx
towerofnix: Or separate them onto different lines, like so:

```js
import {
  getBackpackContents,
  saveBackpackObject,
  deleteBackpackObject,
  soundPayload,
  costumePayload
} from '../lib/backpack-api'; 
```

This wouldn't take any refactoring like your suggestion (where you'd need to rename the functions in backpack-api *and* here), and the function names would stay as englishy as they are now, so probably a little more convenient.

paulkaplan:
@rschamp sorry just got around to fixing the lint errors.

src/components/backpack/backpack.jsx
```diff
@@ -62,18 +66,24 @@ const Backpack = ({contents, error, expanded, loading, onToggle}) => (
...
+                                })}
+                            >
                                 {contents.map(item => (
                                     <SpriteSelectorItem
                                         className={styles.backpackItem}
                                         costumeURL={item.thumbnailUrl}
                                         details={item.name}
                                         dragPayload={item}
                                         dragType={dragTypeMap[item.type]}
+                                        id={item.id}
```
rschamp: I had to do a double-take on this prop name. I checked to remember how it's used and it doesn't cause an issue (we never pass `id` down to the rendered component). But we should consider using a different prop name than `id` for this since it's easily read as the HTML `id` attribute, which would be inappropriate to use in a reusable component.

src/lib/backpack/costume-payload.js
```diff
@@ -0,0 +1,37 @@
+import jpegThumbnail from './jpeg-thumbnail';
+import storage from '../storage';
+
+const costumePayload = costume => {
+    // TODO is it ok to base64 encode SVGs? What about unicode text inside them?
```
rschamp: I think base64 encoding unicode should work fine. I don't see any reason why encoding SVGs this way wouldn't work.

src/lib/backpack/sound-thumbnail.js
```diff
@@ -0,0 +1,3 @@
...
...IGAgIDAQAAAAAAAAABEQAhECAxQVFhQIEwcZGhsfD/2gAMAwEAAhEDEQAAAdUAAAAAIVk2UACFZz2pwvSa1x19ordLrNGVdsVNDvedjlXppXLQy7ril0He87GKzTSuWgKazOOudU5a9RWaXuaAAAAAAAB//9oACAEBAAEFAvGWbFE/E82KBZt8JGuc5zoa/SOdogV1myZzaKq2HE8b30lAX3ZW7BbcW4iMncTyuZSfnxeqV2z2/M2rjyWHjGAWMPUM8dpY9bWJD6X/2gAIAQIAAQUC6Ws0wXBMFx3hvpf/2gAIAQMAAQUC+Af/2gAIAQICBj8CAf/aAAgBAwIGPwIB/9oACAEBAQY/Aur9uSUK5y6VpZApOVFeV8Tn15D60WY3J1moCf5r9qabWFF8KYcPA6+m98O7TNfbphKnBFjLDaLVNyz4pa4G/A0I/wBmjOmQsMHN1OmEAedUUn5XqbmWwS3CD3mhJtZpD0DgElycfmvakXxZdu6oo5BqRQyndTw+UMLYZUzFuN2wv2fBf//aAAgBAQMBPyH4w+6y888e8wK/Yd2OYJamasJkfQXL6ikQPlLqtEOQX1+tJ6/x3VtNruFuMYphTp9AYwnOihP0s1zyk4sfdqu5Y7GJP6GLvTiVE2cZP3XRH9ointmkOrvq1QQW6QXf65Is2rxfoalPUHFc4PKtckUTcIiUEb+UburVx0sIgbb6+F//2gAIAQIDAT8h+OM0UYHBIx0YmuF0zDXC6ZCk04MvC//aAAgBAwMBPyHwpzuQyOQyxkjwv//aAAwDAQACEQMRAAAQkkkkkhkklIrEhkPEnkPEk6Pkkkkkkkkk/9oACAEBAwE/EPjj8Ql0GhYJP6UIgjI3EylYiQdlZgdxTFjQf2wHCn2aU4Yj6glVdVokAKqA31aJdngG0sDpd3oXX8WyOINm3HOLiiIlvAAejFrIbqFisawfSkBz6gD8RlZ90xKenkOY72Ng0hMc/A2cVHA50nk/5TSE7IQn4yF33UxR3IRz6DFBESRsjSwYJBTwiTd34oYrYAQuoCx6/tIGo6gKAkslIWGBBsIIQO1OlGBFumAquhfo8L//2gAIAQIDAT8Q+MTmE6Vy4amgljJtDF30N6vxK2JvpbVbkIUBJpXoYZ8Of//aAAgBAwMBPxD45zLTgNKWhnHVi6YDXHVi6YDXIypKKSiHhf/Z';
```
rschamp: I think we should use webpack for this so it's easier to update.

src/containers/backpack.jsx
```diff
@@ -32,13 +44,61 @@ class Backpack extends React.Component {
             storage._hasAddedBackpackSource = true;
         }
     }
+    componentWillReceiveProps (newProps) {
+        const dragTypes = [DragConstants.COSTUME, DragConstants.SOUND];
+        if (newProps.dragInfo.dragging && !this.props.dragInfo.dragging) {
+            this.dropAreaRect = this.ref && this.ref.getBoundingClientRect();
+        } else if (!newProps.dragInfo.dragging && this.props.dragInfo.dragging && this.state.dragOver) {
```
rschamp: Comments to describe each of these cases would make this more readable

src/components/backpack/backpack.jsx
rschamp: (I didn't mean to change this now, just a sidebar note)

src/containers/backpack.jsx
paulkaplan: Good call. I'll probably extract this in the next pr, when more areas become "drop areas", in a similar way to the `SortableHOC`, it'll probably become the `DropAreaHOC`. But yeah added comments, which i'll keep in if/when it is extracted.

paulkaplan:
@rschamp couldn't figure out how to get that thumbnail to load as a base64 encoded string, but i'm willing to work with you on it if you want? I'd rather replace it with a real waveform thumbnail anyway ;)

paulkaplan:
@rschamp pushed a commit with a jpg + base64 encoder, good call.

Issue #2073, opened 2018-05-15
Changing a variable with a slider monitor from a number to a string moves slider to middle


kchadha:
### Expected Behavior

Changing a variable with a visible slider monitor from a number to a word should not make the slider move to the middle.

### Actual Behavior

Changing a variable with a visible slider monitor from a number to a word makes slider jump to the middle.

### Steps to Reproduce

1. Create a variable and monitor it on the stage.
2. Change the monitor to a slider monitor.
3. Make sure the variable is set to 0 to begin with.
4. Change the variable to a word 'hello'
5. See the slider on the monitor jump to the middle.

####
Notes: In 2.0, a word is always at position 0 in the slider monitor, so if the variable was set to 50, and then set to 'hello,' the slider should reset itself to the beginning.

PR #2419, opened 2018-06-20
Fix booleans not showing up in monitors

### LINKED ISSUE #2409: Variable & list monitors don't display boolean values
### Expected Behavior

Variable monitors that have boolean values (true or false) and list monitors that contain boolean values should display these values correctly (i.e. as the strings "true" or "false").

### Actual Behavior

Variable monitors don't show true/false at all. List monitors only show the value while the list item's input is selected.

### Steps to Reproduce

Create a variable and set its value to a boolean value (e.g. by using a block such as "1 = 1"). Create a list and add a boolean value to it.

Firefox 61.0b14, Debian (testing).

---

I created [this repro project](https://scratch.mit.edu/projects/229846207/) and compared the results in 2.0 and 3.0:

2.0 results 3.0 results

towerofnix:
### Resolves

Fixes #2409.

### Proposed Changes

Makes boolean values (true, false) show up in (list and variable) monitors correctly. Before, they wouldn't show up at all.

### Reason for Changes

Compatibility with 2.0 and convenience.

### Test Coverage

Manually tested with [this project](https://scratch.mit.edu/projects/229992327/). Screenshots below. From left to right: 2.0, 3.0 before this PR, 3.0 with this PR.

"true, true, true, false, false, false, apple" "(blank), (blank), 1, (blank), (blank), 0, NaN "true, true, NaN, false, false, NaN, NaN"

Lists also tested (no screenshots here for brevity). They work right.

### Browser Coverage

N/A (tested on firefox/debian).

### DIFF
--- a/src/containers/slider-monitor.jsx
+++ b/src/containers/slider-monitor.jsx
@@ -15,7 +15,7 @@ class SliderMonitor extends React.Component {
         ]);
 
         this.state = {
-            value: Number(props.value)
+            value: props.value
         };
     }
     componentWillReceiveProps (nextProps) {
--- a/src/lib/monitor-adapter.js
+++ b/src/lib/monitor-adapter.js
@@ -28,6 +28,16 @@ export default function ({id, spriteName, opcode, params, value}) {
     if (typeof value === 'number') {
         value = Number(value.toFixed(6));
     }
-    
+
+    // Turn the value to a string, for handle boolean values
+    if (typeof value === 'boolean') {
+        value = value.toString();
+    }
+
+    // Lists can contain booleans, which should also be turned to strings
+    if (Array.isArray(value)) {
+        value = value.map(item => item.toString());
+    }
+
     return {id, label, category, value};
 }


amazinigmech2418:
The true and false sliders should say 1 and 0 since it should change it to a number, so you should do another conditional for if the monitor is requesting a number or string/boolean.

towerofnix:
Okay, made that change that dewonkifies sliders:

The image for variables in 2.0 that was used earlier The variables in 3.0

This almost matches 2.0, besides the positioning of the slider handle for boolean values. Since, with this PR, booleans are treated like strings, the sliders are always positioned at 1 (see #2073), but like @amazinigmech2418 suggested above, they *should* be positioned at 1 or 0 (for true or false), because that's what their values are when you number-ify them. But that's only true for *boolean* values, and as far as the slider knows, these are *string* values (because this PR stringifies them).

paulkaplan:
Great thanks!

Issue #2101, opened 2018-05-21
Add "Sprite Library" button to sprite selector dropdown. Same as costumes/sounds/backdrops


apple502j:
I saw many comments in feedback that the sprite library is missing. I think It's because they didn't get noticed that the the cat icon is clickable and it shows the library.
I can think two ways:
1. Revert #1759 
2. Add a button to the library to the dropdown and make the cat icon unclickable

hedgehog125:
Maybe the text on the tooltip could be bold and there could be a delay before the pulldown opens? That way people might see it?

carljbowman:
## Add addtional menu item for explicitly adding from Library 

### Overview
We have received feedback (this issue) and observed in workshops that some folks would like to have an explicit menu option for the library. The fact that the add button opens like a tray menu leads people to think they need to select an option from that menu rather than click the main button. 

#### Modification
Add addtional menu item (search icon) to each add menu system that explicitly opens the library. We should allow for redundancy, allowing the main button (circle) to open the library as well.

#### Asset
[search-sprite-library.svg.zip](https://github.com/LLK/scratch-gui/files/2080427/search-sprite-library.svg.zip)

### Add Sprite
![3 0 add sprite](https://user-images.githubusercontent.com/3409578/41099848-be65d5d4-6a2d-11e8-8f7d-80bb80256a7b.png)

### Add Backdrop
![3 0 add backdrop](https://user-images.githubusercontent.com/3409578/41099855-c4d0c56e-6a2d-11e8-8b3e-5cba31bbcdd6.png)

### Add Costume
![3 0 add costume](https://user-images.githubusercontent.com/3409578/41099861-c997bd32-6a2d-11e8-99ce-c958fbd4d564.png)

### Add Sound
![3 0 add sound](https://user-images.githubusercontent.com/3409578/41099880-d21a5bf4-6a2d-11e8-8330-c6876713467c.png)

/cc @ntlrsk @SpeakkVisually @kosiecki

carljbowman:
Assigning to @thisandagain for triage

PR #2424, opened 2018-06-20
Additional menu item for explicitly adding from Library

thisandagain:
### Resolves
Resolves GH-2101

### Proposed Changes
Adds a new item to the action menu for sprites, backdrops, sounds, and costumes which explicitly goes to the library as discussed / described in #2101 

### Reason for Changes
The hope is that this will help resolve some issues we have seen during testing (and in the forums) of Scratchers not being able to find the libraries.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


src/components/sprite-selector/sprite-selector.jsx
```diff
@@ -127,6 +128,10 @@ const SpriteSelectorComponent = function (props) {
                         title: intl.formatMessage(messages.addSpriteFromPaint),
                         img: paintIcon,
                         onClick: onPaintSpriteClick // TODO need real function for this
```
fsih: Do you know what these TODOs mean and if they are still relevant?

thisandagain: I don't!

PR #2433, opened 2018-06-21
Update vm and audio

ericrosenbaum:


gnarf:
On it, will try to fix asap

gnarf:
got the fix into the VM on llk/scratch-vm#1264

gnarf:
test

gnarf:
I went around the merge button here and pushed to develop since I was specifically told to just land it if the tests passed by @rschamp -- we can always reverteroo

PR #2458, opened 2018-06-25
Try to run in full screen on tablets

chrisgarrity:
Using `bowser` to detect tablets, and type of browser to call either `webkitRequestFullScreen()` or `mozRequestFullScreen()`

Full screen request must happen in reponse to a user-action, so it’s included in the ‘try-it’ handler for the preview modal.

We'll need to figure out where this goes when we no longer have the preview-modal.

### Resolves

Parts of the gui UI end up getting cut off on 7-inch tablets due to the amount of browser 'chrome'  (tabs, URL, menus etc). Detecting that it's a tablet and requesting full screen when the user decides to 'Try It' in the preview modal makes the UI much more usable.

### Proposed Changes

Requests that the browser run in full screen mode if it's a tablet

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [x] Chrome

### DIFF
--- a/src/lib/tablet-full-screen.js
+++ b/src/lib/tablet-full-screen.js
@@ -5,10 +5,10 @@ import bowser from 'bowser';
  */
 export default function () {
     if (bowser.tablet) {
-        if (bowser.webkit || bowser.blink) {
+        if ((bowser.webkit || bowser.blink) && document.documentElement.webkitRequestFullScreen) {
             document.documentElement.webkitRequestFullScreen();
         }
-        if (bowser.gecko) {
+        if (bowser.gecko && document.documentElement.mozRequestFullScreen) {
             document.documentElement.mozRequestFullScreen();
         }
     }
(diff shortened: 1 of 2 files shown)


src/lib/tablet-full-screen.js
```diff
@@ -0,0 +1,15 @@
+import bowser from 'bowser';
+
+/**
+ * Helper method to request full screen in the browser when on a tablet.
+ */
+export default function () {
+    if (bowser.tablet) {
+        if (bowser.webkit || bowser.blink) {
+            document.documentElement.webkitRequestFullScreen();
```
paulkaplan: I think we should be a bit more defensive with unstable vendor prefixed APIs like this, could you check that it exists first, like `document.documentElement.webkitRequestFullScreen && document.documentElement.webkitRequestFullScreen()`?

chrisgarrity: 👍

PR #2496, opened 2018-06-27
Translate default project asset names

### LINKED ISSUE #2070: Translating the default sprite/costume/backdrop names
### Expected Behavior

When a user loads the default project in another language all the assets in the project should be translated.

### Actual Behavior

The default sprite is 'sprite1', with 'costume1', and the Stage has 'backdrop1'. These are in the interface, and what ends up in the dynamic menus of the blocks.

See https://github.com/LLK/scratch-blocks/issues/1490

chrisgarrity:
### Resolves

- Resolves #2070 

### Proposed Changes

Use ScratchBlocks.ScratchMsgs to translate the asset names in the default project. Only works on load if the locale has already been set. Once the project has been loaded assets are fixed and will not change names even if the locale is changed.

### Discussion
This is just a workaround for now. In the long term, which module will be responsible for creating the default project, and what locale state will it have? Storage should not know about locale. Either it gets passed already localized json, or it should be passed an object that defines the default names to be substituted when generating the default project. I.e. the caller (whether www or gui) knows about locale.


src/lib/default-project/project-data.js
```diff
@@ -0,0 +1,102 @@
+import ScratchBlocks from 'scratch-blocks';
+
```
chrisgarrity: The alternative approach would be to convert to .jsx like https://github.com/LLK/scratch-gui/blob/develop/src/lib/libraries/decks/index.jsx, and use `FormattedMessage`. I think the ScratchBlocks messages made sense for dynamic menus, but it's not as clear for the default assets.

paulkaplan:
@chrisgarrity this seems good to me, but I think @rschamp did most of the work with the default project and knows more about the way the default project is relied on by some outside contributors, so I'm going to kick this over to him for a specialist worm consult. @rschamp that cool?

src/lib/default-project/project-data.js
rschamp: I think the default project requiring `scratch-blocks` is kind of an odd dependency. Won't this also make pulling out the strings more complicated than if it followed the same process as the rest of the GUI?

chrisgarrity: So I'm mistaken, this can't use the jsx approach used by the libraries because the json will end up with `<FormattedMessage>` React components as data. That's fine for the libraries because they're used by the Library component. But project data is used by the VM, so it won't work with react components.

chrisgarrity:
Closing in favor of https://github.com/LLK/scratch-gui/pull/3145

PR #2537, opened 2018-07-02
Initial version of device connection modals

ericrosenbaum:
This adds a set of modal dialogs to walk the user through connecting to a hardware device via an extension. Eventually these dialogs will contain images and other information specific to individual hardware devices, but for now they are generic. There are many cosmetic details remaining to bring these dialogs to spec.

Other miscellaneous changes include a small change to the base modal component to support more flexible styling, and adding some new colors. 

The connection status indicator button at the top of the blocks category launches the connection modal and also updates to when the device has connected, but there are some known issues with it (the indicator gets out of sync in connection succeeds after closing the modal, and when switching sprites).

### DIFF
--- a/src/components/connection-modal/connected-step.jsx
+++ b/src/components/connection-modal/connected-step.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 import Box from '../box/box.jsx';
+import Dots from './dots.jsx';
 
 import styles from './connection-modal.css';
 
@@ -21,6 +22,10 @@ const ConnectedStep = props => (
                     id="gui.connection.search"
                 />
             </Box>
+            <Dots
+                counter={3}
+                total={3}
+            />
         </Box>
     </Box>
 );
--- a/src/components/connection-modal/connected-step.jsx
+++ b/src/components/connection-modal/connected-step.jsx
@@ -11,17 +11,16 @@ import styles from './connection-modal.css';
 
 const ConnectedStep = props => (
     <Box className={styles.body}>
-        <Box className={styles.buttonRow}>
-            <button
-                className={styles.searchButton}
-                onClick={props.onSearch}
-            >
+        <Box className={styles.activityArea}>
+        </Box>
+        <Box className={styles.bottomArea}>
+            <Box className={styles.instructions}>
                 <FormattedMessage
                     defaultMessage="🔥connected🔥"
                     description="Button in prompt for starting a search"
                     id="gui.connection.search"
                 />
-            </button>
+            </Box>
         </Box>
     </Box>
 );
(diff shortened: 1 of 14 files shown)


rschamp:
Try it out: [feature/device-connection-modals](https://llk.github.io/scratch-gui/feature/device-connection-modals)

src/components/connection-modal/connection-modal.jsx
```diff
@@ -0,0 +1,36 @@
...
+            {props.phase === 'connecting' && <ConnectingStep {...props} />}
+            {props.phase === 'connected' && <ConnectedStep {...props} />}
+            {props.phase === 'error' && <ErrorStep {...props} />}
+        </Box>
+    </Modal>
+);
+
+ConnectionModalComponent.propTypes = {
+    onCancel: PropTypes.func.isRequired,
+    phase: PropTypes.string.isRequired,
```
rschamp: It would be useful to document the phases here so that people know which phases there are.  I would suggest setting the phases to constants, set on the component:
```jsx
import keyMirror from 'keymirror';

// keyMirror creates an object with values the same as keys
ConnectionModalComponent.PHASES = keyMirror({
    'scanning': null,
    'connecting': null,
    'connected': null,
    'error': null
});

ConnectionModalComponent.propTypes = {
    phase: PropTypes.oneOf(Object.keys(ConnectionModalComponent.PHASES))
};

/* containers/connection-modal.js */
const PHASES = ConnectionModalComponent.PHASES;

    this.setState({phase: PHASES.scanning})
```

src/components/connection-modal/connection-modal.jsx
```diff
@@ -0,0 +1,36 @@
...
+import ConnectingStep from './connecting-step.jsx';
+import ConnectedStep from './connected-step.jsx';
+import ErrorStep from './error-step.jsx';
+
+import styles from './connection-modal.css';
+
+const ConnectionModalComponent = props => (
+    <Modal
+        className={styles.modalContent}
+        contentLabel={''}
```
rschamp: I believe this is needed for accessibility, so we should set it to something like (a localized) "Scratch Device Connection Modal".

src/components/connection-modal/connected-step.jsx
```diff
@@ -0,0 +1,57 @@
...
+                        id="gui.connection.disconnect"
+                    />
+                </button>
+                <button
+                    className={styles.connectionButton}
+                    onClick={props.onCancel}
+                >
+                    <FormattedMessage
+                        defaultMessage="go to editor"
+                        description=""
```
rschamp: These descriptions will be the context for translators to translate the text, so it should say something about the context.

paulkaplan:
@rschamp I addressed most of your comments. I think I will follow-up with a PR to address the state in redux vs. the react component later, like you mentioned, if that is ok.

paulkaplan:
@ericrosenbaum instead of removing the currently unused "scanning" state, I just added handlers and a callback for when it is implemented. I set up the listener for `PERIPHERAL_SCAN_TIMEOUT`, fwiw

ericrosenbaum:
I fixed a merge conflict and everything looks fine- back to @rschamp for a final re-review

PR #2575, opened 2018-07-10
Use bitmapAdapter for costume upload

fsih:
### Resolves
Work toward LLK/scratch-svg-renderer#36
Depends on https://github.com/LLK/scratch-svg-renderer/pull/43
Depends on https://github.com/LLK/scratch-vm/pull/1315

Also fixes some places where we use info instead of info/2 for the rotation center. Info[0] appears to be costume width and info[1] appears to be costume height.

### Proposed Changes

Attach bitmapAdapter
Use bitmapAdapter for file upload. The api is slightly different from the old importBitmap

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


fsih:
The branch is up! This branch includes the new loading logic as well as all the bitmap assets, so you can actually load them.
After:
https://llk.github.io/scratch-gui/bitmapLibraryWithBitmapLoad/index.html

For comparison, here is the branch that just adds the bitmap assets.
Before:
https://llk.github.io/scratch-gui/bitmap/

Things to test:
- Anna, Calvert, Cassy, Dan, Football, Lamp are sprites / costumes with bitmap resolution 1, so they are double the size with the new loader
- Test that uploads of all file types and sizes and from camera get resized the same way in Before and After, and don't become blurry (jimp replacement was successful)
- Jodi and Sam have file type gif so they don't load in the paint editor on either version currently
- In Before, when uploading individual costumes (_not_ sprites), the rotation center is set to the bottom right corner instead of the center. info[0] is height and info[1] is width, not rotation center. This has been corrected in After
- I think for other costumes, info[0] does indeed mean rotationCenterX and info[1] is rotationCenterY, because now in After, some sprites are not centered correctly (e.g. unicorn)

/cc @thisandagain

kchadha:
As discussed offline, the rotation center should be updated to divide the info values by 2 for bitmap costumes (and all backdrops?) and leave the values as is for vector costumes.

src/containers/costume-library.jsx
```diff
@@ -26,10 +26,13 @@ class CostumeLibrary extends React.PureComponent {
         ]);
     }
     handleItemSelected (item) {
+        const type = item.md5.split('.')[1];
```
kchadha: Hmmm it's a little dangerous to assume that String.split yields a 2 element array.

fsih:
Made a new issue for the orientation issue:
https://github.com/LLK/scratch-paint/issues/554

PR #2580, opened 2018-07-10
Fix loading projects by hash on player example

rschamp:
### Resolves

Resolves `/player.html#<project id>` not loading the project with id `<project id>`

### Proposed Changes

Pass on uncaptured props from Player component to GUI

### Reason for Changes

Passing on the props down from the HashParserHOC allows the HOC to load the project specified by the hash and give this information to the GUI.

### Test Coverage

I modified the integration tests to gather "performance" logs, which include which requests are made by the browser. This allows us to make sure that the project is actually being requested in the examples.

### Browser Coverage
N/A

### DIFF
--- a/src/playground/player.jsx
+++ b/src/playground/player.jsx
@@ -18,7 +18,7 @@ if (process.env.NODE_ENV === 'production' && typeof window === 'object') {
 
 import styles from './player.css';
 
-const Player = ({isPlayerOnly, onSeeInside, ...props}) => (
+const Player = ({isPlayerOnly, onSeeInside, projectId}) => (
     <Box
         className={classNames({
             [styles.stageOnly]: isPlayerOnly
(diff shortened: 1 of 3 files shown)


gnarf:
Tested this branch and everything seems to work! Thanks!

src/playground/player.jsx
```diff
@@ -28,6 +28,7 @@ const Player = ({isPlayerOnly, onSeeInside}) => (
         <GUI
             enableCommunity
             isPlayerOnly={isPlayerOnly}
+            {...props}
```
gnarf: is there a specific prop this was missing? Curious if you really want to kind of steal the params list from GUI here, or just list the ones you care about, or if it wouldn't make more sense to pass GUI in as a child?

rschamp: Good idea, I updated to use an explicit prop.

paulkaplan:
I made that change @kchadha asked for, but I mucked up pushing it so things got a bit wacky in the commit history, sorry about that

PR #2691, opened 2018-07-20
Fix project loader cleanup function. Safari does not support the use of .finally

kchadha:
### Resolves

- Resolves an issue found during smoke testing where Safari 11.0 was not uploading projects.

### Proposed Changes

Remove the use of .finally which is not supported by Safari or Edge.


rschamp:
Try it out: [fix-finally-bug](https://llk.github.io/scratch-gui/fix-finally-bug)

PR #2706, opened 2018-07-23
[MASTER] Temporarily exclude RTL languages from the menu

chrisgarrity:
Right to left (RTL) languages are not yet fully supported in the editor, so we don't want to give the impression they are by including them in the menu.

Translators can still load the language with the `?locale=he` URL argument.

Currently the excluded languages is just 'he', but I wanted to write it in such a way that it can be used for multiple languages later if we need it.


rschamp:
@chrisgarrity Don't forget to also merge this into `develop` if you haven't already.

Issue #2452, opened 2018-06-25
Implement sharing costumes/sounds to the stage


paulkaplan:
Originally this actually worked, but the change to making the drop code only be called when within the bounding box of the drop area stopped it. So extract the drop area code from the backpack and apply it to the stage area also.

PR #2727, opened 2018-07-25
Share costumes and sounds to stage

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/2452

### Proposed Changes

Extracts the logic for testing if an asset drag is occurring over your component and calling onDrop when it drops. Use this HOC for both the backpack and the stage container. 

### Reason for Changes

So that asset dropping could work on the stage!

drag-to-stage

### DIFF
--- a/src/lib/drop-area-hoc.jsx
+++ b/src/lib/drop-area-hoc.jsx
@@ -1,0 +1,88 @@
+import bindAll from 'lodash.bindall';
+import PropTypes from 'prop-types';
+import React from 'react';
+import omit from 'lodash.omit';
+import {connect} from 'react-redux';
+
+const DropAreaHOC = function (dragTypes) {
+    return function (WrappedComponent) {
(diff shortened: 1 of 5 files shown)


src/lib/drop-area-hoc.jsx
```diff
@@ -0,0 +1,88 @@
...
+                }
+            }
+            setRef (el) {
+                this.ref = el;
+            }
+            render () {
+                const componentProps = omit(this.props, ['onDrop', 'dragInfo']);
+                return (
+                    <WrappedComponent
+                        containerRef={this.setRef}
```
rschamp: I'm curious why you chose to change this to `containerRef`? It makes sense in the context of this PR, but probably will be ambiguous out of context. `dropAreaRef` would clearly show it's originating from the `DropAreaHOC`.

Issue #2719, opened 2018-07-24
Tutorial titles should wrap under thumbnails


paulkaplan:
![image](https://user-images.githubusercontent.com/654102/43159858-8a5de8aa-8f51-11e8-9c6a-23f130dd9bb8.png)

Reported by @chrisgarrity

Issue #2526, opened 2018-06-29
Mini Cards: Gallery card, two options instead of three


kosiecki17:
The final card in each mini deck currently includes three options, leading to three projects in the "How To's" library. We'd like this changed, to include only two options. This will allow for larger thumbnail images and more space for the text underneath the images.
 
At the moment, ignore other discrepancies in design and text between the two sample images below. We're just looking to make a change in the options at the end.

![gallery-card-change](https://user-images.githubusercontent.com/25378759/42102092-f0da5cd0-7b92-11e8-977f-7302adb8c721.jpg)

PR #2729, opened 2018-07-25
Show only 2 other tutorials in last step of each tutorial.

paulkaplan:
Enforce this by using `slice` so that no more are ever shown, which would mess up the layout.

Also fixes the issue where the card text could push out when translated into other languages

### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/2526
- Resolves https://github.com/LLK/scratch-gui/issues/2719

### Proposed Changes

Make the deck preview wider and more padded out from the edges. Also enforce a max width to force titles to wrap.

image

### DIFF
--- a/src/components/cards/card.css
+++ b/src/components/cards/card.css
@@ -153,7 +153,7 @@
     display: flex;
     flex-direction: row;
     justify-content: space-around;
-    padding: 0 0.5rem 0.5rem;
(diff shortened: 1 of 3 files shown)


carljbowman:
Looks good to me! Thanks for making this change @paulkaplan

Issue #2445, opened 2018-06-22
Sprite tile images disappear after changing zoom and switching sprites 


ericrosenbaum:
### Expected Behavior

Sprite tile images shouldn't disappear.

### Actual Behavior

After changing the browser zoom level (e.g. with cmd + or cmd -), clicking a sprite selector tile causes the image to disappear:

![zoom-disappearing-sprite-tiles-bug](https://user-images.githubusercontent.com/567844/41798485-d0b90aca-763b-11e8-8895-dd30e1f879a6.gif)

### Steps to Reproduce

- Add some sprites
- Change browser zoom level
- Switch sprites

### Operating System and Browser

Mac OS 10.13.2 Chrome 66

gnarf:
It's eeven worse now, it seems to lose all the sprites after zoom / any interaction with any of them:

https://clips.twitch.tv/BadMoldyNuggetsTinyFace

PR #2730, opened 2018-07-25
Remove costume canvas and just display tiles using <img> tags

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/2445

### Proposed Changes

Remove the canvas rendering of the sprite tile and instead just show the image directly. The CostumeCanvas component was giving us far more than we needed, but it was only used for the tiles that were not taking advantage of the direction, etc. and was causing problems like #2445 

### Reason for Changes

Much simpler, less code to worry about

### Test Coverage

Updated unit tests around that component


paulkaplan:
@rschamp but 
image

rschamp:
Agreed! It's just it has been here since @tmickel added it in #7, so deserving of a RIP :)

tmickel:
Single digit PR number!!! Cheers RIP :)

PR #2745, opened 2018-07-31
Add from backpack to sprite/stage tiles

paulkaplan:
Make it possible to drop backpack assets on sprite tiles and on the stage.

There is a note in the code about how this does not implement "new sprite from backpack costume", which we can think about later. 

This depends on the vm PR https://github.com/LLK/scratch-vm/pull/1399
It is also based on https://github.com/LLK/scratch-gui/pull/2727, so only the second commit is new here

backpack-add

### DIFF
--- a/src/components/sprite-selector/sprite-list.jsx
+++ b/src/components/sprite-selector/sprite-list.jsx
@@ -53,8 +53,11 @@ const SpriteList = function (props) {
                 // Note the absence of the self-sharing check: a sprite can share assets with itself.
                 // This is a quirk of 2.0, but seems worth leaving possible, it
                 // allows quick (albeit unusual) duplication of assets.
-                isRaised = isRaised || draggingType === DragConstants.COSTUME ||
-                    draggingType === DragConstants.SOUND;
+                isRaised = isRaised || [
+                    DragConstants.COSTUME,
+                    DragConstants.SOUND,
+                    DragConstants.BACKPACK_COSTUME,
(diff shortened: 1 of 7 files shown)


paulkaplan:
bump

rschamp:
@paulkaplan Do you have a build somewhere I can test?

paulkaplan:
Hmm like the other backpack things, due to issues with needing a backpack server, the only way is to pull it locally and run the backpack server and use the fixture/user.json for username/token. 

Maybe this is a good time to think about how to have a development fixture for this?

src/components/sprite-selector/sprite-list.jsx
```diff
@@ -53,8 +53,11 @@ const SpriteList = function (props) {
                 // Note the absence of the self-sharing check: a sprite can share assets with itself.
                 // This is a quirk of 2.0, but seems worth leaving possible, it
                 // allows quick (albeit unusual) duplication of assets.
-                isRaised = isRaised || draggingType === DragConstants.COSTUME ||
-                    draggingType === DragConstants.SOUND;
+                isRaised = isRaised || [
+                    DragConstants.COSTUME,
+                    DragConstants.SOUND,
+                    DragConstants.BACKPACK_COSTUME,
+                    DragConstants.BACKPACK_SOUND].includes(draggingType);
```
rschamp: Seems like this would be more understandable as a `const dragTypes` at the top of the file like `StageSelector` below, but not a big deal.

Issue #2757, opened 2018-08-01
RTL - menu bar


chrisgarrity:
### Expected Behavior

All menubar elements should work in RTL languages.

Splitting out from #1579

PR #2760, opened 2018-08-01
Feature/1579 rtl

### LINKED ISSUE #1579: Support RTL for all interface elements
### Expected Behavior

When the GUI is switched into a RTL language the UI should switch direction as well as the blocks. For example, the stage should be on the left instead of the right.

### Actual Behavior

Since we use flex-box for almost everything - it switches as expected with the `dir` attribute is set to `rtl`. Known exceptions are the `add sprite`, `add backdrop`, and `add extension` buttons. We should check for any use of `position: absolute` and make it work properly in RTL mode.

### Steps to Reproduce

Eventually the direction attribute will be set by the wrapping container (e.g. www) - for now simply set `dir=rtl` on the body of the playground.

I may split these out into separate issues, but starting to note the individual components that need to handle RTL:
- [ ] Blocks - workspace and toolbox
- [ ] Category menu
- [ ] ~Libraries [sprites, etc]~ Moved to #2756
- [ ] ~Tutorials [this may be the same as other libraries]~ Moved to #2756
- [ ] ~How-to cards~ Moved to #2758
- [ ] Action menus (the position of the options is wrong and the tooltip is on the wrong side)
- [ ] Sprite info
  - [ ] margin/padding between labels/ value
  - [ ] show/hide border style (opposite rounded corners)
- [ ] Sprite delete button - fix position
- [ ] ~Modals - preview etc~ Moved to #2759
- [ ] ~menubar menus (drop down to the right instead of left)~ moved to #2757
- [Probably won't fix] - coming soon tooltip when positioned to the side

Questions: 
- backpack?

Paint editor: Tracking in https://github.com/LLK/scratch-paint/issues/574
- fill and stroke selector styles.
- color picker layout
- zoom button borders
- undo/redo button borders
- bitmap button icon padding

Sound Editor 
- [ ] Should the 'play' button point the other direction? No it needs to stay LTR
- [ ] undo/redo borders
- [ ] padding/margins around labels

chrisgarrity:
### Resolves

- Resolves #1579

### Proposed Changes
Makes the basic interface elements work in RTL languages e.g. Hebrew.

Added `isRTL` to the redux locales state. GUI uses it to set the `dir` attribute on the GUI `Box`. Flexbox automatically flows `flex-direction: row` in the correct direction.

Anywhere we had explicit `margin-left|right` the css needed to have rules specifically for `[dir="ltr"]` and `[dir="rtl"]`

### Test Coverage

_To test use `?locale=he` on the URL_

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/stage-selector/stage-selector.css
+++ b/src/components/stage-selector/stage-selector.css
@@ -96,7 +96,6 @@ $header-height: calc($stage-menu-height - 2px);
 .add-button {
     position: absolute;
     bottom: 0.75rem;
-    z-index: $z-index-add-button
 }
 
 .raised, .raised .header {
--- a/src/components/stage-selector/stage-selector.css
+++ b/src/components/stage-selector/stage-selector.css
@@ -1,6 +1,5 @@
 @import "../../css/units.css";
 @import "../../css/colors.css";
-@import "../../css/z-index.css";
 
 $header-height: calc($stage-menu-height - 2px);
 
--- a/src/components/action-menu/action-menu.css
+++ b/src/components/action-menu/action-menu.css
@@ -59,6 +59,10 @@ button::-moz-focus-inner {
     height: calc($main-button-size - 1rem);
 }
 
+[dir="rtl"] .main-icon {
+    transform: scaleX(-1);
+}
+
 .more-buttons-outer {
     /*
         Need to use two divs to set different overflow x/y
--- a/src/components/action-menu/action-menu.css
+++ b/src/components/action-menu/action-menu.css
@@ -1,4 +1,5 @@
 @import "../../css/colors.css";
+@import "../../css/z-index.css";
 
 $main-button-size: 2.75rem;
 $more-button-size: 2.25rem;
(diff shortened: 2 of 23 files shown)


kyleplo:
The language selector causes the page to scroll.
image

chrisgarrity:
@kyleplo thanks, that should get taken care of when I work on #2757

src/components/stage-selector/stage-selector.css
```diff
@@ -96,7 +96,6 @@ $header-height: calc($stage-menu-height - 2px);
 .add-button {
     position: absolute;
     bottom: 0.75rem;
-    z-index: $z-index-add-button
```
kchadha: Is the z-index CSS import still used in this file given this change?

Issue #1966, opened 2018-05-03
Hook up Edit -> Turbo Mode menu item


rschamp:
First, the VM needs to emit events when turboMode changes. Then this button can use those events to display the state of the menu item, and we should update the stage controls to do the same.

PR #2767, opened 2018-08-02
Turbo mode menu 

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/1966

### Proposed Changes

First commit hooks up the new turbo mode events from https://github.com/LLK/scratch-vm/pull/1413 (merged) to the vm listener, and moves the project run state listener out of the controls container to put both related status flags in a single reducer.

Second commit enables the turbo mode menu item, using a render function component as the project saver so it can wrap the menu item.

/cc @carljbowman I implemented the menu item as a simple on/off toggle, not sure if that is what we wanted or not? Here is what it looks like

turbo-mode

### DIFF
--- a/src/containers/turbo-mode.jsx
+++ b/src/containers/turbo-mode.jsx
@@ -1,0 +1,60 @@
+import bindAll from 'lodash.bindall';
+import PropTypes from 'prop-types';
+import React from 'react';
+import {connect} from 'react-redux';
+
+/**
+ * Turbo Mode component passes toggleTurboMode function to its child.
+ * It also includes `turboMode` in the props passed to the children.
+ * It expects this child to be a function with the signature
+ *     function (turboMode, toggleTurboMode, props) {}
+ * The component can then be used to attach turbo mode setting functionality
+ * to any other component:
+ *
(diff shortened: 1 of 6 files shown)


src/containers/turbo-mode.jsx
```diff
@@ -0,0 +1,60 @@
+import bindAll from 'lodash.bindall';
+import PropTypes from 'prop-types';
+import React from 'react';
+import {connect} from 'react-redux';
+
+/**
+ * Turbo Mode component passes toggleTurboMode function to its child.
+ * It also includes `turboMode` in the props passed to the children.
+ * It expects this child to be a function with the signature
+ *     function (turboMode, toggleTurboMode, props) {}
```
rschamp: This doesn't match the example, need to remove the initial `turboMode` argument.

paulkaplan: Thanks! Fixed

paulkaplan:
@rschamp re-assigned after fixing comment. Also brought in the VM update to develop already, so this will work right away

Issue #2540, opened 2018-07-02
Device connection modals need to support different sets of steps 


ericrosenbaum:
Currently the device connection modals support a single sequence of steps, which work for the micro:bit and EV3 connection processes (scanning step with device list, connecting, connected). We need to add support for a modified sequence for the WeDo 2.0 (start button, scanning step with prompt to push button on device, connecting, connected).

PR #2772, opened 2018-08-02
Add device connection steps for Lego WeDo 2.0

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/2540

### Proposed Changes

To handle the device connection flow for Lego WeDo 2.0, which is different from the EV3 and micro:bit, we've added an "auto-scanning step", to use in place of the scanning step. The auto-scanning step includes a "prescan" phase, with a button that initiates the scan, and a "pressbutton" phase, which shows an animation of the button the user is supposed to press on their device. During the pressbutton phase, as soon as a device is found, we connect to it (instead of providing a list of devices to choose from).

Thanks to @paulkaplan for doing a lot of the work on this!

Also note the super cool SVG animation by @paulkaplan (for the step showing you to press the button on the WeDo itself) [here](https://github.com/LLK/scratch-gui/pull/2772/files#diff-aa0be1cf82ecd51bead6a4e6481cebbfR3) 

### Reason for Changes

The WeDo 2.0 only advertises itself for a few seconds after you press its button, so we needed a workflow where you click a button on the screen, followed by the button on the device. It does not have a user-defined unique device name, so we don't provide a device list to choose from.


src/lib/libraries/extensions/index.jsx
```diff
@@ -153,6 +158,20 @@ export default [
             />
         ),
         featured: true,
-        disabled: true
+        disabled: true,
+        launchDeviceConnectionFlow: true,
+        useDeviceList: false,
```
paulkaplan: Since we settled on using the term `autoScan` for the components, maybe we should use `useAutoScan: true` here instead of the device list?

evhan55:
This worked beautifully for me! 👍

LayYuan:
Sorry, I am not sure is it correct to write here. Please kindly delete it if not suitable to post here.
But I searching for the method to connect 2 wedo2.0 in Single scratch project, this feature can be done in LEGO Wedo2.0 App but not in Scratch3.0. 
Single hub in wedo2.0 only support 2 ports, if we can connect 2 wedo2.0 at once, we can control 2 motors and the 2 input sensors, this will be nice and more creative project can be create!
Although LEGO app can connect 2 wedo2.0, but scratch is much more user friendly, which i really hope scratch to connect 2 wedo2.0 at once too!
But this Auto scanning step and connection make it no possible to connect second wedo hub.. 
: (

Issue #2716, opened 2018-07-24
Brokenness in nav bar in にほんご language


fsih:
### Expected Behavior
Text in navigation bar is in navigation bar

### Actual Behavior
Text is vertical
![image](https://user-images.githubusercontent.com/2855464/43155883-95605a40-8f46-11e8-890a-ecd65ddd945f.png)

### Steps to Reproduce

1. Switch language to にほんご
2. Text begins to escape the navigation bar when Chrome is at 1222px across

### Operating System and Browser
Windows 10 Chrome

Kenny2github:
I cannot reproduce this on develop, Windows 10.0.17134.165, Chrome 67.0.3396.99
![image](https://user-images.githubusercontent.com/28599280/43157866-7a55bcb2-8f7e-11e8-97ae-4f857b10e0af.png)

apple502j:
Cannot repro on develop, Windows 7 HomePremium 64bit SP1, Firefox 61.0.1

wufeng87:
So does Chinese, but it is OK for showing Chinese vertically. : )

Kenny2github:
@wufeng87 what operating system are you using? 您在用什么操作系统？

thisandagain:
@BryceLTaylor Could you attempt a repro?

setoyama60jp:
![2018-08-05 8 22 23](https://user-images.githubusercontent.com/1531784/43681226-e22b664e-9888-11e8-95c2-51254b59361a.png)
This happened when narrowing down the width of the browser window.
(I  checked Chrome 67.0.3396.99 on macOS High Sierra 10.13.6)
And did not happen when setting `white-space : nowrap;` to the `span` tag.
```
<span style="
    white-space: nowrap;
">ファイル</span>
```

Kenny2github:
@setoyama60jp Thanks for the repro! And thanks for the suggested solution, as well. Yeah, this seems similar to a ScratchWikiSkin issue I had a while back (which was fixed by [this pull](https://github.com/InternationalScratchWiki/ScratchWikiSkin2/pull/9#issuecomment-375510028), which can provide some more detail).

I also managed to repro this myself just now:
![image](https://user-images.githubusercontent.com/28599280/43690082-9c9c37d0-9904-11e8-9f93-04b23a1e67aa.png)
As you can see, I had inspect element open - this caused the page's width to shrink, and the Chinese characters to stack on top of each other.

[...]

PR #2806, opened 2018-08-07
fix for issue #2716 (brokenness in nav bar in にほんご language)

setoyama60jp:
added a line to manu-bar css so as to prevent brokenness in nav bar in Japanese and Chinese languages.

### Resolves

issue #2716 

- Resolves #

### Proposed Changes

added a line `white-space: nowrap;` to menu-bar-item class in menu-bar.css.

### Reason for Changes

To prevent brokenness in nav bar in Japanese or Chinese language.

### Test Coverage

I did not add any automatic test case and only did visual check manually because I fixed only one style sheet file.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/menu-bar/menu-bar.css
+++ b/src/components/menu-bar/menu-bar.css
@@ -66,6 +66,7 @@
     align-self: center;
     position: relative;
     align-items: center;
+    white-space: nowrap;
     height: $menu-bar-height;
 }
 


fsih:
@chrisgarrity is it fine if we force text in the nav bar not to wrap? Or do some languages need to be allowed to wrap?

fsih:
I confirmed with Chris that it's fine to force not wrapping. To deal with the text being squeezed off the side, we will need additional fixes, such as making some of the text disappear (e.g. Tutorials) if it doesn't fit.

Issue #2132, opened 2018-05-23
StageWrapper and sub-components should not have padding


chrisgarrity:
The stage header and canvas wrapper components include padding to create a .5rem space around the stage. On www the left edge of the stage cannot line up with the left edge of the view because of this padding. 

Currently working around the problem by setting a negative  .5rem margin on the enclosing element on www, but that seems brittle.

paulkaplan:
Yes please. This is also the cause of monitors being slightly misaligned in fullscreen mode.

PR #2807, opened 2018-08-07
padding and margin changes for www project page, monitor positioning

benjiwheeler:
Addresses both the main issue https://github.com/LLK/scratch-gui/issues/2132 , and the issue Paul mentions in a comment on that issue.


src/css/units.css
```diff
@@ -9,6 +9,11 @@ $stage-menu-height: 2.75rem;
 $library-header-height: 3.125rem;
 $library-filter-bar-height: 2.5rem;
 
+$full-screen-top-margin: 0.1875rem;
+/* monitors must be moved down slightly relative to the stage; otherwise
+they appear too high */
+$monitor-padding-top-adjustment: 0.125rem;
```
benjiwheeler: A weakness of this fix is that I don't understand *why* the monitors need to be moved down an additional 2 pixels. Maybe because of the stage border?

chrisgarrity: At 2 pixels they still overlap the border a little, I think it needs to be 3px - and that is exactly the border of the stage-wrapper-overlay.

benjiwheeler:
@chrisgarrity, good points! Revisited this and made deeper changes. This actually fixed other monitor layout issues in the standard size viewer that were wacky. 

Before:

Chrome:
in editor:
screen shot 2018-08-12 at 12 45 39 am
full screen:
screen shot 2018-08-12 at 12 45 49 am


After:

Chrome:
in editor:
screen shot 2018-08-12 at 12 15 38 am
full screen:
screen shot 2018-08-12 at 12 15 45 am

Firefox:
in editor:
screen shot 2018-08-12 at 12 17 46 am
full screen:
screen shot 2018-08-12 at 12 17 32 am

src/lib/screen-utils.js
```diff
@@ -10,8 +10,13 @@ import layout, {STAGE_DISPLAY_SCALES, STAGE_SIZE_MODES, STAGE_DISPLAY_SIZES} fro
  */
 
 const STAGE_DIMENSION_DEFAULTS = {
-    spacingBorderAdjustment: 9,
-    menuHeightAdjustment: 40
+    // referencing css/units.css,
```
benjiwheeler: I think the values below were wrong -- outdated and not updated when the menu and padding increased.

src/lib/screen-utils.js
```diff
@@ -10,8 +10,13 @@ import layout, {STAGE_DISPLAY_SCALES, STAGE_SIZE_MODES, STAGE_DISPLAY_SIZES} fro
...
+    // spacingBorderAdjustment = 2 * $full-screen-top-bottom-margin +
+    //   2 * $full-screen-border-width
+    fullScreenSpacingBorderAdjustment: 12,
```
benjiwheeler: Note that this 12px includes both borders, the 3px padding above the stage's top border, and (this one might be debatable) the 3px padding below the stage's bottom border.

chrisgarrity: I also like the little bit of bottom padding in the browser, but it may not be ideal in a mobile situation (e.g. if it adds scrollbars).

chrisgarrity:
@benjiwheeler I think this is good to go, pending getting Carl's confirmation on whether to have the bottom padding or not.

carljbowman:
Looks good to me. We may need to circle back on the padding when we think about the embed project UI, but I think it's good to keep the padding consistent all the way around for now.

PR #2809, opened 2018-08-07
build dist, even when run without NODE_ENV=production

benjiwheeler:
make development faster by always building assets for dist directory, even when run without NODE_ENV=production .

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -177,42 +177,41 @@ module.exports = [
         ])
     })
 ].concat(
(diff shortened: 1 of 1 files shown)


benjiwheeler:
Turned out this slowed down a bunch of gui developers. If you ran npm start in GUI you couldn’t hit localhost:8601, you had to go to /build/index.html . We want to revisit in the future to see if there's a way to avoid having to do some of the taxing production build steps in development that don't mess with existing workflows.

PR #2824, opened 2018-08-08
Restore deleted sprite.

kchadha:
### Resolves

Towards implementing 'undelete' functionality from 2.0.

### Proposed Changes

- Remove 'Redo' item in the 'Edit' menu
- Replace 'Undo' item in the 'Edit' menu with 'Restore' which restores the last deleted sprite. Restore is disabled in the menu by default, and becomes enabled after a sprite gets deleted. When the option becomes enabled, it reads 'Restore sprite'. Clicking on the option should add the last deleted sprite to the project and then disable the 'Restore' option again, changing the text of the option back to 'Restore'

#### Not Addressed By This PR
- Restoring costumes, sounds, or blocks
- Removal of the confirmation dialog when deleting a sprite (this confirmation is linked to deleting all things: sprites, costumes, sounds, so it has not yet been removed)

### Related PRs
Depends on LLK/scratch-vm#1438, and tests will fail until that PR is merged in.

### Test Coverage

Manual testing. Changes can be tested at https://llk.github.io/scratch-gui/restore-sprite

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -134,13 +134,20 @@ class MenuBar extends React.Component {
     constructor (props) {
         super(props);
         bindAll(this, [
-            'handleLanguageMouseUp'
+            'handleLanguageMouseUp',
+            'handleRestoreOption'
         ]);
     }
     handleLanguageMouseUp (e) {
         if (!this.props.languageMenuOpen) {
             this.props.onClickLanguage(e);
         }
+    }
+    handleRestoreOption (restoreFun) {
+        return () => {
+            restoreFun();
+            this.props.onRequestCloseEdit();
+        };
     }
     render () {
         return (
@@ -283,7 +290,7 @@ class MenuBar extends React.Component {
                                 <DeletionRestorer>{(handleRestore, {restorable, deletedItem}) => (
                                     <MenuItem
                                         className={classNames({[styles.disabled]: !restorable})}
-                                        onClick={handleRestore}
+                                        onClick={this.handleRestoreOption(handleRestore)}
                                     >
                                         {deletedItem === 'Sprite' ?
                                             <FormattedMessage
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -14,6 +14,7 @@ import ProjectLoader from '../../containers/project-loader.jsx';
 import Menu from '../../containers/menu.jsx';
 import {MenuItem, MenuSection} from '../menu/menu.jsx';
 import ProjectSaver from '../../containers/project-saver.jsx';
+import DeletionRestorer from '../../containers/deletion-restorer.jsx';
 import TurboMode from '../../containers/turbo-mode.jsx';
 
 import {openTipsLibrary} from '../../reducers/modals';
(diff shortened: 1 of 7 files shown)


src/containers/deletion-restorer.jsx
```diff
@@ -0,0 +1,67 @@
...
+    }
+    render () {
+        const {
+            /* eslint-disable no-unused-vars */
+            children,
+            /* eslint-enable no-unused-vars */
+            ...props
+        } = this.props;
+        props.restorable = typeof this.props.restore === 'function';
+        props.deletedItem = this.props.deletedItem;
```
rschamp: I think in this case it's ok because `props` is a new object, but assigning to props doesn't look right, since most of the time it should be treated as immutable. Also, isn't `props.deletedItem = this.props.deletedItem` a noop?

I think a more consistent pattern would be something like:

```jsx
return this.props.children(this.restoreDeletion, {
    ...props,
    restorable: typeof this.props.restore === 'function
});
```

src/components/menu-bar/menu-bar.jsx
```diff
@@ -279,24 +280,25 @@ class MenuBar extends React.Component {
...
-                                    <MenuItem>
-                                        <FormattedMessage
-                                            defaultMessage="Redo"
-                                            description="Menu bar item for redoing"
-                                            id="gui.menuBar.redo"
-                                        />
+                                <DeletionRestorer>{(handleRestore, {restorable, deletedItem}) => (
+                                    <MenuItem
+                                        className={classNames({[styles.disabled]: !restorable})}
+                                        onClick={handleRestore}
```
rschamp: It would be nice if the menu would also close `onClick`!

kchadha:
@rschamp, I addressed the comments you made.

One thing I have a question about is that I am getting warnings in the console, and I'm not exactly sure where they're coming from (this was happening prior to the last commit as well). Any ideas?:

image

rschamp:
That warning happens when you use `mapStateToProps` without `mapDispatchToProps`. Without `mapDispatchToProps`, `connect` provides a `dispatch` prop. Then if you do `<Component ...props />` the `dispatch` prop gets passed along until it's added to the final `<div>` (or whatever output component), and React complains about not knowing what to do with a `dispatch` attribute on a `<div>`. To remedy, use `() => {}` as the `mapDispatchToProps` argument to keep `dispatch` from being passed.

rschamp:
I didn't notice the second error referencing `dispatchUpdateRestore`. That is from you beginning to pass the `dispatchUpdateRestore` prop to the component for use in the handler methods, but you haven't prevented passing it down to the child component in render. You should omit it from the props being passed to the child.

kchadha:
@rschamp thanks for explaining those warnings! I fixed the second one, but found that the first one is actually already on develop because there aren't any places in this PR where I'm using `mapStateToProps` without `mapDispatchToProps`.

For that second warning, react was only complaining about the specific instance of `dispatchUpdateRestore` being passed through to the TargetPaneComponent, but I added the analogous change to the deletion-restorer as well. I'm not sure why it wasn't complaining about that instance of the issue...?

PR #2827, opened 2018-08-09
bump render and vm together

gnarf:
These two packages need to be updated, render has new logic that vm uses for touching mouse.


rschamp:
Try it out: [bump-render-vm](https://llk.github.io/scratch-gui/bump-render-vm)

rschamp:
Thanks!

PR #2828, opened 2018-08-09
Positioning custom-procedure blocks in RTL

### LINKED ISSUE #2810: Custom procedure modal centering doesn't work in RTL
The code to keep the block centered doesn't work at all in RTL languages.
screen shot 2018-08-07 at 3 57 36 pm

The solution may be to just right justify for RTL and not try to center.

chrisgarrity:
### Resolves
- Resolves #2810 

### Proposed Changes
This is a temporary fix for the custom procedure modal in RTL. The RTL position calculations have a bunch of special cases because the initial placement of the block in RTL depends on the size of the block, but the position is always reported as 0,0.

### Test Coverage

Try it: https://chrisgarrity.github.io/scratch-gui/feature/2810-custom-proc-rtl/?locale=he

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/custom-procedures.jsx
+++ b/src/containers/custom-procedures.jsx
@@ -58,7 +58,8 @@ class CustomProcedures extends React.Component {
             const dy = (metrics.viewHeight / 2) - (this.mutationRoot.height / 2) - y;
             let dx;
             if (this.props.isRtl) {
-                // HACK!! This is temporary until we can figure out what's going on width
+                // // TODO: https://github.com/LLK/scratch-gui/issues/2838
+                // This is temporary until we can figure out what's going on width
                 // block positioning on the workspace for RTL.
                 // Workspace is always origin top-left, with x increasing to the right
                 // Calculate initial starting offset and save it, every other move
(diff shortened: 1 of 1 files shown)


src/containers/custom-procedures.jsx
```diff
@@ -55,11 +56,46 @@ class CustomProcedures extends React.Component {
...
             const {x, y} = this.mutationRoot.getRelativeToSurfaceXY();
             const dy = (metrics.viewHeight / 2) - (this.mutationRoot.height / 2) - y;
-            let dx = (metrics.viewWidth / 2) - (this.mutationRoot.width / 2) - x;
-            // If the procedure declaration is wider than the view width,
-            // keep the right-hand side of the procedure in view.
-            if (this.mutationRoot.width > metrics.viewWidth) {
-                dx = metrics.viewWidth - this.mutationRoot.width - x;
+            let dx;
+            if (this.props.isRtl) {
+                // HACK!! This is temporary until we can figure out what's going on width
```
kchadha: We should probably file an issue to come up with a real fix for this and reference it here as a TODO comment e.g. `TODO (#blah)` so that we can keep track of this.

PR #2831, opened 2018-08-09
Add hideIntro prop to GUI component

### LINKED ISSUE #2814: Introduce prop to indicate whether to show intro to 3.0 modal
When switching from Community/Project page to editor, Intro Modal should be dropped. After making this change, www should be changed to use this feature.

technoboy10:
### Resolves
This PR resolves #2814.

### Proposed Changes
This PR adds a `hideIntro` prop that can be passed to the GUI element to hide the "Welcome to Scratch 3.0" modal. This does not hide the unsupported browser modal.

### Reason for Changes
As stated in #2814, this change is needed so that the scratch-www project page can opt to not show the intro modal when switching from community to project view.

### Test Coverage

Tested by making sure the intro modal still shows up when no `hideIntro` prop is passed to it.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/preview-modal.jsx
+++ b/src/containers/preview-modal.jsx
@@ -27,6 +27,25 @@ class PreviewModal extends React.Component {
             previewing: false
         };
     }
+
+    /**
+     * Conditionally returns an intro modal depending on the hideIntro prop
+     * @returns { React.Component | null } null if hideIntro is true, the intro modal component otherwise
+     */
+    introIfShown () {
+        if (this.props.hideIntro) {
+            return null; // If hideIntro is true, the intro modal should not appear
+        }
+
+        // otherwise, show the intro modal
+        return (<PreviewModalComponent
+            previewing={this.state.previewing}
+            onCancel={this.handleCancel}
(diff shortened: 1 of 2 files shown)


src/containers/preview-modal.jsx
```diff
@@ -41,7 +41,7 @@ class PreviewModal extends React.Component {
     }
     render () {
         return (supportedBrowser() ?
-            <PreviewModalComponent
+            !this.props.hideIntro && <PreviewModalComponent
```
kchadha: It's not clear what happens here if `this.props.hideIntro` is true...

I think it would be cleaner to add a function elsewhere in this file that conditionally returns `<PreviewModalComponent.../>` or `null` based on `this.props.hideIntro`, and call that function here instead

PR #2869, opened 2018-08-15
Handle RTL in menubar

### LINKED ISSUE #2757: RTL - menu bar

chrisgarrity:
Note: the coming soon tooltip positions are calculated by react-tooltip, and they do not handle RTL.

### Resolves

- Resolves #2757

### Proposed Changes

* The menus open to the left of the menubar item
* Tooltips open on the opposite side - however the tooltip offset is calculated by the react-tooltip library and it is not RTL aware.

### Testing
Try it: 
https://chrisgarrity.github.io/scratch-gui/feature/2757-menubar/?locale=he

- [ ] drop down menus align the right edge with the menubar item and open to the left.
- [ ] the user avatar coming soon tooltip opens on the right and points towards the user
- [ ] coming soon tooltips on menu items open to the left

### DIFF
--- a/src/components/coming-soon/coming-soon.css
+++ b/src/components/coming-soon/coming-soon.css
@@ -62,8 +62,15 @@
 }
 
 .coming-soon-image {
-    margin-left: .125rem;
     width: 1.25rem;
     height: 1.25rem;
     vertical-align: middle;
+}
+
+[dir="ltr"] .coming-soon-image {
+    margin-left: .125rem;
+}
+
+[dir="rtl"] .coming-soon-image {
+    margin-right: .125rem;
 }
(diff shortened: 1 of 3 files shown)


kchadha:
I think the tooltip for the my stuff icon should also get a left/right placement. I think it currently defaults to bottom which can be off (sometimes points to the account menu instead) in both LTR and RTL.

Notice the hover vs. the tooltip placement in the pictures below:

**LTR:**

image



**RTL:**
image

image

chrisgarrity:
I think we should just leave them as they are for now - they're only temporary.

PR #2879, opened 2018-08-16
Update language selection menu

### LINKED ISSUE #2658: Replace language select with scrolling menu
### Expected Behavior

Clicking the language selector should show a menu/list of languages for the user to select from. It should scroll to show all the languages.

### Actual Behavior

The language selector opens a menu that contains an HTML select box that contains the list of languages.

chrisgarrity:
### Resolves

- Resolves #2658 

### Proposed Changes

Initial update of language selection menu to avoid the menu within a menu. The language list is still a `select`, so the list of languages is styled by the browser being used. At some point we probably want to replace the `select` to have a consistent look for the languages across platforms. 
/cc @carljbowman 

Also removes the coming soon wrapper as it's no longer needed.

### Test Coverage
Try it:
https://chrisgarrity.github.io/scratch-gui/issue/2658-lang-select/

Updated language integration test to use the new menu.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/language-selector/language-selector.css
+++ b/src/components/language-selector/language-selector.css
@@ -8,8 +8,8 @@
 /* Position the language select over the language icon, and make it transparent */
 .language-select {
     position: absolute;
-    width: 3rem;
-    height: 3rem;
+    width: $language-selector-width;
+    height: $menu-bar-height;
     opacity: 0;
     user-select: none;
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
--- a/src/components/language-selector/language-selector.css
+++ b/src/components/language-selector/language-selector.css
@@ -7,7 +7,7 @@
 
 /* Position the language select over the language icon, and make it transparent */
 .language-select {
-    position: relative;
+    position: absolute;
     width: 3rem;
     height: 3rem;
     opacity: 0;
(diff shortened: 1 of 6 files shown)


chrisgarrity:
@fsih The changes should address 1, 3 & 4. The language options pop up over the menubar in all browsers, so that's expected (although the edge version is particularly ugly)

src/components/language-selector/language-selector.css
```diff
@@ -5,19 +5,27 @@
...
-}
-
+/* Position the language select over the language icon, and make it transparent */
 .language-select {
-    margin: .5rem;
-    height: 1.85rem;
-    border: 1px solid $motion-primary;
+    position: absolute;
+    width: 3rem;
+    height: 3rem;
```
fsih: For height, use $menu-bar-height from units.css
For width, this should always match language-menu in menu-bar.css. Maybe it would be good to add another constant for that in units.css to make sure they don't get out of sync.

PR #2908, opened 2018-08-17
Add missing propTypes

chrisgarrity:
For some reason the travis eslint did not catch these missing prop validations.

From @rschamp - It may be that there are dependencies that are out of date on travis...


chrisgarrity:
@kchadha I'm just going to merge this because it's blocking my language selector PR

Issue #2762, opened 2018-08-01
Fonts don’t look the same in sprite/costume thumbnails


fsih:
Fonts don’t look the same in sprite/costume thumbnails @kchadha
Mac Safari

The issue here is we are setting an image.src to an SVG, and in order for fonts to render correctly, they need to be inlined. See https://github.com/LLK/scratch-svg-renderer/blob/develop/src/font-inliner.js

Inlining fonts is slow and inefficient, and I think it might be better if we just rendered the SVGs in the thumbnails, instead of converting to images. This would also solve the issue where if you dragged a sprite thumbnail into the backpack, it could maintain the correct transparency. We should discuss which solution would be best.

PR #2948, opened 2018-08-21
Inject fonts thumbnails

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-gui/issues/2762

### Proposed Changes
Inject fonts before setting SVG as image source on thumbnails

### Reason for Changes
This allows thumbnails to render Scratch's fonts correctly

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/sprite-selector-item.jsx
+++ b/src/containers/sprite-selector-item.jsx
@@ -12,6 +12,8 @@ import {SVGRenderer} from 'scratch-svg-renderer';
 import SpriteSelectorItemComponent from '../components/sprite-selector-item/sprite-selector-item.jsx';
 
 const dragThreshold = 3; // Same as the block drag threshold
+// Contains 'font-family', but doesn't only contain 'font-family="none"'
+const HAS_FONT_REGEXP = new RegExp('font-family(?!="none")', 'g');
 
 class SpriteSelectorItem extends React.Component {
     constructor (props) {
@@ -42,9 +44,11 @@ class SpriteSelectorItem extends React.Component {
         // Avoid parsing the SVG when possible, since it's expensive.
         if (asset.assetType === storage.AssetType.ImageVector) {
             const svgString = this.props.vm.runtime.storage.get(this.props.assetId).decodeText();
-            if (svgString.indexOf('font-family') !== -1) {
+            if (svgString.match(HAS_FONT_REGEXP)) {
                 // If the asset ID has not changed, no need to re-parse
-                if (this.svgRendererAssetId === this.props.assetId) return this.cachedUrl;
+                if (this.svgRendererAssetId === this.props.assetId) {
+                    return this.cachedUrl;
+                }
 
                 this.svgRendererAssetId = this.props.assetId;
                 this.svgRenderer.loadString(svgString);
--- a/src/containers/sprite-selector-item.jsx
+++ b/src/containers/sprite-selector-item.jsx
@@ -43,13 +43,13 @@ class SpriteSelectorItem extends React.Component {
         // If the SVG refers to fonts, they must be inlined in order to display correctly in the img tag.
         // Avoid parsing the SVG when possible, since it's expensive.
         if (asset.assetType === storage.AssetType.ImageVector) {
+            // If the asset ID has not changed, no need to re-parse
+            if (this.svgRendererAssetId === this.props.assetId) {
+                return this.cachedUrl;
+            }
+
             const svgString = this.props.vm.runtime.storage.get(this.props.assetId).decodeText();
             if (svgString.match(HAS_FONT_REGEXP)) {
-                // If the asset ID has not changed, no need to re-parse
-                if (this.svgRendererAssetId === this.props.assetId) {
-                    return this.cachedUrl;
-                }
-
                 this.svgRendererAssetId = this.props.assetId;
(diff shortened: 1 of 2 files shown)


test/unit/containers/sprite-selector-item.test.jsx
```diff
@@ -66,4 +67,10 @@ describe('SpriteSelectorItem Container', () => {
         expect(global.confirm).toHaveBeenCalled();
         expect(onDeleteButtonClick).not.toHaveBeenCalled();
     });
+
+    test('Has font regexp works', () => {
+        expect('font-family="Sans Serif"'.match(HAS_FONT_REGEXP)).toBeTruthy();
+        expect('font-family="none"'.match(HAS_FONT_REGEXP)).toBeFalsy();
+        expect('font-family="none" font-family="Sans Serif"'.match(HAS_FONT_REGEXP)).toBeTruthy();
```
cwillisf: It might be a good idea to add things like this:
```js
'fake-font-family="none"'.match(HAS_FONT_REGEXP)
'fake-font-family="some"'.match(HAS_FONT_REGEXP)
'font-family-fake="none"'.match(HAS_FONT_REGEXP)
'font-family-fake="some"'.match(HAS_FONT_REGEXP)
```

src/containers/sprite-selector-item.jsx
```diff
@@ -34,6 +39,34 @@ class SpriteSelectorItem extends React.Component {
...
+        const asset = storage.get(this.props.assetId);
+        // If the SVG refers to fonts, they must be inlined in order to display correctly in the img tag.
+        // Avoid parsing the SVG when possible, since it's expensive.
+        if (asset.assetType === storage.AssetType.ImageVector) {
+            const svgString = this.props.vm.runtime.storage.get(this.props.assetId).decodeText();
+            if (svgString.match(HAS_FONT_REGEXP)) {
+                // If the asset ID has not changed, no need to re-parse
+                if (this.svgRendererAssetId === this.props.assetId) {
+                    return this.cachedUrl;
+                }
```
cwillisf: Should this check move up above `const svgString = ...` to save a bit more time?

test/unit/containers/sprite-selector-item.test.jsx
fsih: What results do you expect for those 4 lines?
I changed your regex a bit to match "font-family" instead of 'font-family="..."', since I think SVG allows adding spaces between words. So currently you would get
no match
match
match
match
(Incidentally, do you know if we should consider capitalization?)
None of these need to match, but false positives are better than missing legitimate cases, and I don't think we have to worry too much about fake-font-family attributes unless.... there were someone on the inside manipulating storage.....🤔

src/containers/sprite-selector-item.jsx
fsih: The svgRendererAssetId is the ID of the asset that has been parsed by SVGRenderer, and by the svgString line we don't know if we will use SVGRenderer. Do you mean cache the result of encodeDataURI if the asset doesn't need to go through SVGRenderer?

test/unit/containers/sprite-selector-item.test.jsx
cwillisf: I was thinking there might be some SVG editors out there that add attributes like `myeditor-alternate-font-family` or `font-family-myeditor-enhanced` or something. I don't know if that happens or if it's even allowed by SVG though... maybe this is a non-issue?

src/containers/sprite-selector-item.jsx
cwillisf: I was just thinking that if `this.svgRendererAssetId === this.props.assetId` then the following must be true:
- we've already seen this asset once
- it passed all the checks at the time
- its ID hasn't changed, so presumably its content hasn't changed and would still pass all the checks
All of those things together imply to me that checking the ID can be used as a short-circuit to the cached result without checking the other things explicitly... unless one of my assumptions is bad, of course.

test/unit/containers/sprite-selector-item.test.jsx
fsih: The font injecting code looks for the "font-family" attribute specifically, so if there were attributes like "myeditor-alternate-font-family" and we did parse the SVG, it still wouldn't inject those fonts. Do you mean maybe it should?

cwillisf: Probably not. I was just thinking that the regex should match those cases where the font injecting code actually makes a change, so probably all of these cases should return false?

fsih: ohh I see. So
(at least 1 whitespace)font-family(any whitespace)=(any whitespace)"(Not none)
?
I need a block-based programming language for regex

cwillisf: Hmm, that's a good point. Maybe it's better to err on the side of false positives (like you said 1.3 million years ago) and refine if we find that this is a performance problem.

PR #2961, opened 2018-08-23
Fix all the little RTL layout bugs

### LINKED ISSUE #2800: RTL layout bugs.
### Expected Behavior

Styles that are asymmetric on the gui should mirror when in Hebrew or any RTL language.

### Actual Behavior
screen shot 2018-08-06 at 3 01 04 pm

### Steps to Reproduce

Switch into Hebrew

chrisgarrity:
### Resolves
- Resolves #2800

Fixes styles for RTL in various places

### Test Coverage

- [ ] borders on toolbox and workspace mirror correctly
- [ ] spacing on stage selector is correct
- [ ] stage size icons are mirrored
- [ ] extension library coming soon text is positioned correctly
- [ ] only mirror blocks icon on the tab (others stay right-handed)
- [ ] drop down icon in create variable modal has correct spacing

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/blocks/blocks.css
+++ b/src/components/blocks/blocks.css
@@ -12,6 +12,13 @@
     border-bottom-right-radius: $space;
 }
 
+[dir="rtl"] .blocks :global(.injectionDiv) {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+    border-top-left-radius: $space;
+    border-bottom-left-radius: $space;
+}
+
 .blocks :global(.blocklyMainBackground) {
     stroke: none;
 }
@@ -31,6 +38,11 @@
     -ms-overflow-style: none;
 }
 
+[dir="rtl"] .blocks :global(.blocklyToolboxDiv) {
+    border-right: none;
+    border-left: 1px solid $ui-black-transparent;
+}
+
 .blocks :global(.blocklyToolboxDiv::-webkit-scrollbar) {
     display: none;
 }
@@ -39,6 +51,12 @@
     border-right: 1px solid $ui-black-transparent;
     box-sizing: content-box;
 }
+
+[dir="rtl"] .blocks :global(.blocklyFlyout) {
+    border-right: none;
+    border-left: 1px solid $ui-black-transparent;
+}
+
 
 .blocks :global(.blocklyBlockDragSurface) {
     /*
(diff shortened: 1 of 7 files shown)


chrisgarrity:
Try it at https://chrisgarrity.github.io/scratch-gui/issue/2800-layout-bugs/?locale=he

kchadha:
I was able to verify all of the tests mentioned in the PR description:
- [x] borders on toolbox and workspace mirror correctly
- [x] spacing on stage selector is correct
- [x] stage size icons are mirrored
- [x] extension library coming soon text is positioned correctly
- [x] only mirror blocks icon on the tab (others stay right-handed)
- [x] drop down icon in create variable modal has correct spacing

I noticed that the sound editor and the paint editor have non-rounded corners for the bottom edge facing the stage in both LTR and RTL, if it's a quick fix, do you want to resolve that issue completely within this PR? Otherwise, we should file it as a separate issue.

screen shot 2018-08-23 at 2 39 23 pm

src/components/blocks/blocks.css
```diff
@@ -12,6 +12,13 @@
     border-bottom-right-radius: $space;
 }
 
+[dir="rtl"] .blocks :global(.injectionDiv) {
```
kchadha: Is there a reason that this doesn't have a corresponding [dir="ltr"] version of the style?

kchadha: Same question in the two sets of changes below.

chrisgarrity: In all the other places I use two versions I was sure that it would only be in the context of the gui - which explicitly sets `[dir="ltr"]`. I'm probably being overly cautious, but on the off chance that it could be outside the gui component, I wanted to leave the default settings so it applies if there's no `dir` attribute set. If the attribute isn't set, the value is "auto" or unknown, so it doesn't match the 'ltr' rule.

kchadha:
@chrisgarrity, I think this looks good now!

I did find one more issue which is that the sprite info pane and the stage no longer line up on the side facing the code/sound/costume editors:

image

I think it's okay if this doesn't go into this PR.

PR #2993, opened 2018-08-27
[Master] Split BrowserModalComponent out of GUI

rschamp:
I know this is pretty ugly, but I think what we want to avoid crashing unsupported browsers

### Resolves

Resolves failing smoke test in #2980 (already merged). _Should_ future-proof us from this type of regression in the future (Hello from the past future us! Sorry!)

### Proposed Changes

This splits the BrowserModalComponent out of the GUI, and protects all GUI-related modules from being imported unless the GUI is actually being mounted.  This protects unsupported browsers from running code that will crash them (like the VM).

### Reason for Changes

If the VM is ever instantiated, which happens whenever GUI is instantiated due to default props, then IE11 will crash. So mount the BrowserModalComponent in a separate code path from the GUI, so that it can be displayed regardless of the code required by the GUI.

### Test Coverage

Failing smoke test now passes!

### Browser Coverage

Mac
 * [X] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [X] IE 11
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -115,8 +115,8 @@ const GUIComponent = props => {
                 dir={isRtl ? 'rtl' : 'ltr'}
                 {...componentProps}
             >
-                {previewInfoVisible ? (
-                    <PreviewModal hideIntro={hideIntro} />
+                {(previewInfoVisible && !hideIntro) ? (
+                    <PreviewModal />
                 ) : null}
                 {loading ? (
                     <Loader />
@@ -282,7 +282,6 @@ GUIComponent.propTypes = {
     costumeLibraryVisible: PropTypes.bool,
     costumesTabVisible: PropTypes.bool,
     enableCommunity: PropTypes.bool,
-    hideIntro: PropTypes.bool,
     importInfoVisible: PropTypes.bool,
     intl: intlShape.isRequired,
     isPlayerOnly: PropTypes.bool,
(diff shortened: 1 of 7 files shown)


rschamp:
Try it out: [hotfix/fix-browser-modal](https://llk.github.io/scratch-gui/hotfix/fix-browser-modal)

BryceLTaylor:
@chrisgarrity Now it's up to you!

src/components/gui/gui.jsx
```diff
@@ -115,8 +115,8 @@ const GUIComponent = props => {
                 dir={isRtl ? 'rtl' : 'ltr'}
                 {...componentProps}
             >
-                {previewInfoVisible ? (
-                    <PreviewModal hideIntro={hideIntro} />
+                {(previewInfoVisible && !hideIntro) ? (
```
chrisgarrity: You're not passing hideIntro into the component any more, so this should just be `previewInfoVisible`

src/lib/app-state-hoc.jsx
```diff
@@ -4,57 +4,80 @@ import {Provider} from 'react-redux';
...
-const enhancer = composeEnhancers(guiMiddleware);
 
 /*
  * Higher Order Component to provide redux state. If an `intl` prop is provided
  * it will override the internal `intl` redux state
  * @param {React.Component} WrappedComponent - component to provide state for
  * @returns {React.Component} component with redux and intl state provided
  */
-const AppStateHOC = function (WrappedComponent) {
+const AppStateHOC = function (WrappedComponent, localesOnly) {
```
chrisgarrity: Should add localesOnly param to the jsDoc

chrisgarrity:
By the way I also verified the unsupported browser on Opera (OSX)

PR #3006, opened 2018-08-28
Update readme with tips

LiFaytheGoblin:
Update readme with tips on what to do when linking or so didn't work correctly

I had no clue what to do when the local version of scratch-gui just wouldn't appear, so Ben and some others helped. This Readme is no updated with what I learned so others won't run into the same problems.


rschamp:
Try it out: [LiFaytheGoblin-update-readme](https://llk.github.io/scratch-gui/LiFaytheGoblin-update-readme)

benjiwheeler:
Looks good!

Issue #2815, opened 2018-08-08
Add title prop, onUpdateTitle, UI for changing title


benjiwheeler:

carljbowman:
Here is a sketch of these states within the Editor's title field
![title states 2x](https://user-images.githubusercontent.com/3409578/44742753-a56a6b80-aace-11e8-998a-bb10199ae734.png)

LiFaytheGoblin:
CSS is at https://github.com/LiFaytheGoblin/scratch-gui/tree/rename_title

Tested for MacOS: Chrome, Safari, Firefox

PR #3016, opened 2018-08-28
make project title editable, both in standalone and www modes

benjiwheeler:
Big help from pairing with @LiFaytheGoblin!!

relies on https://github.com/LLK/scratch-www/pull/2052 and should only be merged after it.

### Resolves

https://github.com/LLK/scratch-gui/issues/2815

Note: we haven't implemented the css changes per @carljbowman's comments on #2815.

### Proposed Changes

Makes project title editable in GUI editor. 

Uses new project title reducer to store and change project title. This is made available to the menu bar, where the project title input uses it to maintain state, and to project saver, which uses the title to determine the downloaded filename.

When run within www, title changes made in the GUI menu bar are updated in www and through it to the server.

Testing:

For testing, we should try unusual unicode characters.

### Related PR

https://github.com/LLK/scratch-www/pull/2052 in scratch-www should be used together with this.

### DIFF
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -13,6 +13,7 @@ import LanguageSelector from '../../containers/language-selector.jsx';
 import ProjectLoader from '../../containers/project-loader.jsx';
 import Menu from '../../containers/menu.jsx';
 import {MenuItem, MenuSection} from '../menu/menu.jsx';
+import ProjectTitleInput from './project-title-input.jsx';
 import ProjectSaver from '../../containers/project-saver.jsx';
 import DeletionRestorer from '../../containers/deletion-restorer.jsx';
 import TurboMode from '../../containers/turbo-mode.jsx';
@@ -393,11 +394,12 @@ class MenuBar extends React.Component {
                     </div>
                     <Divider className={classNames(styles.divider)} />
                     <div className={classNames(styles.menuBarItem)}>
-                        <MenuBarItemTooltip id="title-field">
-                            <input
-                                disabled
-                                className={classNames(styles.titleField)}
-                                placeholder="Untitled-1"
+                        <MenuBarItemTooltip
+                            enable
+                            id="title-field"
+                        >
+                            <ProjectTitleInput
+                                onUpdateProjectTitle={this.props.onUpdateProjectTitle}
                             />
                         </MenuBarItemTooltip>
                     </div>
@@ -521,7 +523,8 @@ MenuBar.propTypes = {
     onRequestCloseEdit: PropTypes.func,
     onRequestCloseFile: PropTypes.func,
     onRequestCloseLanguage: PropTypes.func,
-    onSeeCommunity: PropTypes.func
+    onSeeCommunity: PropTypes.func,
+    onUpdateProjectTitle: PropTypes.func
 };
 
 const mapStateToProps = state => ({
--- a/src/components/menu-bar/project-title-input.css
+++ b/src/components/menu-bar/project-title-input.css
@@ -17,11 +17,17 @@ $title-width: 12rem;
     padding: .5rem;
 }
 
-.title-field,
-.title-field::placeholder {
+.title-field {
     color: $ui-white;
     font-weight: bold;
     font-size: .8rem;
+}
+
+.title-field::placeholder {
+    color: $ui-white;
+    font-weight: normal;
+    font-size: .8rem;
+    font-style: italic;
 }
 
 .title-field:hover {
@@ -30,7 +36,7 @@ $title-width: 12rem;
 
 .title-field:focus {
     outline:none;
-    border: none;
+    border: 1px solid $ui-transparent;
     -webkit-box-shadow: 0 0 0 calc($space * .5) $ui-white-transparent;
     box-shadow: 0 0 0 calc($space * .5) $ui-white-transparent;
     background-color: $ui-white;
--- a/src/components/menu-bar/project-title-input.css
+++ b/src/components/menu-bar/project-title-input.css
@@ -1,0 +1,38 @@
+@import "../../css/colors.css";
+@import "../../css/units.css";
+@import "../../css/z-index.css";
+
+$title-width: 12rem;
+
+.title-field {
+    border: 1px dashed $ui-black-transparent;
+    border-radius: $form-radius;
+    -webkit-border-radius: $form-radius;
+    -moz-border-radius: $form-radius;
+    background-color: $ui-white-transparent;
+    background-clip: padding-box;
+    -webkit-background-clip: padding-box;
+    width: $title-width;
+    height: auto;
+    padding: .5rem;
+}
+
+.title-field,
+.title-field::placeholder {
+    color: $ui-white;
+    font-weight: bold;
+    font-size: .8rem;
+}
+
+.title-field:hover {
+    background-color: hsla(0, 100%, 100%, 0.5);
+}
+
+.title-field:focus {
+    outline:none;
+    border: none;
+    -webkit-box-shadow: 0 0 0 calc($space * .5) $ui-white-transparent;
+    box-shadow: 0 0 0 calc($space * .5) $ui-white-transparent;
+    background-color: $ui-white;
+    color: $text-primary;
+}
--- a/src/components/menu-bar/project-title-input.jsx
+++ b/src/components/menu-bar/project-title-input.jsx
@@ -3,12 +3,21 @@ import {connect} from 'react-redux';
 import PropTypes from 'prop-types';
 import bindAll from 'lodash.bindall';
 import React from 'react';
+import {defineMessages, intlShape, injectIntl} from 'react-intl';
 
 import BufferedInputHOC from '../forms/buffered-input-hoc.jsx';
 import Input from '../forms/input.jsx';
 const BufferedInput = BufferedInputHOC(Input);
 
 import styles from './project-title-input.css';
+
+const messages = defineMessages({
+    projectTitlePlaceholder: {
+        id: 'gui.gui.projectTitlePlaceholder',
+        description: 'Placeholder for project title when blank',
+        defaultMessage: 'Project title here'
+    }
+});
 
 class ProjectTitleInput extends React.Component {
     constructor (props) {
@@ -28,8 +37,8 @@ class ProjectTitleInput extends React.Component {
         return (
             <BufferedInput
                 className={classNames(styles.titleField)}
-                maxlength="100"
-                placeholder=""
+                maxLength="100"
+                placeholder={this.props.intl.formatMessage(messages.projectTitlePlaceholder)}
                 tabIndex="0"
                 type="text"
                 value={this.props.projectTitle}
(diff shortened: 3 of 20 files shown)


src/components/menu-bar/menu-bar.jsx
```diff
@@ -152,6 +155,26 @@ class MenuBar extends React.Component {
             this.props.onRequestCloseEdit();
         };
     }
+    // intercept key presses to detect enter key; make that trigger blur,
+    // and thus project title update
+    handleProjectTitleKeyPress (e) {
```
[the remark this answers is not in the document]
rschamp: [BufferedInputHOC](https://github.com/LLK/scratch-gui/blob/develop/src/components/forms/buffered-input-hoc.jsx) for reference.

LiFaytheGoblin:
CSS is at https://github.com/LiFaytheGoblin/scratch-gui/tree/rename_title

Tested for MacOS: Chrome, Safari, Firefox

LiFaytheGoblin:
Cool looks like this branch here got updated as well, so nevermind my repo link!

benjiwheeler:
@rschamp, we were thinking that we *do* want to always enable editing the title field -- if embedded in www, obviously; but also in beta, as the title for the file when it downloads.

src/components/menu-bar/menu-bar.jsx
```diff
@@ -355,11 +378,17 @@ class MenuBar extends React.Component {
...
+                        >
                             <input
-                                disabled
                                 className={classNames(styles.titleField)}
-                                placeholder="Untitled-1"
+                                placeholder=""
+                                value={this.props.projectTitle}
+                                onBlur={this.handleUpdateProjectTitle}
+                                onChange={this.handleProjectTitleChange}
+                                onKeyPress={this.handleProjectTitleKeyPress}
```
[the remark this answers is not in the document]
benjiwheeler: Done!

src/components/menu-bar/menu-bar.jsx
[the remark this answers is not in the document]
benjiwheeler: Done!

src/reducers/project-title.js
```diff
@@ -0,0 +1,27 @@
+const SET_PROJECT_TITLE = 'projectTitle/SET_PROJECT_TITLE';
+
+const projectTitleInitialState = {
+    projectTitle: 'Untitled-1'
```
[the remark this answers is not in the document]
benjiwheeler: Done!

src/containers/gui.jsx
```diff
@@ -100,12 +109,19 @@ GUI.propTypes = {
...
     previewInfoVisible: PropTypes.bool,
     projectData: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
     projectHost: PropTypes.string,
+    projectTitle: PropTypes.string,
     vm: PropTypes.instanceOf(VM)
 };
 
+GUI.defaultProps = {
+    projectTitle: 'Untitled-1'
+};
```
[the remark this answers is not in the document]
benjiwheeler: Done!

benjiwheeler:
Linda and I confirmed that although several tests fail, when run individually, they pass.

LiFaytheGoblin:
Note for changes to css: Don't use width = outline but box shadow:
border: none;
    border-radius: calc(calc(0.5rem / 2) * 2);
    box-shadow: 0 0 0 0.25rem hsla(0, 100%, 100%, 0.25);

LiFaytheGoblin:
one more css thing

src/components/menu-bar/project-title-input.css
```diff
@@ -0,0 +1,39 @@
...
+}
+
+.title-field:hover {
+    background-color: hsla(0, 100%, 100%, 0.5);
+}
+
+.title-field:focus {
+    outline:none;
+    border: none;
+    border-radius: calc($form-radius * 2); /* double the border width */
```
LiFaytheGoblin: border radius line needs to be removed

src/containers/gui.jsx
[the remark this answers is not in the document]
fsih: It looks like this is still here?

src/containers/project-saver.jsx
```diff
@@ -33,11 +33,7 @@ class ProjectSaver extends React.Component {
         document.body.appendChild(saveLink);
 
         this.props.saveProjectSb3().then(content => {
-            // TODO user-friendly project name
-            // File name: project-DATE-TIME
-            const date = new Date();
-            const timestamp = `${date.toLocaleDateString()}-${date.toLocaleTimeString()}`;
-            const filename = `untitled-project-${timestamp}.sb3`;
+            const filename = `${this.props.projectTitle}.sb3`;
```
fsih: The behavior is kind of weird if the title field is blank (downloads sb3.zip)

src/lib/project-loader-hoc.jsx
```diff
@@ -72,7 +72,6 @@ const ProjectLoaderHOC = function (WrappedComponent) {
                 assetHost,
                 projectHost,
                 projectId,
-                reduxProjectId,
```
fsih: Why is this removed?

src/containers/project-saver.jsx
benjiwheeler: Hm, good point. I'll give it a default in that case.

src/lib/project-loader-hoc.jsx
benjiwheeler: The line is a bug. It was left there by Ray by accident. Sorry, I should have pointed that out!

src/components/menu-bar/project-title-input.css
benjiwheeler: Done

benjiwheeler:
@fsih I made a default title if it's blank, and set a maxLength and tested it on Safari. Thanks for catching those bugs.

src/components/menu-bar/project-title-input.jsx
```diff
@@ -0,0 +1,61 @@
...
+    constructor (props) {
+        super(props);
+        bindAll(this, [
+            'handleUpdateProjectTitle'
+        ]);
+    }
+    // call onUpdateProjectTitle if it is defined (only defined when gui
+    // is used within scratch-www)
+    handleUpdateProjectTitle (newTitle) {
+        this.props.onSetProjectTitle(newTitle);
```
rschamp: I think it'd be more correct if this were something like:
```jsx
        if (this.props.onUpdateProjectTitle) {
            this.props.onUpdateProjectTitle(newTitle);
        } else {
            this.props.onSetProjectTitle(newTitle);
        }
```

That way the redux state won't get out of sync from the incoming prop, and changing the title would work without configuration.

When `onUpdateProjectTitle` is supplied, calling it should cause the parent component (www or the gui playground) to pass in a new title prop to the GUI component, and then that change should dispatch the redux state change, causing everything looking at the redux state to update.

rschamp: I also think there's an issue with the naming here. It's not clear what the difference is between `onUpdateProjectTitle` and `onUpdateProjectTitle`. I think it's because they're the same thing. If we move to the behavior I outlined above, then I think it would be possible to collapse these into a single prop.

src/playground/index.jsx
```diff
@@ -28,6 +28,13 @@ document.body.appendChild(appTarget);
 GUI.setAppElement(appTarget);
 const WrappedGui = HashParserHOC(AppStateHOC(GUI));
 
+// simple example of how you might manage project title externally.
+// Changing project title within GUI interface will update it here.
+let projectTitle = 'Untitled-1';
+const onUpdateProjectTitle = title => {
+    projectTitle = title;
```
rschamp: Does this really cause the GUI to re-render? I would expect this to happen silently, so you would need to have a wrapper component that held this in state (like in the player example).

rschamp:
Also, projects on scratch.mit.edu must have titles 100 characters or shorter. I guess that doesn't matter in standalone mode, but seems like a sensible cap if we want to add one into the title field here.

src/components/menu-bar/project-title-input.jsx
benjiwheeler: Done.

src/playground/index.jsx
benjiwheeler: NO it doesn't! Fixed.

src/playground/index.jsx
```diff
@@ -26,7 +27,39 @@ appTarget.className = styles.app;
...
+// simple example of how you might manage project title externally.
+// Changing project title within GUI interface will update it here.
+class TitledGUI extends React.Component {
+    constructor (props) {
+        super(props);
+        bindAll(this, [
+            'handleUpdateProjectTitle'
+        ]);
+        this.state = {
+            projectTitle: 'Untitled-1'
```
rschamp: Would you please update this default to "Scratch Project"?  This came up in a planning meeting, and is only relevant here, in the standalone playground, not scratch-www.

Also, this string should use `formatMessage` so it can be translated.

benjiwheeler:
@rschamp, please take a look at the titled-hoc.jsx and where it's used in player.jsx, playground/index.jsx, and render-gui.jsx

src/playground/index.jsx
```diff
@@ -27,7 +28,7 @@ if (supportedBrowser()) {
 
 } else {
     BrowserModalComponent.setAppElement(appTarget);
-    const WrappedBrowserModalComponent = AppStateHOC(BrowserModalComponent, true /* localesOnly */);
+    const WrappedBrowserModalComponent = AppStateHOC(TitledHOC(BrowserModalComponent, true /* localesOnly */));
```
rschamp: You shouldn't wrap this one, since it's just the unsupported browser modal. There's also an issue with parenthesis placement (`true` should be passed to AppStateHOC, not TitledHOC), which shows that we should change the API for AppStateHOC to be used like `AppStateHOC(true)(Component))` rather than the kind of odd and hard to read `AppStateHOC(Component, true)`.

src/containers/project-loader.jsx
```diff
@@ -75,6 +76,12 @@ class ProjectLoader extends React.Component {
         if (thisFileInput.files) { // Don't attempt to load if no file was selected
             this.props.openLoadingState();
             reader.readAsArrayBuffer(thisFileInput.files[0]);
+            if (thisFileInput.files[0].name) {
+                const matches = thisFileInput.files[0].name.match(/^(.*)\.sb3$/);
```
rschamp: You could match the first 100 characters here to be consistent with the cap elsewhere

src/containers/project-loader.jsx
```diff
@@ -75,6 +76,12 @@ class ProjectLoader extends React.Component {
...
+                if (matches) {
+                    this.props.onSetProjectTitle(matches[1]);
```
rschamp: Or just truncate to 100 here.

src/playground/index.jsx
benjiwheeler: Sorry, the parenthesis was sloppy of me! Should have caught that.

benjiwheeler: Done.

src/containers/project-loader.jsx
benjiwheeler: Done

src/containers/project-loader.jsx
benjiwheeler: Done

Issue #1771, opened 2018-04-13
Make menubar dropdowns close on action


paulkaplan:
Right now they stay open after downloading / importing projects.

Reported by @BryceLTaylor 

/cc @rschamp

PR #3021, opened 2018-08-28
Save now

### LINKED ISSUE #2968: Allow saving projects and assets
There should be a method for sending `PUT` and `POST` requests to the project and asset servers, so we can create and update projects.

rschamp:
### Dependent PRs
This requires https://github.com/LLK/scratch-projects/pull/76

### Resolves

- Addresses some of #1771 but there is a React issue with the load project menu item
- Resolves #2968 

### Proposed Changes

- Add post/put stores for projects and assets
- Augment the project saver component to allow saving projects to a server in addition to locally
- Add a reducer for the project id, so that the project saver component has access to it
- Hook up the "Save now" menu item to save the project to the store, conditionally if there is an active session. So this will continue to be disabled in the GUI playground, but will be enabled if you're logged in on the preview project page
- Also, while I was there, close the file menu after saving the project to your computer. See 
7e3b53a for why I couldn't do the same for loading a project

### Reason for Changes

We want to start testing how the GUI will be used on the website, and part of that is saving projects.

### Test Coverage

It breaks some tests. I'm fixing that. Hard to test this type of thing since it happens over the network though.

### Browser Coverage

Mac
 * [X] Chrome 
 * [X] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -140,6 +140,7 @@ class MenuBar extends React.Component {
             'handleRestoreOption',
             'restoreOptionMessage'
         ]);
+        this.state = {projectSaveInProgress: false};
     }
     handleLanguageMouseUp (e) {
         if (!this.props.languageMenuOpen) {
@@ -150,6 +151,18 @@ class MenuBar extends React.Component {
         return () => {
             restoreFun();
             this.props.onRequestCloseEdit();
+        };
+    }
+    handleUpdateProject (updateFun) {
+        return () => {
+            this.props.onRequestCloseFile();
+            this.setState({projectSaveInProgress: true},
+                () => {
+                    updateFun().then(() => {
+                        this.setState({projectSaveInProgress: false});
+                    });
+                }
+            );
         };
     }
     restoreOptionMessage (deletedItem) {
@@ -182,8 +195,19 @@ class MenuBar extends React.Component {
         }
     }
     render () {
+        const saveNowMessage = (
+            <FormattedMessage
+                defaultMessage="Save now"
+                description="Menu bar item for saving now"
+                id="gui.menuBar.saveNow"
+            />
+        );
         return (
-            <Box className={styles.menuBar}>
+            <Box
+                className={classNames(styles.menuBar, {
+                    [styles.saveInProgress]: this.state.projectSaveInProgress
+                })}
+            >
                 <div className={styles.mainMenu}>
                     <div className={styles.fileGroup}>
                         <div className={classNames(styles.menuBarItem)}>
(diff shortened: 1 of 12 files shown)


src/components/menu-bar/menu-bar.jsx
```diff
@@ -152,6 +154,24 @@ class MenuBar extends React.Component {
             this.props.onRequestCloseEdit();
         };
     }
+    handleUpdateProject (updateFun) {
+        return () => {
+            this.props.onRequestCloseFile();
+            this.setState({projectSaveInProgress: true},
+                () => {
+                    updateFun().then(() => {
+                        this.setState({projectSaveInProgress: false});
```
chrisgarrity: In the long run this will need error handling, but that's probably a separate task. It may be a little confusing not to have a confirmation that the project actually saved.

src/components/menu-bar/menu-bar.jsx
```diff
@@ -152,6 +154,24 @@ class MenuBar extends React.Component {
...
+                    });
+                }
+            );
+        };
+    }
+    handleCloseFileMenuAndThen (fn) {
+        return () => {
+            this.props.onRequestCloseFile();
+            fn();
```
chrisgarrity: This will also need error handling in the long run. Either as part of the fn(), or here if the fn returns an error.

chrisgarrity:
@rschamp none of the things I flagged are blockers, but it probably does need to have a passing build.

src/components/menu-bar/menu-bar.jsx
```diff
@@ -492,6 +525,7 @@ MenuBar.propTypes = {
 };
 
 const mapStateToProps = state => ({
+    canUpdateProject: typeof (state.session && state.session.session && state.session.session.user) !== 'undefined',
```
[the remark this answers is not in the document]
rschamp: Good point — I guess this should come in as another prop and put into the project info reducer.

src/components/menu-bar/menu-bar.jsx
rschamp: Errors and save confirmation should all be handled by the upcoming messaging system.

src/components/menu-bar/menu-bar.jsx
rschamp: Probably so, which means that it probably shouldn't be in a generic handler like this. I'll give it it's own method.

src/components/menu-bar/menu-bar.jsx
```diff
@@ -246,18 +277,20 @@ class MenuBar extends React.Component {
...
-                                            />
-                                        </MenuItem>
-                                    </MenuItemTooltip>
+                                    <ProjectSaver>{(saveProject, updateProject) => (
+                                        this.props.canUpdateProject ? (
+                                            <MenuItem onClick={this.handleUpdateProject(updateProject)}>
+                                                {saveNowMessage}
+                                            </MenuItem>
+                                        ) : (
+                                            <MenuItemTooltip
```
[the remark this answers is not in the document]
rschamp: Yes, it will need to cover 3 cases: can't save, can save this project, can remix this project.

rschamp:
Thanks @chrisgarrity! Captured these as follow-up items ^

PR #3058, opened 2018-08-31
Made border color of stage selector be same as other border colors

### LINKED ISSUE #3009: Sprite Info Area and Background selector have different colored borders
### Current Overview
The unselected state of the Background selector has a different border color than Sprite Info Area.
image15

These should be the same color. We use a 15% transparent black for our borders, both element uses the correct HSLA... The problem looks to be with how we are setting `box-sizing`.

In the case of the Sprite Info Area the boarder is on the outside of the element, such that the transparent black is appearing over the cool gray background color.

In the case of the Background selector the border is on the inside of the element, so rather than appearing with the cool gray background, it appears over the white.

### Next Steps
Ideally the `box-sizing` positioning of the border on each of these elements should be the same. As far as which is correct, Sprite Info Area appears as it should. The Background selector's border should appear as it does on the Sprite Info Area.

That said, `border-box` has been rather troublesome, so we should be open to alternative solutions.

LiFaytheGoblin:
### Resolves
#3009

### Proposed Changes
Just adds one line of code. Background color white will stop before border, so the border color is set correctly. 

### Test Coverage
Tested on MacOS Chrome
screen shot 2018-08-31 at 4 15 49 pm

### DIFF
--- a/src/components/stage-selector/stage-selector.css
+++ b/src/components/stage-selector/stage-selector.css
@@ -4,6 +4,7 @@
 $header-height: calc($stage-menu-height - 2px);
 
 .stage-selector {
+    background-clip: padding-box;
     display: flex;
     flex-direction: column;
     align-items: center;


LiFaytheGoblin:
Is anyone reviewing this?

LiFaytheGoblin:
Thanks @fsih !!

PR #3099, opened 2018-09-07
Attempt to get monitor labels from VM

cwillisf:
### Resolves

Along with LLK/scratch-vm#1565, this resolves LLK/scratch-vm#693.

### Proposed Changes

When the GUI needs the label for a monitor, allow the VM to have the first attempt at calculating that label. If the VM doesn't handle the request, fall back on the current behavior.

### Reason for Changes

The VM (specifically, the runtime) is the authority on extension blocks, so label decisions regarding extension blocks should be delegated to the VM.


src/lib/monitor-adapter.js
```diff
@@ -11,10 +11,11 @@ const isUndefined = a => typeof a === 'undefined';
  * @param {string} block.opcode - The opcode of the monitor
  * @param {object} block.params - Extra params to the monitor block
  * @param {string|number|Array} block.value - The monitor value
+ * @param {VirtualMachine} block.vm - the VM instance which owns the block
  * @return {object} The adapted monitor with label and category
  */
-export default function ({id, spriteName, opcode, params, value}) {
-    let {label, category, labelFn} = OpcodeLabels(opcode);
+export default function ({id, spriteName, opcode, params, value, vm}) {
+    let {label, category, labelFn} = (vm && vm.runtime.getLabelForOpcode(opcode)) || OpcodeLabels(opcode);
```
fsih: Maybe add a comment here that this is how extensions get opcode labels but otherwise LGTM!

cwillisf:
I just added a comment... let me know if you think of a better or more concise way to explain it.

PR #3102, opened 2018-09-09
Add variables to dropdown menu, and more goodies

towerofnix:
### Resolves

Resolves LLK/scratch-vm#1425: "For this sprite only" (local) variables are now shown in the dropdown.

Resolves LLK/scratch-blocks#1418: This is unintentional (a side-effect of reusing the basic code for the object dropdown), but the property dropdown now accepts blocks.

Fixes backdrop options (backdrop name, backdrop #) from being shown in sprite dropdowns, and vice versa (x position, etc, in stage dropdown). I couldn't spot an issue for this one.

**Depends on LLK/scratch-blocks#1712.** Merge this one after that one!

### Proposed Changes

This PR reworks how the `PROPERTY` dropdown of the "(property) of (object)" sensing reporter block works internally, allowing for the changes listed above.

Note that the property dropdown now accepts blocks. That's not intentional - it's just that I reused most of the code for the object dropdown. I'm not sure how to make it a "square" dropdown (i.e. one which does not accept blocks). In my opinion, letting it accept blocks is a good thing anyways, since it allows and encourages dynamically accessing variables (or even ordinary attributes, like "x position" and "direction"). The argument against it is that an unfamiliar user might try placing a block such as "volume" into that dropdown, and then be confused as to why it always returns 0. But I think that any such user would quickly figure out the error they made; I think the functional benefit greatly outweighs the minor potential difficulty.

That said, I'm okay with making the input not accept blocks, if it'd be preferred to merge this sooner and keep discussion about that topic in its particular issue.

There are some comments I have on specific parts of the code. I'll leave those in a review of the PR.

### Reason for Changes

These are nearly all bug-fixes / enhancements which make the Scratch 3.0 editor match 2.0.

### Test Coverage

I haven't added any new unit tests, but this has been tested manually:

* The "of" block can be dragged out of the flyout with default or changed dropdowns without any errors.
* With the stage selected, the dropdown includes (translated) options for properties relevant to the stage, and the project's global variables.
* With a sprite selected, the dropdown includes (translated) options for properties relevant to sprites, and all variables which are local to that sprite, and no global variables.
* With a block placed inside the object dropdown, the property dropdown includes generic properties applicable to sprites (x position, volume, etc), and no variables.
* Variables are sorted the same way other variable menus are sorted.

### DIFF
--- a/src/lib/blocks.js
+++ b/src/lib/blocks.js
@@ -15,7 +15,7 @@ export default function (vm) {
                     type: 'field_dropdown',
                     name: name,
                     options: function () {
-                        return start.concat(menuOptionsFn());
+                        return start.concat(menuOptionsFn(this));
                     }
                 }
             ],
@@ -110,6 +110,45 @@ export default function (vm) {
         return [[myself, '_myself_']].concat(spriteMenu());
     };
 
+    const variablePropertyMenu = function (thisValue) {
+        const sourceBlock = thisValue.sourceBlock_; // This is the <shadow>.
+        if (sourceBlock) {
+            const ofBlock = sourceBlock.parentBlock_; // This is the "of" block.
+            let block, blocks;
+            if (vm.editingTarget) {
+                blocks = vm.editingTarget.blocks;
+                block = blocks.getBlock(ofBlock.id);
+            }
+            if (!block) {
+                // The block may be in the flyout.
+                blocks = vm.runtime.flyoutBlocks;
+                block = blocks.getBlock(ofBlock.id);
+            }
+            if (block) {
+                // Get the name of the sprite which is selected. This is based on the OBJECT dropdown's selected value,
+                // which may be different from the value returned by a block placed within that input, if present.
+                // TODO: Variables should probably not be fetched at all if the OBJECT dropdown contains a block.
+                const objectInput = blocks.getInputs(block).OBJECT;
+                const objectInputBlock = blocks.getBlock(objectInput.block);
+                const valueField = blocks.getFields(objectInputBlock).OBJECT;
+                const objectName = valueField.value;
+
+                // Don't return options for local variables of the stage, since the stage's local variables are really
+                // the project's global variables.
+                if (objectName !== '_stage_') {
+                    const target = vm.runtime.getSpriteTargetByName(objectName);
+                    if (target) {
+                        // Pass true to skip the stage: we only want the sprite's own local variables.
+                        const variableNames = target.getAllVariableNamesInScopeByType('', true);
+                        return variableNames.map(name => [name, name]);
+                    }
+                }
+                return [];
+            }
+        }
+        return [];
+    };
+
     const soundColors = ScratchBlocks.Colours.sounds;
 
     const looksColors = ScratchBlocks.Colours.looks;
@@ -176,6 +215,13 @@ export default function (vm) {
         const stage = ScratchBlocks.ScratchMsgs.translate('SENSING_OF_STAGE', 'Stage');
         const json = jsonForMenuBlock('OBJECT', spriteMenu, sensingColors, [
             [stage, '_stage_']
+        ]);
+        this.jsonInit(json);
+    };
+
+    ScratchBlocks.Blocks.sensing_of_property_menu.init = function () {
+        const json = jsonForMenuBlock('PROPERTY', variablePropertyMenu, sensingColors, [
+            ['Top value, hard-coded', 'TODO...']
         ]);
         this.jsonInit(json);
     };
--- a/src/lib/blocks.js
+++ b/src/lib/blocks.js
@@ -145,38 +145,41 @@ export default function (vm) {
                 block = blocks.getBlock(ofBlock.id);
             }
             if (block) {
-                // Get the name of the sprite which is selected. This is based on the OBJECT dropdown's selected value,
-                // which may be different from the value returned by a block placed within that input, if present.
-                // TODO: Variables should probably not be fetched at all if the OBJECT dropdown contains a block.
+                // Get the shadow block of the OBJECT dropdown. If the block in the OBJECT slot is not a shadow block,
+                // the slot must be occupied by an actual block; in this case, do nothing, and leave the default list
+                // of sprite properties remaining.
                 const objectInput = blocks.getInputs(block).OBJECT;
                 const objectInputBlock = blocks.getBlock(objectInput.block);
-                const valueField = blocks.getFields(objectInputBlock).OBJECT;
-                const objectName = valueField.value;
+                if (objectInputBlock.shadow) {
+                    // Get the name of the object from the dropdown shadow block.
+                    const valueField = blocks.getFields(objectInputBlock).OBJECT;
+                    const objectName = valueField.value;
 
-                let target;
+                    let target;
 
-                // If the stage is the selected object (of the dropdown), return properties relevant to it.
-                if (objectName === '_stage_') {
-                    const backdropNumber = ScratchBlocks.ScratchMsgs.translate(
-                        'SENSING_OF_BACKDROPNUMBER', 'backdrop #');
-                    const backdropName = ScratchBlocks.ScratchMsgs.translate(
-                        'SENSING_OF_BACKDROPNAME', 'backdrop name');
-                    output = [
-                        [backdropNumber, 'backdrop #'],
-                        [backdropName, 'backdrop name'],
-                        [volume, 'volume']
-                    ];
-                    target = vm.runtime.getTargetForStage();
-                } else {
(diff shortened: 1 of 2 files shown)


src/lib/blocks.js
```diff
@@ -110,6 +110,75 @@ export default function (vm) {
         return [[myself, '_myself_']].concat(spriteMenu());
     };
 
+    const variablePropertyMenu = function (thisValue) {
```
towerofnix: In order to access the "of" block (and then its `OBJECT` input's value), I need to be able to access the "this" value of where this function is run. This argument should probably be named better than `thisValue`, though.

src/lib/blocks.js
```diff
@@ -110,6 +110,75 @@ export default function (vm) {
...
+        const volume = ScratchBlocks.ScratchMsgs.translate('SENSING_OF_VOLUME', 'volume');
+        let output = [
+            [xPosition, 'x position'],
+            [yPosition, 'y position'],
+            [direction, 'direction'],
+            [costumeNumber, 'costume #'],
+            [costumeName, 'costume name'],
+            [size, 'size'],
+            [volume, 'volume']
+        ];
```
towerofnix: The output array is initially created with these values so that if, for any reason, neither a sprite nor the stage can be found based on the object input, the sprite-specific options will be the ones which are shown by default.

src/lib/blocks.js
```diff
@@ -110,6 +110,75 @@ export default function (vm) {
...
+
+        const sourceBlock = thisValue.sourceBlock_; // This is the <shadow>.
+        const ofBlock = sourceBlock && sourceBlock.parentBlock_; // This is the "of" block.
```
towerofnix: This feels a bit hacky. It definitely works, though. There might be a `getParent` option or such, which would at least be better syntax-wise - I haven't checked yet.

src/lib/blocks.js
```diff
@@ -110,6 +110,75 @@ export default function (vm) {
...
+            }
+            if (!block) {
+                // The block may be in the flyout.
+                blocks = vm.runtime.flyoutBlocks;
+                block = blocks.getBlock(ofBlock.id);
+            }
+            if (block) {
+                // Get the name of the sprite which is selected. This is based on the OBJECT dropdown's selected value,
+                // which may be different from the value returned by a block placed within that input, if present.
+                // TODO: Variables should probably not be fetched at all if the OBJECT dropdown contains a block.
```
towerofnix: Making note of this. Actually the (whole) comment here is wrong - if you click the property dropdown while there is a block in the object dropdown, the editor stops working properly (the "of" block gets stuck to your cursor).

src/lib/blocks.js
```diff
@@ -110,6 +110,75 @@ export default function (vm) {
...
+                    // If the stage is the selected object (of the dropdown), only return properties relevant to it.
+                    const backdropNumber = ScratchBlocks.ScratchMsgs.translate(
+                        'SENSING_OF_BACKDROPNUMBER', 'backdrop #');
+                    const backdropName = ScratchBlocks.ScratchMsgs.translate(
+                        'SENSING_OF_BACKDROPNAME', 'backdrop name');
+                    output = [
+                        [backdropNumber, 'backdrop #'],
+                        [backdropName, 'backdrop name'],
+                        [volume, 'volume']
+                    ];
```
towerofnix: Doing `return [...]` would work just as fine as setting `output = [...]` here. Just personal style preference.

src/lib/blocks.js
```diff
@@ -110,6 +110,75 @@ export default function (vm) {
...
+                } else {
+                    const target = vm.runtime.getSpriteTargetByName(objectName);
+                    if (target) {
+                        // Pass true to skip the stage: we only want the sprite's own local variables.
+                        const variableNames = target.getAllVariableNamesInScopeByType('', true);
```
towerofnix: This uses a function which wasn't particularly designed for this purpose, but it's perfectly appropriate for the situation. It returns an array of strings which are the names of the local variables of the sprite - exactly what we need to build the menu options.

towerofnix:
Travis is having a whole bunch of errors - they probably have to do with scratch-blocks not being updated by this PR. There aren't any errors when I do `npm run test:unit` on my computer.

src/lib/blocks.js
```diff
@@ -110,6 +110,75 @@ export default function (vm) {
...
+                // Get the name of the sprite which is selected. This is based on the OBJECT dropdown's selected value,
+                // which may be different from the value returned by a block placed within that input, if present.
+                // TODO: Variables should probably not be fetched at all if the OBJECT dropdown contains a block.
+                const objectInput = blocks.getInputs(block).OBJECT;
+                const objectInputBlock = blocks.getBlock(objectInput.block);
+                const valueField = blocks.getFields(objectInputBlock).OBJECT;
+                const objectName = valueField.value;
+
+                // Only return options for local variables of sprites, since the stage's local variables are really
+                // the project's global variables.
```
joker314: Nice PR! Note that this differs from 2.0 behaviour as global variables were listed in the menu options when `_stage_` was selected.

towerofnix: Thanks, updated.

(I will be away from my computer for a day or two so my responses here could be slow.)

src/lib/blocks.js
towerofnix: This is irrelevant now since we return the stage's local variables (the project's global variables).

src/lib/blocks.js
towerofnix: Resolved - updated the code to not show anything if the slot is occupied by a block. (We find out if it is by checking if the block in the slot is a shadow block: if it is, the dropdown is empty; if it isn't, it's filled with a reporter.)

towerofnix:
Hmm... Tricky problem with this is that existing Scratch 3.0 projects are broken. 2.0 projects adjust properly, but 3.0 projects lose the data previously stored in the PROPERTY field (which is now an input). (Note the console message `Ignoring non-existent field PROPERTY in block sensing_of` from blockly.)

kchadha:
@towerofnix, sorry for the long delay in reviewing this pull request.

I'm afraid since a side effect of this PR is that the property menu of the `sensing_of` block becomes droppable, we cannot accept this PR.

towerofnix:
@kchadha No worries.

Any input on the code which would be needed to implement this, then? The only other dropdown which is not-droppable *and* dynamically-filled is the variable/list/message dropdowns, but these are all implemented in ways which rely on accessing variables either specific the currently selected sprite or shared across all sprites. ([The code for doing so](https://github.com/LLK/scratch-blocks/blob/c9b51592f5bfb3937985b308929583f431661d9c/blocks_vertical/data.js#L607-L626) is in scratch-blocks, which indeed does not have access to other sprites the same way the GUI and VM do.)

And I know the Scratch Team is not obligated to, but I request that you consider the paragraph I've quoted below. And, by necessity, I request that it be considered **soon**: making the block droppable after the release of 3.0 will break many 3.0 projects, but making it before will break virtually none, due to so few existing.

> Note that the property dropdown now accepts blocks. That's not intentional - it's just that I reused most of the code for the object dropdown. I'm not sure how to make it a "square" dropdown (i.e. one which does not accept blocks). In my opinion, letting it accept blocks is a good thing anyways, since it allows and encourages dynamically accessing variables (or even ordinary attributes, like "x position" and "direction"). The argument against it is that an unfamiliar user might try placing a block such as "volume" into that dropdown, and then be confused as to why it always returns 0. But I think that any such user would quickly figure out the error they made; I think the functional benefit greatly outweighs the minor potential difficulty.

towerofnix:
PS, several existing projects would be broken by **not** making the input droppable. The fact is that unless LLK/scratch-vm#1030 - a rather colossal task nobody has gotten anywhere seriously far with since its creation in April - is resolved, any blocks which use hacked arguments will break. Several projects do place arguments inside the "of" block; while I don't have any statistic, it's one of the more common uses of hacked arguments that I've seen. **Not making the input droppable means these projects will break.**

kchadha:
> Any thoughts on the code which would be needed to implement this, then? 

@towerofnix, that's a great question, as you pointed out, it's definitely not straightforward given the current examples we have of making block menus dynamic. @picklesrus and I looked into this a bit and the official blockly way to do it is to replace the "options" part of the JSON block definition for the property field with a function that populates the menu options. From just a preliminary look at what it would take to do this, there are definitely some tricky bits and also some architectural issues to figure out on our end, especially since the function will have to take in information about what is selected on the other menu in the block.

Looking at the original issue that this PR was trying to fix, I noticed that it does not have a help wanted label on it. I am inclined to close this and the related scratch-blocks PRs since the implementation for the issue fix will require some more thought.

PR #3106, opened 2018-09-11
Make title field shrink for smaller screens

### LINKED ISSUE #2641: Toolbar too wide for ipad, cuts off username
### Expected Behavior

The toolbar should contain the username and not scroll

### Actual Behavior

The username is too far to the right and wraps to a new line.  It also overlaps the edge of the screen making it not all visible.  This causes the whole screen to be scrollable to the right.

ipadnamecutoff
ipadcanscrollhorizontally

### Steps to Reproduce

Launch the editor
Look at the upper right of the screen
Scroll to the right

IOS 11.0.3 Safari

LiFaytheGoblin:
### Resolves
#2641 

### Proposed Changes
I did some changes to flex box and items to make the title field shrink

### Reason for Changes
If the title field doesn't shrink the login box will lead to a scrolling condition on smaller screens. Very ugly.

### Test Coverage
MacOS Chrome + Safari, resized screens. Works down to 980px. Npm test passes.

### DIFF
--- a/src/components/menu-bar/project-title-input.css
+++ b/src/components/menu-bar/project-title-input.css
@@ -2,6 +2,11 @@
 @import "../../css/units.css";
 @import "../../css/z-index.css";
 
+/*
+If project-title-input.jsx is part of a menu bar say menu-bar.jsx, it can have additional css classes that
+can set a width for example or what it should do in a flex box (eg. grow).
+*/
+
 .title-field {
     border: 1px dashed $ui-black-transparent;
     border-radius: $form-radius;
--- a/src/components/menu-bar/project-title-input.css
+++ b/src/components/menu-bar/project-title-input.css
@@ -2,8 +2,6 @@
 @import "../../css/units.css";
 @import "../../css/z-index.css";
 
-$title-width: 12rem;
-
 .title-field {
     border: 1px dashed $ui-black-transparent;
     border-radius: $form-radius;
@@ -12,11 +10,8 @@ $title-width: 12rem;
     background-color: $ui-white-transparent;
     background-clip: padding-box;
     -webkit-background-clip: padding-box;
-    width: $title-width;
     height: auto;
     padding: .5rem;
-    flex-grow: 1;
-    width: 2rem;
 }
 
 .title-field {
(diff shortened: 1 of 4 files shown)


src/components/menu-bar/project-title-input.css
```diff
@@ -15,6 +15,8 @@ $title-width: 12rem;
     width: $title-width;
     height: auto;
     padding: .5rem;
+    flex-grow: 1;
```
rschamp: Avoid putting layout properties like `flex-*` on components. The parent component placing it there should control these styles.  Instead allow the parent component to pass it a custom class and allow it to apply the styles to that class.

src/components/menu-bar/project-title-input.css
```diff
@@ -15,6 +15,8 @@ $title-width: 12rem;
...
+    width: 2rem;
```
rschamp: I think it is debatable if even `width` should be controlled by the component (rather than the parent).

src/components/menu-bar/project-title-input.css
LiFaytheGoblin: Oh, that makes sense, will do!

src/components/menu-bar/project-title-input.css
LiFaytheGoblin: Good point I will investigate!

src/components/menu-bar/project-title-input.css
LiFaytheGoblin: Um question: It seems like I would need to import the parents css into the child component. This seems not right. Is there a good way on how to pass a parent's class to it's child?

LiFaytheGoblin: I am doing this right now:

```<ProjectTitleInput
                                className={classNames(styles.titleFieldGrowable)}
                                onUpdateProjectTitle={this.props.onUpdateProjectTitle}
                            />
```

and in the child:

```<BufferedInput
                className={classNames(styles.titleField, this.props.className)}
```

and I did 
```ProjectTitleInput.propTypes = {
    className: PropTypes.string,
```

LiFaytheGoblin: It works today. No idea why it didn't yesterday

LiFaytheGoblin:
Something is up with the css class I'll see what the problem is

PR #3132, opened 2018-09-13
Set lang attribute for screen readers

### LINKED ISSUE #3049: Gui should set lang attribute for accessibility
### Expected Behavior

When we set the gui locale, we should also update the lang attribute to indicate the default language of the page.
```
document.documentElement.lang = ...;
```

It should definitely be done for GUI standalone. It may need some work for everyone to play nice when gui is embedded in www.

LiFaytheGoblin:
### Resolves
#3049 

### Proposed Changes
Sets the lang attribute of the document in the GUI

### Reason for Changes
Setting the correct language is crucial for screen readers so they use the correct pronounciation

### Test Coverage
MacOS, Chrome, not tested with a screen reader

### DIFF
--- a/src/containers/language-selector.jsx
+++ b/src/containers/language-selector.jsx
@@ -27,6 +27,7 @@ class LanguageSelector extends React.Component {
             children,
             ...props
         } = this.props;
+        document.documentElement.lang = this.props.currentLocale; //update the lang attribute of the html for screenreaders
         return (
             <LanguageSelectorComponent
                 onChange={this.handleChange}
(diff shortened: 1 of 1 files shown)


src/containers/language-selector.jsx
```diff
@@ -13,11 +13,13 @@ class LanguageSelector extends React.Component {
         bindAll(this, [
             'handleChange'
         ]);
+        document.documentElement.lang = this.props.currentLocale;
```
chrisgarrity: This line would just be `props.currentLocale`

LiFaytheGoblin:
@chrisgarrity When switching back to www the locale set in the gui is still set as attribute. Is this an issue I should worry about for now?

PR #3143, opened 2018-09-14
Testing a new gui component for alerts

evhan55:
- Basic alert component that displays when a peripheral disconnects in both editor and full screen mode that can be dismissed with an 'x' button
- This component is reusable for other types of alerts in the future with further customization

### DIFF
--- a/src/components/alerts/alert.css
+++ b/src/components/alerts/alert.css
@@ -6,22 +6,30 @@
     position:absolute;
     z-index: $z-index-card;
     margin: 0.5rem 2rem;
+    border: 1px solid #FF8C1A;
+    border-radius: 8px;
+    cursor: move;
+    background: #FFF0DF;
+    display: flex;
+    flex-direction: row;
+    justify-content: center;
+    width: 448px;
+    padding: 16px;
+    font-size: 10pt;
+    box-shadow: 2px 2px 2px 2px rgba(255, 140, 26, 0.25);
 }
 
 .alert {
-    border: 1px solid #FF8C1A;
-    border-radius: 8px;
-    display: flex;
-    flex-direction: column;
-    cursor: move;
-    background: #FFF0DF;
-    width: 448px;
     color: #000;
     z-index: 20;
     overflow: hidden;
-    padding: 16px;
-    font-size: 10pt;
-    box-shadow: 2px 2px 2px 2px rgba(255, 140, 26, 0.25);;
     align-items: left;
+    width: 100%;
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+}
+
+.remove-button {
+    cursor: pointer;
+    color: #000;
+    align-items: center;
 }
--- a/src/components/alerts/alert.css
+++ b/src/components/alerts/alert.css
@@ -1,54 +1,29 @@
 @import "../../css/units.css";
 @import "../../css/colors.css";
 @import "../../css/z-index.css";
+@import "../gui/gui.css";
 
-.alert-container {
-    position:absolute;
-    z-index: $z-index-alert;
-    top: 53px;
-    left: 0;
-    right: 0;
-    margin: auto;
-    border: 1px solid #FF8C1A;
-    border-radius: 8px;
+.alert {
+    width: 100%;
     background: #FFF0DF;
     display: flex;
     flex-direction: row;
-    justify-content: center;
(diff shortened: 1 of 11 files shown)


src/reducers/alerts.js
```diff
@@ -0,0 +1,52 @@
...
+};
+
+const showAlert = function (message) {
+    return {
+        type: SHOW_ALERT,
+        message
+    };
+};
+
+const onReconnect = function () {
```
rschamp: I think this is too tightly connected to the component using the reducer. I don't think this action is necessary when it can just use `closeAlert` when it reconnects.

src/containers/alerts.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+const mapStateToProps = state => ({
+    visible: state.scratchGui.alerts.visible,
+    message: state.scratchGui.alerts.message
+});
+
+const mapDispatchToProps = dispatch => ({
+    onShowAlert: () => dispatch(showAlert()),
+    onCloseAlert: () => dispatch(closeAlert()),
+    onReconnect: () => dispatch(onReconnect())
```
rschamp: Looking at this more, I don't think this is actually used anywhere, so it can be removed altogether.

src/components/gui/gui.css
```diff
@@ -278,3 +278,15 @@ $fade-out-distance: 15px;
...
+/* Alerts */
+
+.alerts-container {
+    width: 448px;
+    z-index: $z-index-alerts;
+    left: 0;
+    right: 0;
+    margin: auto;
+    position: absolute;
+    margin-top: 53px;
```
evhan55: @rschamp , I experimented with putting the alerts container inside of the flexWrapper so that this margin could be more like 5px vs. 53px, but then it seemed very strange to put the alerts container in such a wrapper that seemed to tightly bound to tabs/stage placement.  So, I left the alerts container in the main guiWrapper and made the margin bigger.

src/components/alerts/alert.css
```diff
@@ -0,0 +1,29 @@
+@import "../../css/units.css";
+@import "../../css/colors.css";
+@import "../../css/z-index.css";
+@import "../gui/gui.css";
```
evhan55: This felt a little weird, but seems to be the only way to get the alertsContainer style in gui.css to be accessible in the alert component?

evhan55:
@rschamp Thank you for the review!  I removed the cruft, moved some CSS and left you a couple of notes/questions.

src/components/alerts/alert.css
rschamp: To share styles this way, pass a `className` prop to `Alerts` from GUI.  Then the `Alerts` component can combine the class names passed in from props, and its own.

E.g. `<Box className={classNames(styles.alertsContainer, props.className)}> ... </Box>`

evhan55: Ah, got it! Thank you.

src/components/alerts/alerts.jsx
```diff
@@ -1,26 +1,29 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Box from '../box/box.jsx';
+import Button from '../button/button.jsx';
 
 import styles from './alert.css';
 
-// this is a functional component, declared with arrow syntax
 const Alerts = props => (
```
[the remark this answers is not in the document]
evhan55: Okay, will do.

src/css/z-index.css
```diff
@@ -23,6 +23,7 @@ $z-index-stage-color-picker-background: 2000;
 $z-index-stage-with-color-picker: 2010;
 $z-index-stage-header: 5000;
 $z-index-stage-wrapper-overlay: 5000;
+$z-index-alerts: 5010;
```
evhan55: I had to move this to be above the stage.

src/components/alerts/alerts.jsx
```diff
@@ -0,0 +1,41 @@
...
+import styles from './alert.css';
+
+const Alerts = ({
+    className,
+    message,
+    onCloseAlert
+}) => (
+    <Box
+        bounds="parent"
+        className={classNames(styles.alertsContainer, className)}
```
rschamp: I don't think you need `styles.alertsContainer` anymore, it's not defined in `alert.css`.

rschamp:
Don't worry about the test failure, it's something @BryceLTaylor is working on, unrelated to this.

PR #3145, opened 2018-09-14
Localize default project

### LINKED ISSUE #2070: Translating the default sprite/costume/backdrop names

chrisgarrity:
### Resolves

- Resolves #2070 

### Proposed Changes

Wraps the project loader HOC in injectIntl, and then passes formatMessage as the function to use for translation when generating the project data.

### Reason for Changes

Intermediate step for localizing the default project. Eventually the default project should be built programmatically.

### Test Coverage

Updated project loader test.


src/lib/default-project/project-data.js
```diff
@@ -0,0 +1,150 @@
+import {defineMessages} from 'react-intl';
+
+const messages = defineMessages({
+    backdrop: {
+        defaultMessage: 'backdrop1',
```
rschamp: Seeing it written out like this makes me wonder why we don't use spaces in these names... They aren't variable names or anything. It's definitely too late to change it though :/

src/lib/default-project/project-data.js
```diff
@@ -0,0 +1,150 @@
...
+    }
+});
+
+// use the default message if a translation function is not passed
+const defaultTranslator = msgObj => msgObj.defaultMessage;
+
+/**
+ * Generate a localized version of the default project
+ * @param {function} translateFunction a function to use for translating the default names
+ * @return {json} the project data json for the default project
```
rschamp: This would be `string`, but it doesn't return a string. But it could! But probably it shouldn't. The `object` it returns is more versatile.

src/lib/storage.js
```diff
@@ -9,6 +9,7 @@ import defaultProjectAssets from './default-project';
 class Storage extends ScratchStorage {
     constructor () {
         super();
+        const defaultProjectAssets = defaultProject(this.translator);
         defaultProjectAssets.forEach(asset => this.cache(
```
rschamp: These 2 calls look like they can be deduplicated with one `this.cacheDefaultProject`?

src/containers/target-pane.jsx
```diff
@@ -127,8 +132,9 @@ class TargetPane extends React.Component {
     }
     handleSpriteUpload (e) {
         const storage = this.props.vm.runtime.storage;
+        const costumeSuffix = this.props.intl.formatMessage(sharedMessages.costume, {index: 1});
         handleFileUpload(e.target, (buffer, fileType, fileName) => {
-            spriteUpload(buffer, fileType, fileName, storage, this.handleNewSprite);
+            spriteUpload(buffer, fileType, fileName, storage, this.handleNewSprite, costumeSuffix);
```
rschamp: These upload functions are super overloaded...  I'm pretty confused by `costumeSuffix` here — why does sprite upload need a suffix?

src/lib/file-uploader.js
```diff
@@ -180,7 +180,8 @@ const soundUpload = function (fileData, fileType, soundName, storage, handleSoun
     handleSound(vmSound);
 };
 
-const spriteUpload = function (fileData, fileType, spriteName, storage, handleSprite) {
+const spriteUpload = function (fileData, fileType, spriteName, storage, handleSprite, costumeSuffix) {
```
rschamp: Maybe it would be helpful if there was a jsdoc? I'm not sure what to suggest, this is a lot of arguments.

src/containers/target-pane.jsx
chrisgarrity: Sprite upload creates a costume from the image uploaded. The name of the costume is the filename with `-costume1` appended. Rather than making fileupload locale aware, I added the costume suffix so the caller can take care of localizing. The caller is already where localizing new sprite happens.

PR #3146, opened 2018-09-14
Update paint to version that fixes crash when drawing transparent shapes in bitmap

fsih:
Diff is https://github.com/LLK/scratch-paint/pull/682/files


rschamp:
Try it out: [updatePaintSmoke2](https://llk.github.io/scratch-gui/updatePaintSmoke2)

PR #3170, opened 2018-09-18
Increase timeout in integration test for loading project with url

BryceLTaylor:
### Resolves

Increases the timeout in a test  "Load a project by ID directly through url', which as been flaky

### Proposed Changes

Increases the timeout from 2 seconds to 3.5 seconds while the project loads.

### Reason for Changes

The test keeps failing by timing out, but the project loads, so it shouldn't.  It doesn't always fail, so I'm increasing the timeout.


rschamp:
Try it out: [fix-tests](https://llk.github.io/scratch-gui/fix-tests)

Issue #3173, opened 2018-09-18
Regression: Geometry dash is super slow


fsih:
### Steps to Reproduce
Press green flag on geometry dash
https://llk.github.io/scratch-gui/develop/#105500895

Regression was traced to https://github.com/LLK/scratch-gui/commit/8bfaf69d6e0802917a94121770d9e37debd07996

### Operating System and Browser
all

fsih:
Filed offshoot bug https://github.com/LLK/scratch-gui/issues/3172

Issue #3172, opened 2018-09-18
dragPayload causes many things to rerender every frame


fsih:
### Expected Behavior
1. When nothing is being dragged, dragPayload should not be sent in props
2. When something is being dragged, if the important attributes of dragPayload haven't changed, it shouldn't trigger a rerender

### Actual Behavior
Some performance issues were partially caused by dragPayload changing so frequently whenever the project was playing: https://llk.github.io/scratch-gui/develop/#228228816
![dragpayload](https://user-images.githubusercontent.com/2855464/45719088-392ed500-bb6d-11e8-86c0-79c31700f98c.gif)

### Steps to Reproduce
E.g. put a log statement in render in sprite-selector-item.jsx to see how often it is called

### Operating System and Browser
all

PR #3174, opened 2018-09-18
fixRepeatDecode

fsih:
### Resolves
This (mostly) fixes https://github.com/LLK/scratch-gui/issues/3173 (https://llk.github.io/scratch-gui/develop/#105500895 runs super slow)

### Proposed Changes
decodeText takes a substantial amount of time (making each frame take ~200ms in the problem project). We were caching to prevent too many calls to svgRenderer.loadString. Now include decodeText in the caching too.

Additionally, sprite selector items don't need to rerender every frame. dragPayload is giving a different object every frame, even though every property of the objects match. For now, ignore dragPayload in shouldComponentUpdate. https://github.com/LLK/scratch-gui/issues/3172 is filed as a separate bug to clean this up later.

### Reason for Changes
Performance fix

### Test Coverage
Play geometry dash until you can finish all the levels. Perseverance is key.
Test that dragging still works as expected.

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/sprite-selector-item.jsx
+++ b/src/containers/sprite-selector-item.jsx
@@ -32,7 +32,17 @@ class SpriteSelectorItem extends React.Component {
         ]);
         this.svgRenderer = new SVGRenderer();
         // Asset ID of the SVG currently in SVGRenderer
-        this.svgRendererAssetId = null;
+        this.decodedAssetId = null;
+    }
+    shouldComponentUpdate (nextProps) {
+        // Ignore dragPayload due to https://github.com/LLK/scratch-gui/issues/3172.
+        // This function should be removed once the issue is fixed.
+        for (const property in nextProps) {
+            if (property !== 'dragPayload' && this.props[property] !== nextProps[property]) {
+                return true;
+            }
+        }
+        return false;
     }
     getCostumeUrl () {
         if (this.props.costumeURL) return this.props.costumeURL;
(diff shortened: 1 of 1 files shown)


src/containers/sprite-selector-item.jsx
```diff
@@ -32,7 +32,17 @@ class SpriteSelectorItem extends React.Component {
         ]);
         this.svgRenderer = new SVGRenderer();
         // Asset ID of the SVG currently in SVGRenderer
-        this.svgRendererAssetId = null;
+        this.decodedAssetId = null;
+    }
+    shouldComponentUpdate (nextProps) {
+        // Ignore dragPayload due to https://github.com/LLK/scratch-gui/issues/3172.
+        // This function should be removed once the issue is fixed.
```
cwillisf: Once this is merged, it might be a good idea to add a reminder on #3172 to remove this workaround

src/containers/sprite-selector-item.jsx
```diff
@@ -44,18 +54,20 @@ class SpriteSelectorItem extends React.Component {
         // Avoid parsing the SVG when possible, since it's expensive.
         if (asset.assetType === storage.AssetType.ImageVector) {
             // If the asset ID has not changed, no need to re-parse
-            if (this.svgRendererAssetId === this.props.assetId) {
+            if (this.decodedAssetId === this.props.assetId) {
+                // @todo consider caching more than one URL.
```
cwillisf: This might be easy with something like https://www.npmjs.com/package/lru-cache

src/containers/sprite-selector-item.jsx
fsih: Added!

src/containers/sprite-selector-item.jsx
fsih: Ray was arguing that even the current amount of caching might be premature optimization, so we would have to discuss with him

cwillisf: Ah, interesting. Is the idea that if we can avoid calling `decodeText` every frame by some other means then caching here might be unnecessary? Should we meet about this or just chat informally?

fsih: Ray's argument was ideally sprite-selector-item in its shouldComponentUpdate would only update when assetId changed, which would make the optimization here unnecessary.

However, sprite-selector-item listens on quite a few props, and it's easy to add props without realizing that they might be causing very expensive rerenders (such as dragPayload here), so I think checking the case where the assetID hasn't changed here is worthwhile.

I think that "premature optimization" is only really bad when it's a waste of time or overhead, and this change is small (so far), also here it's not premature because it can cause this big performance issue. I think Ray let it go, possibly because he wanted to go home, but we could all discuss it again if you think that would be good.

PR #3192, opened 2018-09-20
Merge to smoke branch

thisandagain:
### Features
- Added full support for "Right to Left" languages (e.g. Hebrew, Arabic)
- Added support for variable / reporter monitors within extensions

### Bug Fixes
- Minor fixes and block language adjustments to the "micro:bit" extension
- Minor fixes to extensions that connect to hardware ("micro:bit", "LEGO WeDo 2.0", etc)
- Fixed various issues in the costume editor
     - Resizing an object keeps it's ratio if you drag the corners
    - Text in the bitmap editor no longer disappears if you don’t click off of it and change costumes
    - Filled / outlined state and line thickness now stays in sync with the selected shape in bitmap editor if you change options and then undo / redo
    - Text outline thickness matches between stage and editor
    - Pressing backspace or delete in Mozilla Firefox doesn’t try to navigate when paint is open
- Fixed various issues with importing projects from Scratch 2.0
    - Corrupt SVG costumes can no longer cause a crash
    - String and/or numeric reporters plugged into boolean input slots no longer cause a crash

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,6 @@ Instead of `BUILD_MODE=dist npm run build` you can also use `BUILD_MODE=dist npm
 * In the worst case unlink the repositories by running `npm unlink` in both, and start over.
 
 ## Testing
-
 ### Documentation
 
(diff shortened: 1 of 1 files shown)


rschamp:
Try it out: [develop](https://llk.github.io/scratch-gui/develop)

PR #3193, opened 2018-09-20
Remove duplicate deps

fsih:


fsih:
I don't think it changes the build? I figured it was just a bit confusing.

thisandagain:
I don't think so either, but it's kinda gross. :-P

PR #3200, opened 2018-09-20
Expose AppStateHOC component

cwillisf:
### Proposed Changes

For use from `scratch-desktop` and maybe other places.

### Reason for Changes

The `scratch-desktop` repository can use this component to more easily host the "whole" editor.

The main question I have is: is this the right component to use? If not, is there another that would make more sense or should we make a new editor-wrapper component?

### Test Coverage

Since this is just exposing existing code at the module boundary, there's no new code to test.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,5 @@
 import GUI from './containers/gui.jsx';
+import AppStateHOC from './lib/app-state-hoc.jsx';
 import GuiReducer, {guiInitialState, guiMiddleware, initFullScreen, initPlayer} from './reducers/gui';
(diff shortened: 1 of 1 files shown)


towerofnix:
Are there plans to open-source the `scratch-desktop` repository?

cwillisf:
@towerofnix yep! Right now `scratch-desktop` is very early and turbulent so it would be difficult for someone outside the team to contribute, but we'll open-source it as soon as things stabilize a bit. That goes for other repositories as well: the `scratch-link` repository will probably be opened to the public very soon, for example.

Issue #2956, opened 2018-08-22
Alerts for hardware and other info: overall design


kathymakes:
## Overview

We need a way to tell Scratchers when their hardware extension device loses connection.
This design is for an alert message which directs users to the device modal.

## Current Scope

The alert appears when an extension device loses connection.
Clicking "x" dismisses the alert.
![device is disconnected 1](https://user-images.githubusercontent.com/8203939/44667868-f6546400-a9e9-11e8-9c12-3fb5b81a93bf.png)

Clicking "reconnect" triggers the device modal. The original alert should disappear.
![device is disconnected 2](https://user-images.githubusercontent.com/8203939/44667877-ff453580-a9e9-11e8-9264-f1a50c42f142.png)

The alert can also appear in full screen stage mode.
Clicking "reconnect" takes you back to the editor page w/ device modal.
![full screen alerts 1](https://user-images.githubusercontent.com/8203939/44667895-0b30f780-a9ea-11e8-967b-4fddc57bd84d.png)

More than one alert can appear at once. Multiple alerts should stack.
![multiple alerts 3](https://user-images.githubusercontent.com/8203939/44667919-184de680-a9ea-11e8-82b9-b2b2cad9fa41.png)

Stacking behavior persists in full screen stage.
![full screen alerts 2](https://user-images.githubusercontent.com/8203939/44667963-361b4b80-a9ea-11e8-9e21-66b3e5431fdc.png)

Although multiple alerts case is probably rare... the stacking solution isn't ideal. We might revisit a solution that doesn't dominate the entire workspace. 

# Future Considerations

### Extensibility
Alerts could also be useful in other use cases:
- Downloading/uploading confirmations
This case includes a "progress" alert and a "success" alert. 
![download flow 2](https://user-images.githubusercontent.com/8203939/44667987-49c6b200-a9ea-11e8-8741-86dc3039d625.png)
![download flow 3](https://user-images.githubusercontent.com/8203939/44667988-49c6b200-a9ea-11e8-889d-ea41d7cdb56c.png)

(Success alert should disappear on a timer.)

- Lost wifi connection
![wifi is disconnected 1](https://user-images.githubusercontent.com/8203939/44668019-5e0aaf00-a9ea-11e8-9f21-243fe7e9fb28.png)

### Other Extensions Messaging
There are a couple situations where we should communicate that hardware is not connected, with or without alerts.
- What happens if a Scratcher loads a project with hardware extensions, without a device already connected?
[...]

PR #3209, opened 2018-09-22
Show multiple alerts in a stack and allow for custom alert icons.

evhan55:
### Resolves

This is progress towards: https://github.com/LLK/scratch-gui/issues/2956.

### Proposed Changes

- Displays stacks of alerts vertically.
- Shows an icon in an alert if one is provided.

screen shot 2018-09-22 at 5 40 22 pm

### Reason for Changes

To improve on previous alerts that displayed directly on top of each other, and to add further customizability with optional icons.

### Notes

- Is a prerequisite for this PR in scratch-vm: https://github.com/LLK/scratch-vm/pull/1607.

### DIFF
--- a/src/components/alerts/alert.jsx
+++ b/src/components/alerts/alert.jsx
@@ -1,50 +1,39 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import bindAll from 'lodash.bindall';
 
 import Box from '../box/box.jsx';
 import Button from '../button/button.jsx';
 
 import styles from './alert.css';
 
-class AlertComponent extends React.Component {
-    constructor (props) {
-        super(props);
-        bindAll(this, [
-            'handleOnCloseAlert'
-        ]);
-    }
-    handleOnCloseAlert () {
-        this.props.onCloseAlert(this.props.index);
-    }
-    render () {
-        return (
-            <Box
-                className={styles.alert}
-            >
-                <div className={styles.alertMessage}>
-                    {this.props.iconURL ? (
-                        <img
-                            className={styles.alertIcon}
-                            src={this.props.iconURL}
-                        />
-                    ) : null}
-                    {this.props.message}
-                </div>
-                <Button
-                    className={styles.alertRemoveButton}
-                    onClick={this.handleOnCloseAlert}
-                >
-                    {'x'}
-                </Button>
-            </Box>
-        );
-    }
-}
+const AlertComponent = ({
+    iconURL,
+    message,
+    onCloseAlert
+}) => (
+    <Box
+        className={styles.alert}
+    >
+        <div className={styles.alertMessage}>
+            {iconURL ? (
+                <img
+                    className={styles.alertIcon}
+                    src={iconURL}
+                />
+            ) : null}
+            {message}
+        </div>
+        <Button
+            className={styles.alertRemoveButton}
+            onClick={onCloseAlert}
+        >
+            {'x'}
+        </Button>
+    </Box>
+);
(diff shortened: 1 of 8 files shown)


src/components/alerts/alert.jsx
```diff
@@ -0,0 +1,52 @@
...
+import styles from './alert.css';
+
+class AlertComponent extends React.Component {
+    constructor (props) {
+        super(props);
+        bindAll(this, [
+            'handleOnCloseAlert'
+        ]);
+    }
+    handleOnCloseAlert () {
```
evhan55: @rschamp I tried to put this handler in the container, but I found no lint-friendly way to handle passing up the index property to the parent container (no `bind` or arrow functions allowed in the `onClick`).  After lots of Googling, I ended up at this pattern, but let me know if there is a better way to handle this click function in the container somehow.  Thanks!

evhan55: Ah, I see! One level of abstraction I wasn't seeing, thank you!

src/reducers/alerts.js
```diff
@@ -1,36 +1,59 @@
...
     switch (action.type) {
-    case SHOW_ALERT:
+    case SHOW_ALERT: {
+        const newList = state.alertsList.slice();
+        const newAlert = {message: action.data.message};
+        const extensionId = action.data.extensionId;
+        if (extensionId) { // if it's an extension
+            const extension = extensionData.find(ext => ext.extensionId === extensionId);
+            if (extension && extension.name) {
+                // TODO: is this the right place to assemble this message?
```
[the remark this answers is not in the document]
evhan55: Rgr on the FormattedMessage, and yes, in a later PR to handle cleaner alert message APIs, alert types (error/success/etc), and localization all at once.

src/containers/alerts.jsx
```diff
@@ -1,23 +1,51 @@
...
-import AlertsComponent from '../components/alerts/alerts.jsx';
+import Box from '../components/box/box.jsx';
+import Alert from '../containers/alert.jsx';
+
+const Alerts = ({
+    alertsList,
+    className,
+    onCloseAlert
+}) => (
+    <Box
```
rschamp: Since this is a container, technically it should just pass props to a component. I.e., this could get refactored in the reverse way that you refactored the behavior out of the `AlertComponent`. Not a blocker for this PR since you don't do any styling here, but would be nice eventually, maybe in the next round.

src/reducers/alerts.js
```diff
@@ -1,36 +1,71 @@
...
+        type: CLOSE_ALERT,
+        index
+    };
 };
 
-const showAlert = function (message) {
+/**
+ * Function to show an alert with the given input data.
+ *
+ * @param {object} data - data with the following props: {message, extensionId=null}
```
rschamp: These props can be documented like:
```js
/**
 * @param {string} data.message Message for the alert
 * @param {string?} data.extensionId Extension ID for the alert
 */
```
Also not a blocker

src/containers/alert.jsx
```diff
@@ -0,0 +1,35 @@
...
+    }
+    handleOnCloseAlert () {
+        this.props.onCloseAlert(this.props.index);
+    }
+    render () {
+        return (
+            <AlertComponent
+                iconURL={this.props.iconURL}
+                message={this.props.message}
+                onCloseAlert={this.handleOnCloseAlert}
```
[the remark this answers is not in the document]
evhan55: Okay, thanks!

Issue #3235, opened 2018-09-26
bitmaps in vector form cause projects to run slowly when open in paint editor


BryceLTaylor:
### Expected Behavior

Projects should run quickly all of the time.

### Actual Behavior

With a project running, if you open the paint editor and select a costume that has been converted from bitmap to vector in a scratch 2.0 project, the project will slow down tremendously.  

### Steps to Reproduce

I'm still working on repro steps to make a project that has this effect.  There are two projects that I know of that have this behavior already.  [Space Giraffe](https://llk.github.io/scratch-gui/develop/#247578672) (here it is in [2.0](https://scratch.mit.edu/projects/247578672/#editor))and [Geometry Dash](https://llk.github.io/scratch-gui/develop/#105500895) (here it is in [2.0](https://scratch.mit.edu/projects/105500895/#editor))

Load one of the above projects
Click the green flag
Open the Paint Editor
Select a the "background" sprite in Geometry Dash or the "cover" sprite in Space Giraffe
Note: the project runs incredibly slowly

### Operating System and Browser

Chrome on Mac

fsih:
From Z:

The space giraffe gui issue relates to the paint wrapper being rerendered every game frame and part of updating it is re-decoding the svg from binary into utf8. Most of the time is that re-decoding. We can make a quick change to scratch-storage to use browser-implemented TextDecoder and TextEncoder. Some quick testing in Chrome shows the browser-implemented TextDecoder is 10x faster than the JS implemented TextDecoder.

Probably also want to change the gui, to relax re-rendering scratch-paint if it isn’t making changes when the vm is actively running.

https://github.com/LLK/scratch-storage/pull/44 (edited)

PR #3250, opened 2018-09-28
Fix expensive operation in mapStateToProps

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-gui/issues/3235

### Proposed Changes
Move getCostume() out of mapStateToProps. mapStateToProps is called any time the state changes, and getCostume can be very expensive because it has to decode from storage. In addition, I added a shouldComponentUpdate to ignore the props that don't affect rendering.

### Reason for Changes
Projects could run very slowly when the paint editor was open

### Test Coverage
Tested [Space Giraffe](https://llk.github.io/scratch-gui/develop/#247578672) and [Geometry Dash](https://llk.github.io/scratch-gui/develop/#105500895)

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/paint-editor-wrapper.jsx
+++ b/src/containers/paint-editor-wrapper.jsx
@@ -1,7 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import bindAll from 'lodash.bindall';
-import omit from 'lodash.omit';
 import VM from 'scratch-vm';
 import PaintEditor from 'scratch-paint';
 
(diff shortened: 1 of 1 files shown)


src/containers/paint-editor-wrapper.jsx
```diff
@@ -40,9 +46,12 @@ class PaintEditorWrapper extends React.Component {
     }
     render () {
         if (!this.props.imageId) return null;
+        const componentProps = omit(this.props, ['selectedCostumeIndex', 'vm']);
```
chrisgarrity: Instead of including lodash.omit,  most of gui uses object destructuring:
```
const {
    selectedCostumeIndex,
    vm,
    ...componentProps
} = this.props;
```
It gets a little messy if you're not using the prop because you need to handle the eslint unused variables rule. In this case you're using both variables, so it should look cleaner.

PR #3257, opened 2018-09-29
Use defineMessages

### LINKED ISSUE #3252: Make preview modal tooltips translatable by translators
Unfortunately the babel tool that automatically extracts localization messages only recognizes `defineMessages(...)` and `<FormattedMessage ... />`

Uses of `intl.formatMessage(...)` won't throw an error, but the string definition will not get extracted by the tool and therefore will not be available to be translated. For example:
https://github.com/LLK/scratch-gui/blob/2753c00a14e998fe30e2b0f2caa2568f64c8fc96/src/components/preview-modal/preview-modal.jsx#L55-L59

Instead we should move the message definition into a `defineMessages` call, and then call `intl.formatMessage` with the data from the messages object.

apple502j:
### Resolves
#3252 

### Proposed Changes
Change preview modal to use defineMessages

### Reason for Changes
~~To fix mistakes I did in past PR~~
To make preview modal tooltips translatable by translators

### DIFF
--- a/src/components/preview-modal/preview-modal.jsx
+++ b/src/components/preview-modal/preview-modal.jsx
@@ -17,6 +17,21 @@ const messages = defineMessages({
         defaultMessage: 'Welcome to the Scratch 3.0 Beta',
         description: 'Header for Preview Info Modal',
         id: 'gui.previewInfo.welcome'
+    },
+    notNowTooltip: {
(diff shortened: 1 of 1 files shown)


chrisgarrity:
@apple502j it looks good - I'll run all the scripts on Monday to check the execution. Thanks!

PR #3260, opened 2018-10-01
Deploy Smoke to Master

thisandagain:
### Features
- Add new alerts system when a hardware extension is disconnected
- Added support for variable / reporter monitors within extensions
- Added full support for "Right to Left" languages (e.g. Hebrew, Arabic)
- Added localization of library tags
- Added localization of "item # of value in list" block (HT @towerofnix)

### Bug Fixes
- Fixed various issues with importing projects from Scratch 2.0
    - Variable and list name conflicts
    - Improved error logging / reporting
- Fixed various issues with the paint editors
    - Improved hit testing in the reshape tool
    - Transparent stroke & size "0" stroke are now the same
- Fixed various issues with hardware extensions
    - Set default motor power for EV3 and WeDo 2 to 50%
    - Fix issues where EV3 motor position reporter could become inverted
    - Fix issues with WeDo 2 where the "set motor direction" block sometimes stops the motor
    - Fix issue where micro:bit "button pressed?" block should return `true` / `false`
    - Various minor adjustments to the "micro:bit" extension based on testing
- Fixed various issues in the costume editor
     - Resizing an object keeps it's ratio if you drag the corners
    - Text in the bitmap editor no longer disappears if you don’t click off of it and change costumes
    - Filled / outlined state and line thickness now stays in sync with the selected shape in bitmap editor if you change options and then undo / redo
    - Text outline thickness matches between stage and editor
    - Pressing backspace or delete in Mozilla Firefox doesn’t try to navigate when paint is open
- Fixed various issues with importing projects from Scratch 2.0
    - Corrupt SVG costumes can no longer cause a crash
    - String and/or numeric reporters plugged into boolean input slots no longer cause a crash
- Minor fixes and block language adjustments to the "micro:bit" extension
- Minor fixes to extensions that connect to hardware ("micro:bit", "LEGO WeDo 2.0", etc)


rschamp:
Try it out: [smoke](https://llk.github.io/scratch-gui/smoke)

Issue #1368, opened 2018-01-29
Add menu items to play sound blocks for "add sound..." and "record sound..."


ericrosenbaum:
Scratch 2.0 has a menu item "record..." in the play sound blocks:

<img width="167" alt="screen shot 2018-01-29 at 3 57 54 pm" src="https://user-images.githubusercontent.com/567844/35534034-3eb4ca1e-050d-11e8-8383-93111c814f07.png">

We should consider adding options like "add sound..." and "record sound..." to the start sound and play sound blocks, which directly open the sound library or sound recorder modal.

mrjacobbloom:
Should the costume menu parallel that as well?

kyleplo:
Should we include more options:

**Sound**
- record
- add sound
- upload
- surprise

**Costume/Backdrop**
- paint
- add costume/backdrop
- upload
- selfie
- surprise

PR #3309, opened 2018-10-04
Open sound recorder from sound block menu

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/1368
(we can add "add sound..." as a follow-up maybe)

Depends on https://github.com/LLK/scratch-blocks/pull/1731

### Proposed Changes

- When the contents of the sounds menu are generated, append a 'record...' option
- Add a callback triggered by this option to make it switch to the sounds tab and open the sound recorder. 

### Reason for Changes

We'd like to make it more obvious that you can record your own sound, and make the process accessible directly from the play sound blocks themselves (as in Scratch 2.0).

### Test Coverage

Added an integration test

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -21,8 +21,13 @@ import {STAGE_DISPLAY_SIZES} from '../lib/layout-constants';
 import {connect} from 'react-redux';
 import {updateToolbox} from '../reducers/toolbox';
 import {activateColorPicker} from '../reducers/color-picker';
-import {closeExtensionLibrary} from '../reducers/modals';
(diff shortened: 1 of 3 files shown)


paulkaplan:
(marking good to merge, but just to be clear, it does need the package json update)

PR #3320, opened 2018-10-10
Update monitor labels on the stage when language changes

chrisgarrity:
### Resolves

- Resolves https://github.com/LLK/scratch-vm/issues/1279

### Proposed Changes

A complete rewrite of OpcodeLabels. Instead of a function, it’s a class that maintains a current translator function, and updates the opcode labelFn for each opcode when a new translator function is set. The new function `getLabel` provides the original functionality, to return the object with `category` and `labelFn` based on the opcode.

Wrap MonitorList with the `InjectIntl` HOC so that it has access to the intl object. Then pass `intl.formatMessage` as the translator function to `OpcodeLabels`.

### Reason for Changes

Monitors on the stage don't update language correctly when switching language. It seemed worth figuring out how to update the language using the GUI (react-intl) framework instead of ScratchBlocks.ScratchMsgs. It uses the same strategy as storage (passing an translation function) for the default project.

### Test Coverage

Try it: https://chrisgarrity.github.io/scratch-gui/issue/vm1279-monitor-label/

Note: All of the strings are new translation ids in gui. So none of them will be translated until we go through a translation cycle. You can check that it's really using the new ids by looking for a message like `[React Intl] Missing message: "gui.opcodeLabels.costumenumber" for locale: "fr", using default message as fallback.` in the console.

Since most of the strings have already been translated for blocks, they should automatically be translated for gui once they're added. I'd like to add an integration test once we have some known translations.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/lib/opcode-labels.js
+++ b/src/lib/opcode-labels.js
@@ -1,126 +1,247 @@
-import ScratchBlocks from 'scratch-blocks';
+import {defineMessages} from 'react-intl';
 
-const opcodeMap = {
-    // Motion
+const messages = defineMessages({
     motion_direction: {
-        category: 'motion',
-        labelFn: () => ScratchBlocks.ScratchMsgs.translate('MOTION_DIRECTION', 'direction')
+        defaultMessage: 'direction',
+        description: 'Label for the direction monitor when shown on the stage',
+        id: 'gui.opcodeLabels.direction'
     },
     motion_xposition: {
-        category: 'motion',
-        labelFn: () => ScratchBlocks.ScratchMsgs.translate('MOTION_XPOSITION', 'x postion')
+        defaultMessage: 'x postion',
+        description: 'Label for the x position monitor when shown on the stage',
+        id: 'gui.opcodeLabels.xposition'
     },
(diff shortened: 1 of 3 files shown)


src/lib/opcode-labels.js
```diff
@@ -1,126 +1,248 @@
...
+                return this._translator(messages.sensing_current_dayofweek);
+            case 'hour':
+                return this._translator(messages.sensing_current_hour);
+            case 'minute':
+                return this._translator(messages.sensing_current_minute);
+            case 'second':
+                return this._translator(messages.sensing_current_second);
+
+
+            }
```
paulkaplan: Nit: Couple extra newlines above

chrisgarrity: And magically they disappear (force push)

PR #3350, opened 2018-10-11
Add project loading HOC tests

### LINKED ISSUE #3347: Make tests for project state-related higher order components
project-fetcher-hoc:
* Check that it calls onFetchedProjectData with projectData

vm-manager-hoc:
* Mock vm’s load project function, test that it’s called
* test that it then calls project-state reducer's onLoadedProject()
* If it’s called with no data, doesn’t render anything/renders null
* Test that vm is initialized (use a mock for vm)

rschamp:
### Resolves

- Resolves #3347 

### Proposed Changes

Adds tests for some of the new higher order components added in #3303 

### Reason for Changes

We should make sure the things we add work the way we expect

### Test Coverage

These are tests!

### Browser Coverage
N/A

### DIFF
--- a/src/lib/vm-manager-hoc.jsx
+++ b/src/lib/vm-manager-hoc.jsx
@@ -104,6 +104,7 @@ const vmManagerHOC = function (WrappedComponent) {
         onLoadedProject: loadingState => dispatch(onLoadedProject(loadingState))
     });
 
+    // Allow incoming props to override redux-provided props. Used to mock in tests.
     const mergeProps = (stateProps, dispatchProps, ownProps) => Object.assign(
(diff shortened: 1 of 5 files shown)


src/lib/vm-manager-hoc.jsx
```diff
@@ -104,9 +104,14 @@ const vmManagerHOC = function (WrappedComponent) {
         onLoadedProject: loadingState => dispatch(onLoadedProject(loadingState))
     });
 
+    const mergeProps = (stateProps, dispatchProps, ownProps) => Object.assign(
```
benjiwheeler: maybe we should comment on why we're using mergeprops?

PR #3354, opened 2018-10-11
disable onShare button

benjiwheeler:
by putting in an empty default onShare function, I had inadvertently made the button appear enabled in beta/standalone.

### DIFF
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -683,8 +683,7 @@ const mapDispatchToProps = dispatch => ({
     onRequestCloseLogin: () => dispatch(closeLoginMenu()),
     onClickNew: canSave => dispatch(requestNewProject(canSave)),
     onClickSave: () => dispatch(saveProject()),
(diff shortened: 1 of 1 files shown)


src/components/menu-bar/menu-bar.jsx
```diff
@@ -683,8 +683,7 @@ const mapDispatchToProps = dispatch => ({
     onRequestCloseLogin: () => dispatch(closeLoginMenu()),
     onClickNew: canSave => dispatch(requestNewProject(canSave)),
     onClickSave: () => dispatch(saveProject()),
-    onSeeCommunity: () => dispatch(setPlayer(true)),
-    onShare: () => {} // NOTE: implement this
+    onSeeCommunity: () => dispatch(setPlayer(true))
```
chrisgarrity: Need to also remove the onShare prop above in proptypes

benjiwheeler: We agreed that this is ok, we're not completely removing the prop, just leaving it as undefined by default

Issue #2743, opened 2018-07-30
Host tutorials using a service besides Youtube, or host them ourselves?


benjiwheeler:
Using Youtube videos in our in-app tutorials is tricky because many schools block Youtube.

Do we plan to host the videos ourselves or on some other service? 

We may need to nail down a good video player.

thisandagain:
We are considering forking the video player source code developed for one of my other projects and may either use that or work with another 3rd party provider that we are in the process of interviewing / negotiating an agreement with.

apple502j:
Why don't you just use `<video>` tag and host by yourself?

towerofnix:
Right, I'm curious what the reasoning is for using anything but the `<video>` tag. It's certainly long supported by every browser 3.0 targets, and is certain to feel more natural than any custom UI, since browser developers typically make the controls look relatively similar to how the rest of their OS does (see Safari for example).

Going to the trouble of negotiating with a third-party company to *play a video* just does not seem worth it. Presumably/hopefully there's more to the company @thisandagain mentions than purely being a video player provider.

benjiwheeler:
@thisandagain I put this in the November 2018 Milestone because it seems relevant to Hour of Code

MasterOfTheTiger:
If you need to use a third party service for some reason, maybe check out MediaGoblin.

PR #3356, opened 2018-10-12
Migrate Videos from YouTube

thisandagain:
## Resolves
Resolves GH-2743

## Changes
This PR migrates tutorial videos away from YouTube and towards a custom solution that we can put behind our CDN moving forward. All videos have been migrated to the new platform and in the case of the "Getting Started" tutorial an initial set of Closed Captions have been loaded.

/cc @SpeakkVisually @ntlrsk @kosiecki

### DIFF
--- a/src/components/cards/cards.jsx
+++ b/src/components/cards/cards.jsx
@@ -63,17 +63,13 @@ const VideoStep = ({video, dragging}) => (
             <div className={styles.videoCover} />
         ) : null}
         <iframe
-            allowfullscreen
-            allowtransparency="true"
-            frameborder="0"
+            allowFullScreen
+            allowTransparency="true"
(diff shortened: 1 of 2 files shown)


rschamp:
Try it out: [video](https://llk.github.io/scratch-gui/video)

benjiwheeler:
This PR may resolve https://github.com/LLK/scratch-gui/issues/2743

Question: @towerofnix on that issue page brought up the possibility of using the `<video>` tag rather than an iframe, so that native players are used. Not sure if that bears on this.

thisandagain:
The native `<video>` tag player does not support all of the features we need.

thisandagain:
@chrisgarrity R?

PR #3359, opened 2018-10-13
Costume selector RTL

### LINKED ISSUE #3128: Costume dimensions incorrectly formatted in RTL
### Expected Behavior
Figure out the expected behavior for dimensions in RTL: is it still 'width x height', or would 'height x width' be the expected layout.

### Actual Behavior

image20

apple502j:
### Resolves
Resolves #3128 

### Proposed Changes
Adds `dir=ltr` to SpriteSelectorItem

### Reason for Changes
In RTL texts, numbers should be still LTR.

### DIFF
--- a/src/components/sprite-selector-item/sprite-selector-item.jsx
+++ b/src/components/sprite-selector-item/sprite-selector-item.jsx
@@ -45,7 +45,12 @@ const SpriteSelectorItem = props => (
         <div className={styles.spriteInfo}>
             <div className={styles.spriteName}>{props.name}</div>
             {props.details ? (
-                <div className={styles.spriteDetails}>{props.details}</div>
+                <div
+                    className={styles.spriteDetails}
+                    dir="ltr"
+                >
+                    {props.details}
+                </div>
             ) : null}
         </div>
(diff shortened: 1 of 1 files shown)


src/components/sprite-selector-item/sprite-selector-item.jsx
```diff
@@ -45,7 +45,12 @@ const SpriteSelectorItem = props => (
         <div className={styles.spriteInfo}>
             <div className={styles.spriteName}>{props.name}</div>
             {props.details ? (
-                <div className={styles.spriteDetails}>{props.details}</div>
+                <div
+                    className={styles.spriteDetails}
+                    dir="ltr"
+                >
+                    {props.details}
```
chrisgarrity: I'm not sure where `props.details` is getting set, but the reason that it's showing up strangely is that we're using the 'x' character, so the details are latin (ltr) characters, in a rtl context. If we used the unicode multiplication symbol instead (U+00D7) &times;, it would probably display correctly as I don't think it is a strongly directional character.

Anyhow, I think it would be better to solve this through using the multiplication symbol instead of hard-coding the `dir="ltr"` here. The original issue is probably a bit misleading.

apple502j:
I will close it and make another PR.

Issue #3322, opened 2018-10-10
use more explicit boolean props related to capability


benjiwheeler:
In comments on https://github.com/LLK/scratch-gui/pull/3303, re: `menu-bar.jsx`, @rschamp commented:

> Right now we're passing facts like sessionExists into the component and then using them to determine a capability within the component. Since they're both derived in redux already, it'd be clearer if the component received capabilities as props like canCreateProject, and we derived the capability from the facts in mapStateToProps.

PR #3364, opened 2018-10-15
use can__ boolean properties to explicitly set user capabilities

benjiwheeler:
### Resolves:

Resolves https://github.com/LLK/scratch-gui/issues/3322 (and related to https://github.com/LLK/scratch-www/issues/2168 )

### Changes:

sets and uses several variables:
```
canCreateNew
canRemix
canReport
canSave
canSaveAsCopy
canShare
```

Question: are there any booleans we should add? Do all of these make sense?

### Reason for Changes

We want embedding environments to be able to clearly communicate to GUI which capabilities the user should have, rather than have GUI contain knowledge about what various session statuses mean it should enable

### Test Coverage

Added several tests, in new test file `integration/menu-bar.test.js`

### Browser Coverage

Mac Chrome

### DIFF
--- a/src/playground/render-gui.jsx
+++ b/src/playground/render-gui.jsx
@@ -39,7 +39,6 @@ export default appTarget => {
 
     ReactDOM.render(
         <WrappedGui
-            canCreateNew
             backpackOptions={backpackOptions}
         />,
         appTarget);
--- a/src/playground/render-gui.jsx
+++ b/src/playground/render-gui.jsx
@@ -37,5 +37,14 @@ export default appTarget => {
         window.onbeforeunload = () => true;
     }
 
-    ReactDOM.render(<WrappedGui backpackOptions={backpackOptions} />, appTarget);
+    ReactDOM.render(
+        <WrappedGui
+            canCreateNew
+            backpackOptions={backpackOptions}
+            canRemix={false}
+            canSave={false}
+            canSaveAsCopy={false}
+            canShare={false}
+        />,
+        appTarget);
 };
--- a/src/playground/render-gui.jsx
+++ b/src/playground/render-gui.jsx
@@ -41,10 +41,6 @@ export default appTarget => {
         <WrappedGui
             canCreateNew
             backpackOptions={backpackOptions}
-            canRemix={false}
-            canSave={false}
-            canSaveAsCopy={false}
-            canShare={false}
         />,
         appTarget);
 };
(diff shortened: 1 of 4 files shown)


test/integration/menu-bar.test.js
```diff
@@ -0,0 +1,60 @@
...
+    });
+
+    test('File->New should be enabled', async () => {
+        await loadUri(uri);
+        await clickXpath('//button[@title="Try It"]');
+        await clickXpath(
+            '//div[contains(@class, "menu-bar_menu-bar-item") and ' +
+            'contains(@class, "menu-bar_hoverable")][span[text()="File"]]'
+        );
+        await findByXpath('//*[li[span[text()="New"]] and not(@data-tip="tooltip")]');
```
benjiwheeler: So the approach i used was to make strict xpaths that should all be found, where if the capability flag was handled wrong, the tooltip status wouldn't match and the findByXpath will timeout. 

This works, but it seems not ideal. Do you have any suggestions for a better way? I couldn't figure out how to make sense of the promise object that findByXpath returns, but maybe I could somehow use then/catch?

rschamp: Check out what `findByXpath` does — it doesn't actually return anything you can do something with, it just waits. It should probably be renamed `waitUntilXpathFound`.  If you want to do something with the elements, you should use `driver.findElements`.  This worked when I tried it:

```js
        const el = await driver.findElements(By.xpath('//*[li[span[text()="New"]] and not(@data-tip="tooltip")]'));
        expect(el.length).toEqual(1);
```

benjiwheeler:
Updated, take a look!

src/playground/render-gui.jsx
```diff
@@ -37,5 +37,10 @@ export default appTarget => {
         window.onbeforeunload = () => true;
     }
 
-    ReactDOM.render(<WrappedGui backpackOptions={backpackOptions} />, appTarget);
+    ReactDOM.render(
+        <WrappedGui
+            canCreateNew
```
rschamp: Don't need this anymore.

benjiwheeler: It sort of depends on your definition of "need" :)

Isn't it better to explicitly set this, because we want it, rather than assume the default is what we want it to be?

rschamp: Thinking about it more, I think we definitely want the playground to use the defaults, since when you're developing, you don't expect any props to be set, and you expect all the defaults to be used. It would be a frustrating experience to try changing that default and not seeing anything happen, only to find that there is this one prop set at the playground level.

benjiwheeler: Removed, to make usage clearer

PR #3366, opened 2018-10-15
Preload fonts

fsih:
### Resolves

- Resolves https://github.com/LLK/scratch-paint/issues/648

### Proposed Changes
Preload fonts

### Reason for Changes
Prevent imported text from being cut off

### Test Coverage
Tested many browser/OS combos and IE by loading projects 249881325, 228228816, and loading the default project and then uploading those 2 projects.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


fsih:
gui.jsx was refactored in the meantime and I would appreciate help figuring out how to merge with the refactor

src/containers/gui.jsx
```diff
@@ -132,6 +133,7 @@ const ConnectedGUI = connect(
 // ability to compose reducers.
 const WrappedGui = compose(
     ErrorBoundaryHOC('Top Level App'),
+    FontLoaderHOC,
```
rschamp: I think `FontLoaderHOC` should wrap the component `VMManagerHOC`, similar to the way redux's `connect` does:
 - The `fontsLoaded` prop that `FontLoaderHOC` provides is consumed by `VMManager `
 - `VMManager ` can't function without `fontsLoaded`
 - The font loading behavior is an aspect of project loading, more specific to `VMManager` than `GUI`

benjiwheeler: you mean, take it out of this compose list, and put it directly into the VMManagerHOC file?

fsih: I think he means move it down 2 lines

test/unit/util/vm-manager-hoc.test.jsx
```diff
@@ -75,27 +76,38 @@ describe('VMManagerHOC', () => {
...
+            fontsLoaded: true,
+            loadingState: LoadingState.LOADING_VM_WITH_ID,
+            projectData: '100'
+        });
+        expect(vm.loadProject).toHaveBeenLastCalledWith('100');
+        // nextTick needed since vm.loadProject is async, and we have to wait for it :/
+        process.nextTick(() => expect(mockedOnLoadedProject).toHaveBeenLastCalledWith(LoadingState.LOADING_VM_WITH_ID));
     });
-    test('if there is no projectData, nothing is rendered', () => {
+    test('if there is projectData, the child is rendered', () => {
```
benjiwheeler: I'm not sure this test makes sense anymore. Seems like now, if there isn't projectData, the child is still rendered.

fsih: Sure, I can take this test out too

It merged my tests in a weird way. Basically I took out the nothing is rendered test and added a test that load happens when fontsLoaded becomes true. Should there be another test where fontsLoaded doesn't become true, and then vm.loadProject is never called?

src/containers/gui.jsx
benjiwheeler: I originally put it at the top because I thought it might need to be the first request completed, before we fetched; we even tried rendering null in FontLoaderHOC until fonts were loaded, but that doesn't work because (I think??) the fonts don't present themselves until you load the components that use them.

I could put it wherever, just not sure which place Ray means... "similar to the way redux's connect does" makes me think he means, like

`export FontLoaderHOC(VMManagerHOC)` 

or something...

test/unit/util/vm-manager-hoc.test.jsx
benjiwheeler: The problem is there's no real way (that I know) to test "is never called"

src/containers/gui.jsx
fsih: >> because (I think??) the fonts don't present themselves until you load the components that use them.

The fonts should be present in CSS on the page before React lifecycles start happening. I think the issue was that without the stage attached, the renderer isn't given by the GUI to the VM, and something was trying to access the renderer too soon.

test/unit/util/vm-manager-hoc.test.jsx
fsih: I took out the child rendered test and added a doesn't-load-without-fonts test. I think .toHaveBeenCalledTimes(0) works

src/containers/gui.jsx
rschamp: Yes I meant in `VMManagerHOC`, something like:
```jsx
return connect(
         mapStateToProps,
         mapDispatchToProps,
         mergeProps
     )(FontLoaderHOC(VMManager));
```

Although now I changed my mind, I think you should leave it since wrapping it up like that defeats the benefit of HOCs being isolated from each other so they can be tested by themselves.

Issue #3339, opened 2018-10-10
When project is uploaded, call onUpdateProjectTitle


benjiwheeler:
Currently, when an sb* file is uploaded, we are setting the project title in redux, but not calling the `onUpdateProjectTitle` function. We should, so that the project that gui is embedded in can respond.

Issue #3140, opened 2018-09-14
Handle/show title of a 2.0 project when one is loaded in 3.0


benjiwheeler:
### Expected Behavior

Load a 2.0 project in 3.0. I expect to see its title in the project title area.

### Actual Behavior

The project title is unchanged from the previous value.

benjiwheeler:
Fixed in #3379

PR #3379, opened 2018-10-15
sb-file-uploader gets title of any sb.* file; calls onUpdateProjectTitle

benjiwheeler:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/3339 and https://github.com/LLK/scratch-gui/issues/3140

### Proposed Changes

* Handles/shows title of a 2.0 project (and will work for 1.0 projects, and projects missing extension, if it needs to) when one is loaded in 3.0.

* calls the onUpdateProjectTitle passed to it from `MenuBar` with the new title, if it exists. This allows code that encloses or embeds GUI to be notified of the title change when a file is uploaded.

### Reason for Changes

* sb2 file uploads were having titles ignored, for no good reason!
* uploading a Scratch file when editor is under preview/project page should update the title of the project!

### Test Coverage

None. Unclear how to test this well?

### Browser Coverage

tested in Mac Chrome

### DIFF
--- a/src/containers/sb-file-uploader.jsx
+++ b/src/containers/sb-file-uploader.jsx
@@ -41,6 +41,7 @@ class SBFileUploader extends React.Component {
     constructor (props) {
         super(props);
         bindAll(this, [
+            'getProjectTitleFromFilename',
             'renderFileInput',
             'setFileInput',
             'handleChange',
--- a/src/containers/sb-file-uploader.jsx
+++ b/src/containers/sb-file-uploader.jsx
@@ -6,7 +6,6 @@ import {defineMessages, injectIntl, intlShape} from 'react-intl';
 
 import analytics from '../lib/analytics';
 import log from '../lib/log';
-import {setProjectTitle} from '../reducers/project-title';
 import {LoadingStates, onLoadedProject, onProjectUploadStarted} from '../reducers/project-state';
 
 import {
@@ -79,13 +78,11 @@ class SBFileUploader extends React.Component {
             reader.readAsArrayBuffer(thisFileInput.files[0]);
             // extract the title from the file and set it as current project title
             if (thisFileInput.files[0].name) {
-                const matches = thisFileInput.files[0].name.match(/^(.*?)(\.sb[23]?)?$/);
+                // only parse title from files like "filename.sb2" or "filename.sb3"
+                const matches = thisFileInput.files[0].name.match(/^(.*)\.sb[23]$/);
                 if (matches) {
                     const truncatedProjectTitle = matches[1].substring(0, 100);
-                    this.props.onSetReduxProjectTitle(truncatedProjectTitle);
-                    if (this.props.onUpdateProjectTitle) {
-                        this.props.onUpdateProjectTitle(truncatedProjectTitle);
-                    }
+                    this.props.onUpdateProjectTitle(truncatedProjectTitle);
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


benjiwheeler:
@rschamp any suggestions for how to write tests for this? Seems like it's tricky to do it without opening the OS filepicker...

src/containers/sb-file-uploader.jsx
```diff
@@ -79,10 +79,13 @@ class SBFileUploader extends React.Component {
             reader.readAsArrayBuffer(thisFileInput.files[0]);
             // extract the title from the file and set it as current project title
             if (thisFileInput.files[0].name) {
-                const matches = thisFileInput.files[0].name.match(/^(.*)\.sb3$/);
+                const matches = thisFileInput.files[0].name.match(/^(.*?)(\.sb[23]?)?$/);
```
rschamp: To me this reads "Any sequence of characters _optionally_ followed by `.sb`, `.sb2` or `.sb3`".  The file picker we're creating only allows you to pick `.sb2` or `.sb3` files, so I think we could require one of these extensions: `/^(.*)\.sb[23]$/`. Curious why the change to a more expansive set of filenames?

src/containers/sb-file-uploader.jsx
```diff
@@ -79,10 +79,13 @@ class SBFileUploader extends React.Component {
...
                 if (matches) {
                     const truncatedProjectTitle = matches[1].substring(0, 100);
-                    this.props.onSetProjectTitle(truncatedProjectTitle);
+                    this.props.onSetReduxProjectTitle(truncatedProjectTitle);
```
rschamp: When do we not have `props.onUpdateProjectTitle`? This was set up this way before but now It's apparent that this is a little hacky — we're setting it in redux and then asking something different to also update the title to stay in sync. The best way to stay in sync is to always update at the top and let the state go down from there.

src/containers/sb-file-uploader.jsx
benjiwheeler: Here is my reasoning:

If, after some future changes to the code outside of this line, this line runs, and gets a filename like

`Why I sing.sb`
or
`Why I sing`

should it just decide to completely discard "Why I sing" and not change the project title to that? That seems like it would be a mistake.

rschamp: We should update this when we change the code then — what if the code never changes like that? Then this is set up for a situation that never occurs, and confuses the intent. As it is now, it looks like there's possibly a bug in the file input because it's only allowing .sb2 or .sb3 files, but this part of the code can allow .sb files as well. So which one is right? Without context would need to investigate. If we're consistent then there's no ambiguity.

src/containers/sb-file-uploader.jsx
benjiwheeler: It's a little hacky. Maybe gui.jsx could listen for changes to the redux title and call onUpdateProjectTitle, but it's already listening for changes to the incoming projectTitle prop and using that to set the redux project title... would it work to do both in the same place?

rschamp: We shouldn't need to listen for changes to the redux title, `onUpdateProjectTitle` should result in a change to the incoming `title` prop. That would then should cause GUI to update the redux state. Then everything that uses the redux title should read it from there. If `onUpdateProjectTitle` doesn't result in a change to the incoming `title` prop then we don't want the redux state to be different than that prop.

We can't assume what will happen when we call `onUpdateProjectTitle` — maybe there's a sanitizer before it updates the title, or maybe the title shouldn't change in a certain context.

src/containers/sb-file-uploader.jsx
benjiwheeler: I think we have a philosophical difference here, but it's a place that doesn't matter much, and I'm happy to do it your way.

benjiwheeler:
Take a look. I didn't mock up the filepicker input, just made unit tests of the filename extraction. It's a start!

PR #3382, opened 2018-10-15
Add video translation

### LINKED ISSUE #3230: Add support for translated versions of tutorial videos
Build infrastructure for showing different tutorial videos based on locale. 
Define process for managing videos.

chrisgarrity:
### Resolves
- Resolves #3230 

### Proposed Changes
* Look up video id based on locale.
* add japanese video ids
* add test

### Reason for Changes

Show different videos for the tutorials if they have been created for the other language.

### Test Coverage

Added unit test for the new function

Manual testing: only English and Japanese tutorial videos are currently available.

You can try it on my branch: https://chrisgarrity.github.io/scratch-gui/feature/3230-translate-videos/

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/lib/libraries/decks/translate-video.js
+++ b/src/lib/libraries/decks/translate-video.js
@@ -1,0 +1,97 @@
+/**
+ * @fileoverview
+ * Utility function to look up whether a translated video exists for a locale.
+ */
+
+// TODO: ja-Hira are copies of ja, replace with real ja-Hira when available.
+const videos = {
+    'intro-move-sayhello': {
+        'en': 'rpjvs3v9gj',
+        'ja': 'v2c2f3y2sc',
+        'ja-Hira': 'v2c2f3y2sc'
+    },
+    'animate-a-name': {
+        'en': 'pyur30ho05',
+        'ja': 'xrmgno02th',
+        'ja-Hira': 'xrmgno02th'
+    },
+    'Make-Music': {
+        'en': 'ir0j8ljsgm',
+        'ja': 'iybhyobckr',
+        'ja-Hira': 'iybhyobckr'
+    },
+    'Make-A-Game': {
+        'en': '5rp47ys13g',
+        'ja': 'h4q2ea3ojw',
+        'ja-Hira': 'h4q2ea3ojw'
+    },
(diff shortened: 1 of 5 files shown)


src/lib/libraries/decks/translate-video.js
```diff
@@ -0,0 +1,97 @@
+/**
+ * @fileoverview
+ * Utility function to look up whether a translated video exists for a locale.
+ */
+
+// TODO: ja-Hira are copies of ja, replace with real ja-Hira when available.
```
benjiwheeler: weren't you telling me there were some relationships, like brasilian portuguese -> portuguese, that had explicit failover/backup relationships? Was that in scratch-l10n?

Could it make sense to use something like that here?
```
const failovers = {
      'ja-Hira': 'ja'
}
const findWithFailover(object, locale, failovers) {
    if (object.hasOwnProperty(locale)) return object[locale];
    const failover = failovers[locale];
    if (failover) return findWithFailover(object, failover, failovers);
    if (object.hasOwnProperty('en')) return object['en'];
    return null;
}
```
then...
```
const translateVideo = (videoId, locale) => {
    // if the videoId is not recognized in the videos object, assume it's already a video id on wistia
    return findWithFailover(videos[videoId], locale, failovers);
};
```

chrisgarrity: We don't have any fail-overs. I was mentioning it might be nice in the string translations to be able to have the fallback language be something other than English, but it's not supported by react-intl. And it could be problematic - failing over between the two Chinese systems would probably not be acceptable.
Using the ja video for ja-Hira is only temporary. I checked with Yumiko, and we just need to redo the video - her voice over can be used for both. If people start asking for some other language video instead of English, then it might be worth looking into adding something like this. Right now I think it over complicates the videos for a situation that is the exception not the rule.

PR #3409, opened 2018-10-17
Get costume url bugfix

kchadha:
### Proposed Changes

Fix minor issues with watermark:

- Font wasn't actually getting injected into costumes because cachedUrl was getting replaced. This is no longer broken:

image

- Fix issue where watermark could be dragged.

### DIFF
--- a/src/lib/get-costume-url.js
+++ b/src/lib/get-costume-url.js
@@ -26,10 +26,12 @@ const getCostumeUrl = (function () {
                 svgRenderer.loadString(svgString);
                 const svgText = svgRenderer.toString(true /* shouldInjectFonts */);
                 cachedUrl = `data:image/svg+xml;utf8,${encodeURIComponent(svgText)}`;
+            } else {
+                cachedUrl = vm.runtime.storage.get(assetId).encodeDataURI();
             }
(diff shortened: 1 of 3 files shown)


src/lib/get-costume-url.js
```diff
@@ -26,10 +26,12 @@ const getCostumeUrl = (function () {
                 svgRenderer.loadString(svgString);
                 const svgText = svgRenderer.toString(true /* shouldInjectFonts */);
                 cachedUrl = `data:image/svg+xml;utf8,${encodeURIComponent(svgText)}`;
+            } else {
+                cachedUrl = vm.runtime.storage.get(assetId).encodeDataURI();
             }
+        } else {
             cachedUrl = vm.runtime.storage.get(assetId).encodeDataURI();
         }
-        cachedUrl = vm.runtime.storage.get(assetId).encodeDataURI();
```
fsih: Maybe
if (!cachedUrl) cachedUrl = ...
would be easier to reason about than the two elses, but I don't feel strongly about it

kchadha: I think I'm going to leave it as is.

Issue #3430, opened 2018-10-19
extract projectId and projectTitle "sync"ing behavior from gui.jsx


benjiwheeler:
We should extract the projectId and projectTitle sync behaviors out of gui (e.g., in componentDidUpdate) and put them in a wrapper. We should also check for any new incoming projectId, not only on the first construct of project-fetcher-hoc.

PR #3391, opened 2018-10-17
handle many modes of new project creation

benjiwheeler:
Needs https://github.com/LLK/scratch-www/pull/2197 to work

Handles several states in which the user should be able to create a new project:
1. standalone player/beta: file->new just replaces current project data with default data
2. in www, logged in, viewing your project: file->new saves your project, replaces current project data with default data, then saves that data and gets a new id, then informs containing code of the change
3. in www, logged in, viewing someone else’s project:  file->new replaces current project data with default data, then saves that data and gets a new id, then informs containing code of the change
4. in www, not logged in, viewing whatever project:  file->new replaces current project data with default data, then informs containing code of the change
5. in www, logged in, clicking “create” from front page; or, going straight to projects/editor: loads default data, then saves that data and gets a new id, then informs containing code of the change
6. in www, not logged in, clicking “create” from front page: loads default data
7. After #4, when you log in or register while still in the editor, saves current data and gets a new id, then informs containing code of the change

PR #3431, opened 2018-10-19
Refactor IntlProvider in GUI to support integration with www

chrisgarrity:
### Resolves
Required for https://github.com/LLK/scratch-www/issues/2172

### Proposed Changes

This is taking advantage of the fact that react-intl can support nested IntlProviders. GUI will use it’s own ConnectedIntlProvider, that maintains the translations in redux, while www will continue to have it’s own IntlProvider that is initialized when the page loads.

Moving the IntlProvider into GUI means that nothing in the playground can use injectIntl. As a result `TitledHOC` could not use `injectIntl`, and the localization of the default project name needed to be moved into the component.

The Localization HOC is added, it takes an optional onSetLanguage handler, so www can pass a function that will set the cookie that www needs.

### Reason for Changes

Allow localization of GUI when embedded in www

### Test Coverage

Current tests run: these changes should not have any impact on stand-alone GUI.

Manual Testing: 
- [ ] Can switch languages
- [ ] Can load in a different language (e.g. use `?locale=fr` on the URL)
- [ ] Can change the project title
- [ ] The default project title is translated into other languages 

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/menu-bar/project-title-input.jsx
+++ b/src/components/menu-bar/project-title-input.jsx
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import bindAll from 'lodash.bindall';
 import React from 'react';
 import {defineMessages, intlShape, injectIntl} from 'react-intl';
+import sharedMessages from '../../lib/shared-messages';
 
 import BufferedInputHOC from '../forms/buffered-input-hoc.jsx';
 import Input from '../forms/input.jsx';
@@ -16,11 +17,6 @@ const messages = defineMessages({
         id: 'gui.gui.projectTitlePlaceholder',
         description: 'Placeholder for project title when blank',
         defaultMessage: 'Project title here'
-    },
-    defaultProjectTitle: {
-        id: 'gui.gui.defaultProjectTitle',
-        description: 'Default title for project',
-        defaultMessage: 'Scratch Project'
     }
 });
 
@@ -48,7 +44,7 @@ class ProjectTitleInput extends React.Component {
                 type="text"
                 value={this.props.projectTitle ?
                     this.props.projectTitle :
-                    this.props.intl.formatMessage(messages.defaultProjectTitle)
+                    this.props.intl.formatMessage(sharedMessages.defaultProjectTitle)
                 }
                 onSubmit={this.handleUpdateProjectTitle}
             />
--- a/src/components/menu-bar/project-title-input.jsx
+++ b/src/components/menu-bar/project-title-input.jsx
@@ -16,6 +16,11 @@ const messages = defineMessages({
         id: 'gui.gui.projectTitlePlaceholder',
         description: 'Placeholder for project title when blank',
         defaultMessage: 'Project title here'
+    },
+    defaultProjectTitle: {
+        id: 'gui.gui.defaultProjectTitle',
+        description: 'Default title for project',
+        defaultMessage: 'Scratch Project'
     }
 });
 
@@ -41,7 +46,10 @@ class ProjectTitleInput extends React.Component {
                 placeholder={this.props.intl.formatMessage(messages.projectTitlePlaceholder)}
                 tabIndex="0"
                 type="text"
-                value={this.props.projectTitle}
+                value={this.props.projectTitle ?
+                    this.props.projectTitle :
+                    this.props.intl.formatMessage(messages.defaultProjectTitle)
+                }
                 onSubmit={this.handleUpdateProjectTitle}
             />
         );
--- a/src/components/menu-bar/project-title-input.jsx
+++ b/src/components/menu-bar/project-title-input.jsx
@@ -4,7 +4,6 @@ import PropTypes from 'prop-types';
 import bindAll from 'lodash.bindall';
 import React from 'react';
 import {defineMessages, intlShape, injectIntl} from 'react-intl';
-import sharedMessages from '../../lib/shared-messages';
 
 import BufferedInputHOC from '../forms/buffered-input-hoc.jsx';
 import Input from '../forms/input.jsx';
@@ -42,10 +41,7 @@ class ProjectTitleInput extends React.Component {
                 placeholder={this.props.intl.formatMessage(messages.projectTitlePlaceholder)}
                 tabIndex="0"
                 type="text"
-                value={this.props.projectTitle ?
-                    this.props.projectTitle :
-                    this.props.intl.formatMessage(sharedMessages.defaultProjectTitle)
-                }
+                value={this.props.projectTitle}
                 onSubmit={this.handleUpdateProjectTitle}
             />
         );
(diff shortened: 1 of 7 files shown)


src/lib/localization-hoc.jsx
```diff
@@ -0,0 +1,55 @@
...
+            );
+        }
+    }
+    LocalizationWrapper.propTypes = {
+        locale: PropTypes.string,
+        onSetLanguage: PropTypes.func
+    };
+
+    LocalizationWrapper.defaultProps = {
+        onSetLanguage: () => {}
```
rschamp: If someone makes it possible to call this default, they probably don't mean to (they've implemented language switching without persisting the locale anywhere). Maybe we should have the default log a warning to the console?

src/reducers/project-title.js
```diff
@@ -31,6 +21,5 @@ const setProjectTitle = title => ({
 export {
     reducer as default,
     initialState as projectTitleInitialState,
-    defaultProjectTitleMessages,
```
rschamp: This is going to conflict with #3391 — would you coordinate with @benjiwheeler on how this should work?

src/lib/localization-hoc.jsx
chrisgarrity: No, locale switching is still in redux. If Gui is embedded in something that doesn't need it's own handling of locale (e.g. playground), then they don't need to pass in a function.

rschamp: Here's how I understand the way we intended it to work:

We want GUI to maintain its internal notion of `locale` in Redux, but we expect `locale` to come in as a prop from the outside. When we switch locale in GUI, we should call `props.onSetLanguage`. The parent should then set the locale and pass a new `locale` prop to the GUI. At that point something watching the incoming `locale` prop should notice the change and update GUI's locale redux state (maybe that would be this wrapper?).

If all of that were true, then we wouldn't want GUI to set the Redux state except if the incoming `locale` prop were to change, so that's why I'd think we'd want to warn in case of an undefined `onSetLocale`.

src/reducers/project-title.js
chrisgarrity: @rschamp we agreed that @benjiwheeler will land #3391, and then I'll fix up conflicts. I'll move the `defaultProjectTitle` message into `sharedMessages` in gui. Then it can be used in multiple places.

src/lib/shared-messages.js
```diff
@@ -20,5 +20,10 @@ export default defineMessages({
         defaultMessage: 'pop',
         description: 'Name of the pop sound, the default sound added to a sprite',
         id: 'gui.sharedMessages.pop'
+    },
+    defaultProjectTitle: {
+        id: 'gui.gui.defaultProjectTitle',
```
chrisgarrity: Note: I left the id as `gui.gui.defaultProjectTitle` because it's already been translated. If we change the id to `gui.sharedMessages.defaultProjectTitle`, it won't be translated until we update translations.,

src/components/menu-bar/project-title-input.jsx
```diff
@@ -41,7 +42,10 @@ class ProjectTitleInput extends React.Component {
                 placeholder={this.props.intl.formatMessage(messages.projectTitlePlaceholder)}
                 tabIndex="0"
                 type="text"
-                value={this.props.projectTitle}
+                value={this.props.projectTitle ?
```
benjiwheeler: I think we should keep this line as it was before. The placeholder has different text and formatting, and I'm not sure handleUpdateProjectTitle will be called below. Plus, we can already use sharedMessages.defaultProjectTitle where the title is set for a default project.

src/lib/localization-hoc.jsx
```diff
@@ -0,0 +1,55 @@
...
+/*
+ * Higher Order Component to provide localiztion state. Creates a nested IntlProvider
+ * to handle Gui intl context. The component accepts an onSetLanguage callback that is
+ * called when the locale chagnes.
+ * @param {React.Component} WrappedComponent - component to provide state for
+ * @returns {React.Component} component with intl state provided from redux
+ */
+const LocalizationHOC = function (WrappedComponent) {
+    class LocalizationWrapper extends React.Component {
+        componentDidUpdate (prevProps) {
```
benjiwheeler: I'm still a little shaky on the react component lifecycle, but am i right in thinking that componentDidUpdate won't get called on the initial construct/mount/render? If so, onSetLanguage won't get called here when the page initially loads, right? Is that ok?

src/lib/titled-hoc.jsx
```diff
@@ -15,34 +13,24 @@ const TitledHOC = function (WrappedComponent) {
                 'handleUpdateProjectTitle'
             ]);
             this.state = {
-                projectTitle: this.props.intl.formatMessage(defaultProjectTitleMessages.defaultProjectTitle)
+                projectTitle: ''
```
benjiwheeler: I think this should be 'Scratch Project'. (Since I was suggesting we don't set the project-title-input to defaultProjectTitleMessages.defaultProjectTitle

src/reducers/project-title.js
benjiwheeler: Thanks, this is cleaner!

src/lib/titled-hoc.jsx
chrisgarrity: The issue with hard coding 'Scratch Project' is that it will never get translated when the GUI is stand-alone, unless `setReduxTitle` in gui also checks for `newTitle === 'Scratch Project`. If that seems acceptable, then this change makes sense.

chrisgarrity: The better solution is to default projectTitle to null, then the title handling in GUI already handles it properly without having to hard code `'Scratch Project'`.

src/lib/shared-messages.js
benjiwheeler: 👍

src/lib/titled-hoc.jsx
benjiwheeler: Chris came up with a better way, making this `null` and relying on `intl`-injected GUI to detect that and set it.

src/containers/gui.jsx
```diff
@@ -25,14 +25,22 @@ import {
...
+import LocalizationHOC from '../lib/localization-hoc.jsx';
 import ProjectFetcherHOC from '../lib/project-fetcher-hoc.jsx';
 import ProjectSaverHOC from '../lib/project-saver-hoc.jsx';
 import vmListenerHOC from '../lib/vm-listener-hoc.jsx';
 import vmManagerHOC from '../lib/vm-manager-hoc.jsx';
-import {defaultProjectTitleMessages} from '../reducers/project-title';
 
 import GUIComponent from '../components/gui/gui.jsx';
 
+const messages = defineMessages({
```
benjiwheeler: Just a note, Chris and I agreed that when I/we refactor gui.jsx to extract projectId and projectTitle handling (see https://github.com/LLK/scratch-gui/issues/3430), this message should probably come with that logic

src/lib/localization-hoc.jsx
benjiwheeler: I'll understand React better by looking at it running this code after it lands!

Issue #3435, opened 2018-10-19
make gui use single system of displaying errors, connect it to project state


benjiwheeler:
### Expected Behavior

there should be one set of state variables that control whether the project is in an error state. They should be able to be set from anywhere via redux.

### Actual Behavior

They are arbitrarily contained in vm-manager-hoc.

PR #3436, opened 2018-10-19
use project state error for project state related errors

benjiwheeler:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/3435

### Proposed Changes

Uses project-state error state and string instead of vm-manager's error state; adds handling to project-fetcher-hoc

### Reason for Changes

single central project error display system

### Test Coverage

Already put error state test into project-state-reducer.test.js

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/gui.jsx
+++ b/src/containers/gui.jsx
@@ -66,7 +66,7 @@ class GUI extends React.Component {
     render () {
         if (this.props.isError) {
             throw new Error(
-                `Failed to load project from server [id=${window.location.hash}]: ${this.props.error}`);
+                `Error in Scratch GUI [location=${window.location}]: ${this.props.error}`);
         }
         const {
             /* eslint-disable no-unused-vars */
@@ -101,7 +101,7 @@ class GUI extends React.Component {
 GUI.propTypes = {
     assetHost: PropTypes.string,
     children: PropTypes.node,
-    error: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+    error: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
     fetchingProject: PropTypes.bool,
     hideIntro: PropTypes.bool,
     importInfoVisible: PropTypes.bool,
--- a/src/containers/gui.jsx
+++ b/src/containers/gui.jsx
@@ -9,6 +9,7 @@ import {defineMessages, injectIntl, intlShape} from 'react-intl';
 import ErrorBoundaryHOC from '../lib/error-boundary-hoc.jsx';
 import {openExtensionLibrary} from '../reducers/modals';
 import {
+    getIsError,
     getIsShowingProject
 } from '../reducers/project-state';
 import {setProjectTitle} from '../reducers/project-title';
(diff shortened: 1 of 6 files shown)


src/containers/gui.jsx
```diff
@@ -55,16 +56,16 @@ class GUI extends React.Component {
         }
     }
     render () {
-        if (this.props.loadingError) {
+        if (this.props.isError) {
             throw new Error(
-                `Failed to load project from server [id=${window.location.hash}]: ${this.props.errorMessage}`);
+                `Failed to load project from server [id=${window.location.hash}]: ${this.props.errStr}`);
```
rschamp: I see where this change came from but it makes me sad because `errorMessage` is clearer. `errStr` looks like a temporary variable name not a permanent part of the state.  Would it be too much to ask to update `errStr` to be called `errorMessage` everywhere it's defined/used?

benjiwheeler: Not too much to ask! I agree. (I only used a different name when creating errStr because I wasn't yet replacing it everywhere, and they had to coexist!)

benjiwheeler:
I did the renaming...

benjiwheeler:
My additional changes are almost small enough to merge, but since I changed a lot of string locations, I'd appreciate a second look!

src/lib/vm-manager-hoc.jsx
```diff
@@ -51,29 +48,26 @@ const vmManagerHOC = function (WrappedComponent) {
                     this.props.onLoadedProject(this.props.loadingState, this.props.canSave);
                 })
                 .catch(e => {
-                    // Need to catch this error and update component state so that
-                    // error page gets rendered if project failed to load
-                    this.setState({loadingError: true, errorMessage: e});
+                    this.props.onError(`vm-manager-hoc error: ${e}`);
```
rschamp: Did you add this context because we don't have a good stack trace with this way of error handling? Maybe we shouldn't be setting state to the stringified error, but the full error object, so it can contain a stack trace? (If it already stringifies with a stack trace then I would say the added context is unnecessary.)

rschamp: Just realized it had been passing the full error. I think we should continue doing that, and expect the error set with `onError` to be an error object, not a string. As a follow up we should start outputting `e.stack` to get the stack trace.

benjiwheeler: to be clear, for now I'll replace the `errorMessage` prop entirely with an `error` prop, and only get the string from that when outputting it to the user

benjiwheeler: But, there's a bigger issue. I would like to include context that we can display to the user, for help in practically understanding where errors occur, both when they happen to users and to us. Maybe we could pass *both* a string reflecting where the error occured, and the error object itself?

benjiwheeler: that sounds good. i'll make a follow on issue.

benjiwheeler: Made https://github.com/LLK/scratch-gui/issues/3468

Issue #3444, opened 2018-10-22
"Coming soon" disabled UI and tooltips should not be shown in www


benjiwheeler:
### Expected Behavior

When viewing a project in the gui editor through www, the UI choices you see should all be either enabled, or disabled for clear reasons.

### Actual Behavior

UI choices that don't apply to your situation still appear: e.g., the "Share" button or the "Save now" menu item, when you are viewing a project you don't own. Even more confusing, these items show a "Coming soon" tooltip when you mouseover them, even if they're already implemented!

### Proposal

Proposed options when viewing a project you own: ("Save as a copy" is not implemented yet, but it would be here too, if it were!)
<img width="680" alt="screen shot 2018-10-21 at 10 12 53 pm" src="https://user-images.githubusercontent.com/3431616/47276023-4e3ebf80-d582-11e8-84c6-35885dbdbc62.png">

Proposed options when viewing a project you do not own: (note the missing "Save now" and "Share" UI options; "Remix" isn't implemented yet, but it would be here too, if it were!)
<img width="605" alt="screen shot 2018-10-21 at 10 11 58 pm" src="https://user-images.githubusercontent.com/3431616/47276034-5b5bae80-d582-11e8-8b63-50709ce46ec7.png">

PR #3460, opened 2018-10-23
added showComingSoon flag prop for gui

benjiwheeler:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/3444

### Proposed Changes

introduce the showComingSoon prop on GUI, which indicates whether to show "coming soon" tooltips on grayed-out menu options (and header menu bar items); if false, we simply hide them.

### Reason for Changes

In Beta, plenty of stuff is "coming soon". But embedded in www -- or, for that matter, in other players -- we are now able to use the user's session status and project ownership to determine what menu items we do and do not enable.

### Test Coverage

None

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -298,39 +298,39 @@ class MenuBar extends React.Component {
                                     >
                                         {newProjectMessage}
                                     </MenuItem>
-                                ) : (
+                                ) : (this.props.showComingSoon ? (
                                     <MenuItemTooltip
                                         id="new"
                                         isRtl={this.props.isRtl}
                                     >
                                         <MenuItem>{newProjectMessage}</MenuItem>
                                     </MenuItemTooltip>
-                                )}
+                                ) : [])}
                                 <MenuSection>
                                     {this.props.canSave ? (
                                         <MenuItem onClick={this.handleClickSave}>
                                             {saveNowMessage}
                                         </MenuItem>
-                                    ) : (
+                                    ) : (this.props.showComingSoon ? (
                                         <MenuItemTooltip
                                             id="save"
                                             isRtl={this.props.isRtl}
                                         >
                                             <MenuItem>{saveNowMessage}</MenuItem>
                                         </MenuItemTooltip>
-                                    )}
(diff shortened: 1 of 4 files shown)


src/components/menu-bar/menu-bar.jsx
```diff
@@ -298,39 +298,39 @@ class MenuBar extends React.Component {
...
-                                ) : (
+                                ) : (this.props.showComingSoon ? (
                                     <MenuItemTooltip
                                         id="new"
                                         isRtl={this.props.isRtl}
                                     >
                                         <MenuItem>{newProjectMessage}</MenuItem>
                                     </MenuItemTooltip>
-                                )}
+                                ) : [])}
```
rschamp: It doesn't matter too much, but I think we're pretty consistent in this repo in using `prop ? () : null` for conditional rendering.  As an additional aside, I think it might be clearer if we started using `prop && ()` for this, to avoid confusion with nested ternaries.

benjiwheeler: So, there's actually a problem in our code that makes using `null` this way unsafe!

In `menu.jsx`'s `MenuSection`, we're doing:

```
const addDividerClassToFirstChild = (child, id) => (
    React.cloneElement(child, {
        className: classNames(child.className, {
            [styles.menuSection]: id === 0
        }),
        key: id
    })
);

const MenuSection = ({children}) => (
    <React.Fragment>{
        React.Children.map(children, addDividerClassToFirstChild)
    }</React.Fragment>
);
```

It looks like React sees a `null` value as a child, and so we're attempting to get `(null)className`, throwing an error: 
```
Uncaught TypeError: Cannot read property 'className' of null
```

rschamp: Sounds like more reason to use `prop && ()`!

Issue #3034, opened 2018-08-29
Conditionally display Save now or Remix in the File menu


rschamp:
Follow up for https://github.com/LLK/scratch-gui/pull/3021#discussion_r213711451

benjiwheeler:
Should be followed up with correct API calls, https://github.com/LLK/scratch-gui/issues/3197

PR #3461, opened 2018-10-23
made remixing, saving as copy states; project state names are more consistent

benjiwheeler:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/3034

### Proposed Changes

Introduces support for Remixing and Saving as a copy. 

For now, these are done using the general create project functionality in scratch-storage; that means these won't have connections to their parent/original projects.

### Test Coverage

None yet

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


src/components/menu-bar/menu-bar.jsx
```diff
@@ -26,7 +26,9 @@ import {
     getIsUpdating,
     getIsShowingProject,
     requestNewProject,
-    saveProject
+    remixProject,
+    updateProject,
```
rschamp: Why `update` instead of `save`?

src/components/menu-bar/menu-bar.jsx
```diff
@@ -145,16 +149,22 @@ class MenuBar extends React.Component {
         }
     }
     handleClickNew () {
-        // if canSave===true, it's safe to replace current project, since we will auto-save first
-        const readyToReplaceProject =
-            this.props.canSave || confirm('Replace contents of the current project?'); // eslint-disable-line no-alert
+        // if canCreateNew===true, it's safe to replace current project, since we will auto-save first
+        const readyToReplaceProject = this.props.canCreateNew ||
```
rschamp: If we are changing names... I still think "create new" is redundant.

src/reducers/project-state.js
```diff
@@ -49,11 +55,13 @@ const getIsLoadingWithId = loadingState => (
     loadingState === LoadingState.LOADING_VM_NEW_DEFAULT
 );
 const getIsCreating = loadingState => (
-    loadingState === LoadingState.CREATING_NEW
+    loadingState === LoadingState.CREATING_NEW ||
+    loadingState === LoadingState.REMIXING ||
+    loadingState === LoadingState.SAVING_AS_COPY
```
rschamp: ... unless we had `createNew` and `createCopy`?

src/components/menu-bar/menu-bar.jsx
benjiwheeler: I think "save" is ambiguous. Remixing involved saving. Saving as a copy involves saving. Creating involves saving. "Update" tells you that we have a project on the server already, it has an id, and we are just updating that. I think it will be clearer.

Issue #3334, opened 2018-10-10
Add buttons (filters) to Tutorials library menu


ntlrsk:
Add buttons to the top of the Tutorial library window to enable filtering of tutorials.

Here are the initial buttons suggested: 
* Animation
* Art
* Games
* Music 
* Stories

apple502j:
What about adding Extensions?

thisandagain:
@ntlrsk @paulkaplan Proposal for tags and assignments:

| Tutorial | Filter(s) | Tags |
| -------- | -------- | ------ |
| Getting Started | Animation, Art, Games, Music, Stories | help, stuck, how, can, say |
| Animate a Name | Animation | spin, grow |
| Make Music | Music | sound, instrument, play, song, band | 
| Make a Clicker Game | Games | click, clicked, score |
| Make a Chase Game | Games | arrow, keyboard, score |
| Add a Sprite | Animation, Art, Games, Stories | character |
| Add a Backdrop | Animation, Art, Games, Stories | background |
| Change Size | Animation, Art, Games, Stories | scale | 
| Record a Sound | Animation, Art, Games, Music, Stories | music, instrument, play, song, band |
| Make It Spin | Animation, Art, Games, Stories | rotate, rotation |
| Hide & Show | Animation, Art, Games, Stories | appear, disappear | 
| Switch Costume | Animation, Art, Games, Stories | paint, edit, change, character, sprite |
| Use Arrow Keys | Art, Games, Stories | game, keyboard |
| Add Effects | Animation, Art, Games, Stories | fx, color, fisheye, whirl, twist, pixelate, mosaic, 8bit, 8-bit, brightness, ghost, transparency, opacity |

ntlrsk:
Here's how we had listed them before seeing this - it will be good to compare! 


NAME | BUTTONS: FILTERS
-- | --
Getting Started | [None]
Animate a Name | Animation, Art
Make Music | Music
Make a Clicker Game | Games
Make a Chase Game | Games
Video Sensing | Games
Create a Story | Animation, Stories
Animate a Character | Animation, Art
Add a Sprite | Art, Games, Stories
Add a Backdrop | Art, Games, Stories
Change size | Animation, Art
Glide around | Animation, Stories
Record a sound | Music, Stories, Games,
Make It Spin | Animation
Hide & Show | Stories
Switch costume | Animation, Art, Games
Use arrow keys | Games
Add effects | Animation, Art

ntlrsk:
[...]

PR #3478, opened 2018-10-23
Add tutorial filters and tags, and fix translated search in libraries

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/3334

This also resolves an issue where you couldn't search translated libraries, so we were avoiding them. But they now work!

### Proposed Changes

Add tutorial tags and fix the bug with translated library item search by using intl.formatMessage to figure out what the translation is and match against that. 

/cc @thisandagain 

Note some of the filters arent translated in this gif only because it hasn't gone up for translations, and so some of the new ones are using english default.
translated-search


rschamp:
Try it out: [tutorial-tags](https://llk.github.io/scratch-gui/tutorial-tags)

PR #3491, opened 2018-10-24
Use assets stored in the vm instead of storage

rschamp:
Depends on https://github.com/LLK/scratch-vm/pull/1691, and tests will fail until that is merged and this PR is updated with a new version of VM.

### Resolves

Towards LLK/scratch-vm#1577

### Proposed Changes

Stop storing and accessing assets on storage, and begin using assets stored directly on VM costumes and sounds.  Also get rid of lingering uses of `storage.cache`, which won't do anything anymore, now that we don't access assets from the storage cache.

One caveat is that we will continue caching the default project json and assets onto storage.builtinHelper in order for that to keep working offline.

### Reason for Changes

See LLK/scratch-storage#50

### Test Coverage

Existing tests pass with these changes

### Browser Coverage
N/A

### DIFF
--- a/src/containers/record-modal.jsx
+++ b/src/containers/record-modal.jsx
@@ -80,17 +80,19 @@ class RecordModal extends React.Component {
                     sampleCount: clippedSamples.length
                 };
 
-                // Load the encoded .wav into the storage cache and get resulting
-                // md5 from storage
-                const storage = this.props.vm.runtime.storage;
-                const md5 = storage.builtinHelper.cache(
+                // Create an asset from the encoded .wav and get resulting md5
+                const storage = this.props.vm.storage;
+                vmSound.asset = storage.createAsset(
                     storage.AssetType.Sound,
                     storage.DataFormat.WAV,
                     new Uint8Array(wavBuffer),
+                    null,
+                    true // generate md5
                 );
+                vmSound.assetId = vmSound.asset.assetId;
 
                 // update vmSound object with md5 property
-                vmSound.md5 = `${md5}.${vmSound.dataFormat}`;
+                vmSound.md5 = `${vmSound.assetId}.${vmSound.dataFormat}`;
                 // The VM will update the sound name to a fresh name
                 // if the following is already taken
                 vmSound.name = 'recording1';
--- a/src/containers/stage-selector.jsx
+++ b/src/containers/stage-selector.jsx
@@ -106,7 +106,7 @@ class StageSelector extends React.Component {
     }
     render () {
         const componentProps = omit(this.props, [
-            'assetId', 'dispatchSetHoveredSprite', 'id', 'intl', 'onActivateTab', 'onSelect']);
+            'asset', 'dispatchSetHoveredSprite', 'id', 'intl', 'onActivateTab', 'onSelect']);
         return (
             <DroppableStage
                 fileInputRef={this.setFileInput}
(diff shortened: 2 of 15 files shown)


src/containers/record-modal.jsx
```diff
@@ -80,17 +80,19 @@ class RecordModal extends React.Component {
                     sampleCount: clippedSamples.length
                 };
 
-                // Load the encoded .wav into the storage cache and get resulting
-                // md5 from storage
-                const storage = this.props.vm.runtime.storage;
-                const md5 = storage.builtinHelper.cache(
+                // Create an asset from the encoded .wav and get resulting md5
+                const storage = this.props.vm.storage;
+                vmSound.asset = storage.createAsset(
```
rschamp: Instead of keeping the data for an asset in storage, keyed by asset id, we just keep an asset object on the vm objects that use the assets.

src/containers/stage-selector.jsx
```diff
@@ -130,8 +130,8 @@ StageSelector.propTypes = {
     onSelect: PropTypes.func
 };
 
-const mapStateToProps = (state, {assetId, id}) => ({
-    url: assetId && state.scratchGui.vm.runtime.storage.get(assetId).encodeDataURI(),
```
rschamp: Other side of the change — we don't have to pull anything out of storage anymore, it can just come directly from the costume/sound on the VM.

src/lib/storage.js
```diff
@@ -55,7 +55,7 @@ class Storage extends ScratchStorage {
     }
     cacheDefaultProject () {
         const defaultProjectAssets = defaultProject(this.translator);
-        defaultProjectAssets.forEach(asset => this.cache(
+        defaultProjectAssets.forEach(asset => this.builtinHelper._store(
```
rschamp: We still need to do this to have the default project work offline. But I updated this call to avoid deprecation warnings.

src/lib/get-costume-url.js
```diff
@@ -7,30 +8,29 @@ const getCostumeUrl = (function () {
...
             if (svgString.match(HAS_FONT_REGEXP)) {
                 const svgRenderer = new SVGRenderer();
                 svgRenderer.loadString(svgString);
                 const svgText = svgRenderer.toString(true /* shouldInjectFonts */);
                 cachedUrl = `data:image/svg+xml;utf8,${encodeURIComponent(svgText)}`;
             } else {
-                cachedUrl = vm.runtime.storage.get(assetId).encodeDataURI();
+                cachedUrl = asset.encodeDataURI();
             }
+            cachedUrl = asset.encodeDataURI();
```
kchadha: It looks like this line is extra and also overwriting the url data from the font injection case...

src/lib/storage.js
kchadha: Is the plan to have the default project (and its assets) eventually live in storage like the `?` asset for example?

kchadha:
BTW, the recent storage changes also need to be pulled into this PR before it can be merged.

I'm actually confused about how the tests are passing... I got an error when I pulled in these changes that `this.builtinHelper._store` doesn't exist

src/lib/get-costume-url.js
rschamp: Whoops — that's a merge error.

src/lib/storage.js
rschamp: We don't have a plan but we could think about that!

rschamp:
The storage changes are already pulled into develop already! They were backwards compatible, so could go in beforehand. You probably needed to do an `npm install`.

rschamp:
Although it is surprising that all the tests pass without pulling in the VM changes...  I guess it will not crash, but it shouldn't work correctly.

Issue #3197, opened 2018-09-20
When user clicks remix in gui, submit correct request to project server


benjiwheeler:
Note that this will involve some changes in -storage.

Should be done with/after correctly presenting the UI for this: https://github.com/LLK/scratch-gui/issues/3034

Issue #3324, opened 2018-10-10
Display project state error messages using new messaging system


benjiwheeler:
See places in the code with notes like

```
// NOTE: should throw up a notice for user
...onError...
```

PR #3493, opened 2018-10-24
send params in storage request body, so that remixing and saving as copy create links to original projects

benjiwheeler:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/3197

### Proposed Changes

For remix project creation and save as a copy project creation, we need to add params to the POST request. 

The first version of this PR did this by passing an object of params to the scratch-storage library, and letting it assemble the request url.

The current version of this PR doesn't pass the object of params. Instead, it adds these params directly to the outgoing request body.

This requires scratch-projects PR https://github.com/LLK/scratch-projects/pull/83 .

### DIFF
--- a/src/lib/project-saver-hoc.jsx
+++ b/src/lib/project-saver-hoc.jsx
@@ -9,7 +9,9 @@ import {
     createProject,
     doneCreatingProject,
     doneUpdatingProject,
-    getIsCreating,
+    getIsCreatingCopy,
+    getIsCreatingNew,
+    getIsRemixing,
     getIsShowingProject,
     getIsShowingWithoutId,
     getIsUpdating,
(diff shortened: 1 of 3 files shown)


thisandagain:
@benjiwheeler Looks like this is not passing CI (lint errors). Mind taking a look?

src/lib/project-saver-hoc.jsx
```diff
@@ -41,16 +43,39 @@ const ProjectSaverHOC = function (WrappedComponent) {
...
                     .catch(err => {
-                        // NOTE: should throw up a notice for user
                         this.props.onProjectError(`Creating a new project failed with error: ${err}`);
                     });
             }
+            if (this.props.isCreatingCopy && !prevProps.isCreatingCopy) {
+                this.storeProject(null, {
+                    original_id: this.props.reduxProjectId,
+                    is_copy: 1
+                })
```
rschamp: Just noting what we already discussed, that we probably want to provide these to the `body` of our request, rather than configuring query params using storage.

benjiwheeler:
As @rschamp and I talked about, I changed the approach to pass params in the body of the request, not the query.

src/lib/project-saver-hoc.jsx
```diff
@@ -41,16 +43,39 @@ const ProjectSaverHOC = function (WrappedComponent) {
...
             // TODO: distinguish between creating new, remixing, and saving as a copy
-            if (this.props.isCreating && !prevProps.isCreating) {
-                this.storeProject()
+            if (this.props.isCreatingNew && !prevProps.isCreatingNew) {
+                this.storeProject(null)
                     .then(response => {
                         this.props.onCreatedProject(response.id.toString(), this.props.loadingState);
                     })
                     .catch(err => {
-                        // NOTE: should throw up a notice for user
```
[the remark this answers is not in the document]
benjiwheeler: No, but the error is being handled in the main error display code in the gui container, there's an issue for it ( https://github.com/LLK/scratch-gui/issues/3324 ), and an identical comment above. Since I was going to replicate similar code for save as a copy and remix, I didn't want to include this comment so many times.

rschamp:
Marked as needs-work & blocked until https://github.com/LLK/scratch-projects/pull/83 goes in

benjiwheeler:
https://github.com/LLK/scratch-projects/pull/83 is in!

src/lib/project-saver-hoc.jsx
rschamp: Doesn't this still need to be updated to send the params in the body rather than using the API for storage that we decided against?

benjiwheeler: Weird, I could've sworn I had pushed after I rebased on top of develop yesterday. Sorry, please look again now.

rschamp: Oh — no this was my fault, I was misreading this as a call to storage, not our own helper function.

benjiwheeler:
Updated title. (Description was already updated.)

PR #3495, opened 2018-10-24
Add a hook for custom storage init

cwillisf:
### Proposed Changes

This change adds a way for a consumer of `scratch-gui` to customize the `scratch-storage` instance that the GUI creates and shares with the VM and the rest of the system. If the `onStorageInit` property is not provided, the default value makes everything work as it did before.

### Reason for Changes

This allows `scratch-desktop` to inject its own "Helper" to handle loading bundled assets through Electron.

### Test Coverage

The existing tests verify that this change doesn't break current behavior.

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [x] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


cwillisf:
Follow-up filed as #3569.

Issue #3466, opened 2018-10-23
Share button, (inactive) Shared button, or Remix button should appear when appropriate


benjiwheeler:
Follow on to https://github.com/LLK/scratch-gui/issues/3444

Carl shared designs that make the Share button show "Shared" when the project has been shared, and that have a Remix button in its place when remixing is possible and shared is not (that is, when the user is logged in, but does not own the project.)

We should implement these.

Designs: https://projects.invisionapp.com/d/main#/console/15762036/327192535/preview

benjiwheeler:
Thanks for that perspective, @towerofnix -- I definitely wouldn't have thought of that!

I made an issue for that question, at https://github.com/LLK/scratch-gui/issues/3501

benjiwheeler:
Re: design, I removed the label because I spoke to @carljbowman and he said we should not use an icon for Share for the time being (none has been determined yet), but we *should* use the spiral svg for the Remix button.

benjiwheeler:
@carljbowman, one last thought -- could it make sense to have hover styles that alter the appearance of these buttons somewhat?

benjiwheeler:
screenshots of what @paulkaplan and I made:

![screen shot 2018-10-26 at 9 31 46 am](https://user-images.githubusercontent.com/3431616/47569880-cac6fa80-d902-11e8-822c-7ba74e987f2f.png)
![screen shot 2018-10-26 at 9 33 06 am](https://user-images.githubusercontent.com/3431616/47569881-cb5f9100-d902-11e8-8d7f-94c5fb3bfc9f.png)
![screen shot 2018-10-26 at 9 34 18 am](https://user-images.githubusercontent.com/3431616/47569883-cb5f9100-d902-11e8-9692-8d66c3b5b908.png)

PR #3508, opened 2018-10-24
shared and remix buttons

benjiwheeler:
Resolves https://github.com/LLK/scratch-gui/issues/3466

should be merged with https://github.com/LLK/scratch-www/pull/2234

Uses the `isShared` prop, the `canShare` prop, and the `canRemix` prop to conditionally display Share and Remix buttons.

Note that as of the initial PR code version, the Remix button shows the wrong color!


src/components/menu-bar/menu-bar.jsx
```diff
@@ -241,16 +242,43 @@ class MenuBar extends React.Component {
...
+                    {[styles.shareButtonIsShared]: this.props.isShared}
+                )}
                 onClick={this.props.onShare}
             >
-                <FormattedMessage
-                    defaultMessage="Share"
-                    description="Label for project share button"
-                    id="gui.menuBar.share"
-                />
+                {this.props.isShared ? isSharedMessage : shareMessage}
```
paulkaplan: I think with jsx it is generally preferred to inline everything, it makes it easier to read. Could you change this to inline the two formatted messages?

paulkaplan:
@benjiwheeler can you post screenshots of each of the states, and tag @carljbowman to make sure it seems right?

src/components/menu-bar/menu-bar.jsx
```diff
@@ -241,16 +242,43 @@ class MenuBar extends React.Component {
...
+            </Button>
+        );
+        const remixButton = (
+            <Button
+                className={classNames(
+                    styles.menuBarButton,
+                    styles.remixButto
```
paulkaplan: Looks like this className is wrong

PR #3520, opened 2018-10-25
[Develop] Use names for tutorial URL ids

rschamp:
To make URLs more friendly, use e.g., `tutorial=getting-started` instead of `tutorial=1`.  This will also help with some URL handling we're doing for the new project page.

### DIFF
--- a/src/lib/libraries/decks/index.jsx
+++ b/src/lib/libraries/decks/index.jsx
@@ -126,7 +126,7 @@ export default {
             ]
         }
         ],
-        urlId: 'getting-started'
+        urlId: 'getStarted'
     },
 
     'say-it-out-loud': {
@@ -423,7 +423,7 @@ export default {
             ]
         }
         ],
-        urlId: 'animate-a-name'
+        urlId: 'name'
     },
     'Make-Music': {
         name: (
@@ -492,7 +492,7 @@ export default {
             ]
         }
         ],
-        urlId: 'make-music'
+        urlId: 'music'
     },
     'Make-A-Game': {
         name: (
@@ -833,7 +833,7 @@ export default {
                 'switch-costume'
             ]
         }],
-        urlId: 'hide-and-show'
+        urlId: 'hide'
     },
 
     'switch-costume': {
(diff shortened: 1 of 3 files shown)


src/lib/tutorial-from-url.js
```diff
@@ -36,15 +36,10 @@ const getDeckIdFromUrlId = urlId => {
...
-            queryParams.tutorial
-    );
-    if (!isNaN(tutorialID)) {
-        return getDeckIdFromUrlId(tutorialID);
-    }
-    return null;
+    const tutorialID = Array.isArray(queryParams.tutorial) ?
+        queryParams.tutorial[0] :
+        queryParams.tutorial;
+    return getDeckIdFromUrlId(tutorialID);
```
kchadha: Should we also add an early exit in getDeckIdFromUrlId if tutorialID is `undefined`...?

src/lib/libraries/decks/index.jsx
```diff
@@ -423,7 +423,7 @@ export default {
             ]
         }
         ],
-        urlId: 2
+        urlId: 'animate-a-name'
```
chrisgarrity: Currently: `projects/editor/?tip_bar=name`

src/lib/tutorial-from-url.js
kchadha: Would a tutorial that is missing the 'urlId' field get matched to this tutorialId in this case?

src/lib/libraries/decks/index.jsx
```diff
@@ -492,7 +492,7 @@ export default {
             ]
         }
         ],
-        urlId: 3
+        urlId: 'make-music'
```
chrisgarrity: `/projects/editor/?tip_bar=music`

src/lib/tutorial-from-url.js
rschamp: Good catch, we should exit early!

rschamp:
@chrisgarrity I found one more, `hide`. But none of the rest seem to map, so I left them alone. Happy to take any other suggestions.

@kchadha I fixed that bug and added a regression test.

thisandagain:
This looks good. Thanks @rschamp !

chrisgarrity:
👍

PR #3523, opened 2018-10-25
Hack to load a starter project for a specific tutorial

ericrosenbaum:
First of all, let me make clear that @paulkaplan bears no responsibility, moral, ethical, technical or otherwise, for the contents of this PR.

Also, thanks to @paulkaplan for writing the elegant and powerful code in this PR.

The new "Animate an Adventure Game" tutorial requires assets from a starter project. We will generally direct people to it using a URL that includes both the tutorial ID and starter project ID. But, because you can also get to the tutorial by opening it from the library, we need a way to direct people to the starter project. 

After considering several options, we decided to do the following: when you click on this particular tutorial, if your browser is already at a URL containing the starter project ID, open the tutorial as usual; if not, open a new browser window with both the tutorial and starter project open. 

This is a temporary solution, until we have a more robust way to provide assets or starter projects for tutorials.

### DIFF
--- a/src/containers/tips-library.jsx
+++ b/src/containers/tips-library.jsx
@@ -35,6 +35,19 @@ class TipsLibrary extends React.PureComponent {
         ]);
     }
     handleItemSelect (item) {
+        /*
+            Support tutorials that require specific starter projects.
+            If a tutorial declares "requiredProjectId", check that the URL contains
+            it. If it is not, open a new page with this tutorial and project id.
+
+            TODO remove this at first opportunity. If this is still here after HOC2018,
+                 blame Eric R. Andrew is also on record saying "this is temporary".
+        */
+        if (item.requiredProjectId && window.location.href.indexOf(item.requiredProjectId) === -1) {
+            const urlParams = `?tutorial=${item.urlId}#${item.requiredProjectId}`;
+            return window.open(window.location.pathname + urlParams, '_blank');
+        }
+
         this.props.onActivateDeck(item.id);
         analytics.event({
             category: 'library',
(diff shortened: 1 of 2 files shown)


src/containers/tips-library.jsx
```diff
@@ -35,6 +35,19 @@ class TipsLibrary extends React.PureComponent {
...
+            Support tutorials that require specific starter projects.
+            If a tutorial declares "requiredProjectId", check that the URL contains
+            it. If it is not, open a new page with this tutorial and project id.
+
+            TODO remove this at first opportunity. If this is still here after HOC2018,
+                 blame Eric R. Andrew is also on record saying "this is temporary".
+        */
+        if (item.requiredProjectId && window.location.href.indexOf(item.requiredProjectId) === -1) {
+            const urlParams = `?tutorial=${item.urlId}#${item.requiredProjectId}`;
+            return window.open(window.location.pathname + urlParams, '_blank');
```
apple502j: `window.open` is not good because many browsers block it as "pop-up window needs permission"

towerofnix: Agreed. I think a cleaner solution would be to have a link (just a normal `<a>`) within the tutorial, saying "Open Starter Project", which would target `_blank`. I think doing that is more reliable(?) - and it's better UX anyway because they *expect* a new project to open up.

paulkaplan: @apple502j hmm I've tried it out on chrome and safari. In general `window.open` is allowed by browsers as long as it is in response to a direct user action (i.e. you can't just call window.open whenever you want). 

As noted in the comment, this is meant to be very temporary. But if it is being blocked by browsers, of course, then we need to revisit. We will test many different browsers shortly.

PR #3530, opened 2018-10-26
Add reconnect button to hardware disconnect alerts

### LINKED ISSUE #3262: Add peripheral reconnect button to alerts
As specified in https://github.com/LLK/scratch-gui/issues/2956

evhan55:
### Resolves

- Resolves #3262: Add peripheral reconnect button to alerts
- Progress on #2956: Message System - Phase 1

### DIFF
--- a/src/components/alerts/alert.jsx
+++ b/src/components/alerts/alert.jsx
@@ -31,8 +31,8 @@ const AlertComponent = ({
         >
             <FormattedMessage
                 defaultMessage="Reconnect"
-                description="Button to disconnect the device"
-                id="gui.connection.disconnect"
+                description="Button to reconnect the device"
+                id="gui.connection.reconnect"
             />
         </button>
         <Box
--- a/src/components/alerts/alert.jsx
+++ b/src/components/alerts/alert.jsx
@@ -11,7 +11,8 @@ const AlertComponent = ({
     iconURL,
     message,
     onCloseAlert,
-    onReconnect
+    onReconnect,
+    showReconnect
 }) => (
     <Box
         className={styles.alert}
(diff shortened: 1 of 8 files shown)


src/components/alerts/alert.jsx
```diff
@@ -23,19 +25,34 @@ const AlertComponent = ({
...
+        </Box>
     </Box>
 );
 
 AlertComponent.propTypes = {
     iconURL: PropTypes.string,
     message: PropTypes.string,
-    onCloseAlert: PropTypes.func.isRequired
+    onCloseAlert: PropTypes.func.isRequired,
+    onReconnect: PropTypes.func.isRequired
```
rschamp: Similarly, I think we will want it possible to not supply `onReconnect` for alerts that don't have to do with extensions.

src/reducers/alerts.js
```diff
@@ -19,6 +19,7 @@ const reducer = function (state, action) {
             const extension = extensionData.find(ext => ext.extensionId === extensionId);
             if (extension && extension.name) {
                 // TODO: is this the right place to assemble this message?
+                newAlert.extensionId = extensionId;
```
rschamp: Maybe the data about if an alert should show the reconnect button should be set explicitly here, so that we can decide whether to display it or not. It looks like we currently want to display it for every alert associated with an extension, so for now maybe we should set `newAlert.showReconnect` if `extensionId` is truthy?  Then we can pass that as a prop to the alert component.

rschamp: Unrelated — typing out that previous comment made me wonder if `extensionId` can ever end up as `0`? Because then `if (extensionId)` above would be `false`...

src/components/alerts/alert.jsx
```diff
@@ -23,19 +25,34 @@ const AlertComponent = ({
...
+        <button
+            className={styles.connectionButton}
+            onClick={onReconnect}
+        >
+            <FormattedMessage
+                defaultMessage="Reconnect"
+                description="Button to reconnect the device"
+                id="gui.connection.reconnect"
+            />
+        </button>
```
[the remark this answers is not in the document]
evhan55: Yes, thanks!  Tomorrow I will discuss this with @benjiwheeler and @chrisgarrity to see about the best way to incorporate buttons into modular alerts.

src/components/alerts/alert.jsx
evhan55: 👍

src/reducers/alerts.js
evhan55: Yes, I believe this is now done in #3559

evhan55: oh hm that's a good question, I am not sure and will look into it, thank you

benjiwheeler:
@evhan55 and I talked about this PR and agreed that she will clean up one or two small things (in response to @rschamp's comments) and then move to merge it; after it's merged, I'll rebase my pr https://github.com/LLK/scratch-gui/pull/3559 (which makes alerts more general and extensible) on it.

src/reducers/alerts.js
evhan55: extensionId's are strings such as `ev3`, not numbers

benjiwheeler: I'll pick this up in #3559

PR #3532, opened 2018-10-27
Restore IntlProvider for Internet Explorer

chrisgarrity:
Since https://github.com/LLK/scratch-gui/pull/3431 GUI includes its own Intl provider (react-intl supports nested intl providers) that works within the www intl provider.

This PR restores the app-state-hoc intl provider for the stand-alone gui so that IE can function enough to put up the unsupported browser.

This will permit adding localization to things outside what's exported by GUI. We should be very careful about adding localization to any GUI wrapper component because it will work in playground (because of this PR), but possibly break in www.


paulkaplan:
@rschamp @benjiwheeler could you review this? I'm not familiar with the context from #3431 but you two reviewed that one.

benjiwheeler:
This LGTM!

PR #3539, opened 2018-10-29
Cloud Data Provider

kchadha:
### Proposed Changes

Adds a cloud data provider which creates and manages a web socket connection to the cloud data host provided by `www`. The connection is created upon successfully loading a project. 

This is a work in progress towards adding cloud variables to Scratch 3.0. 

This PR does not include:
 - re-attempting a connection upon error
- logic around whether or not the current state indicates that a cloud data connection should be opened (e.g. user info, project info, etc.)
- remix logic
- ability to create new cloud variables in a 3.0 project

#### Related PRs

- LLK/scratch-vm#1703
- LLK/scratch-www#2245

### Reason for Changes

Cloud variables!

### Test Coverage

Manually tested sending and receiving updates to a 2.0 project containing cloud variables when viewing that project from the 3.0 project page.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/lib/cloud-provider.js
+++ b/src/lib/cloud-provider.js
@@ -99,6 +99,26 @@ class CloudProvider {
         this.writeToServer('set', name, value);
     }
 
+    /**
+     * Closes the connection to the web socket and clears the cloud
+     * provider of references related to the cloud data project.
+     */
+    requestCloseConnection () {
+        this.connection.close();
+        this.clear();
+    }
+
+    /**
+     * Clear this provider of references related to the project
+     * and current state.
+     */
+    clear () {
+        this.connection = null;
+        this.vm = null;
+        this.username = null;
+        this.projectId = null;
+    }
+
 }
 
 export default CloudProvider;
--- a/src/lib/cloud-provider.js
+++ b/src/lib/cloud-provider.js
@@ -1,0 +1,104 @@
+import log from './log.js';
+
+
+class CloudProvider {
+    constructor (cloudHost, vm, user, projectId) {
+        this.vm = vm;
+        this.username = user;
+        this.projectId = projectId;
+
+        // Open a websocket connection to the clouddata server
+        this.openConnection(cloudHost);
+    }
+
+    /**
+     * Open a new websocket connection to the clouddata server.
+     * @param {string} cloudHost The cloud data server to connect to.
+     */
+    openConnection (cloudHost) {
+        if (window.WebSocket === null) {
+            log.warn('Websocket support is not available in this browser');
+            this.connection = null;
+            return;
+        }
+
+        this.connection = new WebSocket((location.protocol === 'http:' ? 'ws://' : 'wss://') + cloudHost);
+
+        this.connection.onerror = e => {
+            log.error(`Websocket connection error: ${JSON.stringify(e)}`);
+
+            // TODO Add re-connection attempt logic here
+        };
+
+        this.connection.onmessage = event => {
+            const messageString = event.data;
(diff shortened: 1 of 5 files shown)


src/lib/cloud-provider.js
```diff
@@ -0,0 +1,124 @@
...
+     * @param {string} dataName The name of the cloud variable this message pertains to
+     * @param {string | number} dataValue The value to set the cloud variable to
+     * @param {number} dataIndex The index of the item to update (for cloud lists)
+     * @param {string} dataNewName The new name for the cloud variable (if renaming)
+     */
+    writeToServer (methodName, dataName, dataValue, dataIndex, dataNewName) {
+        const msg = {};
+        msg.method = methodName;
+        msg.user = this.username;
+        msg.project_id = this.projectId;
```
towerofnix: Should this require some form of authentication? (I.e. like `?x-token`.) The absolute last thing anyone wants is for a user to impersonate some other user in the cloud data log by overriding `msg.user` to be something other than their own username.

joker314: @towerofnix This looks the same as 2.0 cloud protocol; where one uses cookies (in the upgrade request, I think) to confirm authenticity.

kchadha: @towerofnix, the messages implicitly use session data, and authentication on the server fails without it.

src/lib/cloud-provider.js
```diff
@@ -0,0 +1,124 @@
+import log from './log.js';
+
+
+class CloudProvider {
+    constructor (cloudHost, vm, user, projectId) {
```
rschamp: Docs would be good here. Also I would suggest `username` rather than `user` since it could be confusing going between www — in www `user` might be an object.

src/lib/cloud-provider.js
```diff
@@ -0,0 +1,124 @@
...
+
+        this.connection.onmessage = event => {
+            const messageString = event.data;
+            log.info(`Received websocket message: ${messageString}`);
+            const message = JSON.parse(messageString);
+            if (message.method === 'set') {
+                const varData = {
+                    projectId: this.projectId,
+                    varUpdate: {
+                        projectId: message.project_id,
```
rschamp: A comment here to explain the two different `projectId`'s would be nice.

src/lib/vm-manager-hoc.jsx
```diff
@@ -54,12 +67,14 @@ const vmManagerHOC = function (WrappedComponent) {
...
             const {
                 /* eslint-disable no-unused-vars */
+                cloudHost,
                 fontsLoaded,
                 loadingState,
                 onError: onErrorProp,
                 onLoadedProject: onLoadedProjectProp,
                 projectData,
                 projectId,
+                userName,
```
rschamp: I think almost everywhere else (including in the new CloudProvider class!) we use `username` without caps — can we stick with that for consistency?

src/lib/cloud-provider.js
kchadha: Oops! That was a remnant of old code that I decided wasn't necessary anymore.

kchadha:
@rschamp I added docs and updated the naming/capitalization of 'username'. I also took out the unnecessary/duplicate projectId in the message being posted to the VM.

Issue #2811, opened 2018-08-07
When user saves project, show save is in process


benjiwheeler:
Somewhat a duplicate of https://github.com/LLK/scratch-gui/issues/3035

When we implement saving in Beta, and user saves, we should give the user some indication that the save is in process.

Some possibilities: we could use some kind of popup modal/dialog, info in the nav bar, some kind of spinner/progress bar, and/or "Saving..." text.

carljbowman:
@kathymakes - Specced out some examples of a flash message system. "Save process" was included as an example of this system. See https://github.com/LLK/scratch-gui/issues/2956 for more info.

@benjiwheeler - Please let us know if you have any addtional questions.

kathymakes:
Some images of a potential "save process" design:

![saving flow 1](https://user-images.githubusercontent.com/8203939/44741225-76ea9180-aaca-11e8-89a1-bb9e1f268643.png)
![saving flow 2](https://user-images.githubusercontent.com/8203939/44741330-b6b17900-aaca-11e8-9681-09b055173775.png)

benjiwheeler:
Question: what tentative text should i use for alerting the user that the project was saved or created?

@kathymakes used "Successfully saved." above, which I like

benjiwheeler:
Oh, and I guess we need svgs for that checkmark, and also for the progress spinner?

kathymakes:
@benjiwheeler Here's the checkmark as a .svg
[checkmark.svg.zip](https://github.com/LLK/scratch-gui/files/2526373/checkmark.svg.zip)

Here's the loading animation I made (click the button, "download")
https://codepen.io/kathymakes/pen/BPXxWB

kathymakes:
@benjiwheeler 
Here is the spinner spec: https://github.com/LLK/scratch-www/issues/2264

Let me know if you have questions :)

PR #3559, opened 2018-10-29
made alert type called standard alerts; used for saving, creating

benjiwheeler:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/2811

### Proposed Changes

* make current SHOW_ALERT, which is extensions-specific, into SHOW_EXTENSIONS_ALERT
* introduce new category of alert: "standard" alerts. They will use SHOW_STANDARD_ALERT.
* introduce new attribute of alerts: alert "level". Values will be 'success' and 'warn' for now. Extensions alerts will all use 'warn' for now, and stay orange. Success-indicating alerts will use 'success', and be green.
* make saving and creating generate "saving/creating" and "successfully saved/created" alerts.

### Reason for Changes

Need to notify users about what's going on!

### Discussion

Obviously, these should not require the user to dismiss them. I'll make that change in a follow-on PR, and provide them a timeout, as well as make the "success" alerts dismiss the "in progress" alerts.

Also, the "in progress" alerts could use an icon -- either an actual progress spinner, or a simulated svg one.

### Screenshots

Creating: (first the top alert appears, then the bottom alert)
screen shot 2018-10-29 at 2 30 35 pm

Saving: (first the top alert appears, then the bottom alert)
screen shot 2018-10-29 at 2 30 44 pm

And FWIW, the original extensions alerts are still working:
screen shot 2018-10-29 at 2 32 17 pm


src/components/close-button/close-button.jsx
```diff
@@ -7,6 +7,8 @@ import closeIcon from './icon--close.svg';
 import closeIconOrange from './icon--close-orange.svg';
 import backIcon from '../../lib/assets/icon--back.svg';
 
+let closeIcons = {};
```
benjiwheeler: Need to initialize this variable so it can be used in the code below, even though it's not actually set until lower down

src/lib/vm-listener-hoc.jsx
```diff
@@ -9,7 +9,7 @@ import {updateTargets} from '../reducers/targets';
 import {updateBlockDrag} from '../reducers/block-drag';
```
benjiwheeler: This file's changes just change "showAlert" to "showExtensionAlert" everywhere

src/lib/project-saver-hoc.jsx
```diff
@@ -4,6 +4,7 @@ import {connect} from 'react-redux';
 import VM from 'scratch-vm';
```
benjiwheeler: This file's changes just use the four new standard alert types

src/lib/alerts/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+import {FormattedMessage} from 'react-intl';
+
+import successImage from '../assets/icon--success.svg';
+
+const AlertLevels = {
+    SUCCESS: 'success',
+    WARN: 'warn'
+};
+
+const alerts = [
```
benjiwheeler: define four standard alert types, to use in project-saver-hoc

evhan55:
This looks good to me!
I have one high level question, do you think it would be possible to just have extension alerts be standard alerts of type warn?  And have standard alerts possibly take buttons that might do something on a callback?  Because these are really just hardware disconnect peripheral alerts, which may or may not be extension related down the line.

paulkaplan:
@benjiwheeler tests are failing here?

benjiwheeler:
Note @rschamp's comment on #3530:

https://github.com/LLK/scratch-gui/pull/3530/files#r228986534

> Maybe the data about if an alert should show the reconnect button should be set explicitly here, so that we can decide whether to display it or not. It looks like we currently want to display it for every alert associated with an extension, so for now maybe we should set newAlert.showReconnect if extensionId is truthy? Then we can pass that as a prop to the alert component.

This PR should address this.

benjiwheeler:
Here is what this looks like with this PR:

save alerts

benjiwheeler:
OK, revised this so it's not so spammy! Take a look at the gif above.

Next change will be to make the alerts auto-dismiss themselves after a set time; this version is a little annoying in requiring alerts to be dismissed, but that's temporary.

src/components/alerts/alert.jsx
```diff
@@ -25,6 +34,7 @@ const AlertComponent = ({
                 />
             ) : null}
             {message}
+            {content}
```
paulkaplan: Looks like you want message and content to be separated by a space:
image

This is what it looks like on beta:
image

Putting these two on the same line and separating with an html entity space like `&nbsp;` will give you the result you want. 

However, this will not work well for translations, where the name of the device may not be at the end of the line. It looks like the message is coming from the VM, so I'm fine with not fixing that here. But can you file an issue on the VM and leave a TODO comment here with the issue number.

src/lib/project-saver-hoc.jsx
```diff
@@ -44,34 +53,40 @@ const ProjectSaverHOC = function (WrappedComponent) {
             }
             // TODO: distinguish between creating new, remixing, and saving as a copy
             if (this.props.isCreatingNew && !prevProps.isCreatingNew) {
-                this.storeProject(null)
+                this.props.onShowCreatingAlert();
+                this.storeProject()
```
paulkaplan: Why did the null go away?

src/reducers/alerts.js
```diff
@@ -1,34 +1,72 @@
...
+        const alertId = action.alertId;
+        if (alertId) {
+            const newAlert = {
+                alertId: alertId,
+                level: AlertLevels.WARN // default level
+            };
+            const alertData = alertsData.find(thisAlertData => thisAlertData.alertId === alertId);
+            if (alertData) {
+                let newList = state.alertsList.slice();
+                newList = newList.filter(curAlert => (
```
paulkaplan: I'm concerned about not having the logic of this reducer in a unit test, would it be a lot to add a unit test for this reducer? Particularly covering the clearList logic?

paulkaplan:
Also there are some pretty rough alignment issues. They were pre-existing, but look much worse with the new alerts because of the icon height. If you want to fix that here that would be good (remove the various top margins, use flex / align-items: center instead) or file a follow up issue for it.

src/components/alerts/alert.jsx
benjiwheeler: I don't know if it's an issue with the VM, beyond the need to translate the message itself -- but there's this issue: https://github.com/LLK/scratch-vm/issues/1725

In the meantime, I put in a TODO to handle RTL (I don't know how you reverse the order of multiple items) and added an &nbsp;

src/lib/project-saver-hoc.jsx
benjiwheeler: Thanks for noticing -- maybe this was just a rebase quirk I didn't catch? (Without the null param, the param is undefined, which has been tested as having the same behavior; that's what we used to use)

src/reducers/alerts.js
benjiwheeler: You're totally right! I added a handful of tests, including of this clearList logic.

benjiwheeler:
@paulkaplan I created https://github.com/LLK/scratch-gui/issues/3686 as a follow-up for styling.

I think your other comments/feedback should be addressed now!

src/containers/alert.jsx
```diff
@@ -41,9 +45,7 @@ class Alert extends React.Component {
     }
 }
 
-const mapStateToProps = state => ({
-    state: state
```
benjiwheeler: this seems to have been completely useless, so I removed it -- it can't be having any effect, right?

evhan55: I vaguely remember putting this in for some lint-like strange reason, but I forget the details, so if it works for you to remove then sounds good!

PR #3565, opened 2018-10-30
Issue/hotfix project load

chrisgarrity:
If you switch language after loading the default project it gets reloaded, losing any changes you may have made. 

This PR checks whether the project id is changing in the reducer, but I think there's a deeper root cause as changing locale should not cause a SET_PROJECT_ID dispatch.

### DIFF
--- a/src/reducers/project-state.js
+++ b/src/reducers/project-state.js
@@ -172,6 +172,10 @@ const reducer = function (state, action) {
         }
         return state;
     case SET_PROJECT_ID:
+        // if the projectId hasn't actually changed do nothing
+        if (state.projectId === action.projectId) {
+            return state;
+        }
         // if setting the default project id, specifically fetch that project
         if (action.projectId === defaultProjectId) {
             return Object.assign({}, state, {


chrisgarrity:
The problem starts with the hash-parser-hoc:
https://github.com/LLK/scratch-gui/blob/96071be18dc0a95c57d0419d8a03f32b768d688b/src/lib/hash-parser-hoc.jsx#L42-L45

I'm not sure why the `hashchange` event is getting fired, but it definitely is.

chrisgarrity:
Not only is `handleHashChange` getting called when the locale changes, but the `HashParserHOC` also gets remounted (the constructor is getting called when the locale changes), so trying to save the current hashId in the HOC doesn't work because it gets re-initialized each time.

benjiwheeler:
Tested this in a bunch of situations. My code could be cleaned up below this check now that we can assume it is false, but that can be a follow-on.

chrisgarrity:
@benjiwheeler yes, I didn't want to change any of the other logic for this PR to make sure I didn't miss an edge case in the saving/loading/remixing/copying.

PR #3583, opened 2018-10-31
Blur language select focus when mouse leaves

### LINKED ISSUE #3302: Language selection keeps keyboard focus if a language is not selected.
### Expected Behavior
If you open the language selection list, but then click away (i.e. do not select a language), focus returns to the editor and the up/down arrows should respond to `when <key> clicked` blocks.

### Actual Behavior

* The language menu keeps focus even after clicking away onto the stage. Upon pressing “up” and “down” keys, the language menu is activated rather than being able to interact with the project. @thisandagain 

I verified that this also happens on Chrome (not just safari), so it's a general (accessibility) issue.

### Steps to Reproduce
* Add `when <up-arrow/down-arrow> pressed` actions to your project.
* Click on the language selector to bring up the list of languages
* Click back on the stage (without selecting a language)
* Pressing up or down arrow reactivates the language selection list instead of the `when ... pressed` block.

chrisgarrity:
### Resolves

- Resolves #3302 

### Proposed Changes
Blur the language select when the user clicks outside the select so that arrow keys go to the stage instead of the select.

### Reason for Changes

Language-select was keeping keyboard focus for arrow keys when it shouldn't

### Test Coverage

Current tests run 

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [x] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/language-selector/language-selector.jsx
+++ b/src/components/language-selector/language-selector.jsx
@@ -7,10 +7,11 @@ import styles from './language-selector.css';
 // supported languages to exclude from the menu, but allow as a URL option
 const ignore = [];
 
-const LanguageSelector = ({currentLocale, label, onChange}) => (
+const LanguageSelector = ({componentRef, currentLocale, label, onChange}) => (
     <select
         aria-label={label}
         className={styles.languageSelect}
+        ref={componentRef}
         value={currentLocale}
         onChange={onChange}
     >
@@ -30,6 +31,7 @@ const LanguageSelector = ({currentLocale, label, onChange}) => (
 );
 
 LanguageSelector.propTypes = {
+    componentRef: PropTypes.func,
     currentLocale: PropTypes.string,
     label: PropTypes.string,
     onChange: PropTypes.func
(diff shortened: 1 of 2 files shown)


benjiwheeler:
Hm, I'm still seeing the problem occur:

blurlang_small

benjiwheeler:
Seems like handleMouseLeave is only being called sometimes, and not always... Seems like you need to move your mouse to prompt the blur.

blurdebug2

chrisgarrity:
@benjiwheeler can we take a look at your issue this morning. You may actually be seeing issues because the debugger is open. If you can suggest an alternative way of getting an event when the user moves outside the language selector - I can give it a try. I didn't want to use onMouseUp on `document` (used by the other menu items) because the`select` is always on the page (with opacity 0), and it means that the event fires on every mouse click.

chrisgarrity:
There are edge cases where neither mouseout nor mouseleave fire. If you click outside the select **without** moving the mouse the select keeps focus until you move the mouse. I think we'll need to rely on the fact that people will usually move the mouse. The real solution is to implement a custom language component (https://github.com/LLK/scratch-gui/issues/3595)

PR #3584, opened 2018-10-31
Add HOC tracking pixels

ericrosenbaum:
Add tracking pixels from code.org to the first page of our two Hour of Code tutorials, "Create Animations That Talk" and "Animate an Adventure Game."

We are now hiding the tracking pixel so that we don't see a broken image, because the pixels are still 404ing.

### DIFF
--- a/src/components/cards/card.css
+++ b/src/components/cards/card.css
@@ -267,3 +267,6 @@
     background: $ui-white;
     box-shadow: 0px 0px 0px 2px $ui-black-transparent;
 }
+.tracking-pixel {
+    display: none;
+}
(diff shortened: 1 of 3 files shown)


src/components/cards/card.css
```diff
@@ -267,3 +267,6 @@
     background: $ui-white;
     box-shadow: 0px 0px 0px 2px $ui-black-transparent;
 }
+.tracking-pixel {
+    display: none;
+}
```
rschamp: Since some browsers won't load images if they have `display: none`, maybe we should just remove the `display: none` style, and just not deploy this until the images exist?

PR #3585, opened 2018-10-31
Add sprite target highlighter on sprite-tile click

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-render/issues/127

### Proposed Changes

Adds a new `highlightedTargetId` and `highlightedTargetTime` to the targets reducer, emit events from the target pane (when selecting a sprite) and show the highlight frame overlayed on the stage

sprite-highlight

This also seems like a good place to use the react `key` property to make sure a new dom element is used any time the time changes, so that even if you click repeatedly on a sprite, a new fade-out animation will start.

### DIFF
--- a/src/containers/target-highlight.jsx
+++ b/src/containers/target-highlight.jsx
@@ -1,0 +1,78 @@
+import bindAll from 'lodash.bindall';
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import {connect} from 'react-redux';
+import VM from 'scratch-vm';
+
+class TargetHighlight extends React.Component {
+    constructor (props) {
+        super(props);
+        bindAll(this, [
+            'getPageCoords'
+        ]);
+    }
+
+    // Transform scratch coordinates into page coordinates
+    getPageCoords (x, y) {
+        const nativeSize = this.props.vm.renderer.getNativeSize();
+        const rect = this.props.vm.renderer.canvas.getBoundingClientRect();
+        return [
+            ((rect.width / nativeSize[0]) * x) + (rect.width / 2),
+            -((rect.height / nativeSize[1]) * y) + (rect.height / 2)
+        ];
+    }
+
+    render () {
+        const {
+            className,
+            highlightedTargetId,
+            highlightedTargetTime,
+            vm
+        } = this.props;
+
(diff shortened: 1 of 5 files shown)


paulkaplan:
@fsih can you check this again on staging? It should have been fixed by the z-index fix we made a few days ago.

src/containers/target-highlight.jsx
```diff
@@ -0,0 +1,78 @@
...
+        super(props);
+        bindAll(this, [
+            'getPageCoords'
+        ]);
+    }
+
+    // Transform scratch coordinates into page coordinates
+    getPageCoords (x, y) {
+        const nativeSize = this.props.vm.renderer.getNativeSize();
+        const rect = this.props.vm.renderer.canvas.getBoundingClientRect();
```
rschamp: Since this forces a reflow, could it cause performance issues if it's being called in render? Also, we already do this and cache it in the stage container. To avoid doing this more than we need to, maybe this component should receive it as a prop? https://github.com/LLK/scratch-gui/blob/develop/src/containers/stage.jsx#L139

src/reducers/targets.js
```diff
@@ -23,6 +26,11 @@ const reducer = function (state, action) {
                 .filter(target => target.isStage)[0] || {},
             editingTarget: action.editingTarget
         });
+    case HIGHLIGHT_TARGET:
+        return Object.assign({}, state, {
+            highlightedTargetId: action.targetId,
+            highlightedTargetTime: action.updateTime
```
rschamp: Seems like you could just put `Date.now()` here and not have to know about that detail when you dispatch.

rschamp:
Wasn't sure if there was something else @ericrosenbaum needed to look at so I'm leaving him assigned with needs review.

src/reducers/targets.js
paulkaplan: I think that would break the backwards/forwards-ness of redux though? I can put it in the action creator though, that would be safe.

paulkaplan:
@rschamp fixed to use precomputed stage dimensions, moving timestamp into action creator and an additional fix to make sure the stage doesn't highlight.

paulkaplan:
Oh didn't need anything from @ericrosenbaum, just thought he might get to reviewing first.

Issue #3601, opened 2018-11-01
Sign in dropdown in editor shows translation ids instead of readable text


benjiwheeler:
When using editor while embedded in www, the sign in dropdown appears as:

![screen shot 2018-11-01 at 2 28 17 pm](https://user-images.githubusercontent.com/3431616/47871411-66b0a480-dde2-11e8-9406-d526a988885e.png)

The problem is that we're actually passing jsx from www to gui, including several www-style `<FormattedMessage />`. These message id's aren't defined within gui, so gui does not know their translations, and within gui's scope, www's intl does not apply.

benjiwheeler:
Reopening because this is fixed in the code (and appears correctly for all languages except English), but needs translations to be built with English included (they haven't been until now) for this to appear fixed in English.

benjiwheeler:
Finally confirmed that this works

PR #3602, opened 2018-11-01
define login dropdown messages, so www component will render with readable text

benjiwheeler:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/3601

### Proposed Changes

Define login dropdown messages in gui's `login-dropdown.jsx` file, so they will be translated in the gui context; then, when www passes its login component to gui via the `renderLogin()` function, they will (hopefully!) translate.

### Reason for Changes

Need readable login text, not weird id strings!

### DIFF
--- a/src/components/menu-bar/login-dropdown.jsx
+++ b/src/components/menu-bar/login-dropdown.jsx
@@ -16,22 +16,22 @@ import styles from './login-dropdown.css';
 // these are here as a hack to get them translated, so that equivalent messages will be translated
 // when passed in from www via gui's renderLogin() function
 const LoginDropdownMessages = defineMessages({ // eslint-disable-line no-unused-vars
-    backdrop: {
+    username: {
         defaultMessage: 'Username',
         description: 'Label for login username input',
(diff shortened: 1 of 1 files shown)


src/components/menu-bar/login-dropdown.jsx
```diff
@@ -7,11 +7,38 @@ eventually be consolidated.
...
+import {defineMessages} from 'react-intl';
 
 import MenuBarMenu from './menu-bar-menu.jsx';
 
 import styles from './login-dropdown.css';
 
+// these are here as a hack to get them translated, so that equivalent messages will be translated
+// when passed in from www via gui's renderLogin() function
+const LoginDropdownMessages = defineMessages({ // eslint-disable-line no-unused-vars
+    backdrop: {
```
chrisgarrity: keys should be something related to the message, e.g., `username:`

benjiwheeler: Sorry, totally missed these! Updated!

Issue #3582, opened 2018-10-31
Some png files won't upload


chrisgarrity:
### Expected Behavior
You should be able to upload a png as a new sprite or new costume. A student at my Scratch Club was trying to upload images they had found on the web. Lots were fine, but some just did nothing. They shared one of the files with me and I was able to reproduce the problem.

### Actual Behavior
Uploading some png files doesn't do anything. It doesn't create the sprite, and doesn't give any error feedback. 

In the console:
```
Failed to load resource: the server responded with a status of 404 ()
raven.js:55 vm  TypeError: Cannot read property 'assetId' of null
    at i (lib.min.js:53)
    at lib.min.js:53
a.(anonymous function) @ raven.js:55
lib.min.js:75 Uncaught (in promise) TypeError: Cannot read property 'name' of undefined
    at e.value (lib.min.js:75)
    at e.set (lib.min.js:75)
    at lib.min.js:133
beta.scratch.mit.edu/:1 Uncaught (in promise) Sprite Upload Error: TypeError: Cannot read property 'bitmapResolution' of undefined
```
### Steps to Reproduce
Try uploading this png as a sprite:
![lion_new_pose_normal_eyes](https://user-images.githubusercontent.com/399209/47795161-63d98500-dcf8-11e8-91e0-7ae47ea8a66f.png)

If uploading/downloading from github has somehow 'fixed' the file, I have the original shared with me.

### Operating System and Browser

Seen on HP Chromebook and MacOS (both Chrome)

SimaKyr:
what resolution original image?

julescubtree:
Hi all,

I've actually wondered if uploading was recently broken. I haven't done much testing, but the only graphics I've been able to upload have been graphics that I've downloaded first from the asset server.

It's definitely not just PNG files; [this SVG won't upload either](https://upload.wikimedia.org/wikipedia/commons/0/02/SVG_logo.svg).

After failed sprite uploads, using the sprite selector to choose and add a sprite results in blank sprites (named after the attempted uploads) showing up. Selecting those blank sprites crashes the GUI.

Please let me know how I can help test things.

SimaKyr:
@julescubtree TypeError: Cannot read property '**** bitmapResolution**** ' of undefined
He try upload png, svg use vector, but console write in error bitmap

fsih:
Note that it's only if you upload it as a new sprite, not as a costume

robertsonadams:
[...]

PR #3603, opened 2018-11-01
Update to VM hotfix version

rschamp:
Fix uploading images as sprites.

This version is a custom build of VM, which has only this change applied to the currently deployed version https://github.com/LLK/scratch-vm/pull/1719

This version should _not_ be pulled into `develop` or `smoke`. Instead we should merge https://github.com/LLK/scratch-vm/pull/1719 and let VM publish a new version on `latest`, and update to that version on `develop`/`smoke`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     "scratch-render": "0.1.0-prerelease.20181024220305",
     "scratch-storage": "1.1.0",
     "scratch-svg-renderer": "0.2.0-prerelease.20181024192149",
-    "scratch-vm": "0.2.0-prerelease.20181025092837",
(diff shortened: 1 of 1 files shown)


rschamp:
Try it out: [hotfix-sprite-upload](https://llk.github.io/scratch-gui/hotfix-sprite-upload)

kchadha:
This PR seems to fix the sprite upload issue. There are still some issues that remain, e.g. uploading a sprite from an image that has the wrong file extension (e.g. a png that got renamed to an svg) and #3608.

However, the first is an issue that is also on develop and this PR fixes the issue where we cannot upload a sprite from an image at all #3582

PR #3614, opened 2018-11-02
Allow code to be dragged directly to sprite tiles

paulkaplan:
Allow the sprite selector items and the stage selector to receive drops from BACKPACK_CODE type draggables. Respond to the drops by adding the code to the target.

I also refactored and commented more on the manual fetching required for sprite arraybuffers and code json. I'm sure we will eventually want to add these types to storage, but not now :)

sharing-code-to-sprites

### DIFF
--- a/src/containers/target-pane.jsx
+++ b/src/containers/target-pane.jsx
@@ -189,7 +189,7 @@ class TargetPane extends React.Component {
                 }, targetId);
             } else if (dragInfo.dragType === DragConstants.BACKPACK_SOUND) {
                 this.props.vm.addSound({
-                    md5: dragInfo.payload.bodiy,
+                    md5: dragInfo.payload.body,
                     name: dragInfo.payload.name
                 }, targetId);
             } else if (dragInfo.dragType === DragConstants.BACKPACK_CODE) {
--- a/src/containers/target-pane.jsx
+++ b/src/containers/target-pane.jsx
@@ -192,6 +192,13 @@ class TargetPane extends React.Component {
                     md5: dragInfo.payload.body,
                     name: dragInfo.payload.name
                 }, targetId);
+            } else if (dragInfo.dragType === DragConstants.BACKPACK_CODE) {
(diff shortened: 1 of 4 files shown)


src/containers/target-pane.jsx
```diff
@@ -189,9 +189,15 @@ class TargetPane extends React.Component {
                 }, targetId);
             } else if (dragInfo.dragType === DragConstants.BACKPACK_SOUND) {
                 this.props.vm.addSound({
-                    md5: dragInfo.payload.body,
+                    md5: dragInfo.payload.bodiy,
```
rschamp: This looks like a typo... or maybe just a stray keypress since I don't think you were trying to do anything here?

src/lib/backpack-api.js
```diff
@@ -73,12 +73,30 @@ const deleteBackpackObject = ({
     });
 });
 
+// Two types of backpack items are not retreivable through storage
+// code, as json and sprite3 as arraybuffer zips.
+const fetchAs = responseType => uri => new Promise((resolve, reject) => {
```
rschamp: Maybe I just need to get used to it if other people have a different feeling about it, but this is too many arrows for me to understand what's going on the first time I read it.  I think it crosses the line of syntax convenience that arrow functions provide and starts treading into confusion, making parsing it a burden again.

I think some line breaks and indentation would help readability, if not using a full-on `function () {}` for the outer function.

paulkaplan: Yeah, honestly this formatting is just a product of the lint auto fixer. I'll fix.

src/containers/target-pane.jsx
rschamp: This still looks like a typo... maybe the GitHub comment-minifier has worked against us already.

paulkaplan:
@rschamp agghhhh sorry you are right it got minimized and I missed it. Fixed. Sorry about that.

PR #3620, opened 2018-11-02
correct wrong assetId prop name

benjiwheeler:
Looks like https://github.com/LLK/scratch-gui/pull/3491 should have changed this old prop name, `assetId`, to `asset`


benjiwheeler:
Incidentally, this is running in local `npm test`:

screen shot 2018-11-02 at 2 08 42 pm

...but not in Travis:

screen shot 2018-11-02 at 2 09 15 pm

PR #3621, opened 2018-11-02
Strip nbsp to fix serializing svg on safari

paulkaplan:
Fixes the issue @thisandagain saw with safari not serializing blocks sometimes, caused by html entities not serializing properly. 

Also made errors in image loading propagate back through the promise chain so it doesn't get stuck on "loading"

### DIFF
--- a/src/containers/backpack.jsx
+++ b/src/containers/backpack.jsx
@@ -114,8 +114,9 @@ class Backpack extends React.Component {
                         contents: [item].concat(this.state.contents)
                     });
                 })
-                .catch(() => {
+                .catch(error => {
                     this.setState({error: true, loading: false});
+                    throw error;
                 });
         });
     }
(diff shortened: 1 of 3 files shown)


src/containers/backpack.jsx
```diff
@@ -107,7 +107,10 @@ class Backpack extends React.Component {
                     username: this.props.username,
                     ...payload
                 }))
-                .then(this.refreshContents);
+                .then(this.refreshContents)
```
rschamp: `refreshContents` doesn't appear to return a Promise, so I'm confused how `catch()` can catch the rejection.

src/containers/backpack.jsx
```diff
@@ -107,7 +107,10 @@ class Backpack extends React.Component {
...
+                .catch(() => {
```
rschamp: It doesn't look like we're logging any of these errors. We probably should, so if/when things start going wrong in the wild, we'll be able to debug remotely.

src/containers/backpack.jsx
paulkaplan: Oh this error in particular is thrown in the payloader, not refreshContents. This stuff changed with the pagination PR though, so I'm going to rebase and push. None of the error stuff will be in this PR

PR #3622, opened 2018-11-02
Add pointer-events:none to sprite img to prevent dragging.

paulkaplan:
Fixes an issue @kathymakes saw where dragging sprite tiles by clicking on the image itself does not work well in firefox, because despite draggable=false on the image, it still drags :( boo firefox. 

@picklesrus I also think this could fix the angle picker on firefox problem


benjiwheeler:
Any existing issue for this?

benjiwheeler:
Here's a gif of this difference in action, FYI!

drag

Issue #3595, opened 2018-11-01
Replace language select with custom component


chrisgarrity:
The language menu is currently using an HTML `<select>...</select>`. Selects behave differently in different browsers. 

To remove inconsistencies we should add a custom component: probably a scrollable menu.

PR #3626, opened 2018-11-02
Revert "Blur language select focus when mouse leaves"

### LINKED ISSUE #3613: Language selecter doesn't keep open on Firefox
### Expected Behavior
It should work!

### Actual Behavior
It doesn't keep open. When I click the glove, the menu appears, and soon disappears.

### Steps to Reproduce
Change language on Firefox.

Windows 7 HomePremium 64bit SP1
Firefox 63

chrisgarrity:
Reverts LLK/scratch-gui#3583
Fixes #3613 

On firefox the mouse is considered to have "left" the select element as soon as it is no longer over the select menubar item. It doesn't treat the options as being part of the select. So this change completely breaks the language selector on Firefox.

Having the language selector keep keyboard focus if you don't select a language (you have to click twice outside the menu to reset focus) seems better than not working at all on Firefox. The permanent solution is https://github.com/LLK/scratch-gui/issues/3595

### DIFF
--- a/src/components/language-selector/language-selector.jsx
+++ b/src/components/language-selector/language-selector.jsx
@@ -7,11 +7,10 @@ import styles from './language-selector.css';
 // supported languages to exclude from the menu, but allow as a URL option
 const ignore = [];
 
-const LanguageSelector = ({componentRef, currentLocale, label, onChange}) => (
+const LanguageSelector = ({currentLocale, label, onChange}) => (
     <select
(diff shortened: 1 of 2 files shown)


rschamp:
Try it out: [revert-3583-issue/3302-language-focus](https://llk.github.io/scratch-gui/revert-3583-issue/3302-language-focus)

Issue #3449, opened 2018-10-22
Watermark appears on top of blocks 


SpeakkVisually:
The watermark of the sprite should appear underneath the block in the scripts area. 

![image](https://user-images.githubusercontent.com/5166432/47307299-b3cf9200-d5fc-11e8-94a6-dc88ad634be5.png)

_e.g. Mac OS 10.11.6 Safari 10.0_

carljbowman:
@SpeakkVisually - Putting the sprite watermark above the blocks is a compromise that we needed to make in the near term.

Surprisingly putting the watermark below the blocks would take a tremendous amount of work. We need to create a system pass, render, and update sprite information within the seperate scratch-blocks codebase. At this point, we do not have time to fully implement that system.

PR #3629, opened 2018-11-05
hide watermark and any other overflow from tab panel

benjiwheeler:
Tiny fix to a problem I noticed, when the screen height is very low:

Before:
screen shot 2018-11-04 at 9 22 36 pm

After:
screen shot 2018-11-04 at 9 22 42 pm

FYI, somewhat related: https://github.com/LLK/scratch-gui/issues/3449

### DIFF
--- a/src/components/gui/gui.css
+++ b/src/components/gui/gui.css
@@ -172,6 +172,7 @@
     position: relative;
     flex-grow: 1;
     display: none;
+    overflow: hidden;
 }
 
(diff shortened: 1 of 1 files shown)


benjiwheeler:
> This fix prevents blocks from being visible when dragged outside the workspace, which is necessary for sharing blocks to other sprites and the backpack.

Ooooohhhhh. Yikes. Hm, maybe this is just a bug we let lie...

PR #3630, opened 2018-11-05
don't require onActivatedBlocksTab prop in TargetPane container

benjiwheeler:
This gets rid of a React console warning introduced by https://github.com/LLK/scratch-gui/pull/3596 .

Note that the proptypes inheritance that was put in place in `containers/target-pane.jsx` two years ago -- where the container duplicates its presentation component's proptypes -- is very confusing to me. I suspect this approach makes mistakes inevitable, and it seems to me to already include a dozen or two unused proptypes declarations.

### DIFF
--- a/src/containers/target-pane.jsx
+++ b/src/containers/target-pane.jsx
@@ -231,6 +231,7 @@ class TargetPane extends React.Component {
 
 const {
     onSelectSprite, // eslint-disable-line no-unused-vars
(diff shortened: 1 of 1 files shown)


paulkaplan:
@benjiwheeler I agree. I think that as that particular component has grown over time, the approach has gotten worse and worse.

rschamp:
+1, the decision to import and use component props in the containers was a mistake. Each component should only include the props that it directly uses.

PR #3638, opened 2018-11-05
automatically save on navigating from page; hide share and community buttons while creating

benjiwheeler:
This mostly resolves https://github.com/LLK/scratch-www/issues/2278

This change makes it so that:
* clicking the Share button triggers a save; once the save completes, user is navigated to the project page
* clicking the See Community button triggers a save; once the save completes, user is navigated to the project page
* navigating away from your project in any other way triggers a save (though this may not complete; it's unclear how effective this is); and
* Share and Community buttons are hidden while project fetching, project loading, and project creation happen (they are shown when the project is showing normally, or when saving)

The biggest outstanding problems are:
* navigating away can't be totally stopped until the save request is sent. So clicking "My Stuff" or the Scratch logo may not send the save request in time before javascript on the page is completely stopped.
* when you do a more complex action that involves saving/creation, such as creating a new project, saving as a copy, or remixing, and you navigate away before that operation has finished (not so unrealistic, on slow connections), you can get into weird states. I somewhat avoid these by hiding the Share and Community buttons while these operations are happening. 

(Of course, there are other ways to navigate away! But since Share and Community will show you the Project Page, these are the ones that can get you into an unexpected state with respect to what's loaded in the vm.)

I wrote up the outstanding problems at: https://github.com/LLK/scratch-gui/issues/3639

### DIFF
--- a/src/components/menu-bar/community-button.css
+++ b/src/components/menu-bar/community-button.css
@@ -1,0 +1,9 @@
+@import "../../css/colors.css";
+
+.community-button {
+    box-shadow: 0 0 0 1px $ui-black-transparent;
+}
+
+.community-button-icon {
+    height: 1.25rem;
+}
--- a/src/components/menu-bar/community-button.jsx
+++ b/src/components/menu-bar/community-button.jsx
@@ -1,0 +1,40 @@
+import classNames from 'classnames';
+import {FormattedMessage} from 'react-intl';
+import PropTypes from 'prop-types';
+import React from 'react';
+import Button from '../button/button.jsx';
+
+import communityIcon from './icon--see-community.svg';
+import styles from './community-button.css';
+
+const CommunityButton = ({
+    className,
+    onClick
+}) => (
+    <Button
+        className={classNames(
+            className,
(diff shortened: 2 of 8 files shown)


paulkaplan:
@benjiwheeler you can't prevent navigation or do anything before it happens, the only thing you can do is warn the user that they have unsaved changes. This will probably require the dirty checking from @rschamp, but you can see how to do that in the  playground/render-gui.jsx file (basically you set `window.onbeforeunload` to a function that returns true when you want to warn the user, and false when navigation is safe.

src/containers/gui.jsx
```diff
@@ -118,7 +125,6 @@ GUI.propTypes = {
     onSeeCommunity: PropTypes.func,
     onStorageInit: PropTypes.func,
     onUpdateProjectId: PropTypes.func,
-    onUpdateProjectTitle: PropTypes.func,
```
benjiwheeler: removing cruft here

rschamp:
+1 to holding off on onbeforeunload until we have dirty state.

PR #3651, opened 2018-11-05
removed save and save copy from beta; better menu section style handling

### LINKED ISSUE #3617: Swap Download and Upload above Save
### Current Behavior
The _Upload from your Computer_ and _Download to your computer_ are at the bottom of the File menu, below the currently non-functioning _Save Now_ and _Save as a copy_ options.

### Desired Behavior
We'd like to move the Upload and Download options above the Save options to increase their visibility unto the Save options are functional.

upload_download

benjiwheeler:
### Resolves

- Resolves #3617 

### Proposed Changes

* removes save and save copy from beta
* better menu section style handling, so that there is still a menu divider

### Reason for Changes

A disabled Save and Save copy make it seem like you can't save, and make users not realize that upload and download are how to "save"

### DIFF
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -337,37 +337,25 @@ class MenuBar extends React.Component {
                                 place={this.props.isRtl ? 'left' : 'right'}
                                 onRequestClose={this.props.onRequestCloseFile}
                             >
-                                <MenuItem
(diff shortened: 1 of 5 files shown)


thisandagain:
@benjiwheeler The logic and test coverage seems reasonable to me. Assigning to @benjiwheeler to take a closer look at the code.

PR #3652, opened 2018-11-05
Remove code that was moved into the VM

paulkaplan:
The variable-utils file was slated to be removed pending adding those methods to scratch-vm. That was done, and now having these separate functions will make it harder to fix issues like the slider monitors not updating global vars.

Better late than never I guess (https://github.com/llk/scratch-vm/issues/1145 was merged in may...)

Hooray removing dead code!


kchadha:
@paulkaplan, why was this PR closed?

paulkaplan:
@kchadha it was only a minor issue: the `getVariableValue` util in GUI is handling which sprite to look for the variable on or whether to look at the stage, whereas the code in the VM doesn't do that. It just means the vm wasn't a drop-in replacement, and a bit of extra logic about which target to look up the variable ID needs to be added into the GUI before calling getVariableValue. It also looks like the GUI util was automatically creating a new array reference, so we just need to add that into the list operations code (most look like they already produce new arrays, so might not be an issue). 

Mostly I just closed it because it didn't work off-the-bat and we didn't need it until we had https://github.com/LLK/scratch-vm/pull/1823 anyway, but now that that is in I can pick this back up

chrisgarrity:
@kchadha I'm reassigning this to you, can you decide if this should just get closed, or be reviewed. Reassign back to me if it should be reviewed.

PR #3655, opened 2018-11-05
added simple spinner component

benjiwheeler:
Implementing spec from https://github.com/LLK/scratch-www/issues/2264

Not actually using this yet, but want to have it in place for alerts.

### DIFF
--- a/src/components/spinner/spinner.jsx
+++ b/src/components/spinner/spinner.jsx
@@ -1,0 +1,26 @@
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import styles from './spinner.css';
+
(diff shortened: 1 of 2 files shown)


src/components/spinner/spinner.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+const SpinnerComponent = function (props) {
+    const {
+        className
+    } = props;
+    return (
+        <div
+            className={classNames(
+                styles.spinner,
+                className
```
fsih: It might be simpler to use props.className instead of deconstructing the object here

benjiwheeler: I didn't understand this before I merged, but @fsih explained it to me and now I do :)

Issue #3608, opened 2018-11-01
SVGs uploaded as costume v. sprite comes in at different sizes


BryceLTaylor:
### Expected Behavior

If you import an image as a sprite it should be the same size as it would if you imported it as a costume.

### Actual Behavior

The image comes in much larger as a costume than it does as a sprite.
![itsfine](https://user-images.githubusercontent.com/1088291/47881462-94a2e280-ddfc-11e8-9d2d-39dff0ff650a.gif)

### Steps to Reproduce

Hover over the new sprite thermometer
Choose upload from file
choose an svg
Note: the new sprite is tiny on the stage
Open the costume editor
choose upload a new costume
choose the same svg
Note: the sprite now appears bigger on the stage, but the new costume is the same size in the paint editor.
select the first costume
edit it slightly
Note: the sprite gets larger on the stage
edit the second costume:
note: it stays the same size

### Operating System and Browser

Mac Chrome

kchadha:
@paulkaplan pointed out that this is probably a bitmap resolution issue. The small version of the svg is about half the size that it is supposed to be.

cc/ @fsih, @cwillisf, @rschamp

thisandagain:
Moved to LLK/scratch-paint#777

kchadha:
@thisandagain, I don't think this is a scratch-paint issue. The phrasing of this issue was a little bit confusing about which sprite is the problem sprite. When uploading the default cat sprite costume-1 svg as a sprite, it comes in really tiny on the stage, whereas it is the correct size in the paint editor.

It seems that this problem is now worse than before where the sprite is now coming in at 1/4 the size it should be on the stage, the costume tile reads half width and height, and the sprite selection highlight seems to correspond w/the half height / width issue.

fsih:
@BryceLTaylor can you test if you can repro? I think this might have been fixed by https://github.com/LLK/scratch-svg-renderer/pull/60

kchadha:
@fsih I can still repro on develop, though it's certainly not as bad as it was on beta (which is what I was testing in my previous comment):

![small-cat](https://user-images.githubusercontent.com/1786240/49169104-0fc9bc80-f307-11e8-8057-5af650ecb436.gif)

PR #3667, opened 2018-11-06
Add tests for uploading sprite and costume by file

paulkaplan:
@kchadha @rschamp this adds tests that fail when I ran them without the vm fix for uploading sprite from file image. It also includes a skipped test that is failing right now due to https://github.com/LLK/scratch-gui/issues/3608, but when that is fixed it should be enabled as a regression test.

To implement these, I just poked around a bit for how to attach local files to file inputs using selenium. Turns out you just "el.sendKeys" the path to the file... who would have guessed that?


test/integration/sprites.test.js
```diff
@@ -74,4 +74,39 @@ describe('Working with sprites', () => {
...
+            .perform();
+        await driver.sleep(500); // Wait for thermometer menu to come up
+        const input = await findByXpath('//input[@type="file"]');
+        await input.sendKeys(path.join(process.cwd(), 'test', 'fixtures', '100-100.svg'));
+        await clickText('100-100', scope.spriteTile); // Sprite is named for costume filename
+
+        // Check to make sure the size is right
+        await clickText('Costumes');
+        await clickText('100-100-costume1', scope.costumesTab); // The name of the costume
+        await clickText('100 x 100', scope.costumesTab); // The size of the costume
```
paulkaplan: this is the line that would fail. Right now it says 50 x 50

paulkaplan:
Thanks for the tip on the problem. I changed to use `path.resolve(__dirname, ...` the same way the build index.html is loaded at the top of the file. Rebuilding, hope that works.

Issue #3597, opened 2018-11-01
implement project creator info in menu bar


benjiwheeler:
Expected:

![screen shot 2018-11-01 at 11 19 28 am](https://user-images.githubusercontent.com/3431616/47860916-09a7f500-ddc8-11e8-9e0a-7f04c59dcc71.png)

Actual:
![screen shot 2018-11-01 at 11 19 59 am](https://user-images.githubusercontent.com/3431616/47860957-1af10180-ddc8-11e8-8cf4-6853fc2e25e9.png)

* show project creator avatar
* show project creator username
* show project name
* hide project title input

PR #3681, opened 2018-11-07
added author-info, user-avatar components

benjiwheeler:
(Needs https://github.com/LLK/scratch-www/pull/2289 before it will work)

### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/3597

### Proposed Changes

uses author id and author username to create a display of info about the project's creator:

screen shot 2018-11-07 at 1 15 25 pm

### Reason for Changes

Display info on the creator! And, have a place to show the title.


src/components/menu-bar/author-info.jsx
```diff
@@ -0,0 +1,60 @@
...
+        )}
+    >
+        <div
+            className={classNames(
+                styles.authorInfo
+            )}
+        >
+            <UserAvatar
+                className={styles.avatar}
+                userId={userId}
```
rschamp: I think we should be passing a full url as `src` from the outside rather than constructing the url from a `userId` here. The details of the path, and even the host are outside the responsibility of GUI. It's possible eventually an environment (offline?) would pass in a dataURI, for instance.  And in the immediate, we would want to use a different thumbnail hosts in local development, on staging, and in production.

src/components/menu-bar/menu-bar.jsx
```diff
@@ -468,17 +469,26 @@ class MenuBar extends React.Component {
...
-                            enable
-                            id="title-field"
-                        >
-                            <ProjectTitleInput
-                                className={classNames(styles.titleFieldGrowable)}
-                                onUpdateProjectTitle={this.props.onUpdateProjectTitle}
-                            />
-                        </MenuBarItemTooltip>
-                    </div>
+                    {(this.props.authorUsername && this.props.authorUsername !== this.props.username) ? (
```
rschamp: It seems like the toggle for this type of display would be whether you should be able to edit the title or not, since in addition to showing the author information, it changes the title field from an editable one to static output.  I'm not seeing any capability flag that already represents that though.

src/components/menu-bar/author-info.css
```diff
@@ -0,0 +1,51 @@
...
+    margin-right: .5625rem;
+}
+
+.author-info .title-author {
+    /* display: flex;
+    align-items: center;
+    justify-content: end; */
+}
+
+.author-info .project-title {
```
rschamp: The beauty of css modules is that it creates localized class names, so you don't have to do this prefixing for specificity's sake. In fact putting this prefixing here makes it more brittle since it couples the style to the element hierarchy unnecessarily. Nothing about a `.project-title` being a child of `.author-info` matters for the font-size, line-height, etc.

src/components/menu-bar/author-info.css
```diff
@@ -0,0 +1,51 @@
...
+    line-height: 1.0625rem;
+    font-weight: bold;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+}
+
+.author-info .username-line {
+    max-width: 12rem;
+    font-size: .75rem;
+    line-height: .9375rem;
```
rschamp: Would any of these numbers make sense to move into `units` so they can be reused?

src/lib/user-thumbnail.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @user-thumbnail
+ * Utility functions to return thumnail-related strings
+ */
+
+/**
+ * Generate a new empty sprite. The caller should provide localized versions of the default names.
```
rschamp: This looks copy-pasted from somewhere else.  But like I said above, this function probably doesn't belong in GUI since it's a detail of www.

src/components/menu-bar/user-avatar.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+import styles from './user-avatar.css';
+
+const UserAvatar = ({
+    className,
+    size,
+    src,
+    userId
+}) => (
+    <React.Fragment>
```
rschamp: This Fragment looks left-over from a previous iteration.

src/components/menu-bar/author-info.jsx
```diff
@@ -0,0 +1,60 @@
...
+    projectTitle,
+    userId,
+    username
+}) => (
+    <div
+        className={classNames(
+            className
+        )}
+    >
+        <div
```
rschamp: The outer wrapper and this one look like they could be combined.

src/components/menu-bar/author-info.jsx
rschamp: Do we want this to link to the author's profile? @carljbowman?

If so, we should accomplish it with an `onAuthorClick` (or something) prop, since we'd want that to be handled differently in the desktop editor.

src/components/menu-bar/author-info.jsx
```diff
@@ -0,0 +1,60 @@
...
+                            <FormattedMessage
+                                defaultMessage="by"
+                                description="Shows that a project was created by this user"
+                                id="gui.authorInfo.by"
+                            />
+                            &nbsp;
+                        </span>
+                        <span className={styles.username}>
+                            {username}
+                        </span>
```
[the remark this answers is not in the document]
benjiwheeler: Looks like there are lots of examples of this, eg components/import-modal/import-modal.jsx:

```
                    <FormattedMessage
                        defaultMessage="To learn more, go to the {previewFaqLink}."
                        description="Invitation to try 3.0 preview"
                        id="gui.importInfo.previewfaq"
                        values={{
                            previewFaqLink: (
                                <a
                                    className={styles.faqLink}
                                    href="//scratch.mit.edu/3faq"
                                >
                                    <FormattedMessage
                                        defaultMessage="FAQ"
                                        description="link to Scratch 3.0 FAQ page"
                                        id="gui.importInfo.previewfaqlinktext"
                                    />
                                </a>
                            )
                        }}
                    />
```

src/components/menu-bar/menu-bar.jsx
benjiwheeler: good point, i felt a little awkwardness here with GUI having the responsibility to interpret that it _should_ show author info instead of title. Maybe I should create a `canEditTitle` flag from www?

src/components/menu-bar/author-info.css
benjiwheeler: That makes sense

src/components/menu-bar/author-info.css
benjiwheeler: Hm, not sure. I'll take a look at other menu bar css and see if there are repeated values

src/components/menu-bar/menu-bar.jsx
rschamp: That makes sense. We still want to be able edit the title in the beta — is that accounted for here?

benjiwheeler: In my current code, the title stays editable in the beta because authorUsername is not set. But if I introduce a canEditTitle flag, I'd have www set it, and the playground code set it too.

src/components/menu-bar/author-info.jsx
benjiwheeler: Could also circle back to this.

src/lib/user-thumbnail.js
benjiwheeler: Oops, should have changed that line. For now, I moved this logic to www, though I hear you that it makes more sense for it to ultimately come from API.

src/components/menu-bar/user-avatar.jsx
benjiwheeler: You're right, it's not necessary now.

src/components/menu-bar/author-info.jsx
benjiwheeler: You're right. I had split them apart when trying to wrest control of the flex width from CSS, but they can be back together now.

src/components/menu-bar/author-info.css
benjiwheeler: Done, consolidated a bunch of them.

src/lib/user-thumbnail.js
benjiwheeler: Done.

src/components/menu-bar/author-info.jsx
benjiwheeler: Done.

src/components/menu-bar/author-info.css
```diff
@@ -0,0 +1,51 @@
...
+}
+
+.author-info .avatar {
+    margin-right: .5625rem;
+}
+
+.author-info .title-author {
+    /* display: flex;
+    align-items: center;
+    justify-content: end; */
```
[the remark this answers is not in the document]
benjiwheeler: Removed.

src/components/menu-bar/author-info.jsx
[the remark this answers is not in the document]
benjiwheeler: Done.

src/components/gui/gui.jsx
```diff
@@ -323,6 +331,9 @@ const GUIComponent = props => {
 GUIComponent.propTypes = {
     accountNavOpen: PropTypes.bool,
     activeTabIndex: PropTypes.number,
+    authorId: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
```
rschamp: When will these be `bool`s? Why not always an (optional) string? If it's not supplied it would be `undefined` which isn't a `bool`.

benjiwheeler: So, the reason is that www's preview.jsx is doing logic like:

```
    const authorId = authorPresent && author.id && author.id.toString();
```

if authorPresent is `false`, the value of authorId will be `false`, not `undefined`.

I could set this differently -- as a ternary, something like

```
    const authorId = authorPresent && author.id && author.id.toString() ? author.id.toString() : undefined;
```
...but my understanding is that it's an anti-pattern to set variables to `undefined` this way?

src/components/menu-bar/author-info.jsx
```diff
@@ -0,0 +1,55 @@
...
+            imageUrl={imageUrl}
+        />
+        <div className={styles.titleAuthor}>
+            <span className={styles.projectTitle}>
+                {projectTitle}
+            </span>
+            <div>
+                <span className={styles.usernameLine}>
+                    <FormattedMessage
+                        defaultMessage="by {username}"
```
chrisgarrity: 👍 this works and allows translators to change the order it that makes more sense for their language.

src/components/gui/gui.jsx
benjiwheeler: Done

PR #3688, opened 2018-11-08
Fix monitor labels for (current [v year]) block.

kchadha:
### Resolves

Fixes a regression introduced by #3320 where the monitor labels for the `(current [v year])` block wasn't displaying.

### Proposed Changes

The sensing_current block has its field values in all caps. The changes in #3320 were no longer calling toLowerCase on them before matching on it.

### Reason for Changes

The monitor label for that block (with all its various arguments) is currently broken.

### Test Coverage

Tested manually.

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

cc/ @paulkaplan, @picklesrus

### DIFF
--- a/src/lib/opcode-labels.js
+++ b/src/lib/opcode-labels.js
@@ -210,7 +210,7 @@ class OpcodeLabels {
         this._opcodeMap.sensing_loudness.labelFn = () => this._translator(messages.sensing_loudness);
         this._opcodeMap.sensing_username.labelFn = () => this._translator(messages.sensing_username);
         this._opcodeMap.sensing_current.labelFn = params => {
-            switch (params.CURRENTMENU) {
+            switch (params.CURRENTMENU.toLowerCase()) {
             case 'year':
                 return this._translator(messages.sensing_current_year);
             case 'month':


kchadha:
@chrisgarrity, one thing I noticed while going over #3320 to fix this bug was that [the code in there](https://github.com/LLK/scratch-gui/pull/3320/files#diff-aa08002866a8bed138d85c86c0feb157R54) seems to think that `tempo` is still a sound category monitor when in fact it's now part of the music extension. I didn't know what should be done about that and didn't want to add those changes to this PR, but I thought you should know about it.

I linked to one of the sections of the code above, but there are more that can be found in that PR.

chrisgarrity:
@kchadha the tempo labels in here are obsolete and can be removed. In the previous PR, I just tried to localize what existed without thinking about whether anything was out of date. You could either remove the tempo opcode, or open a tech dept issue to deal with later.

kchadha:
@chrisgarrity, I think I'll make a separate tech debt issue as you suggested.

PR #3697, opened 2018-11-08
Fix missing fonts in backpack thumbnails

paulkaplan:
Fixes an issue where the backpack thumbnails did not show with the correct fonts.

The solution is to use the `getCostumeUrl` utility, which takes care of inlining font-faces, which is exactly what is needed for this! Thanks @kchadha @fsih for making that utility!


paulkaplan:
Thanks for the tip, I’ll change that in a follow up

PR #3715, opened 2018-11-09
Extensions Library Phase 2

### LINKED ISSUE #3675: Extensions Library Phase 2
Create new extension library gui elements based on this design:

screen shot 2018-11-07 at 10 37 50 am

evhan55:
### Resolves

- Resolves #3675: Extensions Library Phase 2

### Notes

- Might need follow up PR to make sure translations and LTR are handled correctly for CSS styling?


src/components/library-item/library-item.jsx
```diff
@@ -70,13 +73,68 @@ class LibraryItem extends React.PureComponent {
...
+                                        />
+                                    </div>
+                                </div>
+                            ) : null}
+                        </div>
+                        <div className={styles.featuredExtensionCollaboration}>
+                            {this.props.collaborator ? (
+                                <div>
+                                    <div>
+                                        <FormattedMessage
```
apple502j: Why don't you use {} formatting and "values"?

evhan55: oooh, what do you mean by that?  jsx and React are new to me!  Thanks!

apple502j:
Why don't you add Camera to Video Sensing?

src/components/library-item/library-item.jsx
joker314: It's called a _formatted_ message because you can substitute values in. You can then pass in all the named substitutions using a `values` prop.

For example, `"Hello, {name}"` and then a `values` prop of `{name: "Anna"}`

evhan55: Oh, whoa! Thanks so much, I will try it!

joker314: Docs: https://github.com/yahoo/react-intl/wiki/Components#formattedmessage
Note: When passing an object literal as a prop, you need to double up the curly braces

src/components/library-item/library-item.css
```diff
@@ -3,8 +3,9 @@
 
 .library-item {
     display: flex;
+    align-self: stretch;
     flex-direction: column;
-    align-items: center;
+    align-items: left;
```
paulkaplan: This causes the text on for example the sound library to be left aligned:
image

src/components/library-item/library-item.css
```diff
@@ -91,12 +111,49 @@
...
+
 .featured-description {
     display: block;
     font-weight: normal;
     line-height: 2;
 }
 
+.featured-extension-metadata {
+    width: 100%;
+    padding: 0rem 1.25rem 1rem 1.25rem;
```
paulkaplan: Nit: customarily "0" is written with no units in CSS

src/components/library-item/library-item.jsx
paulkaplan: @evhan55 I don't think this would be appropriate here, since we (a) will not be translating the names of the collaborators and (b) are putting the collaborator name in a distinct place, we cannot have the contributor rearranged with the other text.

src/components/library-item/library-item.css
```diff
@@ -48,11 +49,30 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    margin-top: -2.5rem;
+    margin-left: 1rem;
+    align-self: flex-start;
+    background-color: rgb(35, 188, 141);
+    width: 3.5rem;
+    height: 3.5rem;
+    z-index: 10;
```
paulkaplan: We've been trying avoid z-indices because they can complicate interactions with other parts of the UI. In this case, you don't need a z-index because the inset image comes after the featured image, but because the featured image is `position: relative` it still displays over. So you can replace this z-index with `position: relative` to get things to layer correctly without using z-index. The reason is that element ordering without z-index goes:
1. All non-positioned elements, in order of appearance. 
2. All positioned elements, in order of appearance. 
[src](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/Stacking_without_z-index)

src/components/library-item/library-item.css
```diff
@@ -48,11 +49,30 @@
...
 }
 
+.library-item-inset-image-container {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    margin-top: -2.5rem;
+    margin-left: 1rem;
+    align-self: flex-start;
+    background-color: rgb(35, 188, 141);
```
paulkaplan: I think this color is already defined as `$pen-primary`

src/components/library-item/library-item.css
evhan55: Oh got it, will try to address with CSS.

src/components/library-item/library-item.css
evhan55: Okay

src/components/library-item/library-item.css
evhan55: Thanks!

src/components/library-item/library-item.css
evhan55: Cool thanks!

src/components/library-item/library-item.jsx
evhan55: @paulkaplan Actually this line here is for translating the label 'Collaboration with', I think that should be ok, right?  We will not be translating the collaborator names.

joker314: @evhan55, you're right, your change is a good thing... it's just that the thing @apple502 and I mentioned (using curly braces and substitution) doesn't work here because we are putting the collaborator names in a separate tag.

Ideally, we'd use substitution, because in some languages the word order might change. With a long list like this though, I'm sure there's a way of phrasing it that would allow for this word order to be maintained

evhan55: @joker314 Oh, I'm learning all the things!  Thank you for explaining.

evhan55:
I've made an issue about adding RTL to this new GUI layout:
https://github.com/LLK/scratch-gui/issues/3738

evhan55:
@paulkaplan thanks! merging seems blocked on my end: "Merging can be performed automatically once the requested changes are addressed."

Issue #3723, opened 2018-11-12
Handle project upload errors


rschamp:
When we catch errors after attempting to store project or asset data, we should let the user know and attempt to recover if possible.

Scratch 2.0 retries, backing off exponentially if it fails to save. Scratch 2.0 also assumes that the user is logged out when it fails to save sometimes. We should avoid this behavior in 3.0.

This has a few forms:
- [ ] Network timeouts (#3762)
- [ ] The server is returning errors (#3764)
- [ ] User has become logged out (#3764)

But in all cases, we want to prompt the user to download their project ASAP:
- [ ] Modal with project download button (#3761)

PR #3724, opened 2018-11-12
Save 3.0 project JSON to the project server and assets to the assets server

rschamp:
_Depends on https://github.com/LLK/scratch-vm/pull/1750. Tests will fail until that PR is merged and this PR is bumped to the new version of the VM._

Known issues to be addressed in a follow-up:
- Handle failed asset and project json uploads (#3723)
- Auto-saving (#3181) 
- Dirty- (or clean-)state tracking of the project JSON

### Resolves

- Resolves https://github.com/LLK/scratch-vm/issues/1578

### Proposed Changes

Stop saving .sb3 files to the project server, and begin saving assets to the asset server and a project JSON file to the project server.

After assets are saved, mark them as "clean", and only save assets without this "clean" flag. This causes assets not to be saved multiple times unnecessarily.

Also, re-save the default costumes for the cat sprite. Because of line ending settings in Git, these files had diverged from their md5. To avoid this issue going forward, I set .gitattributes to consider all .svg files as binary, so git wouldn't worry about line endings in them.

### Reason for Changes

We want to save flat JSON files rather than zip files to the project server, and dedupe assets on the asset server. This also moves us closer to being able to track the project dirty state.

### Test Coverage

Existing tests will pass, pending bumping the scratch-vm dependency above.

### Browser Coverage

Mac
 * [X] Chrome 
 * [ ] Firefox 
 * [X] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


src/lib/project-saver-hoc.jsx
```diff
@@ -118,25 +119,36 @@ const ProjectSaverHOC = function (WrappedComponent) {
...
-                    // when id is undefined or null, storage.store as we have
-                    // configured it will create a new project with id
-                    return storage.store(
-                        assetType,
-                        dataFormat,
-                        body,
-                        projectId
-                    );
+            requestParams = requestParams || {};
+            return Promise.all(this.props.vm.assets
```
benjiwheeler: Does this assets saving make sense to do in the case where we don't own the current project loaded in the vm? Seems like we're putting the cart before the horse here. I suspect we only want to do this after we've successfully saved the project file, or if we know we own the project.

src/lib/project-saver-hoc.jsx
```diff
@@ -118,25 +119,36 @@ const ProjectSaverHOC = function (WrappedComponent) {
...
+                    ).then(
+                        () => (asset.clean = true)
+                    )
+                )
+            ).then(() => {
+                const body = new FormData();
+                const sb3Json = new Blob([this.props.vm.toJSON()], {type: 'application/json'});
+                body.append('sb3_file', sb3Json, 'sb3_file');
+                for (const key in requestParams) {
+                    if (requestParams.hasOwnProperty(key)) body.append(key, requestParams[key]);
```
benjiwheeler: I vaguely understand that using hasOwnProperty like this is the recommended pattern. But I don't understand why. There are no other properties that actually get iterated in plain old `for (const key in requestParams) {...}`, right? Is there an example of code I can run where I can see this problem occur?

src/lib/project-saver-hoc.jsx
```diff
@@ -118,25 +119,36 @@ const ProjectSaverHOC = function (WrappedComponent) {
...
+                }
+                return storage.store(
+                    storage.AssetType.Project,
+                    storage.DataFormat.JSON,
+                    body,
+                    projectId
+                );
+            })
+                .catch(err => {
```
benjiwheeler: I'm not sure, but I think it's a mistake to intercept this error here, rather than pass it up to the catch() calls that encase the calls to storeProject().

src/lib/storage.js
```diff
@@ -51,6 +56,17 @@ class Storage extends ScratchStorage {
     getAssetGetConfig (asset) {
         return `${this.assetHost}/internalapi/asset/${asset.assetId}.${asset.dataFormat}/get/`;
     }
+    getAssetCreateConfig (asset) {
```
benjiwheeler: I appreciate the comment below (and the one above) but I find them hard to understand. "There is no such thing" meaning, there's no API call for it? But it does exist, as you say in the same sentence? Sorry, I know this stuff is hard to write, I'm just not following!

benjiwheeler:
Tests are failing -- looks like `project-saver-hoc`

src/lib/project-saver-hoc.jsx
joker314: Good question!

Here's some code:

```javascript
class Parent { }

const alice = new Parent()

// inheritted property is also enumerable
Parent.prototype.greeting = "Hi."

for(key in alice) {
	console.log("Enumerable: " + key) // "greeting"
	console.log("Own property? " + alice.hasOwnProperty(key)) // "false"
}
```

Most of the default properties which come with JavaScript are not enumerable by default. However, if you make new ones, they might appear in `for...in` loops. Such loops detect keys which are higher up in the prototype chain, though.

src/lib/project-saver-hoc.jsx
rschamp: Imagine you save the JSON, and then all the assets fail to save. Now your JSON is pointing to a bunch of MD5s that don't exist, so you've effectively corrupted your project. To avoid that, we want to make sure all the assets are stored safely before attempting to save the project JSON. Any asset save failure should cause the whole project save process to fail. And, we don't really care who owns the project as far as the assets go. You just have to be authenticated.

As far as who owns the project, I would hope that's already been determined once we get here? All I'm doing is changing the mechanics of how the project data is saved, not anything to do with ownership. All of that should remain as it was.

src/lib/project-saver-hoc.jsx
rschamp: It's pretty bizarre, I have no recollection of writing this! I meant to leave any existing code unchanged. I'm really at a loss of how this got here 😕 

It's generally safer to use this check whenever you're using `for ... in` for the reason @joker314 said — you only want to enumerate the values you've personally added to the object. But it doesn't matter in this case.

src/lib/storage.js
rschamp: Since an asset is stored as the MD5 of its contents, there's no such thing as "updating" an asset, because if the contents were to change, the filename would change, and then you're not storing that asset anymore, you're creating a new one with a different MD5. So there's literally no "update" route for assets — if you `PUT` to the assets server, you'll get a `404` response.

But, the way storage is written, it tries to decide which way to store based on the presence of an assetId. If it has an ID, then it assumes you want to update the asset with that ID, not store a new one.  This is, once again, a shortcoming of the way storage is designed w/r/t web storage.  So this is a hack to get around that.

I don't know how to put that more concisely though, hence the wordy and confusing comment... any suggestions?

rschamp:
Tests will fail until the VM PR is merged and updated here. This depends on `vm.assets`, which won't exist until that's merged.

The cross-origin error is usually a secondary error from a 500 (or some other error) response, which don't have CORS headers set on it. Can you check out the network tab? Maybe your local API or projects servers need to get updated?

rschamp:
@benjiwheeler I just realized I forgot to push and make a PR for the assets server, which needs updating for this to work. That's probably your CORS issue, sorry!

src/lib/project-saver-hoc.jsx
benjiwheeler: I think I'm understanding better.

Let me clarify. The case where you don't own the project is basically when you're creating a project, from one already loaded in the vm. In that case, we'll get here and start saving assets even though there is not yet any project that you own that's associated with them.

But that's OK! Right? Before, I was assuming that assets are stored with some kind of reference to the project that owns them... but that's not true, right? (The project references the assets, but not the other way around.) If I'm understanding this right, then I totally agree with you that we should save assets, and only afterwards save the project JSON.

src/lib/storage.js
benjiwheeler: Sorry, still don't understand. 

* "Stored as the MD5 of its contents" -- you mean, the MD5 is used as the key name for the file? Or something like that?
* "Which way to store" -- wait, there's another way besides that? Now there's two ways to store that I can't describe...
* "update the asset with that ID, not store a new one" -- here's updating again, which doesn't exist! I'm super lost :( What's the role of the assetId here? Is that different from the MD5?

benjiwheeler:
Got this to run without crashing, using that -assets PR!

How else should I test it? It saves without crashing after I mess around with costumes... anything else I can try?

src/lib/storage.js
rschamp: Yes exactly, the MD5 is the key name on S3 (actually MD5.<data type>, e.g. MD5.svg or MD5.wav)

Maybe also confusing is when talking about assets in general in the context of scratch-storage, I don't mean costumes and sounds, I mean the objects that scratch-storage interacts with, which may also be project sb3 files, project json files, etc. Anything that needs to be stored: https://github.com/LLK/scratch-storage/blob/develop/src/Asset.js and each Asset has a type: https://github.com/LLK/scratch-storage/blob/develop/src/AssetType.js

"Which way to store" — there are two ways to store in general, create or update (POST or PUT on the web). Since scratch-storage is a general tool, it assumes for any asset you will have a config for doing both of these things. It's not set up specifically for the assets server, we have to do that (that's what I'm doing in this PR).

> "update the asset with that ID, not store a new one" -- here's updating again, which doesn't exist! I'm super lost :( What's the role of the assetId here? Is that different from the MD5?

I tried to write this out a bunch of times but I don't think I can concisely describe everything that happens when you store something in this comment. Please take a look again at https://github.com/LLK/scratch-storage/blob/develop/src/WebHelper.js so you understand what `WebHelper.store` does, and after that let's chat offline to figure out what comment would be helpful here.

src/lib/project-saver-hoc.jsx
rschamp: This is kind of here as a placeholder for what we want to do in https://github.com/LLK/scratch-gui/issues/3723. I think we will probably want to handle all of that here, close to where the actual saving is happening.

src/lib/project-saver-hoc.jsx
rschamp: Yes exactly.

rschamp:
@benjiwheeler Let's follow up on the comment. I would like this in for the freeze.

PR #3725, opened 2018-11-12
Handle ?tutorial= query in gui

chrisgarrity:
### Resolves
Prerequisite for https://github.com/LLK/scratch-www/issues/2281

### Proposed Changes

Moved the tutorial id handling into gui so it works when gui is embedded in www as well as standalone.

This seemed better than duplicating the query handling in www, in order to pass an id to gui.

Note, the `?tutorial=all` handling remains in app-state-hoc. Currently it’s only needed in standalone mode for Hour of Code. It’s easy to move into the gui container if we do want it in gui by default.

Question: are there going to be interactions with project save/new... that will lead to ignoring the query string, or otherwise not the results that users expect.

### Reason for Changes
Support opening with a tutorial loaded in www.

### Test Coverage
Current tests pass - it didn't change functionality, just where it was included so it's available in www.

Try it:
https://chrisgarrity.github.io/scratch-gui/issue/www-2281-tutorials/

Manual testing:
- [ ] `/?tutorial=getStarted` should open the getting started tutorial
- [ ] Other tutorials should open as well ('name', 'music', 'chase-game', 'clicker-game', 'animations-that-talk') 
- [ ] 'animate-an-adventure-game' should load with a default project
- [ ] `/?tutorial=all` should load with the tutorials library open
- [ ] if the tutorial isn't recognized it should be ignored

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/gui.jsx
+++ b/src/containers/gui.jsx
@@ -13,6 +13,8 @@ import {
     getIsShowingProject
 } from '../reducers/project-state';
 import {setProjectTitle} from '../reducers/project-title';
+import {detectTutorialId} from '../lib/tutorial-from-url';
+import {activateDeck} from '../reducers/cards';
 import {
     activateTab,
     BLOCKS_TAB_INDEX,
@@ -47,6 +49,7 @@ class GUI extends React.Component {
     componentDidMount () {
         this.setReduxTitle(this.props.projectTitle);
         this.props.onStorageInit(storage);
+        this.setActiveCards(detectTutorialId());
     }
     componentDidUpdate (prevProps) {
         if (this.props.projectId !== prevProps.projectId && this.props.projectId !== null) {
@@ -63,6 +66,11 @@ class GUI extends React.Component {
             );
         } else {
             this.props.onUpdateReduxProjectTitle(newTitle);
+        }
+    }
+    setActiveCards (tutorialId) {
+        if (tutorialId && tutorialId !== 'all') {
+            this.props.onUpdateReduxDeck(tutorialId);
         }
     }
     render () {
@@ -119,6 +127,7 @@ GUI.propTypes = {
     onStorageInit: PropTypes.func,
     onUpdateProjectId: PropTypes.func,
     onUpdateProjectTitle: PropTypes.func,
+    onUpdateReduxDeck: PropTypes.func,
     onUpdateReduxProjectTitle: PropTypes.func,
     previewInfoVisible: PropTypes.bool,
     projectHost: PropTypes.string,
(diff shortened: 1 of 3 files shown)


benjiwheeler:
Hm, seems unfortunate that `/?tutorial=getStarted` shows the "Try it" modal

benjiwheeler:
`https://chrisgarrity.github.io/scratch-gui/issue/www-2281-tutorials/?tutorial=animate-an-adventure-game` does not seem to open a starter project for me... am i doing something wrong?

src/containers/gui.jsx
```diff
@@ -13,6 +13,8 @@ import {
     getIsShowingProject
 } from '../reducers/project-state';
 import {setProjectTitle} from '../reducers/project-title';
+import {detectTutorialId} from '../lib/tutorial-from-url';
```
benjiwheeler: is it possible to move this url detection to an HOC? Seems like that's the design pattern we're trying to follow, and moving towards.

I think we could do the same componentDidMount logic in the HOC, and it should just work, because gui watches the reducer values already.

What do you think? Partially, I'm thinking forward to our needing to mess with the url even more... seems like an HOC wrapping the whole gui at the top level would be good.

src/lib/app-state-hoc.jsx
```diff
@@ -67,16 +66,10 @@ const AppStateHOC = function (WrappedComponent, localesOnly) {
...
-                        if (tutorialId === 'all') {
-                            initializedGui = initTutorialLibrary(initializedGui);
-                        } else {
-                            initializedGui = initTutorialCard(initializedGui, tutorialId);
-                        }
-                    }
+                    // handle ?tutorial=all for beta
+                    // if we decide to keep this for www, functionality should move to
+                    // setActiveCards in the GUI container
+                    if (tutorialId === 'all') initializedGui = initTutorialLibrary(initializedGui);
```
chrisgarrity: This should still disable the preview modal if there's a tutorialId @benjiwheeler

chrisgarrity:
The project id is not automatic. The link from code.org has the project id in the URL. To see it you'd need to add #249143200 to the url. Not suppressing the try-it modal is a problem - I'll fix it.

benjiwheeler:
Thanks, I was able to bring up the project with id and tutorial. But tests are failing...

PR #3749, opened 2018-11-13
Create cloud vars

kchadha:
### Resolves

Towards having fully functional cloud variables in 3.0.

### Proposed Changes

Adds the ability to create cloud variables in a 3.0 project.
- adds a cloud variable option to the variable modal
- refactors cloud data connection logic into a separate HOC. 

#### Known Issues/TODO
- There is still more logic to come (not in this PR) about checking whether the user is a scratcher and about handling the cases where the user is viewing someone else's cloud project in editor mode.
- Adding a tooltip for why the cloud option becomes disabled (e.g. notifying the user about the cloud variable limit)
- Adding alerts/info bubbles about cloud variable usage when creating the first cloud variable in a project (like in 2.0)

### Reason for Changes

Cloud variables!

### Test Coverage

Manual testing:

On a local scratch environment (api, scratchr2, www, projects, assets), create a new account (doesn't yet have to be a scratcher) and ensure the following:
- [ ] navigate to /preview/editor to create a new project, create a cloud variable using the variable modal, see that the cloud variable gets created
- [ ] You should not be able to create more than 8 cloud variables (cloud var option becomes disabled in menu)
- [ ] Create code to modify the cloud variable, save, and share the project.
- [ ] Create another account and interact with the cloud project and see that the value of the cloud variable updates on both accounts (use separate browsers)
- [ ] In one browser, sign out and try to interact with the project (the cloud variable should neither send updates so that they show up on other accounts, nor should they receive updates made from another account).
- [ ] In signed out mode, in the editor, the cloud variable option should not show up in the variable modal
- [ ] Renaming a cloud variable should append a cloud + space if not present at the front of the rename, and should not append an extra one if it is present
- [ ] Cloud variable option should not show up in the variable modal in standalone gui

### Related PRs
- LLK/scratch-vm#1755
- LLK/scratch-blocks#1782

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -380,10 +380,12 @@ class Blocks extends React.Component {
             this.ScratchBlocks.Msg.VARIABLE_MODAL_TITLE;
         p.prompt.varType = typeof optVarType === 'string' ?
             optVarType : this.ScratchBlocks.SCALAR_VARIABLE_TYPE;
-        p.prompt.showMoreOptions =
+        p.prompt.showVariableOptions = // This flag means that we should show variable/list options about scope
             optVarType !== this.ScratchBlocks.BROADCAST_MESSAGE_VARIABLE_TYPE &&
             p.prompt.title !== this.ScratchBlocks.Msg.RENAME_VARIABLE_MODAL_TITLE &&
             p.prompt.title !== this.ScratchBlocks.Msg.RENAME_LIST_MODAL_TITLE;
+        // TODO CLOUD add flag about showing the cloud variable option, which will necessitate
+        // that the flag above is true, e.g. p.prompt.showVariableOptions && canUseCloudData
         this.setState(p);
     }
     handleConnectionModalStart (extensionId) {
@@ -449,11 +451,13 @@ class Blocks extends React.Component {
                     {...props}
                 />
                 {this.state.prompt ? (
+                    /* TODO add the following in the props below:
+                    showCloudOption={this.state.canUseCloudData */
                     <Prompt
                         isStage={vm.runtime.getEditingTarget().isStage}
                         label={this.state.prompt.message}
                         placeholder={this.state.prompt.defaultValue}
-                        showMoreOptions={this.state.prompt.showMoreOptions}
+                        showVariableOptions={this.state.prompt.showVariableOptions}
                         title={this.state.prompt.title}
                         onCancel={this.handlePromptClose}
                         onOk={this.handlePromptCallback}
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -397,11 +397,17 @@ class Blocks extends React.Component {
     handleOpenSoundRecorder () {
         this.props.onOpenSoundRecorder();
     }
-    handlePromptCallback (input, optionSelection) {
+
+    /*
+     * Pass along information about proposed name and variable options (scope and isCloud)
+     * and additional potentially conflicting variable names from the VM
+     * to the variable validation prompt callback used in scratch-blocks.
+     */
+    handlePromptCallback (input, variableOptions) {
         this.state.prompt.callback(
             input,
             this.props.vm.runtime.getAllVarNamesOfType(this.state.prompt.varType),
-            optionSelection);
+            variableOptions);
         this.handlePromptClose();
     }
     handlePromptClose () {
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -431,6 +431,7 @@ class Blocks extends React.Component {
         /* eslint-disable no-unused-vars */
         const {
             anyModalVisible,
+            canUseCloud,
             customProceduresVisible,
             extensionLibraryVisible,
             options,
@@ -460,6 +461,7 @@ class Blocks extends React.Component {
                     /* TODO add the following in the props below:
                     showCloudOption={this.state.canUseCloudData */
                     <Prompt
+                        canUseCloud={canUseCloud}
                         isStage={vm.runtime.getEditingTarget().isStage}
                         label={this.state.prompt.message}
                         placeholder={this.state.prompt.defaultValue}
@@ -491,6 +493,7 @@ class Blocks extends React.Component {
 
 Blocks.propTypes = {
     anyModalVisible: PropTypes.bool,
+    canUseCloud: PropTypes.bool,
     customProceduresVisible: PropTypes.bool,
     extensionLibraryVisible: PropTypes.bool,
     isRtl: PropTypes.bool,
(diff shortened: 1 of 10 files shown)


src/containers/blocks.jsx
```diff
@@ -380,10 +380,12 @@ class Blocks extends React.Component {
             this.ScratchBlocks.Msg.VARIABLE_MODAL_TITLE;
         p.prompt.varType = typeof optVarType === 'string' ?
             optVarType : this.ScratchBlocks.SCALAR_VARIABLE_TYPE;
-        p.prompt.showMoreOptions =
+        p.prompt.showVariableOptions = // This flag means that we should show variable/list options about scope
             optVarType !== this.ScratchBlocks.BROADCAST_MESSAGE_VARIABLE_TYPE &&
             p.prompt.title !== this.ScratchBlocks.Msg.RENAME_VARIABLE_MODAL_TITLE &&
             p.prompt.title !== this.ScratchBlocks.Msg.RENAME_LIST_MODAL_TITLE;
+        // TODO CLOUD add flag about showing the cloud variable option, which will necessitate
```
paulkaplan: hmmm? This is a confusing comment

src/lib/cloud-provider.js
```diff
@@ -65,6 +58,26 @@ class CloudProvider {
...
+        }
+        case 'set': {
+            varData.varUpdate = {
+                name: message.name,
+                value: message.value
+            };
+            break;
+        }
+        }
+        this.vm.postIOData('cloud', varData);
```
paulkaplan: Can you return the varData and put the postIOData in the caller of this function? Seems strange to have a method called "parse_" that actually sends data to the VM.

src/lib/cloud-manager-hoc.jsx
```diff
@@ -0,0 +1,98 @@
...
+                projectId,
+                username,
+                isShowingWithId,
+                /* eslint-enable no-unused-vars */
+                vm,
+                ...componentProps
+            } = this.props;
+            return (
+                <WrappedComponent
+                    canUseCloud={this.canUseCloud}
```
paulkaplan: It would be better to compute the value here instead of relying on a pre-computed value. This will ensure that the value of `canUseCloud` is always up-to-date given the props, whereas setting it in componentDidUpdate seems like it will always be one step behind.

src/lib/cloud-manager-hoc.jsx
```diff
@@ -0,0 +1,98 @@
...
+            if (!canUseCloud) return;
+
+            // TODO add scratcher check
+            this.canUseCloud = canUseCloud;
+
+            if ((this.props.username !== prevProps.username) ||
+                (this.props.projectId !== prevProps.projectId) ||
+                (!this.cloudProvider) ||
+                (this.cloudProvider && !this.cloudProvider.connection)) {
+                // TODO need to add provisions for viewing someone
```
paulkaplan: Because this is in `componentDidUpdate`, it will not be called for the initial render. You probably want to extract lines 35-44 into a separate function that you can call here (during the update) and also in `componentDidMount`. I'm guessing this went unnoticed because the load process would never init the cloud server on mount because the project isn't loaded yet.

src/lib/cloud-manager-hoc.jsx
```diff
@@ -0,0 +1,98 @@
...
+                // else's project in editor mode
+                this.cloudProvider = new CloudProvider(
+                    this.props.cloudHost,
+                    this.props.vm,
+                    this.props.username,
+                    this.props.projectId);
+                this.props.vm.setCloudProvider(this.cloudProvider);
```
paulkaplan: Can you add a componentDidMount that closes the websocket connection?

src/lib/cloud-manager-hoc.jsx
kchadha: @ray, yeah I think that makes sense. This is somewhat temporary because I needed to be able to control when/when not to show the cloud variable option in the variable modal in this set of PRs. I will probably move some of this logic into www when I add more `should-we-connect??` logic, but we should maybe talk offline about what's best to have where.

src/lib/cloud-manager-hoc.jsx
```diff
@@ -0,0 +1,98 @@
...
+        const loadingState = state.scratchGui.projectState.loadingState;
+        return {
+            isShowingWithId: getIsShowingWithId(loadingState),
+            projectId: state.scratchGui.projectState.projectId
+        };
+    };
+
+    const mapDispatchToProps = () => ({});
+
+    // Allow incoming props to override redux-provided props. Used to mock in tests.
```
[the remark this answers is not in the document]
kchadha: hahaha that was a copy paste remnant from the vm-manager-hoc =]

src/lib/cloud-manager-hoc.jsx
```diff
@@ -0,0 +1,98 @@
...
+            this.cloudProvider = null;
+            this.canUseCloud = false;
+        }
+        componentDidUpdate (prevProps) {
+            // If any of the following are not true, do not open a cloud connection
+            const canUseCloud = this.props.cloudHost && this.props.username && this.props.isShowingWithId;
+
+            if (!canUseCloud) return;
+
+            // TODO add scratcher check
```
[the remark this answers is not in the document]
kchadha: I'm going to keep the comment in here for now as a reminder and will remove it when I use updated info from www in here.

src/lib/cloud-manager-hoc.jsx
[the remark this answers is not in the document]
kchadha: @rschamp, @paulkaplan, I think I will end up following up this PR with one that adds a test for the cloud manager hoc. I got stuck trying to figure out how to mock WebSocket (or maybe even CloudProvider) and might need some help on that.

kchadha:
@rschamp, @paulkaplan, I believe I addressed all the PR comments except for the addition of the test. I started work on this but got stuck trying to mock out CloudProvider/WebSocket, so I think I will leave the test for a follow up PR.

@paulkaplan, I modified the cloud-manager-hoc a bit further from when we worked on it together, notable changes are:
- use of componentWillUnmount instead of componentDidUnmount (when I was running the tests, they were complaining that componentDidUnmount is not a real lifestyle method, and it looks like elsewhere throughout the gui we are, in fact, using `componentWillUnmount`.
- the new `canUseCloud` function takes in props so that during componentDidUpdate, we can compare its value to what it would have been. This is important for the initial cloud connection so that it catches the change of loading state from isShowingWithId being false to being true. This is especially important when loading a project's page directly via `/preview/<project id>` for example.

kchadha:
@rschamp, @paulkaplan, I ended up figuring out how to mock a class via the awesome docs here:
https://jestjs.io/docs/en/es6-class-mocks

I added the test to this PR.

src/lib/cloud-manager-hoc.jsx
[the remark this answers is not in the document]
kchadha: JK, I ended up adding a test here after all.

Issue #3710, opened 2018-11-09
Block editors not being hidden by modals when embedded


paulkaplan:
This problem **only impacts gui when it is embedded in the project page**. This does not happen on the standalone beta build.

If you open the note picker or angle picker, then open the extensions library (for example) you still see the field above the modal.

It seems that this code to hide the block editors is not running when opening modals when the GUI is embedded in WWW https://github.com/LLK/scratch-gui/blob/develop/src/containers/blocks.jsx#L133-L135

My best guess is that the state tree may be different and the `anyModalVisible` prop may not be setting correctly in mapStateToProps https://github.com/LLK/scratch-gui/blob/develop/src/containers/blocks.jsx#L558-L561

Thanks to @picklesrus for finding this

paulkaplan:
This is caused by having the GUI consume a prop to hide the preview modal but not updating the store to set the modal.previewInfo to false. So `anyModalVisible` is always true, causing hideChaff to never be called.

paulkaplan:
The solution to this is just to make the default of `modals.previewInfo` false, and have it possible to toggle to true via AppStateHOC.

PR #3768, opened 2018-11-14
Remove the hideIntro override and default preview modal to false

paulkaplan:
The use of an "override" prop that allowed the preview modal to be hidden without actually changing the redux state was causing a problem where the block editors where not closing when modals were opening. That is because it relied on the modal reducer being accurate.

I defaulted the preview state to false to make it simpler to consume the GUI, as www will not want the preview info to be true. I made it true in the playground instead.

This fixes https://github.com/LLK/scratch-gui/issues/3710

I can't write an integration test for that specific issue yet because it only happens when embedded, but the current integration tests ensure that the preview modal is still there in the playground the way we want. I confirmed the fix for #3710 by manually linking and testing.

### DIFF
--- a/src/lib/app-state-hoc.jsx
+++ b/src/lib/app-state-hoc.jsx
@@ -31,10 +31,6 @@ const AppStateHOC = function (WrappedComponent, localesOnly) {
             let reducers = {};
             let enhancer;
 
-            this.state = {
-                tutorial: false
-            };
-
             let initializedLocales = localesInitialState;
             const locale = detectLocale(Object.keys(locales));
             if (locale !== 'en') {
@@ -56,6 +52,7 @@ const AppStateHOC = function (WrappedComponent, localesOnly) {
                     guiMiddleware,
                     initFullScreen,
                     initPlayer,
+                    initPreviewInfo,
                     initTutorialLibrary
                 } = guiRedux;
                 const {ScratchPaintReducer} = require('scratch-paint');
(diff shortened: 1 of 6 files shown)


src/reducers/gui.js
```diff
@@ -102,13 +99,24 @@ const initTutorialLibrary = function (currentState) {
...
         {
             modals: {
-                previewInfo: false,
                 tipsLibrary: true
             }
         }
     );
 };
 
+const initPreviewInfo = function (currentState) {
```
benjiwheeler: I'm not crazy about loading up gui.js with these mini-reducers that mutate state in these arbitrary ways... could it be possible to put this in a reducer? Maybe the modals reducer?

Is this necessary basically because we want this to be the starting state on construct, rather than waiting for a dispatch?

src/reducers/modals.js
```diff
@@ -23,7 +23,7 @@ const initialState = {
     [MODAL_EXTENSION_LIBRARY]: false,
     [MODAL_IMPORT_INFO]: false,
     [MODAL_LOADING_PROJECT]: false,
-    [MODAL_PREVIEW_INFO]: true,
+    [MODAL_PREVIEW_INFO]: false,
```
benjiwheeler: great that you're changing this default.

but, kinda ugly that this logic is spread around multiple reducer files...

src/lib/hash-parser-hoc.jsx
```diff
@@ -77,7 +74,12 @@ const HashParserHOC = function (WrappedComponent) {
         };
     };
     const mapDispatchToProps = dispatch => ({
-        setProjectId: projectId => dispatch(setProjectId(projectId))
+        setProjectId: projectId => {
+            dispatch(setProjectId(projectId));
+            if (projectId !== defaultProjectId) {
+                dispatch(closePreviewInfo());
```
benjiwheeler: nice, this is an elegant solution to the spaghetti code...

src/lib/hash-parser-hoc.jsx
```diff
@@ -9,6 +9,10 @@ import {
     setProjectId
 } from '../reducers/project-state';
 
+import {
```
benjiwheeler: FWIW, this can be one line

src/lib/app-state-hoc.jsx
```diff
@@ -76,6 +77,9 @@ const AppStateHOC = function (WrappedComponent, localesOnly) {
                         } else {
                             initializedGui = initTutorialCard(initializedGui, tutorialId);
                         }
+                    } else if (props.showPreviewInfo) {
```
benjiwheeler: This is a small thing, but would it be possible to rewrite these if/elses to be phrased positively rather than negatively? Like:

```
if (tutorialId === null) {
    if (props.showPreviewInfo) {...}
} else { // tutorial is present
...
}
```

I find it much easier to read code like that... otherwise I have to say to myself, "ok, this will run if tutorial is not null is false, so tutorial IS null, AND showPreviewInfo is true...

paulkaplan:
@benjiwheeler thanks for the comments, restructuring the negative "if" is definitely more clear.

I don't love the mini reducers in GUI either, but I do not really see a better way to have the initial state be something other than what the reducer says it is. It doesn't really make sense to dispatch to make this change, because there is no user action that is happening, it is just that the initial state is different between gui/www.

benjiwheeler:
This looks good, but could you wait for ChrisG to merge the much simpler https://github.com/LLK/scratch-gui/pull/3725 , which also touches hideIntro? Should land today

benjiwheeler:
LGTM

PR #3817, opened 2018-11-16
Update videoState to match default in rendered-target

ktbee:
### Resolves

This will help with https://github.com/LLK/scratch-vm/issues/1635

### Proposed Changes

Set the default project's videoState to 'on'

### Reason for Changes

[Currently the default videoState in the VM is 'on',](https://github.com/LLK/scratch-vm/blob/develop/src/sprites/rendered-target.js#L154) which is helpful for making sure adding a video extension enables the camera by default. Setting the project default to 'on' in the GUI will make sure this default isn't overwritten for default SB3 projects. This change won't automatically turn on the video device, an extension will still need to be added for the video to turn on.

### DIFF
--- a/src/lib/default-project/project-data.js
+++ b/src/lib/default-project/project-data.js
@@ -62,10 +62,7 @@ const projectData = translateFunction => {
                         md5ext: '83a9787d4cb6f3b7632b4ddfebf74367.wav'
                     }
                 ],
-                volume: 100,
-                tempo: 60,
-                videoTransparency: 50,
-                videoState: 'off'
+                volume: 100
             },
             {
                 isStage: false,


kchadha:
@ktbee, I'm not sure I like the idea of having the default project (which is like any normal project, but just happens to be the one we load in by default) have misleading information. 

I think it would be better to remove the video related information from this project since this project doesn't have anything to do with the video extension, and then have the VM do something sensible when there isn't a stage provided e.g. either it shouldn't be looking for video information at all when the stage doesn't exist yet or if it really needs to for some reason, it should maybe default the video state to off, and then when the stage exists, either the stage explicitly specifies the video state or it doesn't have any video state information and the video state defaults to on in that case.

@chrisgarrity, I see that the video information was added to the default project in this PR, was it necessary for some reason? https://github.com/LLK/scratch-gui/pull/3145/files#diff-a47260ca2277c87d339210848ccc7b1aR68

chrisgarrity:
@kchadha it's only there because it was in the default project json at the time ([see the deleted project json file](https://github.com/LLK/scratch-gui/pull/3145/files#diff-dd716c4d9e72e8a8a1ba10fbc1d315fe)). It's not needed for localization - feel free to remove.

ktbee:
@kchadha, all set with removing the default extension settings for sb3 projects!

PR #3818, opened 2018-11-16
Add exponential backoff for cloud provider.

paulkaplan:
Uses randomized duration between [0, 2^k-1] where k is the connection attempt number, maxed out at 5.

I had to rework the tests quite a bit to make this easy to test, to work around the timeouts and the randomization.

### DIFF
--- a/src/lib/cloud-provider.js
+++ b/src/lib/cloud-provider.js
@@ -28,6 +28,8 @@ class CloudProvider {
      * @param {string} cloudHost The cloud data server to connect to.
      */
     openConnection () {
+        this.connectionAttempts += 1;
+
         try {
             this.connection = new WebSocket((location.protocol === 'http:' ? 'ws://' : 'wss://') + this.cloudHost);
         } catch (e) {
@@ -58,14 +60,15 @@ class CloudProvider {
     }
 
     onOpen () {
-        this.connectionAttempts = 0; // Reset because we successfully connected
+        // Reset connection attempts to 1 to make sure any subsequent reconnects
+        // use connectionAttempts=1 to calculate timeout
+        this.connectionAttempts = 1;
         this.writeToServer('handshake');
         log.info(`Successfully connected to clouddata server.`);
     }
 
     onClose () {
         log.info(`Closed connection to websocket`);
-        this.connectionAttempts += 1;
         const randomizedTimeout = this.randomizeDuration(this.exponentialTimeout());
         this.setTimeout(this.openConnection.bind(this), randomizedTimeout);
     }
@@ -79,7 +82,7 @@ class CloudProvider {
     }
 
     setTimeout (fn, time) {
-        log.info(`Reconnecting in ${Math.floor(time / 1000)}s, attempt ${this.connectionAttempts}`);
+        log.info(`Reconnecting in ${(time / 1000).toFixed(1)}s, attempt ${this.connectionAttempts}`);
         this._connectionTimeout = window.setTimeout(fn, time);
     }
 
@@ -167,7 +170,9 @@ class CloudProvider {
             this.connection.readyState !== WebSocket.CLOSING &&
             this.connection.readyState !== WebSocket.CLOSED) {
             log.info('Request close cloud connection without reconnecting');
-            this.connection.onclose = () => {}; // Remove close listener to prevent reconnect
+            // Remove listeners, after this point we do not want to react to connection updates
+            this.connection.onclose = () => {};
+            this.connection.onerror = () => {};
             this.connection.close();
         }
         this.clear();
--- a/src/lib/cloud-provider.js
+++ b/src/lib/cloud-provider.js
@@ -16,23 +16,25 @@ class CloudProvider {
         this.vm = vm;
         this.username = username;
         this.projectId = projectId;
+        this.cloudHost = cloudHost;
 
-        // Open a websocket connection to the clouddata server
-        this.openConnection(cloudHost);
+        this.connectionAttempts = 0;
+
+        this.openConnection();
     }
 
     /**
      * Open a new websocket connection to the clouddata server.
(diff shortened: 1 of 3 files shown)


kchadha:
cc/ @colbygk

src/lib/cloud-provider.js
```diff
@@ -148,7 +168,7 @@ class CloudProvider {
         if (this.connection &&
             this.connection.readyState !== WebSocket.CLOSING &&
             this.connection.readyState !== WebSocket.CLOSED) {
-
+            this.connection.onclose = () => {}; // Remove close listener to prevent reconnect
```
kchadha: Hmmm what's the benefit of having the reconnect logic happen in onClose instead of onError? Is it that onError is too soon to attempt a reconnect?

kchadha: Also maybe keep the log here so that we can be sure that the connection actually closed when we want it to.

paulkaplan: Yeah, error is called before close. They always go together, but error first. Also error is not always called when the socket is closed by the server.

src/lib/cloud-provider.js
```diff
@@ -163,6 +183,10 @@ class CloudProvider {
         this.vm = null;
         this.username = null;
         this.projectId = null;
+        if (this._connectionTimeout) {
+            clearTimeout(this._connectionTimeout);
+            this._connectionTimeout = null;
+        }
```
kchadha: Do we need to reset connectAttempts here?

test/unit/util/cloud-provider.test.js
```diff
@@ -80,8 +98,60 @@ describe('CloudProvider', () => {
...
+        expect(cloudProvider.connectionAttempts).toBe(4);
+
+        cloudProvider.connection.close();
+        expect(timeout).toEqual(15 * 1000); // 2^4 - 1
+        expect(websocketConstructorCount).toBe(5);
+        expect(cloudProvider.connectionAttempts).toBe(5);
+
+        cloudProvider.connection.close();
+        expect(timeout).toEqual(31 * 1000); // 2^5 - 1
+        expect(websocketConstructorCount).toBe(6);
```
kchadha: Should these be changing if we've maxed out at 5? E.g. why is the web socket constructor being called again after maxing out?

paulkaplan: The websocket constructor count is just how many times we've created the websocket connection object (i.e. tried to open a socket). So it keeps increasing because we keep trying, the only thing that maxes out is the time difference between attempts (i.e. connectionAttempt)

src/lib/cloud-provider.js
paulkaplan: Yes

src/lib/cloud-provider.js
```diff
@@ -57,12 +58,31 @@ class CloudProvider {
...
         log.info(`Closed connection to websocket`);
+        const exponentialTimeout = (Math.pow(2, this.connectionAttempts) - 1) * 1000;
+        const randomizedTimeout = this.randomizeDuration(exponentialTimeout);
+        this.setTimeout(this.reconnectNow.bind(this), randomizedTimeout);
+    }
+
+    reconnectNow () {
+        // Max connection attempts at 5, so timeout will max out in range [0, 31s]
+        this.connectionAttempts = Math.min(this.connectionAttempts + 1, 5);
+        this.openConnection();
```
kchadha: should there be a check here that we haven't exceeded the connectionAttempts before opening the connection?

src/lib/cloud-provider.js
paulkaplan:
```suggestion
        }
        this.connectionAttempts = 0;
```

test/unit/util/cloud-provider.test.js
kchadha: I would think that we should stop trying after a while, but I'm not sure... O.o

src/lib/cloud-provider.js
paulkaplan: Similar to comment above, the connection attempt number gets limited just to max the timeout, but it keeps reconnecting ad infinitum

test/unit/util/cloud-provider.test.js
paulkaplan: Never know when it'll come back!

src/lib/cloud-provider.js
```diff
@@ -57,12 +58,31 @@ class CloudProvider {
...
+    }
+
+    randomizeDuration (t) {
+        return Math.random() * t;
+    }
+
+    setTimeout (fn, time) {
+        log.info(`Reconnecting in ${time}ms, attempt ${this.connectionAttempts}`);
```
kchadha: The attempt number in this log message should probably always be increasing, regardless of the fact that we capped out in the decay calculation...

src/lib/cloud-provider.js
kchadha: Not resolving this conversation yet b/c I still think we should continue to log closing a connection, when the close is explicitly requested.

paulkaplan:
Addressed those issues. Feel free to reorganize when merging if there are conflicts with your other PRs

src/playground/render-gui.jsx
```diff
@@ -31,6 +31,11 @@ export default appTarget => {
     if (process.env.NODE_ENV === 'production' && typeof window === 'object') {
         // Warn before navigating away
         window.onbeforeunload = () => true;
+    } else {
+        window.onerror = e => {
```
kchadha: Did you mean for this chunk to be in this PR?

benjiwheeler:
Should there be a maximum number of reconnection attempts? Even something like 20?

colbygk:
I don't think there should be; but don't have a strong opinion about it.

- Colby

PR #3834, opened 2018-11-19
Library 100% height fix

### LINKED ISSUE #3825: Extension library item tiles are too tall in a single row
screen shot 2018-11-15 at 4 33 48 pm
from Eric

evhan55:
### Resolves

- Resolves #3825: Extension library item tiles are too tall in a single row.

### DIFF
--- a/src/components/library/library.css
+++ b/src/components/library/library.css
@@ -10,7 +10,7 @@
     flex-grow: 1;
     flex-wrap: wrap;
     overflow-y: auto;
-    height: calc(100% - $library-header-height);
+    height: auto;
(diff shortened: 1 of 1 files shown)


src/components/library/library.css
```diff
@@ -10,7 +10,7 @@
     flex-grow: 1;
     flex-wrap: wrap;
     overflow-y: auto;
-    height: calc(100% - $library-header-height);
+    height: auto;
```
evhan55: I checked all libraries: costume/library/tips/extensions and they all seem ok to me on Chrome/MacOS with this fix.

PR #3835, opened 2018-11-19
avoid create/save/error loops

benjiwheeler:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/3836

### Proposed Changes

* avoid false success alerts:
  * throws an error if saving assets or project encounters an error (previously, error was being intercepted and not passed on)
  * call to display success alert comes *after* we dispatch success action itself, not before
* do not try to create or save twice in a row (avoids infinite loop of retries)

### Test Coverage

None currently

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/lib/project-saver-hoc.jsx
+++ b/src/lib/project-saver-hoc.jsx
@@ -22,7 +22,6 @@ import {
     getIsShowingWithId,
     getIsShowingWithoutId,
     getIsUpdating,
-    manualUpdateProject,
     projectError
 } from '../reducers/project-state';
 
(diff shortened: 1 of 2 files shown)


src/lib/project-saver-hoc.jsx
```diff
@@ -78,11 +80,11 @@ const ProjectSaverHOC = function (WrappedComponent) {
             }
             return this.storeProject(this.props.reduxProjectId)
                 .then(() => {
+                    // there is nothing we expect to find in response that we need to check here
```
rschamp: What does this comment mean? It kind of reads like "These are not the droids you're looking for".

src/lib/project-saver-hoc.jsx
```diff
@@ -54,14 +54,16 @@ const ProjectSaverHOC = function (WrappedComponent) {
...
+            if (showingCreateable && !prevShowingCreateable && !justTriedCreatingNew) {
                 this.props.onCreateProject();
             } else {
                 // if we're newly able to save this project, save it!
                 const showingSaveable = this.props.canSave && this.props.isShowingWithId;
                 const becameAbleToSave = this.props.canSave && !prevProps.canSave;
                 const becameShared = this.props.isShared && !prevProps.isShared;
-                if (showingSaveable && (becameAbleToSave || becameShared)) {
+                const justTriedUpdating = prevProps.isUpdating; // don't immediately keep trying
+                if (showingSaveable && !justTriedUpdating && (becameAbleToSave || becameShared)) {
```
rschamp: This is getting pretty unwieldy. It's pretty hard to understand both of these conditions. Can we use a guard clause and just avoid getting here in case of the excluded conditions (`!prevShowingCreatable`, `!justTriedCreatingNew`, `!justTriedUpdating`)?

rschamp: ("both of these conditions", meaning this if clause and the one that's been changed above.)

src/lib/project-saver-hoc.jsx
benjiwheeler: Clarified

src/lib/project-saver-hoc.jsx
benjiwheeler: I see what you mean. Take a look now, is this better?

PR #3851, opened 2018-11-20
Alerts CSS width changes and hardware alerts localization

### LINKED ISSUE #3264: Add localization to hardware disconnect alerts

### LINKED ISSUE #3531: Make alerts width 448px / support multiline
Should be 448px instead of 520px, need to make alerts message wrap onto a second line.

evhan55:
### Resolves

- Resolves #3531: 448/multiline
- Resolves #3264: localize hardware alerts

### DIFF
--- a/src/components/alerts/alert.css
+++ b/src/components/alerts/alert.css
@@ -4,8 +4,10 @@
 
 .alert {
     width: 100%;
+    height: auto;
     display: flex;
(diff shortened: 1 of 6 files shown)


benjiwheeler:
LGTM. Only note is that you should think about using the `content: PropType:oneOf(PropType.element, PropType.string)` everywhere.

PR #3859, opened 2018-11-21
updated scratch-l10n in package.json

benjiwheeler:


benjiwheeler:
Hm, do you recognize or understand where the tests are failing?

chrisgarrity:
The menu-bar test is failing because the test is looking for 'Upload from your computer' and 'Download to your computer', but at some point we changed the strings to 'Load...' and 'Save...' Since we had never updated the english strings in l10n, the menus in the built version still used the 'Upload...' and 'Download...' 

The test needs to be updated to look for the correct strings. I'm pretty sure the following failures are just side effects of the first one failing.

Issue #3744, opened 2018-11-13
Don't ask user to "replace contents of the current project" if that's not what they're doing!


benjiwheeler:
signed in to scratch.ly, looking at someone else’s project, File->New currently pops up “replace contents of current project?” but that’s super misleading… maybe “leave without saving?”

Alzter:
When I first used File>New I thought I had erased everything in the project I was on, not created a new one.

benjiwheeler:
@alzter, do you mean in Scratch 2.0, or in the 3.0 Beta?

Alzter:
In Scratch 2.0

PR #3866, opened 2018-11-25
only prompt user to save if project has changed and won’t autosave

benjiwheeler:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/3744

### Proposed Changes

don't warn user with a confirm about leaving project, if user could choose to save the project, or project will autosave, or project is already saved.

### Reason for Changes

Prompt was sometimes unnecessary and confusing.

### Test Coverage

none.

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -159,7 +159,10 @@ class MenuBar extends React.Component {
     handleClickNew () {
         let readyToReplaceProject = true;
         // if the project is dirty, and user owns the project, we will autosave.
-        // but if they don't own this project, user should consider downloading first.
+        // but if they are not logged in and can't save, user should consider
+        // downloading or logging in first.
+        // Note that if user is logged in and editing someone else's project,
+        // they'll lose their work.
         if (this.props.projectChanged && !this.props.canCreateNew) {
             readyToReplaceProject = confirm( // eslint-disable-line no-alert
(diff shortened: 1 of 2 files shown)


paulkaplan:
@benjiwheeler the code LG, but since the flag you are looking to use is introduced in https://github.com/LLK/scratch-gui/pull/3879, you might just want to wait for that to go in and update this PR, instead of having to make a follow-up PR.

benjiwheeler:
Take a look now!

benjiwheeler:
Hm, projectChanged is making use always warn the user... which makes this little better than before

src/components/menu-bar/menu-bar.jsx
```diff
@@ -726,6 +728,7 @@ MenuBar.propTypes = {
     onShare: PropTypes.func,
     onToggleLoginOpen: PropTypes.func,
     onUpdateProjectTitle: PropTypes.func,
+    projectIsDirty: PropTypes.bool,
```
paulkaplan: Probably want this to be projectChanged?

benjiwheeler:
Added several tests of when we will and won't see this confirm.

Issue #3840, opened 2018-11-19
Revise alerts code


benjiwheeler:
- [x] `alertsList.filter` in `containers/alerts.jsx` and `containers/inline-messages.jsx` should be made into a reducer selector
- [x] extended comment defining fields of alerts in alerts reducer should become a jsdoc
- [x] possibly rename SHOW_STANDARD_ALERT since it shows inline alerts too

PR #3867, opened 2018-11-26
Revise alerts code to clarify and reorganize alert type handling

benjiwheeler:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/3840

### Proposed Changes

* alertsList.filter in containers/alerts.jsx and containers/inline-messages.jsx was made into a reducer selector
* extended comment defining fields of alerts in alerts reducer is now a jsdoc
* renamed SHOW_STANDARD_ALERT since it shows inline alerts too

### Reason for Changes

routine code improvement

### DIFF
--- a/src/containers/alerts.jsx
+++ b/src/containers/alerts.jsx
@@ -4,7 +4,8 @@ import {connect} from 'react-redux';
 import {AlertTypes} from '../lib/alerts/index.jsx';
 
 import {
-    closeAlert
+    closeAlert,
+    filterPopupAlerts
 } from '../reducers/alerts';
 
 import AlertsComponent from '../components/alerts/alerts.jsx';
@@ -16,10 +17,7 @@ const Alerts = ({
 }) => (
     <AlertsComponent
         // only display standard and extension alerts here
-        alertsList={alertsList.filter(curAlert => (
-            curAlert.alertType === AlertTypes.STANDARD ||
-            curAlert.alertType === AlertTypes.EXTENSION
-        ))}
+        alertsList={filterPopupAlerts(alertsList)}
         className={className}
         onCloseAlert={onCloseAlert}
     />
--- a/src/containers/alerts.jsx
+++ b/src/containers/alerts.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
-import {AlertTypes} from '../lib/alerts/index.jsx';
 
 import {
     closeAlert,
--- a/src/reducers/alerts.js
+++ b/src/reducers/alerts.js
@@ -21,6 +21,19 @@ const initialState = {
     alertsList: []
 };
 
+const filterPopupAlerts = alertsList => (
+    alertsList.filter(curAlert => (
+        curAlert.alertType === AlertTypes.STANDARD ||
+        curAlert.alertType === AlertTypes.EXTENSION
+    ))
+);
+
+const filterInlineAlerts = alertsList => (
+    alertsList.filter(curAlert => (
+        curAlert.alertType === AlertTypes.INLINE
+    ))
+);
+
 const reducer = function (state, action) {
     if (typeof state === 'undefined') state = initialState;
     switch (action.type) {
@@ -177,6 +190,8 @@ export {
     reducer as default,
     initialState as alertsInitialState,
     closeAlert,
+    filterInlineAlerts,
+    filterPopupAlerts,
     showAlertWithTimeout,
     showExtensionAlert,
     showStandardAlert
--- a/src/reducers/alerts.js
+++ b/src/reducers/alerts.js
@@ -42,7 +42,7 @@ const filterInlineAlerts = alertsList => (
 const reducer = function (state, action) {
     if (typeof state === 'undefined') state = initialState;
     switch (action.type) {
-    case SHOW_ALERT: { // shows standard and inline alerts
+    case SHOW_ALERT: { // intended to show standard and inline alerts, but not extensions
         const alertId = action.alertId;
         if (alertId) {
             const newAlert = {
--- a/src/reducers/alerts.js
+++ b/src/reducers/alerts.js
@@ -12,7 +12,7 @@ const CLOSE_ALERTS_WITH_ID = 'scratch-gui/alerts/CLOSE_ALERTS_WITH_ID';
  *
  * {bool} visible - whether the alerts are visible
  * {array} alertsList - list of alerts, each with properties:
- *  * alert type (required): one of AlertTypes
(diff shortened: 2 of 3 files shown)


src/containers/alerts.jsx
```diff
@@ -16,10 +16,7 @@ const Alerts = ({
 }) => (
     <AlertsComponent
         // only display standard and extension alerts here
-        alertsList={alertsList.filter(curAlert => (
-            curAlert.alertType === AlertTypes.STANDARD ||
-            curAlert.alertType === AlertTypes.EXTENSION
-        ))}
+        alertsList={filterPopupAlerts(alertsList)}
```
rschamp: Previously I suggested using a selector in `mapStateToProps`, to which you replied

> `mapStateToProps` -- the problem (as I understand it) is that if this were in mapStateToProps, it would create a new object every time redux state changes, anywhere, and thus rerender on any redux state change. Whereas like this, the props don't change unless the actual content changes, and thus it only rerenders if the actual content changes.

The solution to this is to memoize the selector:
```js
let _filterPopupAlertsArg, _filterPopupAlertsResult;
const filterPopupAlerts = alertsList => {
    if (alertsList === _filterPopupAlertsArg) return _filterPopupAlertsResult;
    _filterPopupAlertsArg = alertsList;
    _filterPopupAlertsResult = alertsList.filter(curAlert => (
        curAlert.alertType === AlertTypes.STANDARD ||
        curAlert.alertType === AlertTypes.EXTENSION
    ));
    return _filterPopupAlertsResult;
};
```
There are libraries to do this for you, `memoizee` (above is `memoize(filterPopupAlerts, {max: 1})`) and `reselect`.  However, I don't think we should add libraries unnecessarily right now, and this is readable as-is so this is just FYI.

src/reducers/alerts.js
```diff
@@ -2,29 +2,47 @@ import alertsData from '../lib/alerts/index.jsx';
...
+ * {array} alertsList - list of alerts, each with properties:
+ *  * alert type (required): one of AlertTypes
+ *  * closeButton (optional): bool indicating that we should show close button
+ *  * content (optional): react element (a <FormattedMessage />)
+ *  * extentionId (optional): id string that identifies the extension
+ *  * iconURL (optional): string
+ *  * level (required): string, one of AlertLevels
+ *  * message (optional): string
+ *  * showReconnect (optional): bool
+ */
```
rschamp: This isn't valid jsdoc, but I don't think the linter cares because it doesn't know you're intending it to be :)  Previously I mentioned using jsdoc because there is already syntax for denoting required vs optional properties, properties of objects, etc. I think you can leave it (other than the line that doesn't show the actual key name, `alertType`) — we just need it documented, and this does the job well enough for people reading it. 

If you wanted it to be valid, I think you would need two different `@typedef`'s, one for `AlertsInitialState` and one for either `AlertsList` or just `Alert` (and `alertsList` would be `{Array.<Alert>}`). And then finally you would use `@type {AlertsInitialState}` above `initialState`.

src/reducers/alerts.js
```diff
@@ -2,29 +2,47 @@ import alertsData from '../lib/alerts/index.jsx';
...
+    alertsList.filter(curAlert => (
+        curAlert.alertType === AlertTypes.INLINE
+    ))
+);
+
 const reducer = function (state, action) {
     if (typeof state === 'undefined') state = initialState;
     switch (action.type) {
-    case SHOW_STANDARD_ALERT: { // also will show inline alerts
+    case SHOW_ALERT: { // shows standard and inline alerts
```
rschamp: To me this comment implies that `SHOW_ALERT` won't show extension alerts, but that doesn't look like it's true.  This comment doesn't really clarify anything, so I think it could be removed.

```suggestion
    case SHOW_ALERT: {
```

src/reducers/alerts.js
```diff
@@ -2,29 +2,47 @@ import alertsData from '../lib/alerts/index.jsx';
...
 const SHOW_EXTENSION_ALERT = 'scratch-gui/alerts/SHOW_EXTENSION_ALERT';
 const CLOSE_ALERT = 'scratch-gui/alerts/CLOSE_ALERT';
 const CLOSE_ALERTS_WITH_ID = 'scratch-gui/alerts/CLOSE_ALERTS_WITH_ID';
 
+/**
+ * Initial state of alerts reducer
+ *
+ * {bool} visible - whether the alerts are visible
+ * {array} alertsList - list of alerts, each with properties:
+ *  * alert type (required): one of AlertTypes
```
rschamp:
```suggestion
 *  * alertType (required): one of AlertTypes
```

src/containers/alerts.jsx
benjiwheeler: Thanks -- but isn't my current approach nearly, or entirely as effective? I'm calling the selector from the component's render code, not from `mapStateToProps`. So redux should only think `mapStateToProps` has produced a change if the `alertsList` reference changes, which it only will when there's actually a change we want to re-render.

Your code above just checks if if the old and new list references are the same -- that's the same thing that redux already does, right?

(I realize I have only a shaky understanding of what's going on under the hood!)

src/reducers/alerts.js
benjiwheeler: Yeah, I tried to research how you document this kind of stuff, and came up with nothing but the most vague examples.

src/reducers/alerts.js
benjiwheeler: Sorry, I'm confused -- `SHOW_ALERT` certainly isn't intended to show extension alerts. It's true that it does not explicitly check for them, but neither will it show them correctly, if you tried! Is that what you mean?

src/containers/alerts.jsx
rschamp: Your approach has the same effectiveness performance-wise, but organizationally it puts filtering logic into the component, when the filtering could happen one level up. That would make the component less surprising since it would always output what you gave it.

You rightly pointed out that doing that filtering one level up in `mapStateToProps` would incur a performance hit, but that is why the selector-generating library `reselect` does memoization for you, so you don't have to worry about the details of performance to organize the code this way.

src/reducers/alerts.js
rschamp: Yeah it's just a little misleading.

src/containers/alerts.jsx
benjiwheeler: Thanks, that makes sense. I made this follow on: https://github.com/LLK/scratch-gui/issues/3909

src/reducers/alerts.js
benjiwheeler: I still don't totally undertand your point, but I changed the comment to `// intended to show standard and inline alerts, but not extensions` which I think is clearer

Issue #3269, opened 2018-10-01
Blocks stop reacting to key press after adding an asset from one of the libraries.


kchadha:
### Expected Behavior

Running project with if key pressed being checked in a loop should continue responding to key press events when working in other parts of the editor.

### Actual Behavior

If you have a project running e.g. ‘if key pressed’ in a loop under the green flag clicked hat, and you add a backdrop, sprite, costume, or sound **using the library** (this step is very important), the project stops reacting to key presses until you switch back to the blocks tab. 

This is not the case w/adding a surprise backdrop for example.

paulkaplan:
This appears to be another case of things like buttons/inputs having focus (buttons keep focus after they have been clicked) preventing keyboard events from making it to the VM.

PR #3874, opened 2018-11-26
Stop action menu focus stealing keyboard events from the VM

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/3269

### Proposed Changes

Automatically blur the action menu buttons when clicked to prevent it from blocking keyboard events from going to the VM and from reacting to future enter keys.

Tested by doing the following in develop vs. this branch
1. Load the editor
2. Add a "forever =>  if "a" key pressed => rotate"
3. Use it to make sure it is working
4. Open the sprite library
5. Add a new sprite
6. Without clicking anywhere else, try using the "a" key to rotate, it should work. It doesn't work on develop.

### DIFF
--- a/src/components/action-menu/action-menu.jsx
+++ b/src/components/action-menu/action-menu.jsx
@@ -78,6 +78,9 @@ class ActionMenu extends React.Component {
         return event => {
             ReactTooltip.hide();
             if (fn) fn(event);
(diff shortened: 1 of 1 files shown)


paulkaplan:
Hmm couldn't repro.

PR #3879, opened 2018-11-27
Track "project changed" state in Redux

rschamp:
Listen to the new project changed event from the VM and keep track of the state in Redux.  Update it as appropriate.

Known issue is that the project is marked as "changed" immediately after it's loaded. This is due to asynchronous event emitting while the project data is loaded, and needs an update to the VM to fix.

/cc @benjiwheeler @kchadha

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -328,7 +328,10 @@ class Blocks extends React.Component {
             error.message = `Workspace Update Error: ${error.message}`;
             log.error(error);
         }
-        this.workspace.addChangeListener(this.props.vm.blockListener);
+        // All of the changes that happened during the load above are queued with
+        // timeouts, so re-enable the listener in the next tick, so it happens after
+        // the events are already fired.
+        setTimeout(() => this.workspace.addChangeListener(this.props.vm.blockListener));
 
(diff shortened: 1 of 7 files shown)


src/containers/blocks.jsx
```diff
@@ -328,7 +328,10 @@ class Blocks extends React.Component {
             error.message = `Workspace Update Error: ${error.message}`;
             log.error(error);
         }
-        this.workspace.addChangeListener(this.props.vm.blockListener);
+        // All of the changes that happened during the load above are queued with
+        // timeouts, so re-enable the listener in the next tick, so it happens after
+        // the events are already fired.
+        setTimeout(() => this.workspace.addChangeListener(this.props.vm.blockListener));
```
paulkaplan: Does this work? I thought we tried and it didn't...

paulkaplan:
Also @rschamp not sure why the PR hasn't built on travis, is it because of the git conflict? Should be a simple resolution.

src/containers/blocks.jsx
rschamp: Something about this plus pulling out the extraneous targetUpdates in the vm when you switch sprites made it start working for me.

Issue #3418, opened 2018-10-17
In Safari, pressing the arrow keys (particularly “up” and “down”) while playing a project causes the page to scroll.


benjiwheeler:
@thisandagain writes:

> In Safari, pressing the arrow keys (particularly “up” and “down”) while playing a project causes the page to scroll.

benjiwheeler:
Resolved by https://github.com/LLK/scratch-gui/pull/3880

PR #3880, opened 2018-11-27
Prevent arrow/space from scrolling the page

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-www/issues/2210

### Proposed Changes

Prevent default behavior of arrows/spacebar when those keys are used against the document body (i.e. without an input or element focused. If an element is focused, all those keys will still do the thing they are meant to do).

### Reason for Changes

This is needed to make the project page work as expected. In the future, investing in ways of making those accessibility features work would be good (as mentioned by @towerofnix in https://github.com/LLK/scratch-www/issues/2210) But for now, it is broken for everyone because it both sends events to the VM _and_ scrolls the page.

### Test Coverage

We do not currently have any way of doing an integration test for this, but I tested it manually by building it into www and trying to use the spacebar directly when loading the project page.

### DIFF
--- a/src/lib/vm-listener-hoc.jsx
+++ b/src/lib/vm-listener-hoc.jsx
@@ -82,6 +82,12 @@ const vmListenerHOC = function (WrappedComponent) {
                 key: e.key,
                 isDown: true
             });
+
+            // Prevent space/arrow key from scrolling the page.
+            if (e.keyCode === 32 || // 32=space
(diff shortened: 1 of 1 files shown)


benjiwheeler:
Opened https://github.com/LLK/scratch-www/issues/2390 as a follow-up

benjiwheeler:
Also resolves https://github.com/LLK/scratch-gui/issues/3418

PR #3882, opened 2018-11-27
Green flag overlay to start in player mode

paulkaplan:
_note: This PR requires vm https://github.com/LLK/scratch-vm/pull/1783 `RUNTIME_STARTED` events to work correctly._
### Resolves

Not sure if there is an issue tracking this, but this implements an overlay green flag that starts the VM on click when the VM hasn't been started. This also changes when the VM gets started so it does not start automatically if in player mode. It does start if transitioning to editor mode, or when the green flag is clicked.

This PR also removes some cruft in the form of old unused green-flag/stop-all container files, which have seen been absorbed by the <Controls /> container.

In editor mode, so the default local build, you should see no difference. A project that starts automatically (using the when timer > 0 hack) will still start automatically.

In player mode (`/player.html` locally, or embedded in www) it does not start automatically, shows the green flag overlay.

@rschamp updated: 

Example in 2.0

image

Example in this PR

image

Note the missing monitors, but yeah @rschamp I adopted the renderer.draw technique


paulkaplan:
@rschamp Yup came to the same conclusion. Also @kchadha noticed that monitors appear in 2.0, and even run (!!), but not in 3.0 using this method.

PR #3890, opened 2018-11-27
Disconnect from cloud when entering editor mode on someone else's project

kchadha:
### Proposed Changes

- Add logic to disconnect the cloud provider as soon as a user views someone else's project in editor mode. If the user has ever entered editor mode in this session and cannot save the project, they should not be able to modify cloud data.

### Reason for Changes

Prevents users from modifying cloud data when editing someone else's project.

### Test Coverage

Added unit tests for the new functionality.

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


src/lib/cloud-manager-hoc.jsx
```diff
@@ -46,18 +46,22 @@ const cloudManagerHOC = function (WrappedComponent) {
             // This should cover info about the website specifically, like scrather status
             return !!(props.cloudHost && props.username && props.vm && props.projectId);
         }
+        shouldNotModifyCloudData (props) {
```
kchadha: This is not a great name, but writing this function so that the name is not a `negative` check makes the code less readable...

PR #3900, opened 2018-11-28
Add download and retry saving to saving error alert.

paulkaplan:
save-resave-demo

### Proposed Changes

Add the download and retry buttons to the `saveError` alert

I do not like how the alert component is hyper specialized for two distinct use cases. It would be much better to refactor the alerts so that the alerter can just fill the alert with whatever they want. But I do not feel comfortable making that refactor because I know @benjiwheeler and @evhan55  have done most of the work in that area, so I don't think it would be good for me to do that refactoring right now.

To that end, I just followed the pattern and added `showDownload` and `showSaveNow` flags and added the required dispatchers for the actions.

### DIFF
--- a/src/components/alerts/alert.jsx
+++ b/src/components/alerts/alert.jsx
@@ -22,7 +22,11 @@ const AlertComponent = ({
     iconSpinner,
     iconURL,
     level,
+    showDownload,
+    showSaveNow,
     onCloseAlert,
+    onDownload,
+    onSaveNow,
     onReconnect,
     showReconnect
 }) => (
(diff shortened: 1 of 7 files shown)


src/components/alerts/alert.jsx
```diff
@@ -53,6 +57,30 @@ const AlertComponent = ({
                 />
             ) : content}
         </div>
+        {showSaveNow && (
+            <button
+                className={styles.alertConnectionButton}
+                onClick={onSaveNow}
+            >
+                <FormattedMessage
+                    defaultMessage="Try&nbsp;Again" // TODO control wrapping in css
```
rschamp: Maybe we shouldn't localize this yet if we know it's going to change, don't want to unnecessarily ask people to re-translate (or try to translate strings with HTML entities and possibly break things when they do).

paulkaplan:
@rschamp fixed the layout problem that was forcing the `nbsp`. 

@evhan55 I totally agree, I would definitely want to do that. Something like how the different monitor types are separate components that get to totally determine what a monitor looks like. But I think I should wait for @benjiwheeler to be in the office to pair on that?

evhan55:
> @rschamp fixed the layout problem that was forcing the `nbsp`.
> 
> @evhan55 I totally agree, I would definitely want to do that. Something like how the different monitor types are separate components that get to totally determine what a monitor looks like. But I think I should wait for @benjiwheeler to be in the office to pair on that?

Yes, makes sense!

PR #3905, opened 2018-11-29
Send snapshots from the renderer to be saved as project thumbnails

paulkaplan:
This PR relies on https://github.com/LLK/scratch-www/pull/2365 for functionality, but will not break anything without it.

After saving a project, collect a snapshot of the renderer, blobify it and send it up to the onUpdateProjectThumbnail callback.

The structure of this is particularly defensive as to not block any of the project saving process or throw up errors when we do not want them to.

### DIFF
--- a/src/lib/project-saver-hoc.jsx
+++ b/src/lib/project-saver-hoc.jsx
@@ -201,16 +201,15 @@ const ProjectSaverHOC = function (WrappedComponent) {
                     storage.DataFormat.JSON,
                     body,
                     projectId
-                ).then(response => {
-                    this.props.onSetProjectUnchanged();
-                    return response;
-                });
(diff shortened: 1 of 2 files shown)


src/lib/project-saver-hoc.jsx
```diff
@@ -204,11 +206,36 @@ const ProjectSaverHOC = function (WrappedComponent) {
                     return response;
                 });
             })
+                .then(response => {
+                    const id = response.id.toString();
+                    if (id && this.props.onUpdateProjectThumbnail) {
+                        this.storeProjectThumbnail(id);
+                    }
```
rschamp: I think we also want to return `response` here so the create method can use it.

Issue #3684, opened 2018-11-07
Warn user about navigating away from page if project has not been saved


benjiwheeler:
See also https://github.com/LLK/scratch-gui/issues/3243

### Expected Behavior

Logged in, editing my project, I click the "Scratch" icon, or "My Stuff", etc. I get a warning popup that I can cancel out of and save before navigate away from my project.

### Actual Behavior

No warning, so some bad surprises :(

https://github.com/LLK/scratch-gui/pull/3638 will improve the auto-save experience. But there are outstanding problems. Navigating away can't be totally stopped until the save completes. So clicking `Community` and then leaving that page (by, say, hitting the back button, or reloading it) too rapidly will not finish saving first.

What's tricky is that since we don't have dirty tracking yet, there's no way to currently know whether the project has been saved -- so if we used it now, it'd have to fire on _every_ navigation away from the page. Otherwise, you'd be able to save, then try to navigate away, then face an annoying prompt.

When we have dirty tracking, we can introduce this.

PR #3913, opened 2018-11-30
warn user if they leave project that's not saved

benjiwheeler:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/3684

### Proposed Changes

project-saver-hoc now listens for window unload events, and before proceeding with browsing away from the page, checks if the project is dirty and prompts the user if it is.

Issue: since `projectChanged` is sometimes set to true when the vm is loaded -- a know issue, I believe -- this PR so far behaves wrong if you are looking at a project which you do not own; in that case, `projectChanged` seems to be always true, making it prompt you every time you navigate away from such a page, which is annoying.

Also, it's unclear to me whether I should delete the other window unload code in `playground/render-gui.jsx`. It seems to work fine with or without that code, possibly because it's being overridden by the listening in project-saver.

### Test Coverage

Unsure how to test this.

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/lib/project-saver-hoc.jsx
+++ b/src/lib/project-saver-hoc.jsx
@@ -50,7 +50,9 @@ const ProjectSaverHOC = function (WrappedComponent) {
         }
         componentWillMount () {
             if (typeof window === 'object') {
-                window.addEventListener('beforeunload', this.leavePageConfirm);
+                // Note: it might be better to use a listener instead of assigning onbeforeunload;
+                // but then it'd be hard to turn this listening off in our tests
+                window.onbeforeunload = () => this.leavePageConfirm;
             }
         }
         componentDidUpdate (prevProps) {
@@ -92,7 +94,7 @@ const ProjectSaverHOC = function (WrappedComponent) {
         }
         componentWillUnmount () {
             this.clearAutoSaveTimeout();
-            window.removeEventListener('beforeunload', this.leavePageConfirm);
+            window.onbeforeunload = undefined; // eslint-disable-line no-undefined
         }
         leavePageConfirm (e) {
             if (this.props.projectChanged) {
--- a/src/lib/project-saver-hoc.jsx
+++ b/src/lib/project-saver-hoc.jsx
@@ -52,7 +52,7 @@ const ProjectSaverHOC = function (WrappedComponent) {
             if (typeof window === 'object') {
                 // Note: it might be better to use a listener instead of assigning onbeforeunload;
                 // but then it'd be hard to turn this listening off in our tests
-                window.onbeforeunload = () => this.leavePageConfirm;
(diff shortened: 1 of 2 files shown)


benjiwheeler:
Hm, sounds test not passing because of alert prompt

test/integration/sounds.test.js
```diff
@@ -32,6 +32,7 @@ describe('Working with sounds', () => {
 
         // Delete the sound
         await rightClickText('Meow', scope.soundsTab);
+        await driver.sleep(500); // Wait a moment for context menu; only needed for local testing
```
paulkaplan: @benjiwheeler is this still needed,  or was it left in by mistake? Seems strange that the unload code would impact the right click actions

src/lib/project-saver-hoc.jsx
```diff
@@ -44,9 +44,17 @@ const ProjectSaverHOC = function (WrappedComponent) {
...
             bindAll(this, [
+                'leavePageConfirm',
                 'tryToAutoSave'
             ]);
         }
+        componentWillMount () {
+            if (typeof window === 'object') {
+                // Note: it might be better to use a listener instead of assigning onbeforeunload;
+                // but then it'd be hard to turn this listening off in our tests
+                window.onbeforeunload = () => this.leavePageConfirm;
```
paulkaplan: This code looks like it is just returning the function `this.leavePageConfirm`, rather than returning the result of `this.leavePageConfirm(event)`. This will result in the alert always showing–[MDN says](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload#Notes):
> When this event returns (or sets the returnValue property to) a value other than null or undefined, the user is prompted to confirm the page unload.

src/lib/project-saver-hoc.jsx
```diff
@@ -86,6 +94,15 @@ const ProjectSaverHOC = function (WrappedComponent) {
...
             this.clearAutoSaveTimeout();
+            window.onbeforeunload = undefined; // eslint-disable-line no-undefined
+        }
+        leavePageConfirm (e) {
+            if (this.props.projectChanged) {
+                // both methods of returning a value may be necessary for browser compatibility
+                (e || window.event).returnValue = true;
+                return true;
+            }
+            return false; // do not prompt
```
paulkaplan: I don't think this should return false, the spec linked in the other comment says that the function should return undefined, so this return can be removed (i.e. return undefined)

PR #3917, opened 2018-11-30
Update scratch-l10n dependency

chrisgarrity:
New scratch-l10n includes a script for pushing to transifex. Update package.json to use the new script for pushing the source to transfiex..


benjiwheeler:
LGTM

PR #3977, opened 2018-12-06
Updated Scratch Cat!

ericrosenbaum:
Thanks to @carljbowman for the beautiful and subtle update to the Scratch cat:

screen shot 2018-12-06 at 2 33 41 pm

- Update the Cat in the sprite and costume libraries
- Use the new cat in the default project


fsih:
And instantly every Scratch material ever became obsolete

PR #3984, opened 2018-12-07
Treat defaultValue from Blockly prompts properly

### LINKED ISSUE #3983: "Rename variable" prompt should have current name as default value
### Expected Behavior

When a user opens a prompt to rename a variable, the current name of the variable should be the default value in the text box. This would be useful for, for example, variables with long names where the user only wants to edit one part, or variable names with emojis where the user does not want to copy and paste every emoji they used.

### Actual Behavior

There is no default value.

### Steps to Reproduce

1. Make a variable
2. Rename the variable

Windows NT 10.0, Chrome 70

towerofnix:
### Resolves

Together with LLK/scratch-blocks#1807, resolves #3983. Should be merged **before** that PR. (Sorry for making a PR for a not yet help-wanted issue! But I'd already investigated the issue, and a PR is the cleanest way to share my changes.)

### Proposed Changes

Treats the `defaultValue` received from Blockly prompts as the default value for the `<input>`, not the placeholder. Removes the (otherwise unused) `placeholder` property from `<Prompt>` and adds a new `defaultValue` property in its place.

### Reason for Changes

Support for other code.

### Test Coverage

Tested manually (alongside the blocks PR).


towerofnix:
There were some merge conflicts I fixed - hopefully this should merge properly now.

src/containers/blocks.jsx
```diff
@@ -460,10 +460,11 @@ class Blocks extends React.Component {
                 />
                 {this.state.prompt ? (
                     <Prompt
+                        defaultValue={this.state.prompt.defaultValue}
                         isStage={vm.runtime.getEditingTarget().isStage}
                         label={this.state.prompt.message}
-                        placeholder={this.state.prompt.defaultValue}
                         showCloudOption={this.state.prompt.showCloudOption}
+                        showMoreOptions={this.state.prompt.showMoreOptions}
```
kchadha: @towerofnix, it looks like something got messed up during the merge conflict resolution. `showMoreOptions` got renamed to `showVariableOptions` a while ago and it looks like you might not have had those changes when you first started implementing this solution.

I think you just need to remove this line..

kchadha:
```suggestion
```

Issue #3736, opened 2018-11-13
Variable modal css looks different when embedded in www


kchadha:
### Expected Behavior

Styling of things should look the same in www and standalone.

### Actual Behavior

The options in the variable modal display so that the labels for the variable options are too close to the radio buttons/checkbox when the gui is embedded in www.

![image](https://user-images.githubusercontent.com/1786240/48428304-be330680-e738-11e8-8e4c-7f57b6627aac.png)

PR #3998, opened 2018-12-07
Force margin on inputs which is reset by normalizer in www

paulkaplan:
/cc @kchadha 

Fixes https://github.com/LLK/scratch-gui/issues/3736, the issue where the variable modal checkboxes/radios have the wrong padding.


kchadha:
It looks like the custom procedure modal needs a similar fix:

image

paulkaplan:
Included the fix in custom-procedures as well.

PR #4004, opened 2018-12-10
Do not open a new tab when clicking scratch logo

paulkaplan:
### Resolves

- Resolves an issue reported by @kathymakes that the scratch logo should not open a new tab, to try to keep from opening many instances of scratch (is that right @carljbowman?)

### Proposed Changes

Make it navigate in the same tab.


paulkaplan:
@carljbowman i don't know if that is being tracked. If you could collect a specific click-by-click of it happening, we could go through it together and see if it is possible to implement in javascript, which is a bit more strict about things you can do before navigating away from the page. The specific example you mentioned sounds like it would be possible though.

Issue #3277, opened 2018-10-01
After switching languages, the working area's scaling becomes 100%


thisandagain:
### Expected Behavior
![20180830_164819_20180830164945](https://user-images.githubusercontent.com/31377746/44841073-1ea3c300-ac75-11e8-9c5c-c7fdf6e74256.gif)
### Actual Behavior
![20180830_164442_20180830164704](https://user-images.githubusercontent.com/31377746/44841081-23687700-ac75-11e8-9e70-627a0268895a.gif)
### Steps to Reproduce
Do as the image
### Operating System and Browser

Win8, Chrome 68

---
Moved from LLK/scratch-blocks#1706
/cc @cyx2015s

julescubtree:
So it seems like a root cause here is that the component unmounts, and a new component mounts, on a language switch, and the scroll/scale info in the component state is lost.
Two possible solutions, maybe?
1) Switch languages without unmounting the component.
2) Keep workspace info (scroll and scale) in redux state.
I may be able to help with option 2; not sure about 1 (at least until digging into the code more).

thisandagain:
@julescubtree Option 1 is going to be difficult to implement. My leaning would be towards option 2.

epicfaace:
I've implemented option 2 in #4011 and the [zoom setting is able to be passed into the workspace config](https://github.com/LLK/scratch-gui/pull/4011/files#diff-2d67b4b152fdb4487af9345ff9c25487R97); however, it looks like scrollX and scrollY cannot be passed into the workspace config so we cannot set the scrollX and scrollY positions to the initial positions upon changing languages. Is there a way around this?

chrisgarrity:
Closing this issue because it's no longer a problem

PR #4011, opened 2018-12-10
Localization scaling

epicfaace:
### Resolves

After switching languages, the working area's scaling becomes 100% #3277

### Proposed Changes

- Store workspaceMetrics (scrollX, scrollY, scale) in Redux state.
- When the component mounts, pass in the scale parameter to the startScale option for scratch-blocks so that the state

### Reason for Changes

The entire component re-mounts when the language changes, so the zoom needs to be stored in the Redux state so that the component can revert back to the original zoom setting.

### Test Coverage

None

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -536,10 +536,16 @@ Blocks.propTypes = {
         comments: PropTypes.bool,
         collapse: PropTypes.bool
     }),
+    setWorkspaceMetrics: PropTypes.func,
     stageSize: PropTypes.oneOf(Object.keys(STAGE_DISPLAY_SIZES)).isRequired,
     toolboxXML: PropTypes.string,
     updateToolboxState: PropTypes.func,
-    vm: PropTypes.instanceOf(VM).isRequired
+    vm: PropTypes.instanceOf(VM).isRequired,
+    workspaceMetrics: PropTypes.objectOf({
+        scrollX: PropTypes.number,
+        scrollY: PropTypes.number,
+        scale: PropTypes.number
+    })
 };
 
 Blocks.defaultOptions = {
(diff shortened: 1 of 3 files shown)


julescubtree:
(Man, I was hoping to get to this eventually. Oh well, I'll find another bug.)

Last I checked, and I don't know if things'll continue to be this way, you can directly set `workspace.scrollX` and `workplace.scrollY`. So you might be able to do so, with the appropriate `scrollX`/`scrollY` for the active target, after `this.workspace` is set in `componentDidMount()`.

epicfaace:
@julescubtree I tried setting `workplace.scrollX` and `workspace.scrollY` directly, but it didn't work! Take a look at my attempted fix: https://github.com/epicfaace/scratch-gui/tree/feature/localization-scaling-attempted-fix

julescubtree:
Hmm. I took a quick look. It seems to be working part of the time, but not consistently. I've gotten it to more frequently not work on extreme zoom-outs, with a single block close to the edge of the workspace.

epicfaace:
What did you change to make it work better?

julescubtree:
I didn't change anything, actually—just ran a build of your attempted-fix branch and noticed that sometimes the scroll location persists, and sometimes it doesn't.

My hunch is that `scrollX` and `scrollY` are being set, but that scroll position is also set somewhere else, and that's interfering.

picklesrus:
Thanks for doing this @epicfaace! We're getting pretty close to the 3.0 launch and being conservative about what changes we merge in right now since we're starting to run low on testing time.  In January, once 3.0 is out and settled, I'll have some more time to take a closer look at this.  Sorry for the delay and thanks for your patience!

adroitwhiz:
Note that https://github.com/LLK/scratch-gui/pull/4542 also moves workspace metrics into a reducer; maybe compare this PR with that one

chrisgarrity:
I'm closing this PR as other changes appear to have fixed the issue. We can reopen if it re-appears.

PR #4087, opened 2018-12-14
Use 3.0 routes for GETs

rschamp:
This should have happened a while ago, but the Fastly config needed to be updated. It's updated now.

### Resolves

- Needed for https://github.com/LLK/scratch-projects/issues/102

### Proposed Changes

Update storage to use the 3.0 GET route, rather than legacy

### Reason for Changes

This route is simpler, and we can discern whether someone is requesting for 3.0 or 2.0 with this change.

### Test Coverage

They have been amended to support this change

### Browser Coverage
N/A

### DIFF
--- a/src/lib/storage.js
+++ b/src/lib/storage.js
@@ -36,7 +36,7 @@ class Storage extends ScratchStorage {
         this.projectHost = projectHost;
     }
     getProjectGetConfig (projectAsset) {
-        return `${this.projectHost}/internalapi/project/${projectAsset.assetId}/get/`;
+        return `${this.projectHost}/${projectAsset.assetId}`;
     }
     getProjectCreateConfig () {
         return {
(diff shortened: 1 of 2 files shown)


src/lib/storage.js
```diff
@@ -36,7 +36,7 @@ class Storage extends ScratchStorage {
         this.projectHost = projectHost;
     }
     getProjectGetConfig (projectAsset) {
-        return `${this.projectHost}/internalapi/project/${projectAsset.assetId}/get/`;
+        return `${this.projectHost}/${projectAsset.assetId}`;
```
colbygk: This is confusing.  Shouldn't this be referencing `assetHost` or something or are we calling the project JSON an asset too?

paulkaplan:
@rschamp can you port this over to www as well

PR #4125, opened 2018-12-18
Automatically update flyout checkboxes; add "hide" context menu option

### LINKED ISSUE #2032: Show/Hide blocks do not update the checkbox in the flyout
### Expected Behavior

Using the show/hide variable/list blocks should update the checkbox in the flyout

### Actual Behavior

They do not, the checkboxes stay the way they are.

### Steps to Reproduce

Click the "show variable" block to show the monitor for the default variable, and see that the checkbox does not update

n/a

---

@kchadha i'll put this on our "loose ends" list for monitors. The checkboxes are only updated (a) when the toolbox is refreshed (sprite switching) or (b) when the checkboxes are clicked. These blocks introduce a new way for the checkbox state to be toggled, so we need to figure out a way to push these changes to scratch-blocks. We should think about this also in the context of the "Hide" right-click menu item for monitors, which will be another way of changing the checkbox state outside of (a) and (b) above.

### LINKED ISSUE #4116: Add "Hide" option to monitor context menus
### Expected Behavior

In the context (right-click) menus of monitors (e.g. variables, reporter blocks like "answer", lists) there should be an option called "hide", which hides the monitor.

### Actual Behavior

There is no "hide" option! There isn't even a context menu at all on lists.

### Steps to Reproduce

Show a monitor (e.g. by showing a variable), then right-click on it.

N/A.

towerofnix:
### Resolves

Resolves #2032; resolves #4116.

### Proposed Changes

This PR does two things:

* It makes the monitor visibility checkboxes in the scratch-blocks flyout automatically update when variables are shown or hidden (#2032). (1d939b2)

* It adds a "hide" option to the context menus of monitors (#4116). To do this..
  * It adds CSS for "divider" options in `<ContextMenu>`, pulled from react-contextmenu. (f349d9a)
  * It adds a new context menu for list monitors (which previously didn't have one). (b13efcf)
  * It sends an "update monitor" request to the VM when the option is pressed. (e17b675)

### Reason for Changes

First I set out to fix #4116, which is a useful feature brought over from Scratch 2.0. Then I realized that the flyout checkboxes weren't automatically updating, which would be a very surprising and confusing non-effect, so I fixed that, already filed earlier in #2032.

The fix to #2032 is a handy change on its own since it's already a bug you'll notice when toggling variables with the show/hide blocks, but I included it in this PR because it's dramatically more noticeable when there's another manual option to hide the monitor.

### Test Coverage

Tested manually...maybe unit tests are appropriate for the auto-updating checkbox fix? (I don't have much experience with gui tests, though, iirc.)

Since the changes involve calling a function on scratch-blocks (`flyout.setCheckboxState`) for every monitor whenever any monitor is shown or hidden, I was worried there might be a hit on performance when variables are shown or hidden, [so I created a test project to check if there is](https://scratch.mit.edu/projects/271901648/). The result: there isn't! I think that [the "throttle" option in the `updateMonitors` dispatch](https://github.com/LLK/scratch-gui/blob/3a774f8c4cd7dfeca43b06aa7b121b3dfcc3a284/src/reducers/monitors.js#L20-L22) might be helping here.

### Browser Coverage

Tested on Linux Firefox and Chromium.

### DIFF
--- a/src/components/context-menu/context-menu.jsx
+++ b/src/components/context-menu/context-menu.jsx
@@ -13,7 +13,7 @@ const StyledContextMenu = props => (
 const StyledMenuItem = props => (
     <MenuItem
         {...props}
-        attributes={{className: styles.menuItem}}
+        attributes={{className: styles.menuItem, dividerClassName: styles.divider}}
     />
 );
 
--- a/src/components/context-menu/context-menu.css
+++ b/src/components/context-menu/context-menu.css
@@ -24,7 +24,14 @@
     transition: 0.1s ease;
 }
 
-.menu-item:hover {
+.menu-item:not(.divider):hover {
     background: $motion-primary;
     color: white;
+}
+
+.divider {
+    border-bottom: 1px solid $ui-black-transparent;
+    cursor: inherit;
+    margin-bottom: 3px;
+    padding: 2px 0;
 }
--- a/src/components/monitor/monitor.jsx
+++ b/src/components/monitor/monitor.jsx
@@ -54,36 +54,56 @@ const MonitorComponent = props => (
                 })}
             </Box>
         </Draggable>
-        {props.mode === 'list' ? null : ReactDOM.createPortal((
+        {ReactDOM.createPortal((
             // Use a portal to render the context menu outside the flow to avoid
             // positioning conflicts between the monitors `transform: scale` and
             // the context menus `position: fixed`. For more details, see
             // http://meyerweb.com/eric/thoughts/2011/09/12/un-fixing-fixed-elements-with-css-transforms/
-            <ContextMenu id={`monitor-${props.label}`}>
-                <MenuItem onClick={props.onSetModeToDefault}>
-                    <FormattedMessage
-                        defaultMessage="normal readout"
-                        description="Menu item to switch to the default monitor"
-                        id="gui.monitor.contextMenu.default"
-                    />
-                </MenuItem>
-                <MenuItem onClick={props.onSetModeToLarge}>
-                    <FormattedMessage
-                        defaultMessage="large readout"
-                        description="Menu item to switch to the large monitor"
-                        id="gui.monitor.contextMenu.large"
-                    />
-                </MenuItem>
-                {props.onSetModeToSlider ? (
-                    <MenuItem onClick={props.onSetModeToSlider}>
+            props.mode === 'list' ? (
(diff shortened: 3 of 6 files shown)


src/components/context-menu/context-menu.jsx
```diff
@@ -13,7 +13,7 @@ const StyledContextMenu = props => (
 const StyledMenuItem = props => (
     <MenuItem
         {...props}
-        attributes={{className: styles.menuItem}}
+        attributes={{className: styles.menuItem, dividerClassName: styles.divider}}
```
towerofnix: This line is causing a test to fail, for some reason (see travis).. I'm not really sure what's wrong though, since this (existing) "className" attribute doesn't cause a similar error, and as far as I can tell these are basically identical.

The code I took note of, BTW, is: https://github.com/vkbansal/react-contextmenu/blob/7580dc0d74b8cbd9ff7ff7f816c71f997f03717f/src/MenuItem.js#L55-L59

Would a better approach to this, rather than specifying `<MenuItem divider />`, be to have a new `StyledMenuItemDivider` / `<MenuItemDivider>` component? That component would specify `attributes.dividerClassName` while `StyledMenuItem` would not.

paulkaplan:
Hi @towerofnix! Thanks for this PR. We are busy getting ready for the 3.0 release, so there will be a bit of a delay reviewing this, but it'll get looked at in January.

src/components/context-menu/context-menu.jsx
towerofnix: @joker314 OK, sounds good - the diff of that does make sense in fixing it (`+ dividerClassName={undefined}`). That said, I'm curious *why* it's undefined, especially considering that `className` is set properly (`"react-context-menu-item"`). Pushed fix anyway, but, thoughts?

joker314: @towerofnix Not too sure, but you'll probably want to update https://github.com/LLK/scratch-gui/blob/develop/test/unit/components/sprite-selector-item.test.jsx (and then the snapshot too) to make it not undefined.

towerofnix:
@paulkaplan Sounds good. Breaking it up like that will also make it easier to work into your new PR #4146, which also adds a context menu to lists but with slightly different syntax. It'll take me some time to split it up but this looks like a good plan.

towerofnix:
Closing in favor of #4147 and #4148 (and a future PR for the visual divider).

Issue #4120, opened 2018-12-18
Upload from your computer does not save the current project


BryceLTaylor:
### Expected Behavior

If you have made changes to your project and you go to File>upload from your computer it should save your current changes and then make a new project with the data from the uploaded project

### Actual Behavior

Currently it does not save the old project before uploading the new one.

### Steps to Reproduce

Create a new project
Add a backdrop
Note: The upper right corner should say "Save Now"
Select File>Upload from your computer
Choose a project to upload from your computer
Click "My Stuff" folder
Open the first project
Note: the added backdrop is not present in the project.

### Operating System and Browser

Mac Chrome

BryceLTaylor:
This project is related to https://github.com/LLK/scratch-gui/issues/4123

PR #4134, opened 2018-12-18
When user uploads project file, project will auto-save before loading

benjiwheeler:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/4120

### Proposed Changes

* adds several states to `project-state` reducer to handle file upload
* introduces ability to indicate that a project load failed, and then to return to normal state with currently showing project; we should discuss this
* refactors `sb-file-uploader` so that it syncs with `project-state` state machine

Note that the titles still aren't set right; we could add that here or do it as a follow on. But, at least now the old project's title isn't overwritten.

Basic functionality: uploading a project saves current project first
main

The original project is indeed saved correctly:
orig-is-saved

Uploading a second project file, with no page reload (so keeping the same `FileReader` object), works correctly:
second

trying to upload an invalid file fails without taking you away from your current project:
invalid

### Reason for Changes

current project wasn't being saved before uploading project and saving as new project

### Test Coverage

Adjusted and added project state reducer tests

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/sb-file-uploader.jsx
+++ b/src/containers/sb-file-uploader.jsx
@@ -97,6 +97,7 @@ class SBFileUploader extends React.Component {
     onload () {
         if (this.reader) {
             this.props.onLoadingStarted();
+            const filename = this.fileToUpload && this.fileToUpload.name;
             this.props.vm.loadProject(this.reader.result)
                 .then(() => {
                     analytics.event({
@@ -109,10 +110,11 @@ class SBFileUploader extends React.Component {
                     this.props.onLoadingFinished(this.props.loadingState, true);
                     // Reset the file input after project is loaded
                     // This is necessary in case the user wants to reload a project
+                    if (filename) {
+                        const uploadedProjectTitle = this.getProjectTitleFromFilename(filename);
+                        this.props.onUpdateProjectTitle(uploadedProjectTitle);
+                    }
                     this.resetFileInput();
-                    // disable title handling for now; @todo reenable, with correct logic
-                    // const uploadedProjectTitle = this.getProjectTitleFromFilename(thisFileInput.files[0].name);
-                    // this.props.onUpdateProjectTitle(uploadedProjectTitle);
                 })
                 .catch(error => {
                     log.warn(error);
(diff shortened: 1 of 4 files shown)


src/containers/sb-file-uploader.jsx
```diff
@@ -60,35 +87,41 @@ class SBFileUploader extends React.Component {
...
+                    });
+                    // Remove the hash if any (without triggering a hash change event or a reload)
+                    history.replaceState({}, document.title, '.');
+                    this.props.onLoadingFinished(this.props.loadingState, true);
+                    // Reset the file input after project is loaded
+                    // This is necessary in case the user wants to reload a project
+                    this.resetFileInput();
+                    // disable title handling for now; @todo reenable, with correct logic
+                    // const uploadedProjectTitle = this.getProjectTitleFromFilename(thisFileInput.files[0].name);
+                    // this.props.onUpdateProjectTitle(uploadedProjectTitle);
```
paulkaplan: Why can't this line be uncommented?

paulkaplan:
@benjiwheeler I noticed that this causes a regression where the project title does not get updated in the standalone editor. If possible, I think we should get this in without introducing a regression?

src/containers/sb-file-uploader.jsx
benjiwheeler: I was trying to keep this change focused to fixing this one issue and not the title-switching one; I can't keep the current functionality. But let me try fixing that too, here -- it's probably not so hard

benjiwheeler:
I was trying to keep this change focused to fixing this one issue and not the title-switching one; I can't keep the current functionality. But let me try fixing that too, here -- it's probably not so hard.

paulkaplan:
@benjiwheeler yeah, it's just it is hard to revert PRs that break one thing in favor of another. We have to fix both before either goes in I think, so let's just include both here

benjiwheeler:
Just waiting for tests to pass, then will merge

PR #4144, opened 2018-12-19
On modal mount, push history if current history state doesn't match this modal's id

benjiwheeler:
### Resolves

- Resolves https://github.com/LLK/scratch-www/issues/2129

### Proposed Changes

Currently, on modal mount, we adjust the `window.history`. If `history.state === null`, we push a new history entry; if it's not null, we just replace history.

The problem with this is that if we've already navigated within the single page application, say from the project page to editor, then `history.state` is not null; so we just replace history when the library modal is open. Then if the user hits the back button, the browser navigates back to the previous `history` entry -- in this case, the project page.

This change instead checks `history.state === null || history.state !== this.id`, which is more permissive; if the `history.state` entry exists but is not the same as the current view, we push, instead of replacing.

### Reason for Changes

Browsing was behaving in an unexpected, frustrating way.

### Test Coverage

I added a test of the back button functionality, but note that it doesn't actually fail if this code change isn't included. Still, I think it's good to have.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/test/helpers/selenium-helper.js
+++ b/test/helpers/selenium-helper.js
@@ -14,6 +14,8 @@ class SeleniumHelper {
             'clickText',
             'clickButton',
             'clickXpath',
+            'elementIsVisible',
+            'elementIsNotVisible',
             'findByText',
             'findByXpath',
             'getDriver',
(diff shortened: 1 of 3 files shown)


test/helpers/selenium-helper.js
```diff
@@ -24,6 +26,13 @@ class SeleniumHelper {
         ]);
     }
 
+    elementIsVisible (element) {
+        return this.driver.wait(until.elementIsVisible(element));
+    }
+    elementIsNotVisible (element) {
```
kchadha: We should probably remove this since it's not being used anywhere

Issue #2054, opened 2018-05-14
List monitors: importing from CSV, exporting


paulkaplan:
This is an existing feature in 2.0. 

![image](https://user-images.githubusercontent.com/654102/40000044-bf2afa2c-5758-11e8-8d9d-da1269d1fa8a.png)

If the csv has more than one column, you are prompted to specify a column number.

Exporting is done as a "CSV" with one column, i.e. a newline separated list, it is saved as a `.txt` file. 

@carljbowman already has mockups of this, just has not been prioritized yet.

griffpatch:
This is still a missing feature that I use a lot... can't imagine doing without the import and export of lists to/from file.

PR #4146, opened 2018-12-19
Implement list import and export from files

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/2054

### Proposed Changes

This PR implements the import and export context menu items for lists with the same functionality as 2.0
- import csv, allow user to choose the column if there is more than one (bonus, tsv, since the parser module autodetects delimiter)
- export txt file, newline delimited (include both kinds of line endings for windows users)

There are some regression tests covering the context menu for the other variable types, so i'm confident this doesn't introduce regressions, but it is hard to make selenium test for this, I'll give it a try.

I know this is additional functionality so close to launch, so I'd be fine holding this if @kchadha you do not have time to review, etc.

### DIFF
--- a/src/components/monitor/monitor.jsx
+++ b/src/components/monitor/monitor.jsx
@@ -68,7 +68,7 @@ const MonitorComponent = props => (
                             id="gui.monitor.contextMenu.default"
                         />
                     </MenuItem>}
-                {props.onSetModeToDefault &&
+                {props.onSetModeToLarge &&
                     <MenuItem onClick={props.onSetModeToLarge}>
                         <FormattedMessage
                             defaultMessage="large readout"
--- a/src/components/monitor/monitor.jsx
+++ b/src/components/monitor/monitor.jsx
@@ -54,35 +54,52 @@ const MonitorComponent = props => (
                 })}
             </Box>
         </Draggable>
-        {props.mode === 'list' ? null : ReactDOM.createPortal((
+        {ReactDOM.createPortal((
             // Use a portal to render the context menu outside the flow to avoid
             // positioning conflicts between the monitors `transform: scale` and
             // the context menus `position: fixed`. For more details, see
             // http://meyerweb.com/eric/thoughts/2011/09/12/un-fixing-fixed-elements-with-css-transforms/
             <ContextMenu id={`monitor-${props.label}`}>
-                <MenuItem onClick={props.onSetModeToDefault}>
-                    <FormattedMessage
-                        defaultMessage="normal readout"
-                        description="Menu item to switch to the default monitor"
-                        id="gui.monitor.contextMenu.default"
-                    />
-                </MenuItem>
-                <MenuItem onClick={props.onSetModeToLarge}>
-                    <FormattedMessage
-                        defaultMessage="large readout"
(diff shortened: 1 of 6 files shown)


src/components/monitor/monitor.jsx
```diff
@@ -54,35 +54,52 @@ const MonitorComponent = props => (
...
-                {props.onSetModeToSlider ? (
+                {props.onSetModeToDefault &&
+                    <MenuItem onClick={props.onSetModeToDefault}>
+                        <FormattedMessage
+                            defaultMessage="normal readout"
+                            description="Menu item to switch to the default monitor"
+                            id="gui.monitor.contextMenu.default"
+                        />
+                    </MenuItem>}
+                {props.onSetModeToDefault &&
```
towerofnix: Should this be `props.onSetModeToLarge &&`?

paulkaplan: Yes it should! Thanks!

benjiwheeler:
A few quick thoughts from testing this:
* it works great
* is there any way to tell the user which file formats are supported when importing? Maybe the `import` menu item could be `import (.csv, .tsv)`? Or a confirm could give info before the filepicker opens?
  * Just thinking of a curious kid who notices that context menu and hasn't thought through how this could work; I remember being very confused as a kid by app interfaces that assumed I knew what I was looking for and immediately opened a filepicker.
* is there any way the exported files could be used for import, too? Maybe as `.csv`? It seems a shame not to allow that, and the file content itself works if I rename the `.txt` file to `.csv`.
* exporting from google sheets to csv
* is there any way we could enable pasting from a copied selection? (Am I remembering that you made that work as a demo?)

paulkaplan:
@benjiwheeler I added an allow for txt files to the file input, you are right symmetry is important with this. I didn't realize it, but scratch actually allows you to try importing any file, and does some pretty awesome/bad stuff when you import, say, a PDF file.

image

(supporting this in scratch3 would be a lot more work, btw)

I would love to follow up on some usability re: wording, file formats, etc post launch, since I'm sure there are some specific things from classrooms (i.e. science/data). I checked some basic things, like making sure that you can import from google sheets if you do "download as -> csv" (or tsv), but scratch2 doesn't support any of the excel-ish formats and nor does this.

paulkaplan:
@benjiwheeler I didn't see that when exporting from google sheets... did you have a bunch of rows selected or something like that?

towerofnix:
>is there any way the exported files could be used for import, too?

Just going to bring up that this is a critical feature of list export/import. It's the same concept as "export a sprite" and "import a sprite", and people use exporting/importing lists a *lot* as a means of exporting the save data for a game (and other types of data generated by a project). (nb: I haven't tried this PR so I might be misinterpreting here)

paulkaplan:
@towerofnix I already allowed importing txt files, so you can go back and forth. I don't want to export as a csv because it will not technically be a csv (it does not handle escaping commas)

benjiwheeler:
I was exporting the whole Google sheet -- but it formatted one column as drop-downs, maybe that was part of why there were so many empty rows.

PR #4147, opened 2018-12-19
Automatically update flyout checkboxes

### LINKED ISSUE #2032: Show/Hide blocks do not update the checkbox in the flyout

towerofnix:
Cherry-picked from #4125.

### Resolves

Resolves #2032.

### Proposed Changes

When the GUI sees (as emitted from the VM) that a variable has been updated, automatically update the checkboxes for all variables.

This isn't a perfect solution, since whenever *any* monitor is updated, a loop must call a function on *all* monitor checkboxes. This obviously indicates we're doing more than necessary. That said, there's no event for showing or hiding a specific monitor; only one for updates to any monitor (which includes only a map of the current monitor state, not a list of changes). Would it make sense to update the VM to add such monitor show/hide events, which would be used exclusively by this new code?

### Reason for Changes

Variable monitors can be shown/hidden by the "show" and "hide variable" blocks, and with #4116, by using the "hide" option in monitor context menus. It's confusing to users when the corresponding flyout checkbox doesn't automatically reflect that the monitor has been shown or hidden.

The flyout checkbox is intended to *control* whether or not a monitor of a block is visible, but also to *reflect* its visibility; this PR adds that "reflect" part.

### Test Coverage

Tested manually with the show/hide variable blocks, but once the code here is figured out (i.e. see "proposed changes"), there should be a new automatic test added.

### Browser Coverage

Tested on Linux Firefox and Chromium.


thisandagain:
@paulkaplan Bump

Issue #4159, opened 2018-12-19
Second set of sprites (and stage) get created


benjiwheeler:
To repro:

1. on staging, load the project page of a large project like Bouncy Heroes or Geometry Dash
2. before it fully loads, click See inside
3. the loading screen will take a long time
4. A few seconds or so after it loads... the sprites seem to duplicate themselves (the stage duplicates itself too!)

Here's me doing this with bouncy heroes, [https://scratch.ly/projects/1300000810](https://scratch.ly/projects/1300000810) :
![bh1](https://user-images.githubusercontent.com/3431616/50252195-37ff8500-03b4-11e9-9078-2e3e21a542fe.gif)

(continued)
![bh2](https://user-images.githubusercontent.com/3431616/50252196-37ff8500-03b4-11e9-99f8-370dcd65829c.gif)

...note the doubled sprites.

benjiwheeler:
Update after investigating:

So, in the project state sequence that happens when we load projects, there’s actually a key step where two different loading tasks happen simultaneously and asynchronously:

1. project is fetched from server, and
2. fonts are loaded.

the font loading isn’t actually a part of the project state reducer at all; it’s just kicked off by the GUI on construction.

where it relates to project state is that the vm manager hoc’s `componentDidUpdate` waits until project state becomes some form of LOADING_VM, **and** font loaded state is done. When either one becomes true, and both are now true, the vm loads in the project data.

Which works great! …when you are just in the editor, or just on the project page, or when you let either one fully load and then switch to the other.

But, if you switch between them right when the vm is already loading and not done yet, GUI is unmounted and reconstructed — and practically the first thing it does is set the fonts loaded state var to false.

And then loads fonts again, setting fonts loaded to true.

Along comes vm manager hoc — which is already busy loading the project data — and says, “Hey look, fonts just got loaded! And the current project state is LOADING_VM ! I know what that means… we’re ready to load project data into the vm!”

…so it loads it into the vm again.

PR #4160, opened 2018-12-20
refactored font loading state to be in redux

benjiwheeler:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/4159

### Proposed Changes

instead of maintaining the state of font loading in the `font-loader-hoc`, put it into redux.

### Reason for Changes

We need font loading state to be shared across project page and editor, so that fonts are not loaded a second time.

The second font load was triggering a second vm load of the project data.

### Test Coverage

none... not obvious what we would test

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/reducers/project-state.js
+++ b/src/reducers/project-state.js
@@ -452,7 +452,7 @@ const onLoadedProject = (loadingState, canSave, success) => {
                 type: DONE_LOADING_VM_WITHOUT_ID
             };
         default:
-            break;
+            return;
         }
     }
     return {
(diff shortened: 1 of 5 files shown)


src/reducers/project-state.js
```diff
@@ -452,7 +452,7 @@ const onLoadedProject = (loadingState, canSave, success) => {
                 type: DONE_LOADING_VM_WITHOUT_ID
             };
         default:
-            break;
+            return;
```
fsih: I don't understand this change. It might help to add a comment explaining how you could get into this state, and what you are doing in that case.

Issue #3721, opened 2018-11-10
Sprite info number inputs accept text


towerofnix:
### Expected Behavior

You shouldn't be able to type text in the sprite info inputs which are expected to only accept numbers, e.g. "size" or "x".

### Actual Behavior

You can type text in those inputs. Once you press enter or de-focus the input, the value will reset to whatever it was before.

### Steps to Reproduce

Type text into any of the number-based inputs in the sprite info area above the sprite palette (any of the ones in the screenshot below).

![The x, y, size and direction inputs in the sprite info area](https://user-images.githubusercontent.com/9948030/48307078-ded24500-e51b-11e8-8da3-ea79081b0eca.png)

### Operating System and Browser

Firefox 64.0b6, Debian (Linux).

---

See also LLK/scratch-blocks#1779, which is similar but not quite the same.

ericrosenbaum:
Update, literally years later: we're looking at this again, and it turns out that our team decided that the current validation behavior of these fields is what we want (i.e. allowing people to type in letters etc. even though they are later validated away).

PR #4192, opened 2018-12-24
Fix issue #3721 by checking if input is numeric (and letting BufferedInput be parameterized over a validation function)

dosaylazy:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/3721

### Proposed Changes

This commit gives the BufferedInput HOC a prop where the user can
provide a validation function. It also modifies the SpriteInfo and
DirectionPicker components to use this new prop to check that the
input is a number.

### Reason for Changes

This PR ensures that users can only type numbers into numeric fields. It parameterizes BufferedInput over a validation predicate so that clients can provide their own custom logic, making the BufferedInput component more modular.

### Test Coverage

Sorry, I haven't wrote tests yet. I tested my code by building Scratch-GUI locally and using the sprite inputs. I can write some tests later if you want.

### Browser Coverage

GNU/Linux Mint
 * [X] Chromium
 * [x] Firefox


ericrosenbaum:
Thanks for your contribution, and sorry about the delay. We've decided not to make the change suggested in https://github.com/LLK/scratch-gui/issues/3721 so we're going to close this one.

PR #4196, opened 2018-12-26
Remove Vivaldi block

### LINKED ISSUE #2370: Remove block for Vivaldi
[Edited by @thisandagain]

## Issue Scope
Remove block logic for Vivaldi as it is Chromium based and *should* work with Scratch 3.0. The Scratch Team will not officially support Vivaldi, but (for now) there is no reason to explicitly block it (unlike Opera Mini, Internet Explorer, and Silk which we know have breaking bugs).

Also update text in [the modal](https://github.com/LLK/scratch-gui/blob/a991e106087506ac89f2fd2f7d096c1dfc526e32/src/components/browser-modal/browser-modal.jsx#L35) to read:

>We are very sorry, but Scratch does not support this browser. We recommend trying a supported browser such as Google Chrome, Mozilla Firefox, Microsoft Edge, or Apple Safari.

## Original Description by @python-megapixel 
The warning for unsupported browsers claims that users on Vivaldi or Silk should switch to "a newer browser" such as Chrome or Firefox. This is an inaccurate description because Vivaldi was released in 2015 and Silk was released in 2011, while Chrome was released in 2008 and Firefox in 2002.

Cqoicebordel:
### Resolves

- Resolves #2370 

### Proposed Changes

Remove the case of Vivaldi in the blocking browsers logic

### Reason for Changes

Because there is no reason to block Vivaldi, as it's always based on a recent version of Chromium, and Scratch supports Chromium.

### Test Coverage

No test coverage but the change is trivial.

### DIFF
--- a/src/components/browser-modal/browser-modal.jsx
+++ b/src/components/browser-modal/browser-modal.jsx
@@ -32,7 +32,7 @@ const BrowserModal = ({intl, ...props}) => (
                 <p>
                     { /* eslint-disable max-len */ }
                     <FormattedMessage
-                        defaultMessage="We're very sorry, but Scratch 3.0 does not support Internet Explorer, Vivaldi, Opera or Silk. We recommend trying a newer browser such as Google Chrome, Mozilla Firefox, or Microsoft Edge."
+                        defaultMessage="We're very sorry, but Scratch 3.0 does not support Internet Explorer, Opera or Silk. We recommend trying a newer browser such as Google Chrome, Mozilla Firefox, or Microsoft Edge."
                         description="Unsupported browser description"
                         id="gui.unsupportedBrowser.description"
                     />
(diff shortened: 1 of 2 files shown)


thisandagain:
Thanks @Cqoicebordel! Can you please update the `gui.unsupportedBrowser.description` text with the copy I provided in #2370?

Cqoicebordel:
Done :)

Note that I didn't remove Silk from the browsers blocked. It was mentioned here and there that's why I'm writing this note, but the bug was clearly around Vivaldi, and the result of the discussion on Silk wasn't clear enough. I can easily remove the block for Silk if necessary though. The code is quite easy.
The remaining blocked browsers are MSIE, Opera and Silk, just FYI.

thisandagain:
@Cqoicebordel Just an FYI that we'll hold this to be merged after January 2nd.

Cqoicebordel:
Thanks for the info, and no problem, I'm just glad it will :)

Kenny2github:
@thisandagain January 2nd has come and gone, FYI if you weren't subscribed to the thread

Issue #2134, opened 2018-05-23
Player only mode (stage) should have loading animation


chrisgarrity:
The loading animation only shows on the full editor. When the player/stage is loaded by itself it should still have a loading animation.

PR #4210, opened 2018-12-27
Use the loading screen in player mode

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/2134

### Proposed Changes

Use the loading animation in player mode as well as editor mode.

### Reason for Changes

The lack of loading animation makes the project page view feel like it isn't working.

Here is what it currently looks like
loading-animation-without

With this change it looks like this
loading-animation

I simply passed in the loading prop to the stage wrapper so that it can be shown in player only mode.

I also changed the CSS for the loading animation so that it is position absolute instead of fixed, so that it can be contained within the stage size in player mode (but is still full screen in editor mode, which I made sure to test)

/cc @thisandagain @rschamp  I made this because i was concerned about the way the project page felt using larger projects. We can talk offline about whether you think this should go in before release or not.

### DIFF
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -141,6 +141,7 @@ const GUIComponent = props => {
             <StageWrapper
                 isRendererSupported={isRendererSupported}
                 isRtl={isRtl}
+                loading={loading}
                 stageSize={STAGE_SIZE_MODES.large}
                 vm={vm}
             >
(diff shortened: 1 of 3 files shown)


thisandagain:
Thanks @paulkaplan. This looks minimal enough, but I'll defer to you and @rschamp re: launch risk. This certainly would help initial loading feel more responsive to the user.

PR #4232, opened 2019-01-03
Wrap the stage canvas so that styles are not directly applied to it, fixing bounds calculations

adroitwhiz:
### Resolves

- Resolves https://github.com/LLK/scratch-render/issues/381

### Proposed Changes

This adds a wrapper to the stage's `<canvas>` element, which soaks up all the styles previously applied directly to the `<canvas>`.

It also rounds off the dimensions returned by `getStageDimensions()`.

### Reason for Changes

The first change makes it so that calling `getBoundingClientRect()` on the stage canvas will return the *inner* dimensions of the canvas, excluding the grey border applied to the stage.

The aforementioned function is used by `updateRect()` in containers/stage.jsx, which is then called by `componentDidUpdate()` to tell the renderer what size to render the stage to.

The problem with the old code arose with the fact that the renderer should be given the *inner* dimensions (excluding borders). By giving it the outer dimensions, the renderer will render the canvas too large. The canvas will then be downsized to the "proper" dimensions via the `width` and `height` CSS attributes, causing resampling and thus blurriness.

The second change also prevents resampling via CSS--the CSS scale of the canvas is determined by `getStageDimensions()` and non-integer values cause blurriness on some browsers.

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [x] Edge

Linux
 * [x] Chrome 
 * [x] Firefox 
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


paulkaplan:
@adroitwhiz thanks for this! It looks great. That css issue was definitely the problem for the regular size player.

PR #4304, opened 2019-01-09
Revise how unsupported browsers are handled

### LINKED ISSUE #4297: Change behavior for "Browser Incompatible" state
## Background
We are seeing lots of reports of fatal errors from users who (upon further investigation) are using older versions of browsers that are not supported (e.g. Safari 9, Chrome 49, etc). At the same time, we have had discussions with various Scratchers that are frustrated by us being fairly conservative with what platforms with support (e.g. Opera, Vivaldi). 

## Proposed Change
We should remove the current "Browser Unsupported" modal which blocks specific browsers and browser versions that are below our "supported" version threshold. Instead, we should only display this error state in the event of a fatal (e.g. BSOD) exception.

### Current State
- User loads a project running Safari 9
- User is immediately presented with a "Browser Unsupported" modal

### Future State
- User loads a project running Safari 9
- The application throw a fatal exception
- User is presented with a "Browser Unsupported" message

### Current State
- User loads a project running Opera 57
- User is immediately presented with a "Browser Unsupported" modal

### Future State
- User loads a project running Opera 57
- User can use Scratch

### Current State
- User loads a project running Chrome 61
- User is immediately presented with a "Browser Unsupported" modal

### Future State
- User loads a project running Edge on XBox
- User can use Scratch until they potentially encounter a feature / action which results in a fatal exception
- User is presented with a "Browser Unsupported" message

### Headline
```
An Error Occurred
```

### Body
```
We are very sorry, but it looks like you are using a browser that Scratch does not support. We recommend updating to the latest version of a supported browser such as Google Chrome, Mozilla Firefox, Apple Safari, or Microsoft Edge. 
```

chrisgarrity:
### Resolves

- Resolves #4297

### Proposed Changes

Browsers other than IE will be allowed to load the GUI.
However, if there is an error, the error boundary will display an unsupported browser message recommending they update to the latest version of a supported browser.

For the purposes of this PR, ‘supported’ means you can load the gui (anything other than IE), and recommended is just the browsers that meet our minimum requirements for Chrome, Firefox, Edge and Safari.

### Reason for Changes

Allow unsupported browsers (other than IE) to run the gui. If an unsupported browser gets an error, don't log it. Instead suggest the user install a supported version.

### Testing

- [ ] IE should still show unsupported browser without loading the gui
- [ ] Unsupported browsers should load the gui
- [ ] An error in an unsupported browser should show the message to install a supported version
- [ ] An error in a supported version should log the error, and let the user know that the error has been logged

e.g. Gui loaded in Opera:
screen shot 2019-01-09 at 1 14 17 pm

Error in Opera:
screen shot 2019-01-09 at 1 13 25 pm

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 * [x] Opera
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome
* [x] Silk


src/lib/supported-browser.js
```diff
@@ -1,15 +1,34 @@
...
- * Helper function to determine if the browser is supported.
+ * Helper function to determine if the browser is supported at all.
  * @returns {boolean} False if the platform is definitely not supported.
  */
-export default function () {
-    if (bowser.msie ||
-        bowser.opera ||
-        bowser.silk) {
+const supportedBrowser = () => {
+    if (bowser.msie) {
```
chrisgarrity: This could be `return !bowser.msie;`. I left it since there may be browsers other than IE that can't load gui at all.

PR #4359, opened 2019-01-16
Adds BorderedMenuItem and use it for delete menu

### LINKED ISSUE #4345: UX: Duplicate and Delete are right next to each other, causing deleted sprites.
### Expected Behavior

I recently ran a scratch camp with a bunch of kids using 3.0.  A lot of my old tutorials included 'stamping' sprites to duplicate them.  E.g. we would code one ghost sprite, then duplicate it to have multiple bad guys running along the screen.  

In 3.0 it is very easy to accidentally delete the ghost you just coded. 
image

Suggest that in the ContextMenu for the sprite selector item, a horizontal rule be added between actions that create/duplicate and actions that destroy.
    
https://github.com/LLK/scratch-gui/blob/57aed74998a4bd41df3caa13a0f9cab5fbe703bc/src/components/sprite-selector-item/sprite-selector-item.jsx

### Actual Behavior

Several kids accidentally deleted their sprites during camp because the stamp tool is no longer in 3.0.  

### Steps to Reproduce

Drink 4 cups of coffee
New scratch project
Right click on cat
Try to select duplicate, but accidentally choose delete

apple502j:
### Resolves
Resolves #4345 

### Proposed Changes
* Adds BorderedMenuItem, which has a border at the top and becomes red when hovered
* Use it for Delete menu (also reorder it)

Red for delete

### Reason for Changes
Some people accidentally delete sprites because there was no difference between this and the others

### Test Coverage
* Lint passed
* Checked on browser

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/context-menu/context-menu.css
+++ b/src/components/context-menu/context-menu.css
@@ -17,23 +17,20 @@
     z-index: $z-index-context-menu;
 }
 
-.menu-item, .menu-item-bordered {
+.menu-item {
     padding: 8px 12px;
     white-space: nowrap;
     cursor: pointer;
     transition: 0.1s ease;
 }
 
-.menu-item-bordered {
-    border-top: 3px solid $text-primary;
-}
-
-.menu-item:hover, .menu-item-bordered:hover {
+.menu-item:hover {
+    background: $motion-primary;
     color: white;
 }
 
-.menu-item:hover {
-    background: $motion-primary;
+.menu-item-bordered {
+    border-top: 2px solid $text-primary-transparent;
 }
 
 .menu-item-bordered:hover {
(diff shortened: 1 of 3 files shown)


carljbowman:
Sorry I was not able to respond quick enough yesterday, before this was merged.

@apple502j - Thanks for working on this. Great to have a divider style. We def need one in out Context Menu component. Also nice to shift the menu items specifically for this menu.

The separator is a little chunky and feels slightly inconsistent with the rest of the UI. It'd be great to reduce the border down to 1px and make it a transparent black, we should have one in GUI, possibly a 15% or 25%. @benjiwheeler - Do you mind making those changes?

We can keep the red hover for now, but it is something that we should continue to monitor.

PR #4362, opened 2019-01-16
Improve layout/paint performance in editor while running a project

paulkaplan:
This PR introduces a couple of layout/css tricks in order to vastly improve the layout+paint+composition performance of the editor while a typical projects is running. The main improvement is to apply the same layout trick that is used in the sprite library to allow the sprite selector images to change frequently without causing a layout of the target pane as a whole. That makes it so much less layout/painting has to happen:

birds-develop
birds-fixed

This improves the composition time by almost a half on both low-power machines (chromebook composition time goes from 20ms to 10ms) and high-power machines (my laptop goes from 5ms to 2.5ms).

I see an almost 50% improvement in frame times on the low-power chromebook using the "Birds" project (id=198711193) which was chosen from the front page.

Prior to this change, the react updates were a smaller part of frame time than the layout+paint+comp part, but this change makes the react portion the dominant blocker for frame time (on this project). 

FYI the changes to the selector/backpack is to make those work with the style changes on the sprite-selector-item, since that component is used in those places as well.

### DIFF
--- a/src/components/asset-panel/selector.css
+++ b/src/components/asset-panel/selector.css
@@ -56,7 +56,7 @@ $fade-out-distance: 100px;
 
 .list-item {
     width: 5rem;
-    min-height: 5rem;
+    height: 5rem;
     margin: 0.5rem auto;
 }
 
--- a/src/components/backpack/backpack.css
+++ b/src/components/backpack/backpack.css
@@ -69,9 +69,17 @@
 }
 
 .backpack-item {
-    min-width: 4rem;
-    max-width: 6rem;
+    width: 4rem;
+    height: 4.5rem;
     margin: 0 0.25rem;
+
+    /* Need to hide overflow because of background setting below */
+    overflow: hidden;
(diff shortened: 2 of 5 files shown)


src/components/asset-panel/selector.css
```diff
@@ -56,7 +56,7 @@ $fade-out-distance: 100px;
 
 .list-item {
     width: 5rem;
-    min-height: 5rem;
+    height: 5rem;
```
kchadha: should I be worried about this change from `min-height` to `height`? Was there a particular reason we had min-height before?

src/components/backpack/backpack.css
```diff
@@ -69,9 +69,17 @@
 }
 
 .backpack-item {
-    min-width: 4rem;
-    max-width: 6rem;
+    width: 4rem;
+    height: 4.5rem;
```
kchadha: Same question here

PR #4374, opened 2019-01-17
Revert "Adds BorderedMenuItem and use it for delete menu"

paulkaplan:
Reverts LLK/scratch-gui#4359

After talking with @carljbowman @benjiwheeler, we are going to bring this back out and work on the design before bringing it back. We don't want to block the high priority bugfixes that are going to production. Thanks for the submission @apple502j, sorry about the confusion.

### DIFF
--- a/src/components/context-menu/context-menu.css
+++ b/src/components/context-menu/context-menu.css
@@ -28,11 +28,3 @@
     background: $motion-primary;
     color: white;
 }
-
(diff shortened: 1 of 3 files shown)


rschamp:
Try it out: [revert-4359-delete-select](https://llk.github.io/scratch-gui/revert-4359-delete-select)

PR #4376, opened 2019-01-17
Fix delete button on sprite tiles

paulkaplan:
This fixes a bug I introduced in https://github.com/LLK/scratch-gui/pull/4362, by putting an absolutely positioned image above the "close" button that deletes sprites. 

I fixed the issue and added an integration test. Previous integration tests only tested the right-click -> delete sprite for some reason.


benjiwheeler:
LGTM, tested and delete buttons on sprites and costumes are visible and usable

PR #4384, opened 2019-01-18
Add tag for Centaur-a and Centaur-b

colbygk:
### Resolves

N/A

### Proposed Changes

Adds tag for Centaur-a and Centaur-b increasing searchability.

### Reason for Changes

Increases searchability

### Test Coverage

N/A


thisandagain:
These files are generated. Work with Eric and KO to update.

PR #4387, opened 2019-01-18
Throttle target updates to 500ms

paulkaplan:
Mostly this helps with longer sprite lists with rotating costumes, but generally removes the react bottleneck while running the project.

Had to update the direction picker to act like a "buffered input", i.e. store a temporary state to make dragging the dial smooth, and take updates from the VM when they come.

This brings the costume updates in the sprite list more in line with the scratch2 behavior, where it didn't change the tile image every frame.

/cc @picklesrus 

Before merging, I'd like to talk to @carljbowman about the design implications of this. I'll add it to the agenda for next team meeting too.

### DIFF
--- a/src/containers/direction-picker.jsx
+++ b/src/containers/direction-picker.jsx
@@ -12,11 +12,16 @@ class DirectionPicker extends React.Component {
             'handleClosePopover',
             'handleClickLeftRight',
             'handleClickDontRotate',
-            'handleClickAllAround'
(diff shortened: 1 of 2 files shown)


paulkaplan:
We are going to take a different approach to this because it can slow down certain interactions that we expect to be shown immediately, like drag reordering and updating properties while the project is running. We are going to see about a different approach that only throttles the costume updates.

PR #4396, opened 2019-01-19
Improve sound editor performance with longer sounds

paulkaplan:
Optimize the waveform render and the playhead animation to improve FPS while playing a sound, as well as overall usability of the sound editor when dealing with larger sounds.

The individual commits give more info on the specific optimizations, but TL;DR don't render so many points, don't render so often, and when you render don't cause a layout or composite.

This does cause a visual change in the waveform display, but only for sounds that are longer than anything we have in the library. It starts skipping points conservatively, at the point where there are more svg path data points than pixels :) 

Here is the display with a long sound on develop, I could only get 10fps while playing the sound
image

Smooth as butter at 60fps with these optimizations. Shows the difference in the waveform also
image

### DIFF
--- a/src/components/audio-trimmer/audio-trimmer.css
+++ b/src/components/audio-trimmer/audio-trimmer.css
@@ -12,6 +12,9 @@ $hover-scale: 2;
     left: 0;
     width: 100%;
     height: 100%;
+
+    /* Force the browser to paint separately to avoid composite cost with waveform */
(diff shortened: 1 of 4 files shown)


paulkaplan:
@picklesrus when reviewing this I recommend using the "diff options => hide whitespace changes" because most of the waveform change is indentation.

PR #4410, opened 2019-01-22
Delete Preview Modal

### LINKED ISSUE #4263: Remove preview modal from codebase
### Expected Behavior

The repository should I believe be for the most recent version of Scratch 3.0

### Actual Behavior

On open of http://localhost:8601/, Scratch 3.0 beta popup appears.

### Steps to Reproduce

-Follow instructions for Scratch 3.0 development version install
-Go to http://localhost:8601/
-Observe

My browser / operating system: Windows NT 10.0, Chrome 71.0.3578.98

nikhiljha:
### Resolves

- Resolves #4263 

### Proposed Changes

- Deleted code related to the startup modal.

### Reason for Changes

- See issues.

### Test Coverage

- Deleted code, no relevant tests were found.

### Browser Coverage

Mac
 * [X] Chrome 
 * [x] Firefox 
 * [X] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


apple502j:
You need to modify all tests.

paulkaplan:
Hi @nikhiljha, thanks for submitting your first contribution to scratch-gui. A couple of things:

1. When submitting a PR, try to make sure it only solves one problem at a time. This change removes old code and adds a new feature. Those are unrelated, so should go in two separate PRs.
2. The tests are failing as @apple502j mentioned. You can check the results by clicking "details" in the bottom section that says "All checks have failed". We can't merge a PR until it is passing.

src/components/gui/gui.jsx
```diff
@@ -157,9 +156,6 @@ const GUIComponent = props => {
                 dir={isRtl ? 'rtl' : 'ltr'}
                 {...componentProps}
             >
-                {previewInfoVisible ? (
```
paulkaplan: You need to remove `previewInfoVisible` from the list above on line 110. That is one of the things causing the lint errors.

src/components/menu-bar/save-status.jsx
```diff
@@ -40,6 +40,14 @@ const SaveStatus = ({
         </div>
     ));
 
+// Listen for Control/Command + S.
+document.addEventListener("keydown", function(e) {
```
paulkaplan: This cannot go here, because `onClickSave` is only defined inside the render function above. This listener can be attached in the menu-bar.jsx file, add a componentDidMount to add the listener and a componentWillUnmount where you remove the listener. 

[This guide](https://reactjs.org/docs/react-component.html#componentdidmount) might be helpful when dealing with React components

nikhiljha:
Thank you for explaining everything! I know you didn't have to, and I really appreciate it.

I tried to get the builds to pass, but for some reason Travis finds the sound board check to fail...

> Test Suites: 1 failed, 28 passed, 29 total
> Tests:       1 failed, 2 skipped, 203 passed, 206 total
> Snapshots:   1 failed, 4 passed, 5 total

but on my local machine

> Test Suites: 28 passed, 28 total
> Tests:       2 skipped, 198 passed, 200 total
> Snapshots:   5 passed, 5 total

do you have any idea of why this may be? Node and NPM versions on both systems are identical.

nikhiljha:
Whew, finally got all the tests to pass!

nikhiljha:
The tests are super inconsistent now, even though there should (?) be nothing that causes them to do that. I'll look at it more in a bit but right now it looks like things are failing because the tests have race conditions or something...

nikhiljha:
That's... pretty annoying actually.

I guess I could just add timer delays. An ideal solution would be to `await loadFinished()` or something, but oh well.

paulkaplan:
@nikhiljha interesting issue! I'll look into that, see if there is anything else we can use to know the editor is done loading

paulkaplan:
@nikhiljha actually it looks like there is one remaining "click try it" in the `costumes.test.js` in test at line 161. That is causing that test to fail, which I think sometimes causes a cascade of failures with selenium for some reason. Try removing that line and seeing if the tests are able to run again. I'll try locally, if you want to push that commit and we can see if travis will run it.

In general it shouldn't be a problem that the try it button isn't there because the way the selenium tests work they will usually wait a certain amount of time if an element is not yet visible, and keep retrying until that timeout expires. The only time this doesn't work is if the element is on the page but not clickable, in which case it fails right away. So there may still be failures after this fix, but let's see!

Thanks again for putting so much work into this. Sorry that the feedback time is slow, but this is very helpful for the scratch team, so many thanks.

nikhiljha:
Well that was easy, I just merged develop (and removed the new "Try Its" that that brought) and everything magically fixed itself.

I still have no idea why it was erroring in the first place though (and builds still fail on my local machine).

nikhiljha:
Reverse engineered the tests...

```
if (random(1,10) > 3) {
    test.fail();
}
```

[this](https://github.com/LLK/scratch-gui/pull/4410/commits/781c682e60b6c2326075c5d55550752e2ad0e249) should not cause tests to fail :(

paulkaplan:
@nikhiljha sorry about the trouble. The issue is that as this was in progress, more integration tests have been added to develop, all of which still click the "try it". So when you merged develop, you got some tests that still expect that button.

paulkaplan:
I pushed a commit to your branch that removes the new "try it" steps that had entered develop since you started. If all is well on travis my plan is to merge this and then update the current PRs against develop on top of this. Thanks for putting in the work on this.

Issue #2858, opened 2018-08-13
Uploading a sprite without any sounds while the sound tab is visible crashes sound editor


kchadha:
### Expected Behavior

Editor tabs should not crash.

### Steps to Reproduce

1. Open a new project in 3.0
2. Click on the sounds tab
3. Use the 'add sprite' action menu, file upload button to upload a sprite that does not have any sounds (The bananas sprite in the 2.0 library does not have any sounds, so is a good example of this -- export that sprite from 2.0 and upload it into this 3.0 project).
4. Observe that the sound editor crashes.

This also affects restoring the last deleted sprite if the last deleted sprite did not have any sounds.

PR #4415, opened 2019-01-23
Remove the redux throttle for targets update.

paulkaplan:
This fixes the root cause of #2858, which is caused because we are rendering the sound editor based on the GUI editing target data, but retrieving the sound based on the vms editing target data. The throttle of the redux action was causing those to get out of sync while a project was running.

The throttle is not needed because the vm runtime already batches those updates at 30 fps. The updates that come from top-level actions like switching sprites go out immediately.

This is also a minor performance boost, because there was a few ms being spent in setting and checking timers that were not needed.

/cc @picklesrus I'm glad we had just been talking about this, because i'm much more confident that this can be safely removed. 

@ericrosenbaum tagging you to just test the functionality since you found that bug.

### DIFF
--- a/src/reducers/targets.js
+++ b/src/reducers/targets.js
@@ -39,10 +39,7 @@ const updateTargets = function (targetList, editingTarget) {
     return {
         type: UPDATE_TARGET_LIST,
         targets: targetList,
-        editingTarget: editingTarget,
-        meta: {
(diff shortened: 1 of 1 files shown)


picklesrus:
Fixes https://github.com/LLK/scratch-gui/issues/4416

PR #4417, opened 2019-01-23
Throttle costume and watermark updates

paulkaplan:
### Resolves

Greatly improves performance of projects, specifically on low-power machines.

This is a more specific follow up to #4387, after talking to people about ux implications

### Proposed Changes

Throttle updates to sprite list costumes to 500ms. Does not throttle any other sprite info/list property. Also throttle the watermark updates.

Introduces a new HOC to handle this specific behavior, where we want to throttle updates on a specific prop.

### Reason for Changes

Costume updates come in too quickly and are not very useful.

@carljbowman this avoids the UX problems identified in our group meeting. Specifically:
- it always allows updates to all the sprite info values, so interacting with the sprite info values or reordering sprites is always real-time.
- it is throttled on a per-sprite-tile basis, so costume changes are instantaneous if you make them manually on a sprite that isn't already updating.

### Test Coverage

Added unit test coverage for the new HOC

### Browser Coverage

Tested performance using "Birds" [198711193] on one of the older chromebooks.

---
Paired on this with @picklesrus

### DIFF
--- a/src/components/sprite-selector/sprite-list.jsx
+++ b/src/components/sprite-selector/sprite-list.jsx
@@ -12,7 +12,7 @@ import ThrottledPropertyHOC from '../../lib/throttled-property-hoc.jsx';
 
 import styles from './sprite-selector.css';
 
-const ThrottledSpriteSelectorItem = ThrottledPropertyHOC('asset', 500, SpriteSelectorItem);
+const ThrottledSpriteSelectorItem = ThrottledPropertyHOC('asset', 500)(SpriteSelectorItem);
 
 const SpriteList = function (props) {
     const {
--- a/src/components/sprite-selector/sprite-list.jsx
+++ b/src/components/sprite-selector/sprite-list.jsx
@@ -8,8 +8,11 @@ import Box from '../box/box.jsx';
 import SpriteSelectorItem from '../../containers/sprite-selector-item.jsx';
 import SortableHOC from '../../lib/sortable-hoc.jsx';
 import SortableAsset from '../asset-panel/sortable-asset.jsx';
+import ThrottledPropertyHOC from '../../lib/throttled-property-hoc.jsx';
 
 import styles from './sprite-selector.css';
+
+const ThrottledSpriteSelectorItem = ThrottledPropertyHOC('asset', 500, SpriteSelectorItem);
 
 const SpriteList = function (props) {
     const {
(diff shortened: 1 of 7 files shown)


src/components/sprite-selector/sprite-list.jsx
```diff
@@ -8,9 +8,12 @@ import Box from '../box/box.jsx';
 import SpriteSelectorItem from '../../containers/sprite-selector-item.jsx';
 import SortableHOC from '../../lib/sortable-hoc.jsx';
 import SortableAsset from '../asset-panel/sortable-asset.jsx';
+import ThrottledPropertyHOC from '../../lib/throttled-property-hoc.jsx';
 
 import styles from './sprite-selector.css';
 
+const ThrottledSpriteSelectorItem = ThrottledPropertyHOC('asset', 500, SpriteSelectorItem);
```
rschamp: This doesn't really matter but I think the convention for configuring HOCs is `ThrottledPropertyHOC('asset', 500)(SpriteSelectorItem)`

src/lib/backpack/sprite-payload.js
```diff
@@ -1,23 +1,29 @@
...
+        const payload = {
+            type: 'sprite',
+            name: target.sprite.name,
+            mime: 'application/zip',
+            body: zippedSprite,
+            // Filled in below
+            thumbnail: ''
+        };
+
+        const costumeDataUrl = target.sprite.costumes[target.currentCostume].asset.encodeDataURI();
```
rschamp: This is a suspiciously deep reference, seems like some guards would be prudent before assuming we can `.` our way all the way here.

src/lib/backpack/sprite-payload.js
```diff
@@ -1,23 +1,29 @@
...
-).then(zippedSprite => {
-    const payload = {
-        type: 'sprite',
-        name: sprite.name,
-        mime: 'application/zip',
-        body: zippedSprite,
-        // Filled in below
-        thumbnail: ''
-    };
+const spritePayload = (id, vm) => {
```
rschamp: I don't understand this change, is it an additional performance bump, or related to the throttling?

paulkaplan: Good question. This was a necessary bugfix that we needed to make before the throttledpropertyhoc could work. There was an overly greedy change to the dragPayload which was causing shouldComponentUpdate to always return true.

src/lib/backpack/sprite-payload.js
paulkaplan: It is using things that are always coming from the VM, so I feel ok about it. Also if it fails, the payloading should fail which is caught elsewhere anyway

src/components/sprite-selector/sprite-list.jsx
paulkaplan: Right! Fixed

PR #4433, opened 2019-01-25
Control/Command+S to save

### LINKED ISSUE #4262: Pressing Ctrl-S should save the project
### Expected Behavior

Ctrl+S is a really common keyboard shortcut in almost any program, so pressing it should cause Scratch to save the project (just like pressing the "Save Now" button).

### Actual Behavior

Ctrl+S doesn't do anything in Scratch 3.0 (it just opens the browser save-page-as dialog which, yes, [can be overridden](https://stackoverflow.com/a/14180949/4633828)).

### Steps to Reproduce

Press Ctrl+S while in the 3.0 editor.

towerofnix:
### Resolves

Resolves #4262.

### Proposed Changes

Makes it so that when Control+S (Command+S, if on macOS) is pressed, the "save now" button is effectively pressed, and the browser's save-as dialog is not displayed.

### Reason for Changes

It's a useful and familiar keybinding to introduce to Scratch.

### Test Coverage

No new tests (there didn't seem to be any for the menu bar component yet).

I tested this manually by confirming that the browser "Save page as..." dialog is prevented from appearing. Because "save now" is not a feature in non-www scratch-gui, this PR has no effect on the standalone editor (Scratch Desktop or llk.github.io, etc). If wanted, I could probably make it so that pressing Ctrl+S opens up the "save to your computer" dialog if "save now" is unavailable. This isn't as smooth as simply overwriting an existing file, what Ctrl+S does on any native app, but it may be better than nothing. (Of course file overwriting is outside the scope of this PR - once it's done, the relevant code added by this PR could be modified.) Thoughts?

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Linux
* [x] Firefox

### DIFF
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -227,7 +227,8 @@ class MenuBar extends React.Component {
         };
     }
     handleKeyPress (event) {
-        if (event.key === 's' && event.ctrlKey) {
+        const modifier = (window.navigator.platform === 'MacIntel') ? event.metaKey : event.ctrlKey;
+        if (modifier && event.key === 's') {
             this.props.onClickSave();
             event.preventDefault();
         }
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -151,10 +151,17 @@ class MenuBar extends React.Component {
             'handleClickSeeCommunity',
             'handleClickShare',
             'handleCloseFileMenuAndThen',
+            'handleKeyPress',
             'handleLanguageMouseUp',
             'handleRestoreOption',
             'restoreOptionMessage'
         ]);
+    }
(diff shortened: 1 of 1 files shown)


paulkaplan:
@cwillisf do you think this would break anything with scratch desktop? Should we only bind these shortcuts if it isn't in desktop?

cwillisf:
@paulkaplan I think it should be OK in Scratch Desktop. The replacement save dialog is is implemented by hooking Electron's (Chromium's) "will save" event so it should just work. It'll need testing, of course, but I don't think we need to block this PR for that reason.

towerofnix:
@paulkaplan Sure, pushed: 528b3412d6f3810754a40e30ddbce64b21ec0a15

PR #4438, opened 2019-01-27
Export costumes and sounds!

### LINKED ISSUE #4437: Export assets
_Thanks to all Scratchers who found this problem, and thanks to kenny2github for finding out that this issue is not created on GitHub yet!_

### Expected Behavior
All assets should be exportable.

### Actual Behavior
There is no way to do so.

### Steps to Reproduce
1. Right-click costumes, sounds or backdrops. No way to export it.

any

### See also
ceebee: ["You can’t export sounds or costumes: This is another feature we do plan to implement in the future."](https://scratch.mit.edu/discuss/post/3388941/)

NOTE: I'll submit a pr for this soon.

apple502j:
### Resolves
Resolves #4437

### Proposed Changes
Adds export feature for costumes/backdrops/sounds! :tada: :laughing: 

### Reason for Changes
Everyone wanted it!

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


IamTregsthedev:
I can Help you with the chromebook part.

i have a chromebook, and will be willing ot help

apple502j:
@IamTregsthedev You should make another issues if it has problem with Chromebook.
First, this should be merged early, because so many people want it to be implemented.
Second, some other actions (saving projects or lists) may be glitching, not only this.

This code should work for Chromebook, so I can't help you unless you have a problem on Chromebook.

paulkaplan:
This looks great! I pushed a commit to your branch to remove the duplication between the `download-text` and `download-url` utilities, using your new utility instead since it is more generic.

Works on a chromebook from my testing. My only concern would be iOS, but haven't tried that and know that there are existing issues with downloading anything on iOS, so I'm not too concerned about that.

@carljbowman @thisandagain My only question is really whether or not we want to inline the fonts in the downloaded SVG files (which is what `getCostumeUrl` does). It would allow people to edit the SVGs in other programs while seeing the correct fonts, and I think we are ok to distribute these fonts with downloaded assets, but I think it is still an open question because it can make the downloads much bigger. I'll chat offline with the team about this, but I'm going to merge your PR for now, and if we feel differently about it later we can change it.

I'm working on some integration tests for this as well, but currently headless chrome doesn't play nicely with local file downloads, so I am a bit stuck :(

thisandagain:
🎉 Great. Thanks @apple502j and @paulkaplan.

Re: inline fonts. My inclination is to keep the scope of this as-is and then follow-up should we see significant demand for it from the community in the future.

paulkaplan:
@thisandagain sorry i may have been unclear. As implemented, this *does* inline the fonts into downloaded SVG costumes that use text, since that was actually simplest in this case.

thisandagain:
@paulkaplan Oh sorry. I should have looked at the code. 😅 

That actually seems ideal. When these assets are then imported back into Scratch I would just want to make sure that this data is removed before the asset is saved to Scratch's servers.

PR #4447, opened 2019-01-28
Move rounding into sprite info to improve performance

paulkaplan:
For projects with large numbers of sprites, the way we use object spread in mapStateToProps for targetPane is causing both a lot of garbage collection and a lot of extra work because of the way object spread is transpiled to vanilla JS.

It was costing almost 10ms (about 100% of the frame budget for GUI updates) on [this project](https://scratch.mit.edu/projects/247892564/).

### DIFF
--- a/src/containers/target-pane.jsx
+++ b/src/containers/target-pane.jsx
@@ -248,22 +248,18 @@ TargetPane.propTypes = {
     ...targetPaneProps
 };
 
-const mapStateToProps = state => ({
-    editingTarget: state.scratchGui.targets.editingTarget,
-    hoveredTarget: state.scratchGui.hoveredTarget,
-    sprites: Object.keys(state.scratchGui.targets.sprites).reduce((sprites, k) => {
-        let {direction, size, x, y, ...sprite} = state.scratchGui.targets.sprites[k];
-        if (typeof direction !== 'undefined') direction = Math.round(direction);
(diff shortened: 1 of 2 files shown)


src/containers/target-pane.jsx
```diff
@@ -248,22 +248,18 @@ TargetPane.propTypes = {
...
+const mapStateToProps = state => {
+    const props = {
+        editingTarget: state.scratchGui.targets.editingTarget,
+        hoveredTarget: state.scratchGui.hoveredTarget,
+        sprites: state.scratchGui.targets.sprites,
+        stage: state.scratchGui.targets.stage,
+        raiseSprites: state.scratchGui.blockDrag,
+        spriteLibraryVisible: state.scratchGui.modals.spriteLibrary
+    };
+    return props;
```
rschamp: Maybe cruft, but this seems like it could be simplified to the same syntax as it was:
```js
    editingTarget: state.scratchGui.targets.editingTarget,
    hoveredTarget: state.scratchGui.hoveredTarget,
    sprites: state.scratchGui.targets.sprites,
    stage: state.scratchGui.targets.stage,
    raiseSprites: state.scratchGui.blockDrag,
    spriteLibraryVisible: state.scratchGui.modals.spriteLibrary
});
```

paulkaplan: Good point! fixed

PR #4449, opened 2019-01-28
Only update the toolbox when the blocks tab is visible.

paulkaplan:
Need to store the rendered xml because the props can change while the blocks tab is hidden, but we re-render based on what was rendered, not the previous props (which may not have been rendered).

This is similar to the `ThrottledPropertyHOC` in that there may be prop changes that do not get rendered, but in this case for a different reason.

This increases the performance of switching between code and other editor tabs, as well as improving the "See Inside" performance. It makes switching to the code tab as fast as switching between sprites.

@picklesrus and I talked through most of the situations we could imagine, and we have an integration test that covers switching sprites both while visible and not visible. But @BryceLTaylor we'll still want to make sure to pay special attention to this when testing, because this is an area that is easy to get wrong. We'll know we did it wrong if, for example, the toolbox has blocks that have the wrong padding (tried rendering while invisible) or if the block menus or blocks are not correct for the sprite you are on.

Performance-wise, this makes switching between code/costume/sound tabs much faster, by about 100ms on my machine (more on lower power machines).


kchadha:
Haven't tested the changes myself, but things to think about/test:

- adding an extension
- changing something about blocks (e.g. the name of a the first costume of a sprite) while not in the blocks tab and switching back
- changing things about variables or broadcast messages (though I think the blocks tab should always be visible for this)
- switching languages
    - while in the blocks tab
    - while not in the blocks tab

chrisgarrity:
> * switching languages
>   * while in the blocks tab
>   * while not in the blocks tab
Yes, I remember there were issues around switching language while not in the blocks tab. The update toolbox needed to happen with the block caching turned off. I think blocks updated, but any loaded extensions didn't, or maybe vice versa.

kchadha:
Amendment to the 'adding an extension' suggestion:

- Try exporting a sprite with an extension, create a new project, switching to the costumes tab, upload the exported sprite

src/containers/blocks.jsx
```diff
@@ -155,7 +165,6 @@ class Blocks extends React.Component {
                 this.setLocale();
             } else {
                 this.props.vm.refreshWorkspace();
-                this.updateToolbox();
```
kchadha: Do you know what this case was covering before? I'm confused about the `if (this.props.isVisible === prevProps.isVisible)` check we're in.

kchadha: One small bug that I'm noticing is that if I'm in the blocks tab and click on a category (e.g. Variables), switch to the costumes tab, resize the browser window (e.g. make it smaller), and switch back to the blocks tab, the variables category is still selected but the toolbox is scrolled back up to the motion category.

This seems relatively small and can probably be filed as an issue for later.

paulkaplan: @kchadha that seems to be a pre-existing issue, i can repro that on current production.

src/containers/blocks.jsx
```diff
@@ -297,12 +308,32 @@ class Blocks extends React.Component {
...
+        }
+    }
     onWorkspaceUpdate (data) {
         // When we change sprites, update the toolbox to have the new sprite's blocks
-        if (this.props.vm.editingTarget) {
-            const target = this.props.vm.editingTarget;
-            const dynamicBlocksXML = this.props.vm.runtime.getBlocksXML();
-            const toolboxXML = makeToolboxXML(target.isStage, target.id, dynamicBlocksXML);
+        const toolboxXML = this.getToolboxXML();
+        if (toolboxXML) {
```
kchadha: What gets displayed when this is not true?

paulkaplan: It doesn't change the toolbox in this case. This would be a pretty strange condition though, it only happens if certain expectations aren't met (a sprite with no costumes, there is no stage, etc.)

test/integration/blocks.test.js
```diff
@@ -174,4 +174,58 @@ describe('Working with the blocks', () => {
         const logs = await getLogs();
         await expect(logs).toEqual([]);
     });
+
+    test('Renaming costume changes the default costume name in the toolbox', async () => {
```
kchadha: 🎉

test/integration/blocks.test.js
```diff
@@ -174,4 +174,58 @@ describe('Working with the blocks', () => {
...
+
+        // Make sure it is updated in the block menu
+        await clickText('Code');
+        await clickText('Looks', scope.blocksTab);
+        await driver.sleep(500); // Wait for scroll to finish
+        await clickText('newname', scope.blocksTab);
+    });
+
+    // NOTE: This test describes the current behavior so that changes are not
+    // introduced inadvertly, but I know this is not the desired behavior
```
kchadha: Do we have an issue filed for the desired behavior? If so, maybe add todo comments here?

src/containers/blocks.jsx
kchadha: @paulkaplan the behavior is not the same on production. On production, the toolbox scrolls to the top and the selected category also switches to 'Motion', with these changes, the category stays as 'Variables' even though you're seeing 
'Motion' blocks

Issue #4458, opened 2019-01-30
Dragging blocks from backpack should generate new IDs


kchadha:
Dragging code from the backpack into a workspace should generate new block IDs for all the blocks in the stack.

Currently it doesn't which causes bugs like:
- cannot drag a block from the backpack into the same workspace twice
- issues with multiple sprites having the same block ID (this will create issue with two *different* variables having the same ID), also creates issues with project saving, where switching to two workspaces that have the same code from the backpack emits a project changed state

PR #4468, opened 2019-02-01
Give code from the backpack new IDs before loading into the VM

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/4458

Note that I based this on the commit from https://github.com/LLK/scratch-gui/pull/4467 since it is hard to see what is going on without it, but the code is completely separate, so that PR should be merged first then I can remove it from this PR.

### Proposed Changes

Code stored in the backpack has a new set of IDs from the code it was originally generated from, but in order to add the same code multiple times it needs to get new IDs each time. 

I wrote a simple utility for mutating the blocks array in two passes to give new IDs then update the input references and next/parent references to heal the connections.

### Reason for Changes

This is a unique problem that is not currently addressed by the VM, where code as JSON needs to be uniquified on the block level (rather than the variable ID reference level, e.g.), so I put the utility in the GUI and added tests for each type reference that might need to be healed. 

I generated a bunch of block JSON and then looked at all the fields that could contain ID references, and added a test for each of them. 

@kchadha could you confirm that the only block properties that contain IDs are:
- block.next
- block.parent
- block.inputs[inputName].block
- block.inputs[inputName].shadow

### Test Coverage

Added a unit test for the new utility.

Manually I tested a few cases I was concerned might be particularly bad:
- Add a new variable, drag the "change by" into the backpack, switch to the stage and drag it back out onto the stage. Confirmed new block changes the same variable as the old blocks.
- After above test, change the name of the new variable to something else. Drag the "change by <old name>" block out of the backpack again, and see that it magically updates to be "change by <new name>". Cool! I hope...
- Put a custom block definition in the backpack that contains usage of its parameters in the definition stack. Drag that stack out of the backpack to a new sprite or new project. Make sure the param references still work right.
- Pulling a custom block definition out of the backpack more than once... creates multiple definitions, and multiple call blocks in the toolbox, but all the call blocks only use a single definition stack. This is a pre-existing issue with custom blocks that can have the same names (and can currently be replicated with share-the-love), so I don't think it needs fixing here.

### DIFF
--- a/src/lib/new-block-ids.js
+++ b/src/lib/new-block-ids.js
@@ -1,0 +1,31 @@
+const uid = () => String(Math.floor(1000000 * Math.random()));
+
+
+export default blocks => {
+    const oldToNew = {};
+
+    // First update all top-level IDs and create old-to-new mapping
+    for (let i = 0; i < blocks.length; i++) {
+        const newId = uid();
+        const oldId = blocks[i].id;
+        blocks[i].id = oldToNew[oldId] = newId;
+    }
+
+    // Then go back through and update inputs (block/shadow)
+    // and next/parent properties
+    for (let i = 0; i < blocks.length; i++) {
+        for (const key in blocks[i].inputs) {
+            const input = blocks[i].inputs[key];
+            input.block = oldToNew[input.block];
+            input.shadow = oldToNew[input.shadow];
+        }
+        if (blocks[i].parent) {
+            blocks[i].parent = oldToNew[blocks[i].parent];
+        }
+        if (blocks[i].next) {
+            blocks[i].next = oldToNew[blocks[i].next];
+        }
+    }
+
+    return blocks;
+};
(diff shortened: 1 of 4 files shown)


kchadha:
@paulkaplan I haven't looked at the changes closely yet, but I think it might make more sense to move this code into the VM as it will need to be reused for duplicating a sprite as well.

src/lib/new-block-ids.js
```diff
@@ -0,0 +1,31 @@
+const uid = () => String(Math.floor(1000000 * Math.random()));
```
kchadha: This uid function should be the same as the one we use for the VM.

kchadha:
Another reason for moving this into the VM (and I can't decide if this is important to fix now or not) is that we're not handling variable IDs correctly. 

Currently, if you add a variable block from the backpack into a new project that already has a variable of the same name, there is a warning in the console about conflicting IDs. The project seems to be working regardless of this warning so it's probably not critical to fix now, but definitely worth tracking in an issue. We would probably need to fix this in the VM if we do end up fixing it.

paulkaplan:
Closing this in favor of https://github.com/LLK/scratch-vm/pull/1979

Issue #4476, opened 2019-02-04
Many blocks untranslated when loaded in another language


chrisgarrity:
### Expected Behavior

_Please describe what should happen_
When the editor is loaded with a language other than English, the blocks should be translated.

### Actual Behavior

_Describe what actually happens_
If you have your language set to something other than english about 30% of the blocks still show up in the toolbox in English. When you drag them to the workspace they appear translated.

![untranslated-onload](https://user-images.githubusercontent.com/399209/52200977-681ebc80-286a-11e9-9390-dfe36817411d.gif)

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_
Set your language to something other than English for loading the editor. I.e. change your language on the website before opening the editor, or load the editor with a locale: https://llk.github.io/scratch-gui/develop/?locale=es

All of the basic blocks should be translated, but many will come up in English. Switch to another language, then back to the original, and the blocks will be translated.

I think this regression may be due to the perf changes. /cc @paulkaplan @picklesrus

paulkaplan:
Oh no! That definitely looks like a regression. I'll look into this right away.

thisandagain:
Resolved in #4478

PR #4478, opened 2019-02-04
[master] Hotfix/toolbox language

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/4476

### Proposed Changes

1. Make sure to load the initial locale into scratch blocks messages before rendering toolbox.
2. Mark the `locale` as a required prop, since it is being used on mount. This works because it is passed in from mapStateToProps, and the reducer has a default value (en).
3. Add an integration test to make sure this is working.

### Reason for Changes

The toolbox was not being refreshed correctly if the initial locale was not english. Make sure to load the initial locale so that the toolbox does not have to first be generated in english, then again in your language.

### Test Coverage

Added an integration test to `localization`. The previous test checked the blocks, but only after switching languages. This new test checks the blocks using the `locale=...` url, to ensure it works on initial load. 

---

@chrisgarrity I updated master to be the version that is currently deployed on WWW and made this as hotfix against master, so that we can publish a new GUI version and update WWW as a hotfix. This issue seems big enough to want to hotfix, and it seems like the fix is low risk. We can talk about whether or not to hotfix this offline though.

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -87,6 +87,7 @@ class Blocks extends React.Component {
     componentDidMount () {
         this.ScratchBlocks.FieldColourSlider.activateEyedropper_ = this.props.onActivateColorPicker;
         this.ScratchBlocks.Procedures.externalProcedureDefCallback = this.props.onActivateCustomProcedures;
+        this.ScratchBlocks.ScratchMsgs.setLocale(this.props.locale);
 
         const workspaceConfig = defaultsDeep({},
             Blocks.defaultOptions,
@@ -526,7 +527,7 @@ Blocks.propTypes = {
     extensionLibraryVisible: PropTypes.bool,
     isRtl: PropTypes.bool,
     isVisible: PropTypes.bool,
-    locale: PropTypes.string,
+    locale: PropTypes.string.isRequired,
     messages: PropTypes.objectOf(PropTypes.string),
     onActivateColorPicker: PropTypes.func,
     onActivateCustomProcedures: PropTypes.func,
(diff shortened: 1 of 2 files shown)


paulkaplan:
(the failing test has to do with the sauce test which we haven't been using, so i think it is ok to ignore)

src/containers/blocks.jsx
```diff
@@ -87,6 +87,7 @@ class Blocks extends React.Component {
     componentDidMount () {
         this.ScratchBlocks.FieldColourSlider.activateEyedropper_ = this.props.onActivateColorPicker;
         this.ScratchBlocks.Procedures.externalProcedureDefCallback = this.props.onActivateCustomProcedures;
+        this.ScratchBlocks.ScratchMsgs.setLocale(this.props.locale);
```
chrisgarrity: It would be nice for `setLocale` to handle the special cases, but it doesn't appear to be simply checking `isVisible` in `setLocale` because it would still need to be called twice in `componentDidMount`, before the blocks are injected, and after the vm is attached.

chrisgarrity: I'm not sure that it does need to be called after `attachVM`, it does appear to work, I'd want a lot more tests, like loading a project with an extension before making the change.

test/integration/localization.test.js
```diff
@@ -52,4 +52,15 @@ describe('Localization', () => {
         const logs = await getLogs();
         await expect(logs).toEqual([]);
     });
+
+    // Regression test for #4476, blocks in wrong language when loaded with locale
```
chrisgarrity: 👍

PR #4485, opened 2019-02-04
Sort variable options in "of" block

### LINKED ISSUE #4397: Variables in "(attribute) of (sprite)" menu not sorted
### Expected Behavior

Variables in the "(attribute) of (sprite)" menu should be sorted alphabetically, just like all other menus that contain variables.

### Actual Behavior

They aren't sorted.

### Steps to Reproduce

Create, in order, two "for this sprite only" variables: first "Z", then "A". Note that when you open the attribute menu of "(attribute) of (that sprite)", they are sorted "Z" then "A", i.e. by creation date, not alphabetically.

---

Code fix goes here: https://github.com/LLK/scratch-gui/blob/ff473e8fb0610565edfd783d7f17d9133e0bdb42/src/lib/blocks.js#L267

(Or in the VM definition for `getAllVariableNamesInScopeByType`, so all uses of the function are automatically given the sorted list.)

towerofnix:
### Resolves

Fixes #4397.

### Proposed Changes

Sorts the variable options in the "(attribute) of (sprite)" block:

The dropdown attribute of an "of" block with a sprite selected; the options read the usual sprite properties, then "a, Ax, b, Bx, z".

Also sorts variables on the stage (i.e. "for all sprite" variables) when the stage is selected:

The same menu but with the stage selected; the options read "a global, b global, my variable, z global".

### Reason for Changes

To make the "of" block considerably more useful in projects where there are lots of variables; and to make it more consistent with other blocks that have variable dropdowns.

### Test Coverage

No new unit tests. Tested manually.

### DIFF
--- a/src/lib/blocks.js
+++ b/src/lib/blocks.js
@@ -247,6 +247,7 @@ export default function (vm) {
                 }
                 // Get all the stage variables (no lists) so we can add them to menu when the stage is selected.
                 const stageVariableOptions = vm.runtime.getTargetForStage().getAllVariableNamesInScopeByType('');
+                stageVariableOptions.sort(ScratchBlocks.scratchBlocksUtils.compareStrings);
                 const stageVariableMenuItems = stageVariableOptions.map(variable => [variable, variable]);
                 if (sensingOfBlock.inputs.OBJECT.shadow !== sensingOfBlock.inputs.OBJECT.block) {
                     // There's a block dropped on top of the menu. It'd be nice to evaluate it and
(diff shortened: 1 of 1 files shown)


towerofnix:
@paulkaplan OK, pushed (and tested to make sure it works - it does): 9166cb12d5243adb9e1fcc4a9bd032ebaba0862e

Although I'm inclined to believe that this is rather a less clean solution... defining an equivalent utility function to only be used twice seems worse than reusing an existing function. At least doing it this way means we don't rely on scratch-blocks's API stay in the same..but then, we were already doing that with the rest of the code in this file, right?

PR #4495, opened 2019-02-06
Wait until loader becomes stale to continue in project loader tests

paulkaplan:
This is a test cleanup PR that tries to use a dynamic instead of a fixed wait for project loader tests. Instead of waiting a fixed time, use the [`until.stalenessOf`](https://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/lib/until.html#stalenessOf) webdriver method to wait until the loader element is removed.

/cc @BryceLTaylor @rschamp

### DIFF
--- a/test/helpers/selenium-helper.js
+++ b/test/helpers/selenium-helper.js
@@ -21,7 +21,8 @@ class SeleniumHelper {
             'getSauceDriver',
             'getLogs',
             'loadUri',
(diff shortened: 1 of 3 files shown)


paulkaplan:
@rschamp fixed! turns out the blocks example doesn't show a loader screen. i was getting to excited :)

PR #4505, opened 2019-02-07
Use blob download instead of data-uri to enable large downloads

paulkaplan:
This was already being used for sb3s and sprite3s, but the code was repeated in different places. Make a single helper that downloads blobs, and make all download paths use it.

Fixes an issue we found in testing that downloading large costumes/sounds and downloading on Edge was broken. This is because the data-uri download method doesn't work in those two cases. Use the method we were using for sprite3s and sb3s, which is to download as a Blob. Works in Edge and works for large assets.

/cc @apple502j note that this is an update to your patch, it is interesting to see the limitations of data-uri-based downloading!


src/containers/costume-tab.jsx
```diff
@@ -154,7 +154,8 @@ class CostumeTab extends React.Component {
     }
     handleExportCostume (costumeIndex) {
         const item = this.props.vm.editingTarget.sprite.costumes[costumeIndex];
-        download(`${item.name}.${item.asset.dataFormat}`, item.asset.encodeDataURI());
+        const blob = new Blob([item.asset.data], {type: item.asset.contentType});
```
rschamp: Is there `asset.contentType`? I think you want `asset.assetType.contentType`.

src/containers/sound-tab.jsx
```diff
@@ -90,8 +90,8 @@ class SoundTab extends React.Component {
 
     handleExportSound (soundIndex) {
         const item = this.props.vm.editingTarget.sprite.sounds[soundIndex];
-        const soundDataURL = item.asset.encodeDataURI();
-        download(`${item.name}.${item.asset.dataFormat}`, soundDataURL);
+        const blob = new Blob([item.asset.data], {type: item.asset.contentType});
```
rschamp: Same here.

PR #4507, opened 2019-02-07
Initial version of translated tutorial gifs

### LINKED ISSUE #3632: Add localization support for Tutorial images in 3.0
### Expected Behavior

When viewing tutorial cards in other languages, any image that includes blocks should be in the current language (assuming that we have generated the images for that language)

The gui should dynamically load the correct set of images based on the locale. Adding the localized gifs should not substantially change the initial size of the gui.

### Actual Behavior

Images are always in English even when headers have been translated

Note: animated gifs for pt-BR, ja, it, and es are already available on the Scratch Resources

chrisgarrity:
### Resolves

- Resolves #3632 (Note, if this seems like a good approach there are several more languages that can be added)

### Proposed Changes

Changes the cards content to have 'ids' for the images instead of loading the default English images.
Adds utility functions that will provide the image based on locale, or the default English image if we don't have images for the locale. Also adds a function that dynamically loads the alternate images when the locale is switched.

Currently only implements Spanish.

Refactors the way step images are referenced and named.
* step image refers to an `id` rather than an image file
* stepIds use a 'standard' naming convention - use camel case, compose the id from a short form of the tutorial name followed by the step action. e.g. `chaseGameChangeScore`
* step files should all be in the steps folder - the name should correspond to the step id converted to 'kebab' case, followed by the locale and file type. e.g. `chase-game-change-score.en.gif`
* webpack is configured to create different chunks for different languages and load them dynamically.
* chunks are created in the `dist/chunks` directory, and the publicPath is set to static so that www can copy chunks to `static` and have everything work.

### Reason for Changes

Tutorials should be available in other languages.

### Test Coverage

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


src/lib/libraries/decks/translate-image.js
```diff
@@ -0,0 +1,45 @@
...
+const translations = {
+    es: () => loadSpanish()
+};
+
+const loadImageData = locale => {
+    if (translations.hasOwnProperty(locale)) {
+        translations[locale]()
+            .then(newImages => {
+                savedImages = newImages;
+                savedLocale = locale;
```
rschamp: This is usually something we do in redux. I know that it works, but just noting it now in case we see issues. I'm not sure exactly what a bug from this would look like, maybe a race condition?

PR #4520, opened 2019-02-08
Remove prune step from deployment for now

rschamp:
It doesn't affect greenkeeper subdirectories, which are the ones that cause the most issues. And now the prune step itself is breaking builds. So until it works better, remove this step.


rschamp:
Try it out: [prune-removal](https://llk.github.io/scratch-gui/prune-removal)

PR #4521, opened 2019-02-08
Extract css from the JS

mzgoddard:
### Resolves

Loading feels faster. And a small memory and speed performance improvement.

### Proposed Changes

- Extract CSS from the output JS

### Reason for Changes

Extracting the CSS provides a few benefits:
- The CSS can then style the body tag before JS executes.

  Chrome and Firefox (seem to?) render the page when it reaches the first javascript to evaluate. Since that is at the bottom of the body tag, the page will render the empty body tag. If the CSS set a background color on body we would see that rendered until JS evaluates and the first render completes with JS produced DOM. (This can be achieved other ways, but this is kind of a free win with the other benefits.)
- The CSS appears fewer times in memory. Loading the CSS with JavaScript is useful during development. In production extracting the CSS we remove two copies of it from JS memory. We save its appearance in the full script source code. We save its appearance as the string to the module that is evaluated as a function.
- Skips evaluating as JS before evaluating as CSS. Skips creating style tags for each css file.

Here are some examples with the CSS setting the background color:

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "lodash.omit": "4.5.0",
     "lodash.pick": "4.4.0",
     "lodash.throttle": "4.0.1",
+    "mini-css-extract-plugin": "^0.5.0",
     "minilog": "3.1.0",
     "mkdirp": "^0.5.1",
     "papaparse": "4.6.2",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -5,12 +5,39 @@ var webpack = require('webpack');
 // Plugins
 var CopyWebpackPlugin = require('copy-webpack-plugin');
 var HtmlWebpackPlugin = require('html-webpack-plugin');
+var MiniCssExtractPlugin = require('mini-css-extract-plugin');
 var UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 
 // PostCss
 var autoprefixer = require('autoprefixer');
 var postcssVars = require('postcss-simple-vars');
 var postcssImport = require('postcss-import');
+
+const cssLoader = {
+    loader: 'css-loader',
+    options: {
+        modules: true,
+        importLoaders: 1,
+        localIdentName: '[name]_[local]_[hash:base64:5]',
+        camelCase: true
+    }
+};
+
+const postcssLoader = {
+    loader: 'postcss-loader',
+    options: {
+        ident: 'postcss',
+        plugins: function () {
+            return [
+                postcssImport,
+                postcssVars,
+                autoprefixer({
+                    browsers: ['last 3 versions', 'Safari >= 8', 'iOS >= 8']
+                })
+            ];
+        }
+    }
+};
 
 const base = {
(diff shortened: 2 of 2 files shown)


paulkaplan:
@rschamp do you have thoughts on this? My feeling is that this makes it harder to use as an node module (which for WWW/offline editor/future hybrid apps is how we use GUI). I'm not exactly sure how we would want to integrate this with WWW. Thoughts?

gnarf:
"harder to use as a node module" - isn't this 100% a browser application and not a module?  I get that www might need to be updated to include the dist/bundle.css or something but I think this is a longer term performance issue anyway, we should not be doing "css bundled in js".

mzgoddard:
@paulkaplan oh yeah that. Hmm.

For this PR we can limit the change to the build folder output since the package uses two builds.

In the long term I think we want to extract the css in gui, www, desktop, etc.

I can think of a few ways off the top of my head to how we can extract the css here for downstream use. The first would be to consume scratch-gui the way gui consumes vm, render, etc. The next would be to mark css files as external and state the loaders to consume the css in downstream webpack configs. After that we could store the webpack build's stats or records output and use that metadata to track the css file for scratch-gui.js, though well with the current configuration that'd be easy to compute either way. The last option would be a webpack plugin or loader that shims the entry of scratch-gui.js with a require('./output.css') that in a downstream build would use configured loaders.

-----

Thinking on the loading experience part, loading a local build of the gui takes 1.6s on a MBP 2017 15" laptop, and 3.6s on a SMS Chromebook 2017. During that time the browser will render a white box. And the desktop app shows nothing, which feels kind of broken. Loading examples of some current builds:

<details>
<summary>Local build loading GUI</summary>



</details>

<details>
<summary>/projects/ route on scratch.mit.edu</summary>



</details>

rschamp:
I agree we can't merge this type of change until we solve how to integrate it into scratch-www. I know that we can manually add a step on www's side, but it would be nice if there was some way to automate it, since e.g. we would want to do the same thing in scratch-paint, which should propagate to scratch-gui and then to scratch-www.

I like the first option @mzgoddard listed where scratch-www would do the building of gui the way gui currently builds everything else. But if that is too large of a project, it shouldn't gate this PR (but if it's not that huge then it probably should).

mzgoddard:
- Only extract CSS for `build/` folder items, do not extract CSS for `dist/` folder items

That should make this individual PR easier. And we can work towards the larger goal of using this technique in www/desktop/etc.

benjiwheeler:
@seotts and I looked at this in July 2021.

Labeled with medium severity, because memory and speed have an outsized effect on older and lower-performance devices, and thus an outsized obstacle to equity in our user experience.

Labeled with high impact, because this would affect all users.

benjiwheeler:
Closing this because it's unlikely we'll review and merge it any time soon. See issue we filed: https://github.com/LLK/scratch-gui/issues/7368 , which links to this in case someone works on it in the future.

PR #4527, opened 2019-02-11
Add bounding box to prevent sprite overflow

### LINKED ISSUE #1791: Dragging a sprite to the (right-side) edge of the screen shows scrollbar
### Expected Behavior

No scrollbar should appear when the sprite is dragged to the right-side of the screen; the sprite should just be cut off.

### Actual Behavior

A horizontal scrollbar appears at the bottom of the screen.

### Steps to Reproduce

Grab a sprite and drag it so that part of it is past the right-side edge of the screen.

Debian (testing); Firefox 60.0b11.

towerofnix:
### Resolves

Resolves #1791, where dragging a sprite past the edge of the screen would cause a scrollbar to show up.

### Proposed Changes

Adds a wrapper `<div>` around the canvas element, which covers the entire screen and has `overflow: hidden`, so the translated canvas doesn't show up past the screen edges. (Z-indexes should still work; I had to move the `z-index` property onto the wrapper, but this matches all the original behavior as far as I can tell - sprites still get dragged over the menu bar, etc.) I also adjusted the math for positioning the canvas, since it's now placed relative to the top-left of the screen instead of to the stage rectangle.

### Reason for Changes

Fixes a peculiar (though basically harmless) issue, to make the editor feel a little more bug-free.

### Test Coverage

Tested manually. Everything still works, even when the screen is very small (which makes you have to scroll to navigate the entire gui).

### Browser Coverage

Linux Firefox and Chromium.

### DIFF
--- a/src/containers/stage.jsx
+++ b/src/containers/stage.jsx
@@ -312,9 +312,11 @@ class Stage extends React.Component {
         this.dragCanvas.style.display = 'none';
     }
     positionDragCanvas (mouseX, mouseY) {
-        // mouseX/Y are relative to stage top/left, and dragCanvas is already
-        // positioned so that the pick location is at (0,0).
-        this.dragCanvas.style.transform = `translate(${mouseX}px, ${mouseY}px)`;
+        // mouseX/Y are relative to stage top/left, but dragCanvas is
+        // positioned at top left of the whole viewport, so account for this.
+        const x = this.rect.x + mouseX;
+        const y = this.rect.y + mouseY;
+        this.dragCanvas.style.transform = `translate(${x}px, ${y}px)`;
     }
     onStartDrag (x, y) {
         if (this.state.dragId) return;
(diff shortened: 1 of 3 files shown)


src/containers/stage.jsx
```diff
@@ -312,9 +312,11 @@ class Stage extends React.Component {
         this.dragCanvas.style.display = 'none';
     }
     positionDragCanvas (mouseX, mouseY) {
-        // mouseX/Y are relative to stage top/left, and dragCanvas is already
-        // positioned so that the pick location is at (0,0).
-        this.dragCanvas.style.transform = `translate(${mouseX}px, ${mouseY}px)`;
+        // mouseX/Y are relative to stage top/left, but dragCanvas is
+        // positioned at top left of the whole viewport, so account for this.
+        const x = this.rect.x + mouseX;
+        const y = this.rect.y + mouseY;
```
adroitwhiz, line 318:
```suggestion
        const x = this.rect.left + mouseX;
        const y = this.rect.top + mouseY;
```

Legacy (i.e. not Blink-based) Edge doesn't have `x` or `y` properties on the rectangle returned from `getBoundingClientRect`.

chrisgarrity:
Thanks for submitting this PR. Unfortunately it's been long enough that it's accumulated conflicts and is still a very low priority, so I'm closing.

PR #4531, opened 2019-02-11
Prompt before file upload replaces project

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-desktop/issues/15

### Proposed Changes

Show a confirmation before uploading a file if you are about to replace the contents of the current project (i.e. offline or not logged in)

### Reason for Changes

Because there is a danger of losing your project

### Test Coverage

Added two integration tests to cover uploading projects using the file menu.

---

@benjiwheeler this fix is in GUI but it will fix the problem with desktop as well. As I mentioned on that issue, this is an existing bug in a few situations in GUI, not just offline.


src/containers/sb-file-uploader.jsx
```diff
@@ -42,6 +43,11 @@ const messages = defineMessages({
         id: 'gui.projectLoader.loadError',
         defaultMessage: 'The project file that was selected failed to load.',
         description: 'An error that displays when a local project file fails to load.'
+    },
+    uploadWarning: {
```
benjiwheeler: `menu-bar.jsx` already has this message:
```
    confirmNav: {
        id: 'gui.menuBar.confirmNewWithoutSaving',
        defaultMessage: 'Replace contents of the current project?',
        description: 'message for prompting user to confirm that they want to create new project without saving'
    }
```
Could you pass that in instead?

src/containers/sb-file-uploader.jsx
```diff
@@ -107,7 +128,11 @@ class SBFileUploader extends React.Component {
                         nonInteraction: true
                     });
                     // Remove the hash if any (without triggering a hash change event or a reload)
-                    history.replaceState({}, document.title, '.');
+                    try { // Can fail e.g. when GUI is loaded from static file (integration tests)
```
benjiwheeler: interesting

src/containers/sb-file-uploader.jsx
paulkaplan: @benjiwheeler good idea, moved it to sharedMessages and used it in both places.

src/containers/sb-file-uploader.jsx
paulkaplan: Yeah I only realized it when I saw that the integration test was failing because it was hitting the "catch" alert condition below.

PR #4534, opened 2019-02-12
Show an "Importing..." alert during file import

paulkaplan:
### Resolves

- Resolves a UX issue with importing where importing multiple files, or importing gifs, would leave the editor in a loading state for an indeterminate amount of time depending on the files, but gave no indication that something was happening. After chatting with @carljbowman, we decided the existing alert system with a spinner would be a good first step.

With sprites

sprite-multi-import-alert

With costumes

costume-multi-import-alert

### Proposed Changes

Make the file uploader return information about which file has been processed, and use that to decide when to close the alert. Add an HOC for the show/close to avoid code repetition.

---

@carljbowman just an FYI, this only creates a single alert that is closed after all files have been processed, instead of the stacking multi alerts that I showed you in the prototype. I assume this is strictly better, but just wanted to get your thoughts.

@rschamp This is a WIP since we talked about different ways to refactor this, and i wanted to show you what it would look like to just add the index info to handleFileUpload. The biggest functionality issue to me is that there is no error handling. If the file fails to parse, or if the costume/sprite fails to upload, there is no way to close the importing spinner. Promises definitely make error propagation easier in async situations, so I would definitely like to refactor just for that. Maybe we can pair on this?

### DIFF
--- a/src/containers/costume-tab.jsx
+++ b/src/containers/costume-tab.jsx
@@ -215,7 +215,7 @@ class CostumeTab extends React.Component {
                         this.props.onCloseImporting();
                     }
                 });
-            });
+            }, this.props.onCloseImporting);
         }, this.props.onCloseImporting);
     }
     handleCameraBuffer (buffer) {
--- a/src/containers/costume-tab.jsx
+++ b/src/containers/costume-tab.jsx
@@ -161,13 +161,12 @@ class CostumeTab extends React.Component {
     handleNewCostume (costume, fromCostumeLibrary) {
         const costumes = Array.isArray(costume) ? costume : [costume];
 
-        costumes.forEach(c => {
+        return Promise.all(costumes.map(c => {
             if (fromCostumeLibrary) {
-                this.props.vm.addCostumeFromLibrary(c.md5, c);
-            } else {
-                this.props.vm.addCostume(c.md5, c);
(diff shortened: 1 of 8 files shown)


carljbowman:
@paulkaplan - Thanks for pulling this together. 

I agree, a single alert is better. We could obviously imagine many variations of feedback specificity and a subsequent "Cancel" action, but I think this is a **great** first step.

src/containers/costume-tab.jsx
```diff
@@ -204,12 +204,17 @@ class CostumeTab extends React.Component {
...
+        this.props.onShowAlert('importingAsset');
+        handleFileUpload(e.target, (buffer, fileType, fileName, fileIndex, fileCount) => {
             costumeUpload(buffer, fileType, storage, vmCostumes => {
                 vmCostumes.forEach((costume, i) => {
                     costume.name = `${fileName}${i ? i + 1 : ''}`;
                 });
-                this.handleNewCostume(vmCostumes);
+                this.handleNewCostume(vmCostumes).then(() => {
+                    if (fileIndex === fileCount - 1) {
+                        this.props.onCloseAlert('importingAsset');
```
rschamp: Can either `handleFileUpload` or `handleNewCostume` throw errors/reject? If so, we probably should close the alert in that case as well?

PR #4537, opened 2019-02-12
Fixes #4253 by telling the vm the langauge when it is initialized.  B…

### LINKED ISSUE #4253: Translate extension "language" block returns "English" until you go into editor
### Expected Behavior

If your language in Scratch is set to something other than English, the "language" block in the translate extension should return that language from both the project page and editor. 

### Actual Behavior

If you view a project from the project page, the block returns "English." If you go into the editor, it will return the correct language and if you go back to the project page it will continue to return the correct language. 

### Steps to Reproduce

Switch your language to something other than English. Go to [this test project](https://scratch.mit.edu/projects/277135194/). View it from the project page-- language should be wrong. Go into editor-- language should be correct. Go back to project page-- language should be correct. 

picklesrus:
…efore it didn't get the language until the blocks component mounted.

### Resolves

Resolves #4253 

### Proposed Changes

Tells the VM the locale when initializing it.

### DIFF
--- a/src/lib/vm-manager-hoc.jsx
+++ b/src/lib/vm-manager-hoc.jsx
@@ -33,6 +33,7 @@ const vmManagerHOC = function (WrappedComponent) {
                 this.props.vm.attachAudioEngine(this.audioEngine);
                 this.props.vm.setCompatibilityMode(true);
                 this.props.vm.initialized = true;
+                this.props.vm.setLocale(this.props.locale, this.props.messages);
             }
             if (!this.props.isPlayerOnly && !this.props.isStarted) {
                 this.props.vm.start();
@@ -105,6 +106,8 @@ const vmManagerHOC = function (WrappedComponent) {
         isLoadingWithId: PropTypes.bool,
         isPlayerOnly: PropTypes.bool,
         loadingState: PropTypes.oneOf(LoadingStates),
+        locale: PropTypes.string,
+        messages: PropTypes.objectOf(PropTypes.string),
         onError: PropTypes.func,
         onLoadedProject: PropTypes.func,
         onSetProjectUnchanged: PropTypes.func,
@@ -119,6 +122,8 @@ const vmManagerHOC = function (WrappedComponent) {
         return {
             fontsLoaded: state.scratchGui.fontsLoaded,
             isLoadingWithId: getIsLoadingWithId(loadingState),
+            locale: state.locales.locale,
+            messages: state.locales.messages,
             projectData: state.scratchGui.projectState.projectData,
             projectId: state.scratchGui.projectState.projectId,
             loadingState: loadingState,
--- a/src/lib/vm-manager-hoc.jsx
+++ b/src/lib/vm-manager-hoc.jsx
@@ -79,6 +79,8 @@ const vmManagerHOC = function (WrappedComponent) {
                 /* eslint-disable no-unused-vars */
                 fontsLoaded,
                 loadingState,
+                locale,
+                messages,
                 isStarted,
                 onError: onErrorProp,
                 onLoadedProject: onLoadedProjectProp,
--- a/test/unit/util/vm-manager-hoc.test.jsx
+++ b/test/unit/util/vm-manager-hoc.test.jsx
@@ -19,11 +19,16 @@ describe('VMManagerHOC', () => {
                 projectState: {},
                 mode: {},
                 vmStatus: {}
+            },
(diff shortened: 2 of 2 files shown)


picklesrus:
Adding @kchadha @cwillisf, and @chrisgarrity because I'm not totally sure we're okay with implications of this change. Here are my questions:
1. The call to setLocale in vm will cause a call to 
extensionManager.refreshBlocks in a place where it didn't before.  In the case of calling this from player mode, that seems to work fine because refreshBlocks runs over _loadedExtensions and there are none yet. Is that something we can/should count on? And should it matter?
2.  Why does vm's setLocale always call refreshBlocks but only conditionally call formatMessage.setup when the locale has actually changed?
3. Should I conditionally call vm.setLocale from gui only when the language is changed?  I didn't, because, iiuc, the vm.inInitialized is essentially keeping track of that for me.

src/lib/vm-manager-hoc.jsx
```diff
@@ -33,6 +33,7 @@ const vmManagerHOC = function (WrappedComponent) {
                 this.props.vm.attachAudioEngine(this.audioEngine);
                 this.props.vm.setCompatibilityMode(true);
                 this.props.vm.initialized = true;
+                this.props.vm.setLocale(this.props.locale, this.props.messages);
```
chrisgarrity: This may be overkill, but you could check whether the locale is different before setting it. `this.props.vm.getLocale()` will return the current vm locale. If it doesn't match `this.props.locale`, then set it.

cwillisf: IMO it would be better for `setLocale` to check internally if it's being asked to make a change, and early-out or otherwise do nothing if the new locale matches the old one. That way we only write that `if` statement in one place instead of in every place that calls `setLocale`.

picklesrus: Yeah. That's why I'm curious about why vm's setLocale does some things conditionally, but always calls refreshBlocks.

chrisgarrity: So checking is definitely overkill as vm does the check. There were times when the language would change and the extension blocks would end up out of sync (one language behind). I think calling refreshBlocks ensured that the extensions were always in sync. However, it does seem like we could put the refreshBlocks inside the check and not call it if the language hasn't changed.

chrisgarrity: I did a quick check to see what would happen if the refreshBlocks wasn't called. When you switch from player to `see inside` the extension blocks are not rendered properly (red blocks). So even though vm is initialized, and the locale is set, we do need to refreshBlocks when switching between player and editor views. Basically, always calling refreshBlocks ensures that the extensions get rendered correctly regardless of the path to get here (load project with extension, add extension, see inside, etc)

kchadha: @cwillisf, @chrisgarrity, I'm curious what your thoughts are about calling `vm.setLocale()` before or after `vm.start()`? I think it makes sense, but I want to make sure because there are not too many things we call before `vm.start()`

src/lib/vm-manager-hoc.jsx
```diff
@@ -119,6 +122,8 @@ const vmManagerHOC = function (WrappedComponent) {
         return {
             fontsLoaded: state.scratchGui.fontsLoaded,
             isLoadingWithId: getIsLoadingWithId(loadingState),
+            locale: state.locales.locale,
```
kchadha: locale and messages should also be added where we are omitting props to prevent unnecessary props from being passed down to children (in `render` between the two eslint disable comments)

picklesrus: Fixed!

src/lib/vm-manager-hoc.jsx
cwillisf: @kchadha are you asking about theoretical principles or concrete bugs? ;)

IMO it's morally fine to call `vm.setLocale()` before `vm.start()` in that I think of the locale as a setting that can be changed at any time. You can paint your car whether or not it's moving... in fact, some would argue it's better to paint a stationary car. Those people have no sense of adventure.

In terms of whether or not it's currently safe... uh... I'm not sure, to be honest. I would expect it to be safe and I'm not aware of anything on the VM side that would make it unsafe, but I'm less familiar with the side effects on the scratch-blocks side of things. Any thoughts on that, @chrisgarrity?

Issue #4400, opened 2019-01-20
Changing  sprite of a "(attribute) of (sprite)" in block palette rearranges blocks in the workspace (sometimes)


towerofnix:
### Expected Behavior

The "(attribute) of (sprite)" block does not destroy your project.

### Actual Behavior

It does.

Case in point:

![Many complex scripts all layered onto each other.](https://user-images.githubusercontent.com/9948030/51441373-1e00dd00-1ca7-11e9-80e0-23ffaca8d6e4.png)

Every script is positioned at (0, 0).

Zip file of this project attached (see "Battle Controller" sprite): [Untitled-297.zip](https://github.com/LLK/scratch-gui/files/2776769/Untitled-297.zip)

### Steps to Reproduce

I can't reliably repro this, unfortunately, but it's happened to me several times while developing that large project. This only ever occurs immediately after I change the sprite selected in the dropdown menu. It's a huge pain and always causes either lost progress (restoring from a page reload) or rearranging all the scripts back to where they're supposed to be.

### Operating System and Browser

Firefox 65.0b11, Linux (Debian).

towerofnix:
I'm not sure if it's relevant or not (absolutely could be a red herring), but I think I've always been zoomed in whenever the wonkiness happens.

kchadha:
@towerofnix Could you explain how this issue is related to the`(attribute) of (sprite)` block? Are you seeing this issue only in workspaces that have that block? What browser(s) are you seeing this issue in?

towerofnix:
@kchadha Sorry for the lack of a consistent repro case. I'm having trouble reproducing this myself, even following my earlier notes (i.e. the comments here) and working in the same project.

Yes, it's tied to the "of" block. See my notes in "Steps to Reproduce":

> This only ever occurs immediately after I change the sprite selected in the dropdown menu.

Just the same, the browser I used is noted.

I believe that selecting an option from the sprite menu is causing the workspace to reload, which is presumably capable of causing a mess with this, although I cannot repro this lately. Or perhaps it's not a workspace reload; it needs more investigation.

Here's what I can reliably reproduce while working inside https://scratch.mit.edu/projects/284710366/editor/:

* Changing the selected sprite in the "of" block's menu causes a noticeable amount of lag (unresponsive editor time). (Is this because there are a lot of blocks in the workspace; or, maybe more likely, is it because there's a lot of sprite-local variables? Needs more testing.)

[...]

PR #4542, opened 2019-02-13
Position blocks added from "share the love" within the viewport

### LINKED ISSUE #1739: Change defaults and ordering for > < = blocks
To help new learners more quickly recognize and use the > < and = operators, suggestion to add default value of 100 on the right-hand side. 
 
In addition, may help to order so that > is first (since that's used most often in projects). One suggestion is to put in the order that they are often said aloud  (at least in U.S.): > < = greater than, less than, equal to. (Currently equal to was put second to help minimize visual confusion so could order that way instead.)

### LINKED ISSUE #4270: Dropping a script from the backpack should make it appear where you drop it
### Expected Behavior

Dragging a script from the backpack and dropping it onto the code workspace should make it appear where you dropped it.

### Actual Behavior

The script shows up elsewhere (at the origin of the workspace, I think). Seriously a pain to move it somewhere more sensible..!

### Steps to Reproduce

Drag a script out of the backpack.

towerofnix:
### Resolves

Resolves LLK/scratch-vm#1739. Hopefully resolves #4270, too, but I can't test that to be certain (I don't know how to use the backpack while offline).

*(Tangentially related to #4400; I noticed while testing the repros I described [here](https://github.com/LLK/scratch-gui/issues/4400#issuecomment-462994133) that switching the sprite no longer messed with the workspace scroll while zoomed in, and it still does in scratch-gui/develop. There's still lag, presumably from generating the menu item elements, though. /cc @kchadha)*

### Proposed Changes

Takes `workspaceMetrics`, previously stored in the `<Blocks>` container's state, and changes it to be stored in global state via a reducer. Changes the `<TargetPane>` container so that, when it handles dropped blocks, it repositions them to fit within the viewport of the target sprite's workspace. (This behavior is in a new function that is also used when receiving blocks dropped from the backpack, so it should work in that case too, but I don't know how to test the backpack offline.)

### Reason for Changes

To make "share the love" (and backpacked scripts) a much more useful feature when working in large projects (sprites with lots of scripts).

### Test Coverage

Tested manually. (Would unit tests be helpful here? There's quite a bit of tweaked code, but it was mostly just simple substitutions and the typical react/redux stuff for making use of global state; and I'm not sure if the added behavior should take unit tests or not.)

### Browser Coverage

Linux Firefox, Chromium.

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -14,7 +14,7 @@ import ExtensionLibrary from './extension-library.jsx';
 import extensionData from '../lib/libraries/extensions/index.jsx';
 import CustomProcedures from './custom-procedures.jsx';
 import errorBoundaryHOC from '../lib/error-boundary-hoc.jsx';
-import {STAGE_DISPLAY_SIZES} from '../lib/layout-constants';
+import {BLOCKS_DEFAULT_SCALE, STAGE_DISPLAY_SIZES} from '../lib/layout-constants';
 import DropAreaHOC from '../lib/drop-area-hoc.jsx';
 import DragConstants from '../lib/drag-constants';
 import defineDynamicBlock from '../lib/define-dynamic-block';
@@ -509,7 +509,7 @@ class Blocks extends React.Component {
             });
     }
     render () {
-        /* eslint-disable no-unused-vars, no-shadow */
+        /* eslint-disable no-unused-vars */
         const {
             anyModalVisible,
             canUseCloud,
@@ -528,11 +528,11 @@ class Blocks extends React.Component {
             onRequestCloseExtensionLibrary,
             onRequestCloseCustomProcedures,
             toolboxXML,
+            updateMetrics: updateMetricsProp,
             workspaceMetrics,
-            updateMetrics,
             ...props
         } = this.props;
-        /* eslint-enable no-unused-vars, no-shadow */
+        /* eslint-enable no-unused-vars */
         return (
             <React.Fragment>
                 <DroppableBlocks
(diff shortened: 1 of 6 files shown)


apple502j:
Does it have something to do with 1739? It's comparing operator's issue which is already closed.

towerofnix:
@apple502j Oh, thanks for reminding me - I was referring to an issue in a different repo. Fixed now!

thisandagain:
@paulkaplan Bump

src/containers/blocks.jsx
```diff
@@ -284,14 +284,17 @@ class Blocks extends React.Component {
     onWorkspaceMetricsChange () {
         const target = this.props.vm.editingTarget;
         if (target && target.id) {
-            const workspaceMetrics = Object.assign({}, this.state.workspaceMetrics, {
-                [target.id]: {
+            // Dispatch updateMetrics later, since onWorkspaceMetricsChange may be (very indirectly)
```
benjiwheeler, line 304: Does this look kosher to others?

benjiwheeler:
@chrisgarrity and I think this is now good to merge -- but we have one outstanding question.

@rschamp and @picklesrus: does the use of setTimeout here look kosher to you? https://github.com/LLK/scratch-gui/pull/4542/files#r471669610

Update: see my comment in the code

src/containers/blocks.jsx
benjiwheeler: Got feedback from others -- sounds like this is ok, though there is some feeling that it might not be necessary.

I assume it's necessary for the reasons @towerofnix wrote in the comment -- that you don't want to dispatch a reducer action while already handling a reducer action.

benjiwheeler:
Thanks @towerofnix!!!!!!!

PR #4564, opened 2019-02-15
[master] Request a toolbox refresh after switching between tabs to fix #4561

### LINKED ISSUE #4561: Cannot add variables after switching to costume tab
### Expected Behavior

I should be able to switch to the costumes tab and switch back to the code tab and then add variables.

### Actual Behavior

The variables get added to the VM but it appears that the toolbox does not update to show them. You can tell they are really being added because switching to another sprite and switching back makes them reappear.

### Steps to Reproduce

Explained above. I also wrote an integration test to repro it that should be included in a fix:

```
    test('Creating variables after switching tabs updates the toolbox', async () => {
        await loadUri(uri);
        await clickText('Costumes');
        await clickText('Code');
        await clickText('Variables', scope.blocksTab);
        await driver.sleep(500); // Wait for scroll
        await clickText('Make a List');
        const el = await findByXpath("//input[@name='New list name:']");
        await el.sendKeys('list1');
        await clickButton('OK');
        await clickText('list1', scope.blocksTab);
    });
```

paulkaplan:
A hotfix version of https://github.com/LLK/scratch-gui/pull/4563

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -162,6 +162,7 @@ class Blocks extends React.Component {
                 this.setLocale();
             } else {
                 this.props.vm.refreshWorkspace();
+                this.requestToolboxUpdate();
             }
 
             window.dispatchEvent(new Event('resize'));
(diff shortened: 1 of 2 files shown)


paulkaplan:
Going to merge this one so we can get it up on staging for hotfix testing.

PR #4572, opened 2019-02-19
Check conditions before autoscrolling flyout when custom procedures dialog is closed

towerofnix:
### Resolves

Depends on LLK/scratch-blocks#1901. That PR should be merged first!

Fixes LLK/scratch-blocks#1856. Also fixes the issue where selecting "Cancel" would scroll the flyout. I can't find an issue filed for that, though. (But I could've sworn one existed..)

### Proposed Changes

Adds two conditions which are checked when the custom procedure dialog ("Make a Block") is closed:

1. Check if the dialog was cancelled.
2. Check if the dialog was for editing an existing custom block (rather than creating a new one).

If either of these are true, the flyout is *not* automatically scrolled. (Also, if the former, the workspace isn't updated, either, since no changes were made.)

### Reason for Changes

To make the custom procedure dialog and related autoscrolling feel more intuitive!

### Test Coverage

Tested manually.

### Browser Coverage

Linux Firefox, Chromium.

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -445,6 +445,12 @@ class Blocks extends React.Component {
     }
     handleCustomProceduresClose (data) {
         this.props.onRequestCloseCustomProcedures(data);
+
+        // No data: the user cancelled the dialog, so don't do anything to the workspace.
+        if (!data) {
+            return;
+        }
+
         const ws = this.workspace;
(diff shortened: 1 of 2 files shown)


kchadha:
Closing this PR, the corresponding scratch-blocks pr, and the corresponding issue since the risk of this change far outweighs the benefits of fixing this very low severity bug.

Issue #4428, opened 2019-01-24
Alert message div blocked some menu bar dropdown buttons


JinXJinX:
### Expected Behavior

Alert message should not block partial menu bar dropdown buttons.

### Actual Behavior

menu bar dropdown buttons are blocked by alert message div
![scratch 3 0 gui](https://user-images.githubusercontent.com/12160296/51706671-3184ae80-1fed-11e9-9937-d79540080436.gif)

### Steps to Reproduce

1. trigger a alert message
2. click 'file' button in menu bar
3. you can see 'new' button and 'load from computer' button is covered by alert's div

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_

### Operating System and Browser

Mac OS 10.14.2 Safari 12.0.2/ Chrome 71.0.3578/ FireFox 64.0.2

thisandagain:
@carljbowman Any thoughts on the desired z-index here? I can see arguments both ways.

thisandagain:
Thanks @carljbowman. Sounds good.

benjiwheeler:
Related to https://github.com/LLK/scratch-gui/issues/4226

benjiwheeler:
From my checking, it appears this issue was already fixed before #4916 went in. Possibly it was fixed by #4452.

apple502j:
@benjiwheeler Not for me on Firefox 76:
1) On sound tab, upload non-sounds (this creates an alert that cannot disappear)
2) You cannot click Code/Costume tab
3) On develop you can

benjiwheeler:
@apple502j I totally agree about the Code/Costume tabs. But this issue is about menu bar dropdown menus.

PR #4576, opened 2019-02-20
Merge latest changes (#2)

### LINKED ISSUE #4262: Pressing Ctrl-S should save the project

### LINKED ISSUE #4263: Remove preview modal from codebase

### LINKED ISSUE #4561: Cannot add variables after switching to costume tab

davidaylaian:
* chore(package): update scratch-vm to version 0.2.0-prerelease.20190118221822

Closes #4390

* Remove transition on sprite tiles to make placholder not flash black.

Also use white as the background since it gives the same impact and will not look bad if the filter isn't applied or takes a while to be applied

* Optimize Waveform component to reduce renders and cap max nodes in svg.

The waveform component was being re-rendered while the playhead changed, which is a costly operation because it has to calculate the full svg path data string.

* Improve paint/composite time for sound editor while sound is playing.

Forcing the playhead onto its own painting layer removes an expensive composite with the waveform. Animate using transforms instead of left prevents layouts of the rest of the page while playing a sound.

* Update WeDo 2.0 extension name

* Disable target updates while sound recording

* Fix #4263: Remove Preview Modal

* Fix #4262: Saving with Ctrl/Cmd + S

* Fully delete preview modal.

* chore(package): update scratch-render to version 0.1.0-prerelease.20190122223537

Closes #4378

* chore(package): update scratch-l10n to version 3.1.20190123150639

Closes #4408

* Move the mask on the audio meter outside the svg for repaint performance

* Revert "Fix #4262: Saving with Ctrl/Cmd + S"

This reverts commit e2ee0e06f7293795e7f6685ed1bb5e99bee52eb8.

* Remove the redux throttle for targets update. 

This fixes the root cause of #2858, which is caused because we are rendering the sound editor based on the GUI editing target data, but retrieving the sound based on the vms editing target data. The throttle of the redux action was causing those to get out of sync while a project was running.

The throttle is not needed because the vm runtime already batches those updates at 30 fps. The updates that come from top-level actions like switching sprites go out immediately.

This is also a minor performance boost, because there was a few ms being spent in setting and checking timers that were not needed.

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190123164824

* Fix issue where dragPayload was being invalidated for sprites

Use PureComponent instead since it is a simple shouldComponentUpdate check now.

* Remove dead code

* Throttle updates to assets on sprite selector item via HOC

Adds unit tests for this HOC.

* Throttle the watermark updates using the same HOC

* Throttle updates to stage images also

* chore(package): update scratch-blocks to version 0.1.0-prerelease.1548272895

* Use curried style for HOC configuration

* Revert "chore(package): update scratch-l10n to version 3.1.20190123150639"

* Update l10n

* Add container with overflow hidden to prevent scrollbars from playhead

* Actually fully remove preview modal.

* Fix test broken in #4422 by updating the snapshot.

* Use fontInliner instead of entire SVGRenderer for showing costume tiles.

The fontInliner from the scratch-svg-renderer was changed recently to work entirely on strings, so we can use that instead of parsing/loading/possibly drawing/stringifying the entire SVG. 

This does not address the size-1 cache issue where the cache is invalidated any time there is more than 1 sprite. But that can be handled later.

Using the 4x CPU slowdown in chrome dev tools, on a simple costume that just has a single text element the getCostumeUrl time goes from 10ms to 1ms, very important for updating in a timely way.

* chore(package): update scratch-blocks to version 0.1.0-prerelease.1548354812

Closes #4426

* fixed mouse click issuue

* Remove Irrelevant Tests

* chore(package): update scratch-render to version 0.1.0-prerelease.20190125023926

* Save when Ctrl+S is pressed

* Detect Cmd+S on macOS

* chore(package): update scratch-svg-renderer to version 0.2.0-prerelease.20190125192231

* rg "Try It" -l | xargs gsed -i "/Try It/d"

* Sometimes local machine says no errors, but travis finds them...

* Export costumes and sounds

* Avoid double toolbox update when setting the locale

* chore(package): update scratch-render to version 0.1.0-prerelease.20190128154859

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190128155421

* chore(package): update scratch-blocks to version 0.1.0-prerelease.1548691077

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190128162704

Closes #4443

* Move rounding into sprite info to avoid generating lots of garbage

It also avoids repeating the costly object spread, which becomes a problem for projects with a lot of sprites

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190128192220

* Remove cruft

* Remove References

* Fix Lint

* Different tests fail each time I run them.

* Only update the toolbox when the blocks tab is visible.

Need to store the rendered xml because the props can change while the blocks tab is hidden, but we re-render based on what was rendered, not the previous props (which may not have been rendered).

This is similar to the `ThrottledPropertyHOC` in that there may be prop changes that do not get rendered, but in this case for a different reason.

This increases the performance of switching between code and other editor tabs, as well as improving the "See Inside" performance. It makes switching to the code tab as fast as switching between sprites.

* Add an integration test to cover renaming costumes updating the blocks

It looks like an "test.only" was left in on one of these tests by mistake.

* chore(package): update scratch-l10n to version 3.1.20190130142816

* Add test for menus not updating when adding new costumes

I know we will want to change this eventually, so I added comments about that. But the next commit originally contained an inadvertent change to this functionality, so I wanted to add an integration test so that when we do change it, we

* Make the toolbox xml reflect the costume, sound and backdrop name

This invalidates the toolbox XML when switching back to the code tab after e.g. renaming a costume, making the code tab

* Menus cover monitors

fixes https://github.com/LLK/scratch-www/issues/2587
and https://github.com/LLK/scratch-gui/issues/4428

changed menu-bar z-index to one more than monitors so drop down menus cover monitors when opened.

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190130160436

* Include blocks update that improves toolbox layout perf

* chore(package): update scratch-blocks to version 0.1.0-prerelease.1548864869

* Update project changes test to test a project change that should actually emit a change event. Clicking on blocks should not emit a project changed event.

* Update package.json

* chore(package): update scratch-blocks to version 0.1.0-prerelease.1548885087

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190130220715

* Make uploaded sprites not draggable

* Call updateToolbox after adding new blocks from the backpack.

This fixes the issue where you could not see any new variables or custom block callers, as well as fixing the issue where the toolbox did not update from then on.

* Use more generic download-url and remove older download-text utility

* Use "stable" tag on npm for master builds and "latest" for develop/smoke

This is different of what it is currently, where master/smoke goes to "latest" and develop goes to "develop"

* Export costumes without inlined fonts for now.

Since the common use-case is download/edit/re-upload, and we do not want to explode the storage size of assets (which are considerably larger with fonts inlined), just do what scratch2 did and do not inline the fonts. This is not ideal if kids want to use the costumes for other purposes, but until we can create a way to strip inlined fonts before uploading to the asset server, this is the conservative way to go.

* Use bowser for macOS detection

* try building again

* remove "try it"

* retain showBranding after SET_FULL_SCREEN

* add mode reducer unit tests

* add empty object to abject assign

* Initialize locale on ScratchBlocks before initial workspace injection

This fixes a problem where having a non-default locale set during the initial load would display the blocks in the default toolbox in the wrong language, because the toolbox recycling toggle in setLocale does not work on initial load because there is no editing target to refresh.

* Include a regression test for #4476

* Remove beforeunload callback that was breaking tests

* remove isAdmin from mode reducer

* Move blocks z-index up to match gui

Override the z-index for the blocks drag layer. At 1000, it will be over everything in the gui other than the navbar itself (monitors on the stage, tutorials, alerts, dropdown menus, etc.)

* Sort variable options in "of" block

* Escape special characters in user defined strings that can appear in the toolbox xml.

* fix(Typo): change 'dserver' to 'server'

* Fix toolbox updating after see-inside.

We can't rely on refreshWorkspace always causing a toolbox update, so make sure to update the toolbox manually if it is needed.

* Add an integration test for toolbox updating after see-inside.

I needed to update the player css because the editor side wasn't large enough to be useable.

* require text-encoding only if it is needed

* Wait until loader becomes stale to continue in project loader tests

* Include the `waitUntilGone` helper in the other project load tests

* Don't refer to scratchBlocksUtils from scratch-gui

* Make file uploader call callback for each uploaded file

* Add integration tests for multiple file upload

Including sprites, backdrops, costumes and sounds inputs

* There is no loading screen for blocks example, dont wait for it

* Load files one at a time to enforce load order

* chore(package): update scratch-blocks to version 0.1.0-prerelease.1549376808

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190205221329

Closes #4480

* chore(package): update scratch-l10n to version 3.1.20190206143031

* set an initial message number to skip rendering a second time

* chore(package): update scratch-render to version 0.1.0-prerelease.20190206213754

* chore(package): update scratch-blocks to version 0.1.0-prerelease.1549534784

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190207134238

Closes #4503

* Use blob download instead of data-uri to enable large downloads

This was already being used for sb3s and sprite3s, but the code was repeated in different places. Make a single helper that downloads blobs, and make all download paths use it.

* Get the asset type correctly

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190207191220

Closes #4506

* Revert "Update scratch-render to the latest version 🚀"

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190207224121

Closes #4510

* chore(package): update scratch-l10n to version 3.1.20190207224638

* chore(package): update scratch-blocks to version 0.1.0-prerelease.1549643185

* chore(package): update scratch-render to version 0.1.0-prerelease.20190208165820

Closes #4513

* Gif upload v1: sprites, costumes

* Pass function for naming costumes, add integration tests

* Fix namer conventions to be less bizarre

* Remove namer from costume and sound upload helpers to reduce spaghetti

* Remove prune step from deployment for now

It doesn't affect greenkeeper subdirectories, which are the ones that cause the most issues. And now the prune step itself is breaking builds. So until it works better, remove this step.

* Name uploaded backdrops based on the name

* Remove extra parens

* Fix backdrop uploading from gif

* Tag NPM releases

Check if an NPM release just happened. And if so, tag the commit and push to GitHub.

Thanks to @paulkaplan for pairing on this

* Simplify disposal method handling

* chore(package): update scratch-l10n to version 3.1.20190211142555

* Put the ask blocks text input above the monitors on the stage.

* Support the isDiscrete flag on slider monitors

* Add unit test to make sure isDiscrete becomes correct step size

* Prompt user about file upload replacing the current project

This only shows when you are on a project that is SHOWING_WITHOUT_ID and has changes. This is only the case when logged out, or in the desktop editor.

* Make sure to put history in try/catch so it can be used with selenium

Selenium loads the playground using the file:// protocol, where history.replaceState throws an error

* Add integration test for uploading a file and seeing the prompt

* Remove try-it step from more integration tests

* Remove more try-it steps

* Update to develop and remove try-it steps from new tests

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190212150857

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190212160344

* chore(package): update scratch-blocks to version 0.1.0-prerelease.1549990124

Closes #4540

* Share the replace project confirmation string between new and upload

* chore(package): update scratch-l10n to version 3.1.20190213142844

Closes #4536

* Show an "Importing..." alert during file import

* Make sure to only close import alert after import is done

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190213162739

* chore(package): update scratch-render to version 0.1.0-prerelease.20190213183713

Closes #4544

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190213190040

Closes #4548

* Refactor to handle errors from file reader

* Add error handling for the costumeUpload function

* chore(package): update scratch-l10n to version 3.1.20190213193054

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190213210403

* Fix loading single frame gifs by removing onDone in favor of indices

Because the `onFrame` callback could do async work, it was possible for onDone to try to submit the costumes before the onFrame had produced any. This was subject to browsers different timing treatments of setTimeout vs. promise resolution. This makes the gif-decoder use the same style of progress API as the file uploader

* move when the dragging drawable is extracted from the renderer

Extracting the drawable art data is very expensive. We should delay
doing that work as long as possible. As such it should wait until a few
more branching return statements before we extract it.

* Request a toolbox refresh after switching between tabs to fix #4561

* chore(package): update scratch-vm to version 0.2.0-prerelease.20190215190223

* chore(package): update scratch-l10n to version 3.1.20190218084652

### Resolves

- Resolves #

### Proposed Changes

### Reason for Changes

### Test Coverage

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


davidaylaian:
Oops, my bad. Didn't mean to make this.

Issue #4241, opened 2019-01-03
Can't middle/ctrl/cmd-click account dropdown to open links in a separate tab


towerofnix:
### Expected Behavior

Holding control or command while clicking, or middle-clicking, should open links in the account dropdown in a separate tab.

### Actual Behavior

All those cases load the link in the tab used by the editor.

### Steps to Reproduce

Open this dropdown menu, then click it in one of the ways described above.

![The account dropdown in the top right of the online editor, including options "Profile", "My Stuff", "Account settings", "Sign out"](https://user-images.githubusercontent.com/9948030/50641847-ab220300-0f3f-11e9-8e01-47f3dfcaa3e0.png)

Note that the "my stuff" *icon* is an actual `<a>` link, and so it has all the usual expected behavior of `<a>` links (e.g. also having `cursor: pointer` by default, something the account dropdown is glaringly missing). As such, options in the account dropdown should all be made to be `<a>`s.

danfarrow:
I think this is fixed / no longer an issue. 

The markup for the account dropdown now uses `<a>` tags and has this structure:

```<ul class="dropdown production open">
  <li>
    <a href="/users/my-username/">
      <span>Profile</span>
    </a>
  </li>
  <li>
    <a href="/mystuff/">
      <span>My Stuff</span>
    </a>
  </li>
  <!-- etc -->
</ul>
```

towerofnix:
@danfarrow That's only the case on scratch-www, not for the menu in scratch-gui, where the structure is like this:

```
<ul class="..">
  <li class="..">
    <span>Profile</span>
  </li>
  <li class="..">
    <span>My stuff</span>
  </li>
  <!-- etc -->
</ul>
```

danfarrow:
Oh whoops I was not aware of scratch-www! I had been following this project thinking it was the scratch 3.0 web gui. Forgive my ignorance but how does scratch-gui differ/where is it deployed?

towerofnix:
@danfarrow No worries! The LLK repositories are a bit of a labyrinth to navigate. :)

The scratch-www repository is used as the implementation for all the React-based pages on the Scratch website; for example, the homepage, the privacy policy page, and the messages page. (But not pages that are yet to be "migrated" to the more modern React, such as the My Stuff page; those are in a private repository called scratchr2.)

[...]

PR #4589, opened 2019-02-25
fix(UI): fix cursor property when hovering account nav items

### LINKED ISSUE #4584: Cursor could turn into a hand when hovering account dropdown
### Expected Behavior

I think cursor should keeps same format when hovering any nav dropdown
like this:
img_0446

### Actual Behavior

But when you hovering account dropdown, cursor is this:
img_0457

and this:
img_0447

### Steps to Reproduce

1. Login
2. open account dropdown
3. hovering over menu item

MacOS Chrome

JinXJinX:
### Resolves

- Resolves #4584 

### Proposed Changes

Change cursor to a hand when hovering account nav items

### Reason for Changes

make the UI look better and consistent

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [x] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/menu/menu.css
+++ b/src/components/menu/menu.css
@@ -37,6 +37,10 @@
     background-color: $ui-black-transparent;
 }
 
+.menu-item.hoverable {
+    cursor: pointer;
+}
+
 .menu-section {
     border-top: 1px solid $ui-black-transparent;
 }
--- a/src/components/menu/menu.jsx
+++ b/src/components/menu/menu.jsx
@@ -39,7 +39,11 @@ const MenuItem = ({
     onClick
 }) => (
     <li
-        className={classNames(styles.menuItem, className)}
+        className={classNames(
+            styles.menuItem,
+            styles.hoverable,
+            className
+        )}
         onClick={onClick}
     >
         {children}


towerofnix:
This is not the appropriate fix here, because the root of the problem is that the elements are not `<a>`s when they should be. That causes a problem arguably worse than the cursor being wrong - the elements do not actually behave like links. I've filed that in #4241, which is also labelled help wanted. Fixing that (by making the links `<a>`s) would both make them behave as links middle-click-wise as well as make `cursor: pointer` happen automatically (since that's the default browser style for `<a>`s).

JinXJinX:
> This is not the appropriate fix here, because the root of the problem is that the elements are not `<a>`s when they should be. That causes a problem arguably worse than the cursor being wrong - the elements do not actually behave like links. I've filed that in #4241, which is also labelled help wanted. Fixing that (by making the links `<a>`s) would both make them behave as links middle-click-wise as well as make `cursor: pointer` happen automatically (since that's the default browser style for `<a>`s).

@towerofnix  Thanks for the advice. I agree that use < a> tag can solve both #4589 and #4241, and it will also reduce code complexity by remove the MenuItemContainer component. 

However, on the other hand, I'm not sure about the usage scenario of middle/ctrl/cmd-click account dropdown. Because only the user who read the scratch-gui source code, knows some buttons are links, others are not. for the most of users, they are all menu items. 

Therefore, If we add that implicit method. Does this make users confused? menu items' inconsistent behavior

joker314:
HTML is supposed to describe a document. It should not be the website's decision how a website should be navigated, instead it should leave that to the customisable browser.

The advantages of using an `<a>` tag for links are:
- Many screen readers have an option to list all the links on a page.
- Many screen readers will tell you if an element is a clickable link (and this _is_ a link, and not just a button, especially because it's taking you to another page)
- Browser extensions can target links specifically
- Search Engine Optimisation
- Some people disable CSS, or override it, and they should still have obvious clues that they are dealing with clickable elements
- etc.

JinXJinX:
@joker314 for example, those three dropdown don't contain any link
screen shot 2019-02-25 at 10 24 50 pm

benjiwheeler:
About using `<a>` tags -- one of the tricky things is that some menu items (profile, my stuff, etc.) are links that will take you to another page, while others (undo, turbo, save, download, upload, etc.) are triggers for actions within the page. Even the menu items that will navigate the user away from the page often first do things like trigger an auto-save. It's not totally clear to me what the middle click (or command-click on Mac) behavior should be, or how to implement it.

benjiwheeler:
I'm going to merge this PR because it's an improvement on where we are, even if it doesn't solve the problem on a deeper level. Thanks for the submission, @JinXJinX !

PR #4595, opened 2019-02-28
Fix issue #2256 by disable text input auto capitalize feature on IOS devices

### LINKED ISSUE #2256: New variables default to capitalized on iOS Safari
(Note from smoketest doc: _to fix: see https://stackoverflow.com/questions/35513968/disable-auto-correct-in-safari-text-input_)

### Expected Behavior

Variables should be lowercased by default

### Actual Behavior

The Safari input defaults to capitalizing the first letter

### Steps to Reproduce

Create a new variable while using iOS Safari

iOS 11 Safari 11

JinXJinX:
### Resolves

- Resolves #2256 

### Proposed Changes

added two attributes to every text input (4 places: project title, assets filter, variable name, list item) to disable text input auto capitalize feature on IOS devices

### Test Coverage

updated unit/components/sound-editor.test.jsx

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [x] Safari

Android Tablet
* [x] Chrome

### DIFF
--- a/src/components/prompt/prompt.jsx
+++ b/src/components/prompt/prompt.jsx
@@ -46,6 +46,8 @@ const PromptComponent = props => (
             <Box>
                 <input
                     autoFocus
+                    autoCapitalize={'none'}
+                    autoCorrect={'off'}
                     className={styles.variableNameTextInput}
                     defaultValue={props.defaultValue}
                     name={props.label}
--- a/src/components/prompt/prompt.jsx
+++ b/src/components/prompt/prompt.jsx
@@ -46,8 +46,8 @@ const PromptComponent = props => (
             <Box>
                 <input
                     autoFocus
-                    autoCapitalize={'none'}
-                    autoCorrect={'off'}
+                    autoCapitalize="none"
+                    autoCorrect="off"
                     className={styles.variableNameTextInput}
                     defaultValue={props.defaultValue}
                     name={props.label}
--- a/test/unit/components/sound-editor.test.jsx
+++ b/test/unit/components/sound-editor.test.jsx
@@ -6,6 +6,8 @@ describe('Sound Editor Component', () => {
     let props;
     beforeEach(() => {
         props = {
+            autoCapitalize: 'none',
+            autoCorrect: 'off',
             canUndo: true,
             canRedo: false,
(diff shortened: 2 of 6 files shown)


joker314:
Thanks for the changes, they look like they solve the issue effectively! I'm not sure I agree that auto-capitalisation should be turned off completely, because it's a useful feature. 

For example, titles of projects usually start with a capital letter.

Though I agree that variables (and lists) maybe should not start (by default) with a capital letter, mostly because other (i.e. native) Scratch blocks don't and it's nice to have nothing look out of place.

I have no opinion on the asset filter.

test/unit/components/sound-editor.test.jsx
```diff
@@ -6,6 +6,8 @@ describe('Sound Editor Component', () => {
     let props;
     beforeEach(() => {
         props = {
+            autoCapitalize: 'none',
```
paulkaplan: Is this needed?

src/components/prompt/prompt.jsx
```diff
@@ -46,6 +46,8 @@ const PromptComponent = props => (
             <Box>
                 <input
                     autoFocus
+                    autoCapitalize={'none'}
```
paulkaplan: With JSX you can just use `autoCapitalize="none"` instead of the curly brackets + single quotes. Could you change that in this file and the others?

JinXJinX: I updated this : )

test/unit/components/sound-editor.test.jsx
JinXJinX: you are right, I have removed this line.

JinXJinX:
hi @paulkaplan . I updated the code style and removed redundant part. please check this PR again when you have time.

chrisgarrity:
Thanks for your contribution. We ended up deciding that input capitalization is actually expected on iOS and we'd leave Scratch as is, so the issue was closed.

PR #4599, opened 2019-02-28
Only change monitor mode on primary (left) double click

### LINKED ISSUE #2089: Double-right/middle-clicking toggles monitor modes
### Expected Behavior

Only double-left-clicking toggles a monitor's mode. (This is the behavior in Scratch 2.0.)

### Actual Behavior

Using the right-mouse or middle-mouse to double click also toggles a monitor's mode.

### Steps to Reproduce

Show a monitor (click a checkbox in the block palette), then double-click it using the right or middle mouse buttons. Note that the monitor mode changes (to "large" or "slider").

Firefox 61.0b5, Debian (testing).

### LINKED ISSUE #3863: Variable status changes by double-rightclicking
### Expected Behavior
The status should not change when we double-right-click the variable. 

### Actual Behavior

It changes when double-right-clicking the monitor.

### Steps to Reproduce
Double-right-click the monitor

Windows 7, Firefox 63

towerofnix:
### Resolves

Fixes #2089 (also fixes #3863).

### Proposed Changes

Makes double-clicking a monitor only change the mode if done with the primary (left) mouse button.

### Reason for Changes

Other buttons (right, middle/scrollwheel) usually aren't used for double clicking, so generally, users don't expect programs to respond to those buttons acting as double clicks.

### Test Coverage

Tested manually.

### Browser Coverage

Linux Firefox, Chromium.

### DIFF
--- a/src/containers/monitor.jsx
+++ b/src/containers/monitor.jsx
@@ -108,7 +108,10 @@ class Monitor extends React.Component {
             y: newY
         }));
     }
-    handleNextMode () {
+    handleNextMode (event) {
+        if (event && event.button !== 0) {
+            return;
+        }
         const modes = availableModes(this.props.opcode);
         const modeIndex = modes.indexOf(this.props.mode);
         const newMode = modes[(modeIndex + 1) % modes.length];


towerofnix:
@paulkaplan Agreed about that reasoning but I'm not 100% sure what that code should look like - should I be putting an arrow function in the monitor component?

```jsx
<Box
    ...
    onDoubleClick={event => {
        // props.mode === 'list' || !props.draggable ? null : props.onNextMode
        ...logic...
    }}
>
```

Or should I define that as a separate function at the top of that component file?

```jsx
const handleDoubleClick = function (event) { ... };
...
    onDoubleClick={handleDoubleClick}
```

(If the former, there's also the question of how to get `props` into scope...)

Or something else?

towerofnix:
apparently no longer needed since the original issue doesn't affect more recent versions of Firefox, so closing!

PR #4600, opened 2019-02-28
Call hideChaff when user mouses-down outside Blockly

### LINKED ISSUE #1081: Right click menu doesn't close when clicking in another area
### Expected Behavior

If you left click while the scripting area's right click menu is open, the menu should close.

### Actual Behavior

The menu only closes if you left click inside the scripting area.

### Steps to Reproduce

1. Right click in the scripting area
2. Left click in the sprites area
3. The right click menu is still there

Tested on Chrome 63.0.3239.108 and Firefox 57.0.1, both on Linux

towerofnix:
### Resolves

Fixes #1081.

### Proposed Changes

Calls `hideChaff` whenever the user mouses-down anywhere outside of Blockly.

### Reason for Changes

Users don't expect context menus to stay visible when they click outside the menu. It'll already be hidden if the user clicks anywhere inside of Blockly, but outside, it isn't.

We explicitly check that the mouse-down doesn't happen inside Blockly (or any related elements) because Blockly handles relevant logic when it does.

### Test Coverage

Tested manually, but not *super* extensively. Clicking menu options still works, for certain, and the PR does call `hideChaff` at the proper times.

### Browser Coverage

Linux Firefox, Chromium.

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -66,6 +66,7 @@ class Blocks extends React.Component {
             'onBlockGlowOff',
             'handleExtensionAdded',
             'handleBlocksInfoUpdate',
+            'onDocumentMouseDown',
             'onTargetsUpdate',
             'onVisualReport',
             'onWorkspaceUpdate',
@@ -112,6 +113,8 @@ class Blocks extends React.Component {
         // @todo change this when blockly supports UI events
         addFunctionListener(this.workspace, 'translate', this.onWorkspaceMetricsChange);
         addFunctionListener(this.workspace, 'zoom', this.onWorkspaceMetricsChange);
+
+        document.addEventListener('mousedown', this.onDocumentMouseDown);
 
         this.attachVM();
         // Only update blocks/vm locale when visible to avoid sizing issues
@@ -177,6 +180,7 @@ class Blocks extends React.Component {
         this.detachVM();
         this.workspace.dispose();
         clearTimeout(this.toolboxUpdateTimeout);
+        document.removeEventListener('mousedown', this.onDocumentMouseDown);
     }
 
     setLocale () {
@@ -271,6 +275,26 @@ class Blocks extends React.Component {
                 block.inputList[0].fieldRow[0].setValue(value);
             }
         });
+    }
+
+    onDocumentMouseDown (evt) {
+        // When the user clicks not inside the Blockly injection div *or* a widget/etc, the click certainly happened
+        // outside of Blockly, so call hideChaff to hide inputs, menus, etc. (If the click happened inside one of those
+        // elements, Blockly will handle relevant logic itself.
+
+        const injectionDiv = this.workspace.getInjectionDiv();
+        let el = evt.target;
+        while (el) {
+            if (el === injectionDiv) {
+                return;
+            }
+            if (Array.from(el.classList).find(cls => /blockly(.*)Div/.test(cls))) {
+                return;
+            }
+            el = el.parentElement;
+        }
+
+        this.ScratchBlocks.hideChaff();
     }
 
     onTargetsUpdate () {


src/containers/blocks.jsx
```diff
@@ -273,6 +277,26 @@ class Blocks extends React.Component {
...
+        // outside of Blockly, so call hideChaff to hide inputs, menus, etc. (If the click happened inside one of those
+        // elements, Blockly will handle relevant logic itself.
+
+        const injectionDiv = this.workspace.getInjectionDiv();
+        let el = evt.target;
+        while (el) {
+            if (el === injectionDiv) {
+                return;
+            }
+            if (Array.from(el.classList).find(cls => /blockly(.*)Div/.test(cls))) {
```
towerofnix: I don't really like this kludge, but it's the best I can come up with. Blockly doesn't seem to have any "does this element belong to Blockly" function, and it seems it would be an awfully specific function to only be used here.

picklesrus: It is a bit brittle to assume from gui that the blocks will have this specific dom structure. How about doing something like what the action menu is doing?  It binds a listener that then checks whether click is inside this.blocks.   Here's a link to the action menu code: https://github.com/LLK/scratch-gui/blob/develop/src/components/action-menu/action-menu.jsx#L68

towerofnix: I'd like to do something like that, but the trouble is that the element may not even be a child of the injection div. `blocklyWidgetDiv` (contains Blockly context menu) and `blocklyDropdownDiv` (contains block dropdown menus) are key here, since they are appended straight onto `document.body`, not within the injection div.

I originally was doing something very similar to the code you linked, but I ended up just using a `for` loop since I had to do this additional classList test too.

picklesrus: Hmm, but if the click is outside of the blocks div, it seems safe to call hide chaff regardless of the structure of blocks html tree?  Or maybe I am misunderstanding how contains works? Does contains check inside the rectangle or does contains check this element is a child of in the dom?

towerofnix: The `contains` function returns whether or not the passed element is a descendant - nothing to do with bounding boxes, which we'd have to check with some other method (`getBoundingClientRect` and see if the click was within that range). But it's sort of irrelevant anyway, since such a bounding-box check *doesn't* guarantee that it's safe to call `hideChaff()`. For example, suppose the Blockly context menu is opened in such a position that it is partially outside the Blockly injection div:

The Blockly context menu, partially covering the Scratch stage.

Then suppose the user mouse-downs "Add Comment" - but on the area outside the Blockly workspace (e.g. over the stage). If we did an "is the click contained within the bounding box of Blockly?" test, it'd be false, so we would call `hideChaff` - instead of leaving control up to the Blockly widget div. This would hide the menu early and prevent the user from actually clicking on the menu option.

As far as I can tell, to decide whether to call `hideChaff` in a totally safe way, we have to check if any of the element's ancestors are a Blockly div; and that's what my code above does (brittle-y, albeit successfully).

picklesrus: Ah, got it.  I misunderstood how contains worked.

ScratchBlocks.WidgitDiv.DIV is how you can get a handle on that div.  I don't love digging this deep into the blocks from GUI, but it is at least a public interface as opposed to walking the DOM structure and looking at things that the blocks library essentially assumes are private.  So what if you check whether the blocks div contains the target click and the widget div (which you should be able to get using ScratchBlocks.WidgetDiv.DIV) doesn't contain it?

towerofnix: I can do that, but there's at least one other case where we want Blockly to take control of the click event (i.e. don't call `hideChaff`) - the block dropdown element, `.blocklyDropdownDiv`. So we would have to check for that element too. This feels like a lot of special-casing, but then so is checking for the class name, which is basically private to boot as you mentioned, so do you think checking for both is worth doing?

I'm somewhat inclined to make a public function in scratch-blocks for checking if an element is part of Blockly, so at least we could move all this special-casing into scratch-blocks. Which means, if any more elements like these are added, maintenance would need to be kept up in that repo rather than here. That could be useful since it wouldn't mean having a PR in scratch-gui whenever a new element is added in scratch-blocks (definitely not often, but still). It'd be nice if there were some way to avoid having to do special-casing anywhere at all (e.g. an inherent "register out-of-injection-div element" function which would need to be called to append to the body, or something similar). But maybe that would be overkill, considering the "is element part of Blockly" function only has an immediate use in this PR.

thisandagain:
@picklesrus Can you please re-review when you have a moment?

chrisgarrity:
Thanks for your contribution. Note that we decided that the risks associated with fixing the bug outweigh the benefits, so we closed the issue.

Issue #4580, opened 2019-02-21
Bootstrap project loading


mzgoddard:
### Expected Behavior

Render anything, load project data and load assets as soon as possible.

### Actual Behavior

For the full editor it can be seconds before anything is visible on screen. Loading project data is delayed until about when the first onscreen render of the produced html. Loading project assets is delayed until fonts have been officially loaded which is delayed until the google analysis pixel is loaded.

Here is an example summary breakdown of loading the editor (https://llk.github.io/scratch-gui/develop/#169401431) on a MBP 2017 15" laptop.

1. Compile JS [1117ms, can be cached]
2. Evaluate JS [775ms]
2.1 Evaluate VM
2.2 Evaluate GUI
3. Render React Application [790ms]
3.5 Render Scratch-Blocks [456ms]
3.6. Rerender parts [62ms]
3.6.5 Load project json
4. Idle [potentially seconds]
4.1 Blockly.Events.fireNow_ / blocklyListen [50ms]
4.2 Rerender toolbox with same xml [152ms, can be de-duplicated]
4.3 Loading react rendered assets [can be cached]
4.4 Loading google's analysis pixel [cannot be cached]
5. Ready state change, Call Font.load for all fonts [brief]
6. Fonts loaded [brief]
7. Deserialize project [383ms]
8.1 Compile scratch-parser schema [54ms]
8.2 Validate data [2ms]
8.3 Transform Data into Objects [123ms]
8.4 Load assets [196ms]
9. Idle, waiting for first asset data to load [varies, can be cached]
10. Transform raw asset data into costumes and sprites [varies]
11. Render Sprite Panel [307ms]
12. Render toolbox for selected Sprite [266ms]

These numbers are from a pretty powerful computer, other machines like chromebooks, phones and other devices will have longer times.

### Direction

I think this issue isn't about making scratch spend less cpu time on loading tasks but breaking up and reordering some of the steps the gui takes loading. This will make the early part of loading more responsive and decrease overall loading time by reducing idle time that happens before a project by id or default project is deserialized and costumes and sounds are created.

I think a possible order may look like (I've left out some details from above for brevity):

1. Compile JS
2. Evaluate HOCs, VM, Storage, (render-fonts), and Loader
3. Render React Application
3.6. Rerender parts
3.6.5 Load project json
4. Intentionally idle for some frames
5. Evaluate and render rest of GUI (excluding Scratch-Blocks)
6. Idle, wait on project.json
7. Deserialize project
[...]

PR #4611, opened 2019-03-01
wrap some parts of gui in a hoc to delay rendering if possible

mzgoddard:
This is a Proof of Concept PR for https://github.com/LLK/scratch-gui/issues/4580.

I don't think this PR will be merged but instead we'll pick apart the pieces we want or replace them with entirely new work but probably fulfilling the same goal.

This PR works towards the goal of improving the initial page load time by targeting four points during the loading process.

1. Project data file is requested.
2. Data starts deserializing and project assets are requested.
3. Loader component is renderered.
4. Project is fully loaded.

As an example of how this Proof of Concept improves loading, the first step on develop can take a 2017 Macbook Pro 1.5 seconds to execute. This PR separates the work so the first step can take about 0.3 seconds on the same machine. Step 2 with this change can take 0.2 seconds. So when the data request is made if it completes in 0.2 seconds we can be fetching assets by the time develop is making the first data request. Since the current loading process has network caused idle time, this PR uses that idle time to render the GUI in pieces. So we can see improved load times by 1 second on the same machine. Then when we get to other devices like a Chromebook we can see larger improvements since that first 1.5 seconds on develop may take 3 seconds on a Chromebook.

This PR breaks apart the work performed before the data is requested into 17 parts. The first is a technical minimum (optimistically it could be further decomposed) that creates a vm, storage, redux store, and renders some of the top level HOCs with a blue background. This is what is needed in this PR to perform the first loading point.

The next 5 are needed to deserialize the project data and start asset requests. The first of these 5 steps renders the remaining top level HOCs and creates the AudioEngine. The next two evaluate scratch-svg-renderer and scratch-render separately. The next builds the renderer. In develop, the Stage container creates the renderer. If we can move that point earlier to here we can render less before deserializing the project data. The last part prepares the next set of code to load, it needs to do this so that once the data is deserliazing the next group is able to immediately render after that.

The next 3 render the animated Loader component. The first two evaluate scratch-paint and scratch-gui/src/components/gui/gui.jsx separately. The last renders the GUIComponent, thus rendering the Loader and many other components. If the project data is deserialized before these steps, they are immediately executed once deserialization completes. This way the Loader component is render no later than when the assets are requested, but if the data request is taking too long, the Loader will still be rendered to assure scratchers that gui is in fact loading.

The next 7 evaluate and render 4 parts of the GUI: MenuBar, TargetPane, CostumeTab, and SoundTab.

Up to this point, the 15 parts after the first are allowed to render as long as JS seems to be idle. The final part renders scratch-blocks. It is left unrendered until after the data is deserialized and the assets are requested.

All together this lets the project data be requested as soon as possible, after the first part. Then the data can be serialized after the next 5. The Loader component will render as soon as it can or immediately after the data has been deserialized. And the remaining parts can fill in over time or immediately when the project is fully loaded.

```js
<StageWrapper
    isRendererSupported={isRendererSupported}
    isRtl={isRtl}
    stageSize={stageSize}
    vm={vm}
/>
```

This is a use of the work in this PR. The react element is actually left unmodified.

For this to load on demand or asynchronously when not yet needed, this PR currently performs a few steps to wrapp StageWrapper.

```js
const LoadStageWrapper =
    delayHOC.loadComponent(() => require('../../containers/stage-wrapper.jsx'));
```

First we wrap it in a HOC that takes a callback that called returns the intended to component. Then it renders that lazily evaluated component by passing the properties that are passed to StageWrapper in its use as a ReactElement.

```js
const AsyncStageWrapper = delayHOC({
    ready: true,
    stall: delayHOC.loading,
    weight: 2
})(LoadStageWrapper);
```

Next we wrap the Load component. This next component uses three values to determine if and when it renders. If ready is true and stall is falsy, the wrapped component renders immediately which causes the load component to evaluate StageWrapper and render it. If stall is true, weight determines the order the wrapper is eventially rendered in relation to other delayed elements in the React application.

Ready, stall, and weight can be functions that are given the redux state. Those functions should return booleans for ready and stall and numbers for weight based on changes in the redux state. With these functions the wrapper can determine when to render the wrapped component based on values and not just time passing.

The next step the work takes is to separate evaluating the component's module and rendering the component.

```js
const EvalStageWrapper =
    delayHOC.loadNull(() => require('../../containers/stage-wrapper.jsx'));
```

This component when rendered will evaluate stage-wrapper.jsx but renders nothing.

```js
const AsyncEvalStageWrapper = delayHOC({
    ready: true,
    stall: true,
    weight: 1
})(EvalStageWrapper);
```

And this will render Eval which will evaluate stage-wrapper.jsx shortly after being inserted into the DOM.

We can add this to the AsyncStageWrapper component.

```js
const AsyncStageWrapper = delayHOC({
    ready: true,
    stall: delayHOC.loading,
    weight: 2,
    placeholder: () => <AsyncEvalStageWrapper />
})(LoadStageWrapper);
```

All together this delays evaluation and rendering the StageWrapper while stall is true. During that the placeholder is rendered instead of LoadStageWrapper. The placeholder will then shortly after evaluate stage-wrapper.jsx. And after that Async will render Load, and so render the full component.

Most of the uses of delayHOC in this PR follow this pattern to break evaluating modules and rendering the components of those modules. There is one other important use.

https://github.com/mzgoddard/scratch-gui/blob/poc-20190222/src/containers/gui.inner.jsx#L99-L121

This function and its lazy execution serve two purposes. As mentioned we need four objects to load a project. The last one created in the GUI is the Renderer. In the develop branch this is done in containers/stage.jsx. If we move its creation to this point we can delay rendering components/gui/gui.jsx and thereby delay containers/stage.jsx until after we start fetching assets. That way we are fetching the assets sooner.

The second purpose is reusing the canvas that tests is the Renderer is supported. Testing for support and creating the Renderer need to construct a webgl context with the same details to make sure its supported. As isSupported(canvas) creates the WebGL context, passing that canvas to the Renderer constructor reuses the WebGL context. This is an additional small performance improvement by de-duplicating that work.

### Other Things

This PR includes a behaviour like delayHOC called conditionHOC.

There are two properties to components/gui/gui.jsx that are more active than the other properties: loading and targetIsStage. When only these properties change the GUI component and its children are re-rendered or considered for rendering. This means when we show or hide the Loader component, much of GUI re-renders.

Similarly targetIsStage changes when we clear the VM add the Stage to the VM, select the editingTarget if it isn't the Stage, or when a scratcher selects the stage or a sprite after selecting the stage. These cases cause much of GUI to re-render.

If we move those properties out of the set passed to components/gui/gui.jsx we can remove these extra re-renders and update smaller parts of the application for these state changes. Most of the remaining properties passed to GUI are far less likely to change (especially while loading).

conditionHOC offers a way we could use to test if something should be rendered based on redux state. It's a small wrapper like delay's ready, to test a redux selector on whether the wrapped component should be rendered.

#### Extract CSS

This change benefits from extracted CSS. In development builds, the lazy evaluated modules insert their CSS into style tags. This forces the page to recalculate style and layout even if the rest of the page's DOM remains absolutely unchanged.

With the CSS extracted, or other method that puts the CSS into the page at one time, those re-calculations and layouts do not happen.

In practice while the network is busy and JS is idle, the async work will take more time than with the css extracted. But if the network is quick, loading time will have little difference with or without CSS extracted. This is in regards to the work proposed here. There are still other reasons to extract the css.

### Other Other Things

I wanted to give some mentions to changes that I think we can more immediately implement that I discovered by looking into this work.

<details>
<summary>onProjectChanged repeats</summary>

```js
handleProjectChanged () {
    if (this.props.shouldEmitUpdates) {
        if (!this._projectChangedPromise) {
            this._projectChangedPromise = Promise.resolve()
                .then(() => {
                    this._projectChangedPromise = null;
                })
                .then(() => this.props.onProjectChanged());
        }
    }
}
```

Deserializing projects and sprites from the backpack create a lot of block instances in each Sprite's Blocks. Each block created emits a project changed event. Connected to vm-listener-hoc, it then dispatches a redux change for each event. Redux will re-render components connected to the projectChange state once, but it still has to check all of the currently rendered Redux connected components, create their state derived props, merge those props, and shallow compare them. This is pretty fast, but multiply that by the hundreds, thousands, tens of thousdands of blocks created deserializing a project and it adds a obvious amount of time.

A simple option we might do here is debounce until the end of the microtask queue, that redux dispatch. A slightly more involved change would for vm-listener-hoc to connect to projectChanged and not dispatch the change if the project is already marked changed.

</details>

<details>
<summary>projectChanged is state only</summary>

```js
handleClickNew () {
    let readyToReplaceProject = true;
    // if the project is dirty, and user owns the project, we will autosave.
    // but if they are not logged in and can't save, user should consider
    // downloading or logging in first.
    // Note that if user is logged in and editing someone else's project,
    // they'll lose their work.
    if (this.props.projectChanged && !this.props.canCreateNew) {
```

Relatedly, components connected to projectChanged seem to use it for behaviour and not for rendering. But they still re-render as they don't have a shouldComponentUpdate or other behaviour to separate the application state change from the rendering state. MenuBar as one example re-renders much of its hierarchy when projectChanged changes. MenuBar uses projectChanged to determine behaviour instead of rendering. We probably don't need to hunt down all the cases of this. Many of them are pretty small. MenuBar is one example we should probably change as its at a high enough level it causes enough work to be done to clearly show up in a flame chart.

</details>

<details>
<summary>border-radius layers</summary>

```css
.stage {
    /*
        Fixes a few extra pixels of margin/padding, that adds on to the bottom
        of the element, which messes up the chrome padding consistency
    */
    display: block;

    border-radius: $space;
    border: $stage-standard-border-width solid $ui-black-transparent;
```

We can reduce the time browsers spend compositing layers by making this stage element with border-radius its own layer with will-change or translatez.

</details>

<details>
<summary>scratch-parser Buffer.toString</summary>

```js
// Validate input type
var typeError = 'Input must be a Buffer or a string.';
if (!Buffer.isBuffer(input)) {
    try {
        input = new Buffer(input);
```

This is part of scratch-parser but related to loading performance changes around step 2. In the case that input is a utf8 encoded string of json, we are doing an expensive copy her. Many projects have json in the hundreds of thousands of KB in Scratch 3. Instead we only need the first 3 bytes to determine how to handle input.

```js
return callback(null, [input.toString('utf-8'), null]);
```

In one case once we can assume the input is a utf-8 string we are using a function that calls a function called slowToString. It is a slow function. Like many other places we can replace this with TextDecoder and in the case we are on a system that doesn't support that (Edge) we can use the polyfill that 4 other scratch repos are using.

</details>

<details>
<summary>workspaceMetrics repeats</summary>

```js
const workspaceMetrics = Object.assign({}, this.state.workspaceMetrics, {
    [target.id]: {
        scrollX: this.workspace.scrollX,
        scrollY: this.workspace.scrollY,
        scale: this.workspace.scale
    }
});
this.setState({workspaceMetrics});
```

Blocks.onWorkspaceMetricsChange seems to be called 4 times after a drag in blocks.jsx during the same mouseup even. The metrics values haven't changed from a drag block, and probably never changes while handling one mouseup event.

I'm not sure why those four times are happening but I'd look first to the setState call here. We can easily test if the values have changed. And then only setState if they have. So even if we get extra calls to the method we won't cause a partial react render each time.

</details>

<details>
<summary>Stage makes renders for block drags</summary>

```js
onMouseUp (e) {
    const {x, y} = getEventXY(e);
    const mousePosition = [x - this.rect.left, y - this.rect.top];
    this.cancelMouseDownTimeout();
    this.setState({
        mouseDown: false,
        mouseDownPosition: null
    });
```

Stage.mouseUp is bound to the document. So after a mouseUp from dragging a block, it also runs. I don't think we need to change the bind, but the logic in cancelMouseDownTimeout and the logic to setState here, should check that they will change the state first. setState is called here when the state will not change and there are also two setStates happening in quick succession. As cancelMouseDownTimeout also calls setState the is rendered twice when it should be once.

First thing would be to check that the values are changing before calling here. We don't need to always check. But in the case of here in Stage, Blocks, and others they execute enough unnecessary work that we probably improve the situation.

</details>

<details>
<summary>Backpack makes renders for block drags</summary>

```js
handleBlockDragUpdate (isOutsideWorkspace) {
    this.setState({
        blockDragOutsideWorkspace: isOutsideWorkspace
    });
}
```

Same thing here in Backpack.handleBlockDragUpdate. I'm not sure if we want to use some kind of common helper for this.

</details>

### DIFF
--- a/src/components/bootstrap-loader/bootstrap-loader.jsx
+++ b/src/components/bootstrap-loader/bootstrap-loader.jsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import {FormattedMessage} from 'react-intl';
+import classNames from 'classnames';
+import styles from './loader.css';
+import PropTypes from 'prop-types';
+
+const mainMessage = (
+    <FormattedMessage
+        defaultMessage="Bootstrapping"
+        description="Main loading message"
+        id="gui.bootstapLoader.headline"
+    />
+);
+
+class LoaderComponent extends React.Component {
+    render () {
+        return (
+            <div
+                className={classNames(styles.background, {
+                    [styles.fullscreen]: this.props.isFullScreen
+                })}
+                dir={this.props.dir}
+            >
+                <div className={styles.container}>
+                    {this.props.children}
+                </div>
+            </div>
+        );
+    }
+}
+
+LoaderComponent.propTypes = {
+    isFullScreen: PropTypes.bool
+};
+LoaderComponent.defaultProps = {
+    isFullScreen: false
+};
+
+export default LoaderComponent;
--- a/src/components/bootstrap-loader/loader.css
+++ b/src/components/bootstrap-loader/loader.css
@@ -1,0 +1,50 @@
+@import "../../css/colors.css";
+@import "../../css/z-index.css";
+
+.background {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    z-index: $z-index-loader;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    background-color: $motion-primary;
+    font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+    text-align: center;
+    color: white;
+}
+
+.fullscreen {
+    /* Break out of the layout using position: fixed to cover the whole screen */
+    position: fixed;
+    /* Use the fullscreen stage z-index to allow covering full-screen mode */
+    z-index: $z-index-stage-wrapper-overlay;
+}
+
+.block-animation {
+    width: 125px;
+    height: 150px;
+    margin: 50px auto 0px;
+}
+
+.block-animation img {
+    display: block;
+    position: relative;
+    height: 30%;
+    margin-top: -4px;
+}
+
+.top-block,
+.middle-block,
+.bottom-block {
+    visibility: hidden;
+}
+
+.title {
+    font-size: 2rem;
+    font-weight: bold;
+    margin: 0.75rem 0;
+}
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -10,14 +10,8 @@ import tabStyles from 'react-tabs/style/react-tabs.css';
 import VM from 'scratch-vm';
 import Renderer from 'scratch-render';
 
-import Blocks from '../../containers/blocks.jsx';
-import CostumeTab from '../../containers/costume-tab.jsx';
-import TargetPane from '../../containers/target-pane.jsx';
-import SoundTab from '../../containers/sound-tab.jsx';
-import StageWrapper from '../../containers/stage-wrapper.jsx';
 import Loader from '../loader/loader.jsx';
 import Box from '../box/box.jsx';
-import MenuBar from '../menu-bar/menu-bar.jsx';
 import CostumeLibrary from '../../containers/costume-library.jsx';
 import BackdropLibrary from '../../containers/backdrop-library.jsx';
 import Watermark from '../../containers/watermark.jsx';
@@ -33,6 +27,8 @@ import TelemetryModal from '../telemetry-modal/telemetry-modal.jsx';
 
 import layout, {STAGE_SIZE_MODES} from '../../lib/layout-constants';
 import {resolveStageSize} from '../../lib/screen-utils';
+import delayHOC from '../../lib/delay-hoc.jsx';
+import conditionHOC from '../../lib/condition-hoc.jsx';
 
 import styles from './gui.css';
 import addExtensionIcon from './icon--extensions.svg';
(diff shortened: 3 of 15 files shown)


mzgoddard:
Crud. I wanted to make this a draft PR and forgot. Probably too late now.

rschamp:
To summarize our offline discussion: This is a great proof of concept, so we'd like to break it down into a few smaller pieces and improve the ergonomics of using the delay HOC. We spoke about doing one or more preliminary PRs that would make this work easier to do in a more ergonomic way. Secondly, trying to start with a minimal change so that we just update the interaction between project fetching and rendering, and then revisiting more after that.

PR #4613, opened 2019-03-03
Use validator to update "of" block's property field when object changes

### LINKED ISSUE #4400: Changing  sprite of a "(attribute) of (sprite)" in block palette rearranges blocks in the workspace (sometimes)

towerofnix:
### Resolves

Fixes #4400. Also resolves LLK/scratch-vm#1787 (more or less - you still have to tweak the value in both places, but they're in the same file; we could move the arrays to a constant at the top of the file if wanted though, and not worry about exporting, which would've been required before).

Depends on LLK/scratch-blocks#1908 to add support for passing a validator to dropdown fields through `jsonInit`. Should be merged at the same time as LLK/scratch-vm#2034, which removes the old "of"-switching behavior.

### Proposed Changes

In current Scratch 3.0, when you change the `OBJECT` input of an "of" block, the `PROPERTY` field is updated correspondingly, so that it has an appropriate value selected. (If you switch to a sprite, it is "x position"; if you switch to the stage, it's "backdrop #".) The VM handles this; but since the VM only keeps track of representations of blocks, not the actual Blockly blocks themselves, it doesn't have the capability to simply update a field's selected value. Instead, it requests that the gui [update the whole workspace](https://github.com/LLK/scratch-vm/blob/ff59ebd7c5cb0cca3a22efae15dda1e273457cd8/src/engine/blocks.js#L588-L595).

This PR makes it so that, rather than updating the whole workspace from the VM, only the affected "of" block is reloaded. It uses a validator function to do this; validator functions are basically callbacks that are run every time a field changes. Typically, validator functions are used to assert that a selected value is correct (and otherwise undo the change); in this PR, we instead make use of one to update another dropdown. (This is very similar in principle to the "stop" block, [which changes whether or not a block can be placed below it based on the selected input](https://github.com/LLK/scratch-blocks/blob/f53412b38ee769bc48c5f34312ddb3287a0e678e/blocks_vertical/control.js#L193-L205).)

In order to actually change the field input, we serialize the block into XML/DOM, tweak the field input, then recreate the block in the correct place. (This is essentially the same behavior [as in my "relabel" PR](https://github.com/LLK/scratch-blocks/blob/272337288390ad278a2b9d6d9422d5962b174c82/core/scratch_blocks_utils.js#L260-L271).) We have to recreate the block like this because scratch-blocks doesn't provide us any way to update the attribute field's value - in fact, the attribute field isn't actually stored in any accessible way on the "of" block object.

### Reason for Changes

The current behavior, updating the entire workspace when an object dropdown changes, is very costly performance-wise, and has a host of other frustrating side-effects as well. These are described in #4400 - a particularly annoying one is that it sometimes seems to reload the scroll position of the workspace. With these issues, the "of" block is virtually unusable in any kind of large project.

The new behavior refreshes only the modified "of" block, and does so in a way that doesn't have any side-effects. This greatly improves the performance of the block, and makes working with it useful in any project.

### Test Coverage

Tested manually by switching the values in the dropdown, then confirming that the correct value is returned, and that the block is placed in the right position or input.

### Browser Coverage

Linux Firefox, Chromium.

### DIFF
--- a/src/lib/blocks.js
+++ b/src/lib/blocks.js
@@ -7,7 +7,7 @@ import ScratchBlocks from 'scratch-blocks';
  */
 export default function (vm) {
 
-    const jsonForMenuBlock = function (name, menuOptionsFn, colors, start) {
+    const jsonForMenuBlock = function (name, menuOptionsFn, colors, start, validator) {
         return {
             message0: '%1',
             args0: [
@@ -16,7 +16,8 @@ export default function (vm) {
                     name: name,
                     options: function () {
                         return start.concat(menuOptionsFn());
-                    }
+                    },
+                    validator: validator
                 }
             ],
             inputsInline: true,
@@ -207,7 +208,49 @@ export default function (vm) {
         const stage = ScratchBlocks.ScratchMsgs.translate('SENSING_OF_STAGE', 'Stage');
         const json = jsonForMenuBlock('OBJECT', spriteMenu, sensingColors, [
             [stage, '_stage_']
-        ]);
+        ], function (choice) {
+            const input = this; // eslint-disable-line no-invalid-this
+
+            // Only update property field when object input actually changes.
+            if (choice === input.getValue()) {
+                return;
+            }
+
+            // Determine the new value to put in the property field; this is based on which object is selected, and
+            // should be equal to the first value in the dropdown that would show up when you click the field while
+            // that object is selected.
+            let newValue;
+            if (choice === '_stage_') {
+                newValue = 'backdrop #';
+            } else {
+                newValue = 'x position';
+            }
+
+            const oldBlock = input.sourceBlock_.parentBlock_;
+            const ws = oldBlock.workspace;
+
+            // Since this is a validator function, the value of the object input won't actually be set yet.
+            // We need to set it before we get the XML of the original block; otherwise the XML (and hence the new
+            // block) will have the old object value selected.
+            input.setValue(choice);
+
+            // Get the XML of the original block, then tweak the returned PROPERTY field so that it has the intended
+            // new value. Then generate the new block.
+            const xml = ScratchBlocks.Xml.blockToDom(oldBlock);
+            const field = xml.querySelector('field[name=PROPERTY]');
+            field.removeChild(field.firstChild);
+            field.appendChild(document.createTextNode(newValue));
+            const newBlock = ScratchBlocks.Xml.domToBlock(xml, ws);
+
+            // Restore the new block to its original output connection or position, then discard the original block.
+            if (oldBlock.outputConnection.targetConnection) {
+                newBlock.outputConnection.connect(oldBlock.outputConnection.targetConnection);
+            } else {
+                const xy = oldBlock.getRelativeToSurfaceXY();
+                newBlock.moveBy(xy.x, xy.y);
+            }
+            oldBlock.dispose();
+        });
         this.jsonInit(json);
     };
 


kchadha:
@picklesrus, want to pair review this one?

kchadha:
@towerofnix, thank you for your pull requests! Unfortunately I don’t think we’ll be able to accept these PRs because the solution has a lot of unexplored assumptions that we think could cause problems with other parts of the infrastructure.

One part of the issue is that the code in lib/blocks.js is already a piece of technical debt that we want to resolve in the future. This solution we use for handling dynamic menus is not ideal since the GUI is overwriting pieces of the scratch-blocks code. Taking this code would mean increasing our technical debt.

Another part of the issue is that scratch-blocks does not currently have this relationship with blocks where one menu depends on the value of another menu. The "validator" feature was designed to validate properties about an individual field, not to maintain a relationship between two fields. I think instead of working around this, we would want to build a solution that adds this feature to scratch-blocks. The problem right now is that’s probably a large amount of work and we don’t have concrete ideas for a reasonable solution to this problem yet.

I do think #4400 is an important issue, so I think it would be good to continue to explore a reliable repro for that issue as it might be easier to figure out what is causing the blocks to lose their position information rather than fixing this issue of updating the whole workspace vs. updating just one block.

towerofnix:
@kchadha Thanks for the review. I understand this is a considerable amount of code for a fix in an area that is already tech debt to an extent. (Should that file be modified to include a header at the top stating that adding code to it should be avoided whenever possible? After all it's a pain to contributors to find their code simply can't be merged in part because a code file was never really meant to be modified.)

>I do think #4400 is an important issue, so I think it would be good to continue to explore a reliable repro for that issue as it might be easier to figure out what is causing the blocks to lose their position information rather than fixing this issue of updating the whole workspace vs. updating just one block.

I'm very hesitant to investigate further since that issue has only occurred once during all my use of the "of" block. Although this could totally be bogus, it seems as though it may just be a rare race condition, since it's only happened the one time. Regardless, it would be surprising if it were *not* connected to reloading the workspace, and this is something we should fix regardless, because of its dramatic performance cost whenever the object input is changed.

See discussion here (yet to complete, admittedly) regarding the specific "duplicate and replace block" behavior which is also used in this PR: https://github.com/LLK/scratch-blocks/pull/1907#issuecomment-472251967

>I think instead of working around this, we would want to build a solution that adds this feature to scratch-blocks. The problem right now is that’s probably a large amount of work and we don’t have concrete ideas for a reasonable solution to this problem yet.

I agree that what we have here isn't the cleanest solution. Changing scratch-blocks to have a feature for this is probably out of scope, and would better be approached when we move the entirety of the dynamic menu code into scratch-blocks (i.e. get rid of this tech debt), but it's worth thinking about nonetheless, since the inherent performance issue (among other side-effects of reloading the workspace, like #4400) does affect anyone who tries to use the "of" block.

Please do note this line:

>(This is very similar in principle to the "stop" block, which changes whether or not a block can be placed below it based on the selected input.)

In existing code, the stop block is modifying the shape of the block when an input changes. With this PR, the "of" block changes another inputs value within the same block. Is this really so different?

Interested in your thoughts.

kchadha:
> @kchadha Thanks for the review. I understand this is a considerable amount of code for a fix in an area that is already tech debt to an extent. (Should that file be modified to include a header at the top stating that adding code to it should be avoided whenever possible? After all it’s a pain to contributors to find their code simply can’t be merged in part because a code file was never really meant to be modified.)

I think the issue at hand is more about this specific solution rather than the fact that this particular file is being edited. In general GUI should not be reaching into scratch-blocks code and be knowledgable of scratch-blocks internal apis (like modifying XML). We implicitly limit use or edits to that file through our open source contributor guidelines (e.g. we would not mark an issue as help-wanted if it required some sort of solution similar to the one we currently have for dynamic menus).

We really value your contributions, and we understand that it’s definitely a pain to submit a big PR and then be told that the solution is not correct for what might seem like vague reasons. It might be better in the future to discuss potential solutions for a bug with the Scratch Team on the issue itself, that way we can have a conversation about whether or not we think something is a good approach before getting a pull request.

> I’m very hesitant to investigate further since that issue has only occurred once during all my use of the “of” block. Although this could totally be bogus, it seems as though it may just be a rare race condition, since it’s only happened the one time. Regardless, it would be surprising if it were _not_ connected to reloading the workspace, and this is something we should fix regardless, because of its dramatic performance cost whenever the object input is changed.

I agree, it does seem like it might be a race condition of some sort. I am concerned about why the blocks are resetting to a different space. That makes it seem like the VM has out of sync information or is not sending block coordinates along with the XML when refreshing the workspace.

You are also right that reloading the whole workspace when using that block is a performance issue. We are definitely looking into improving performance across the editor and also specifically how to improve performance of scratch-blocks in general so that loading a workspace is not so much of a performance bottleneck.

I agree with your solution that the ability to reload a single block is a great idea! I don’t think validators for block fields are necessarily the right way to achieve that though. As you know, every piece of code we introduce in our codebase serves as an example to others of how to do things. This is another reason why I don’t think this solution is optimal to include. In general it is good to maintain a strong separation of concerns when designing or manipulating an API. Scratch-blocks has an assumption, inherited from Blockly, that field validators only change things about the field they are validating. I want to make sure that we are using the API according to its assumptions because when we don’t, that usually is the source of bugs and unintentional side-effects.

The `sensing_of` block is one case where the value of one field affects the value of a completely different field, and since this is a case in the core (non-extension) scratch blocks, it certainly calls for creating a concrete system around handling this within scratch-blocks and scratch-vm itself.

> Please do note this line:
>
> > (This is very similar in principle to the “stop” block, which changes whether or not a block can be placed below it based on the selected input.)
>
> In existing code, the stop block is modifying the shape of the block when an input changes. With this PR, the “of” block changes another inputs value within the same block. Is this really so different?

You brought up a great point about mutations, and it seems like maybe we might be able to do something with mutations to solve this problem, but of course without fully investigating that solution, I can’t say whether it would work or not.

PR #4627, opened 2019-03-06
Run pify package through babel loader.

kchadha:
### Resolves

Resolves an issue where es6 source files make it into built GUI preventing www from building correctly.

### Proposed Changes

Run the 'pify' package (introduced as a dependency of scratch-parser in LLK/scratch-parser#52) through the babel loader.


kchadha:
cc/ @mzgoddard

Issue #1350, opened 2018-01-26
Color picker block does not work with touch devices


paulkaplan:
### Expected Behavior

_Please describe what should happen_

Expect to be able to use the eyedropper on a touch device. It isn't 100% clear what the interaction should be, but at the very least it seems like touching down on the sprite should select a color. It doesn't seem to even do that. Testing on an iPad, I can't use the eyedropper in either the blocks/stage or the costume editor. 

### Actual Behavior

_Describe what actually happens_

It seems to dismiss the eyedropper tool without actually choosing a color.

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_

iPad, try using the eyedropper. 

I'm putting "needs discussion" on this to see if there is a slightly better interaction we can imagine for touch, and what the priority of that is. /cc @carljbowman @thisandagain 

----

Updated title because the eyedropper for the costume editor was fixed to work with touch. Now it is just the stage

hedgehog125:
I can repo in chrome devtools device toolbar.

kyleplo:
@hedgehog125 Me too! There is also a warning in the console.

hedgehog125:
Maybe it should only select when you release your finger from the screen and while you drag it behaves as if you move a mouse.

mrjacobbloom:
That sounds too easy to screw up, I'd rather it keeps the zoom open until you do some other interaction to close the dropper

hedgehog125:
Hmm... maybe there could be a tick underneath the highlighter circle?

PR #4628, opened 2019-03-06
Make the stage color picker work with touch.

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/1350

### Proposed Changes

Make the color picker on the stage work the same way the costume editor one does: show the picker on _either_  mouse down or mouse move, then submit the color on mouse up. This allows it to work on touch, where you can touch down to see the loupe and drag around to pick your color.

### Reason for Changes

The touching color block couldn't be used on touch!

----

A lot of this change is just moving blocks of code around.

This PR can be tested on touch devices at https://llk.github.io/scratch-gui/touch-picker/

### DIFF
--- a/src/containers/stage.jsx
+++ b/src/containers/stage.jsx
@@ -175,9 +175,11 @@ class Stage extends React.Component {
         const {x, y} = getEventXY(e);
         const mousePosition = [x - this.rect.left, y - this.rect.top];
 
-        // Set the pickX/Y for the color picker loop to pick up
-        this.pickX = mousePosition[0];
-        this.pickY = mousePosition[1];
+        if (this.props.isColorPicking) {
+            // Set the pickX/Y for the color picker loop to pick up
+            this.pickX = mousePosition[0];
+            this.pickY = mousePosition[1];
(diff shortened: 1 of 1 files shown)


rschamp:
Try it out: [touch-picker](https://llk.github.io/scratch-gui/touch-picker)

paulkaplan:
@rschamp thanks for catching that. Fixed.

paulkaplan:
I think we should file an issue for the growing complexity of `stage.jsx`. I don't like that modifying the logic for the color picker behavior is right next to modifying the drag behavior

src/containers/stage.jsx
```diff
@@ -112,7 +112,9 @@ class Stage extends React.Component {
     }
     startColorPickingLoop () {
         this.intervalId = setInterval(() => {
-            this.setState({colorInfo: this.getColorInfo(this.pickX, this.pickY)});
+            if (typeof this.pickX === 'number') {
```
rschamp: What situation is this not true, before it's initialized?

PR #4638, opened 2019-03-11
Remove most of the GA events

paulkaplan:
/cc @thisandagain @rschamp 

Per discussion offline, these were causing too much traffic for not enough usefulness. Leaving stuff having to do with tips library/tutorial card navigation as well as hardware extension connectivity, as those help us find issues with Scratch.

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -85,6 +85,7 @@ class Blocks extends React.Component {
         this.toolboxUpdateQueue = [];
     }
     componentDidMount () {
+        window.vm = this.props.vm;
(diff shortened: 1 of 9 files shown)


src/containers/blocks.jsx
```diff
@@ -85,6 +85,7 @@ class Blocks extends React.Component {
         this.toolboxUpdateQueue = [];
     }
     componentDidMount () {
+        window.vm = this.props.vm;
```
rschamp: Looks like you maybe left this in by accident?

paulkaplan:
Thanks, fixed!

PR #4640, opened 2019-03-12
Scratch desktop update

### LINKED ISSUE #4598: scratch-desktop branch needs to be merged in to develop
We are building scratch-desktop using a non-develop branch (gui/scratch-desktop). @cwillisf mentioned needing to follow up on getting the changes there merged in. I think this should be a priority, as there are some important features there, like showing the library using local assets.

Here is [the comparison](https://github.com/LLK/scratch-gui/compare/scratch-desktop). A few highlights
- Reconciling the changes to the way Library items are shown, both in terms of resource location (storage vs. direct url) and structure (container/component for library items)
- Allowing GUI consumer to listen for project telemetry events [(related issue)](https://github.com/LLK/scratch-vm/issues/2036)
- Reconciling the asset paths in the webpack config for static assets [(related issue)](https://github.com/LLK/scratch-gui/issues/4606)

Related issues:

/cc @chrisgarrity

cwillisf:
### Resolves

Resolves #4598 

### Proposed Changes

This PR merges the features originally built for Scratch Desktop into the main development branch. These include:
- Add telemetry events and optional hooks to collect those events
- Load library images, such as the costumes in the sprite and costume libraries, through `scratch-storage` by using the new `ScratchImage` class
- Load library images only when visible
- Support relative paths for static assets
   - Note that this change may currently be incompatible with `scratch-www`

### Reason for Changes

1. Merging these features into `develop` decreases ongoing development effort
2. Other Scratch projects would like to use these features in order to operate without a network connection

### Test Coverage

Existing tests have been updated for compatibility with these changes.

Note that `test/integration/backpack.test.js` seems unreliable locally but passes on Travis CI. This is consistent with the `develop` branch.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox


rschamp:
Try it out: [scratch-desktop-update](https://llk.github.io/scratch-gui/scratch-desktop-update)

package.json
```diff
@@ -100,6 +100,7 @@
     "react-test-renderer": "16.2.0",
     "react-tooltip": "3.8.0",
     "react-virtualized": "9.20.1",
+    "react-visibility-sensor": "^5.0.2",
```
paulkaplan: @cwillisf I think we usually pin user-facing dependencies to prevent untraceable update issues. (i.e. without a package-lock, it isn't possible to have reproducible builds without pinning to specific version)

webpack.config.js
```diff
@@ -208,7 +208,7 @@ module.exports = [
                         loader: 'file-loader',
                         options: {
                             outputPath: 'static/assets/',
-                            publicPath: '/static/assets/'
+                            publicPath: 'static/assets/'
```
chrisgarrity: This is definitely going to break scratch-www. I've been looking (without success) for a way to configure webpack that would allow both www and desktop/android to be happy.

src/containers/library-item.jsx
```diff
@@ -75,21 +75,17 @@ class LibraryItem extends React.PureComponent {
...
         }
-        return this.props.iconMd5;
+        return this.props.iconSource;
     }
     render () {
-        const iconMd5 = this.curIconMd5();
-        const iconURL = iconMd5 ?
-            `https://cdn.assets.scratch.mit.edu/internalapi/asset/${iconMd5}/get/` :
-            this.props.iconRawURL;
+        const iconSource = this.curIconSource();
```
chrisgarrity: It seems that we lose the ability to load the assets from the cdn when we merge this, and loading the library images gets slower for any online versions. If there is a way to know whether it's the desktop version or not, maybe we could set the iconSource to {uri: <cdn link>}  instead of {assetId, assetType}

cwillisf:
Closing this in favor of #4644 and #4645

zhengyongxiong:
Can scratch-desktop branches build exes executed on win10 or omgs executed on MacOS?I tried this to run dist, but I didn't succeed.

PR #4643, opened 2019-03-12
proposed BASE_PATH for webpack config.

chrisgarrity:
I made these changes to the webpack config, and tried them with my local-www. So that part works. I didn't get to trying it with android (rebuilding gui with `BASE_PATH=""` as part of the build)

This is a PR on a PR, so I'm not sure what to label it :)

Also feel free to close if you find a better solution.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -12,6 +12,8 @@ var autoprefixer = require('autoprefixer');
 var postcssVars = require('postcss-simple-vars');
 var postcssImport = require('postcss-import');
 
+const BASE_PATH = process.env.BASE_PATH || '/';
+
 const base = {
     mode: process.env.NODE_ENV === 'production' ? 'production' : 'development',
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -12,6 +12,8 @@ var autoprefixer = require('autoprefixer');
 var postcssVars = require('postcss-simple-vars');
 var postcssImport = require('postcss-import');
 
+const BASE_PATH = process.env.BASE_PATH || '/';
```
cwillisf: Judging by the way it's used, it looks like this might be more convenient / descriptive:
```suggestion
const STATIC_PATH = process.env.STATIC_PATH || '/static';
```

I defer to {@chrisg, @paulkaplan, @rschamp}'s opinion though ;)

chrisgarrity:
Closing in favor of https://github.com/LLK/scratch-gui/pull/4646

Issue #4606, opened 2019-03-01
Allow project saving method to be configured


paulkaplan:
Currently project saving is handled in GUI/project-saver-hoc, which basically does the following on a save:
1. Use `storage.store` on all dirty assets
2. Use `xhr` module to POST/PUT against `storage.projectHost/projectId`
    - The `storeProject` function is called with request params like title, is_remix, is_original depending on if it is creating a new project, remixing, copying, etc.
    - Expects {id: projectId} in response
3. Use (passed in) `props.onUpdateProjectThumbnail(projectId, thumbnailBlob)` to update the thumbnail

Step 3 is easily configurable externally. I _believe_ step 1 can be configured by customizing a storage helper to do this (although I haven't tried it, so not 100% on this). But step 2 is not configurable.

Depending on how we want this to look in the future, we could do a few things to make project saving more configurable. 
1. Use scratch-storage to store project JSON, so it can be configured the same way asset saving can be configured. There is a consistency to this, but project JSONs would be pretty different to save on the web because of authentication, so might be a fair bit of work that wouldn't really be used elsewhere. 
2. Add a `onUpdateProjectData` (or similar) prop that can be called that acts like [this promise](https://github.com/LLK/scratch-gui/blob/develop/src/lib/project-saver-hoc.jsx#L227-L267). Then www can pass in that same function, but other consumers of GUI can handle project data saving differently.
3. ... other ideas?

/cc @rschamp @chrisgarrity @cwillisf since you probably have thoughts on the approach for this.

paulkaplan:
@cwillisf I looked into configuring step (1) above, i.e. configuring the asset storage, and it looks like `storage.store` doesn't cycle through the helpers the same way `storage.load` does, and it is hard-coded to use the web helper and nothing else. So turns out configuring (1) isn't really possible right now either.

PR #4659, opened 2019-03-18
Move project saving to server to helper, allow external save action

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/4606

### Proposed Changes

Move the current save code to a helper that can be eventually moved to WWW, and allow a GUI consumer to pass in a `onUpdateProjectData` method to handle project JSON saves.

### Reason for Changes

See linked issue.

### DIFF
--- a/src/lib/project-saver-hoc.jsx
+++ b/src/lib/project-saver-hoc.jsx
@@ -232,16 +232,8 @@ const ProjectSaverHOC = function (WrappedComponent) {
                         () => (asset.clean = true)
                     )
                 )
-            ).then(() => {
-                const {onUpdateProjectData} = this.props;
-                const saveReturnValue = onUpdateProjectData(projectId,
-                    savedVMState, requestParams);
-                // Allow onUpdateProjectData to either return immediately or a promise.
-                if (!(saveReturnValue.then && saveReturnValue.catch)) {
-                    return Promise.resolve(saveReturnValue);
-                }
-                return saveReturnValue;
-            })
+            )
+                .then(() => this.props.onUpdateProjectData(projectId, savedVMState, requestParams))
                 .then(response => {
                     this.props.onSetProjectUnchanged();
                     const id = response.id.toString();
--- a/src/lib/project-saver-hoc.jsx
+++ b/src/lib/project-saver-hoc.jsx
@@ -1,15 +1,14 @@
 import bindAll from 'lodash.bindall';
 import React from 'react';
 import PropTypes from 'prop-types';
-import queryString from 'query-string';
 import {connect} from 'react-redux';
 import VM from 'scratch-vm';
-import xhr from 'xhr';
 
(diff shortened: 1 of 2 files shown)


src/lib/project-saver-hoc.jsx
```diff
@@ -413,7 +383,8 @@ const ProjectSaverHOC = function (WrappedComponent) {
     ProjectSaverComponent.defaultProps = {
         autoSaveIntervalSecs: 120,
         onRemixing: () => {},
-        onSetProjectThumbnailer: () => {}
+        onSetProjectThumbnailer: () => {},
+        onUpdateProjectData: saveProjectToServer
```
rschamp: We should remove this eventually, although adding the default makes this PR seamless, so I think it's the way to go now.

src/lib/project-saver-hoc.jsx
```diff
@@ -234,45 +233,14 @@ const ProjectSaverHOC = function (WrappedComponent) {
...
-                } else {
-                    Object.assign(opts, {
-                        method: 'put',
-                        url: `${storage.projectHost}/${projectId}${qs}`
-                    });
+                const {onUpdateProjectData} = this.props;
+                const saveReturnValue = onUpdateProjectData(projectId,
+                    savedVMState, requestParams);
+                // Allow onUpdateProjectData to either return immediately or a promise.
+                if (!(saveReturnValue.then && saveReturnValue.catch)) {
```
rschamp: Can we rely on the supplier of `onUpdateProjectData` to ensure it returns a promise? If not, maybe we should wrap the entire thing in a promise that gets returned?  Something like:
```js
return new Promise((resolve, reject) => {
    try {
        const saveReturnValue = onUpdateProjectData(projectId, savedVMState, requestParams);
    } catch (e) {
        return reject(e);
    }
    // if saveReturnValue is a promise, this wrapper promise will resolve to what saveReturnValue resolves to
    return resolve(saveReturnValue); 
});
```

paulkaplan: Yeah i wasn't quite sure whether to enforce the promise requirement. That code block actually started as `if (!returnValue.then) { throw new Error('You must return a promise'); }` but then i realized i could just resolve at that point. Also you don't need to try/catch because it is already inside a promise chain, so throwing in that function would automatically go to the next `.catch` in the chain.

paulkaplan: Actually i think you can just `return Promise.resolve(onUpdateProjectData(...))` since promises unwrap the resolve value from other promises

rschamp: That looks good to me.

paulkaplan: Oh wait, even simpler, it is in a promise chain, it doesn't matter if it is a promise or not, either way the next then will get the same result at the right time

PR #4664, opened 2019-03-18
replace Blocks.utils.is3dSupported with constant value function

mzgoddard:
### Resolves

Faster initial render of ScratchBlocks.

### Proposed Changes

Always return true from `ScratchBlocks.utils.is3dSupported`.

### Reason for Changes

The browsers that Scratch supports will always support 3D CSS Transform.

This change provides a minor performance improvement but one that can accumulate with other changes to reduce style and layout recalculations.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari

### DIFF
--- a/src/lib/blocks.js
+++ b/src/lib/blocks.js
@@ -331,5 +331,15 @@ export default function (vm) {
         return collator.compare(str1, str2);
     };
 
+    // Blocks wants to know if 3D CSS transforms are supported. The cross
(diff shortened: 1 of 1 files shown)


picklesrus:
@mzgoddard  This looks good!  If you resolve the conflicts I'm happy to merge it today.

mzgoddard:
- Fix conflicts

PR #4663, opened 2019-03-18
Precache cache text widths for blocks

mzgoddard:
### Proposed Changes

Add a function that is called with the toolbox and workspace xml before calling the standard Blocks API to create or update the presentation of that information. This function walks the same data and follows some of the Blocks workflow or makes assumptions about how it will construct `<text>` elements. It then adds all of the requisite text elements to the dom and measure them all together. Then it caches the values to use later.

### Reason for Changes

Blocks currently caches the text lengths but it changes the DOM each time before it measures the length of some text. This DOM change causes a recaluculation of the style and layout. We cannot avoid the recalc, either it happens on demand or we wait until the next frame when the browser will have recalculated anyways to render our DOM changes. Instead we want to reduce the number of recalculates by making all the DOM changes we want before measuring text length or heights of objects, etc.

Having one generally larger recalc than a lot of small ones we cost us less overall time by avoiding the context switching and other work the browser needs to do to perform this work.

### Benchmark Data

<dl>
<dt>project</dt>
<dd>tested on an empty project with an empty workspace or 173918262 with a workspace full of blocks and variables in the toolbox.</dd>
<dt>operation</dt>
<dd>toolbox 1, 2, and 3 are the three toolbox renderings that happen during an initial load currently. toolbox 2 experiences no change in this PR as it has no new text to measure lengths for so it has been excluded. workspace renders the scripts in the target sprite.</dd>
<dt>change</dt>
<dd>the change in time spent between develop and this PR</dd>
</dl>

|  project | operation | change |
| --- | --- | --- |
|  empty | toolbox 1 | 28.87% |
|  empty | toolbox 3 | 31.66% |
|  173918262 | toolbox 1 | 21.11% |
|  173918262 | toolbox 3 | 27.82% |
|  173918262 | workspace | 34.02% |

<details>
<summary>Detailed Chart</summary>

The `toolbox 1` operation takes place during the initial render and appears to have a large deviation in recorded times outside the changes of this PR. The chromebook and pi3b+ times for the `toolbox 1` operation have been excluded.

While `toolbox 1` has a lot of noise in its performance metrics due to its surrounding work, `workspace` does seem to decrease in performance due to this change directly. However, The difference seems fixed and small enough to only be recordable on the pi3b+ on an empty project. Otherwise it cannot be detected or the improvement outweighs the cost.

<dl>
<dt>device</dt>
<dd>hardware and browser tested. chrome, safari, and firefox were tested on a 2017 15" MBP.</dd>
<dt>project</dt>
[...]

PR #4669, opened 2019-03-19
measure blocks text widths with measureText

mzgoddard:
### Resolves

This works in the same area as https://github.com/LLK/scratch-gui/pull/4663. We could choose one or both of these PRs. I think we'll select this one.

### Proposed Changes

Create a set of 2D canvas rendering contexts with their `font` member set to the font used by the text elements for each class. Call measureText and use the returned width value in place of the computed text length from the text element.

### Reason for Changes

After writing some PRs yesterday I recalled another API we can use to measure the text width. [CanvasRendering2DContext.measureText](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText), appropriately named, measures some given text according to the font member value. This way we can measure the width of some text on demand without causing DOM style and layout work. The cache is still useful performance wise so we should keep that.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
  
Chromebook
 * [x] Chrome
 
iPad
* [x] Safari

### Benchmark Data

I'm going to save making this section at the moment. It takes a lot of time to build the data and tables. This should have slightly better to or the same performance as #4663 since it does not need to walk the data before blocks does.

### DIFF
--- a/src/lib/blocks.js
+++ b/src/lib/blocks.js
@@ -331,5 +331,188 @@ export default function (vm) {
         return collator.compare(str1, str2);
     };
 
+    // Default font returned from CanvasRendering2DContext.font.
+    const SPEC_DEFAULT_FONT = '10px sans-serif';
+
+    // Convert a CSSStyleDecoration into an object without hyphenated keys for
+    // iterating font strings.
+    const fontMembers = ({
+        font: font,
+        'font-style': style = 'normal',
+        'font-varient': variant = 'normal',
+        'font-weight': weight = 'normal',
+        'font-stretch': stretch = 'normal',
+        'font-size': size,
+        'line-height': lineHeight = 'normal',
+        'font-family': family
+    }) => ({
+        font,
+        style,
+        variant,
+        weight,
+        stretch,
+        size,
+        lineHeight,
+        family
+    });
+
+    // Set of functions that produce font strings for
+    // CanvasRendering2DContext.font.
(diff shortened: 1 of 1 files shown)


picklesrus:
My main worry about this PR is overriding ScratchBlocks.getCachedWidth from GUI. Could this PR be made in scratch-blocks instead?  Would we have enough information at Scratch Blocks' inject time to run the text measuring code to build the cache? Feel free to ping me when you're around if you want to chat.

mzgoddard:
- Fix conflicts

mzgoddard:
> Could this PR be made in scratch-blocks instead?

I think so. This code does not explicitly depend on any GUI details. The bit of info the code needs is the font's for the class names scratch-blocks uses. And that is retrieved with computed style that could (and is) used in scratch-blocks.

> Would we have enough information at Scratch Blocks' inject time to run the text measuring code to build the cache?

I'm not sure I understand the question. The text measuring behaviour in scratch-blocks needs the same information to measure text that this version needs. They both need the text to measure and the font decided by css and the given className. If we do not have that information at inject time the current text measurements made by scratch-blocks should be wrong.

> Feel free to ping me when you're around if you want to chat.

I'll ping you tomorrow.

picklesrus:
Closing in favor. https://github.com/LLK/scratch-blocks/pull/1937

Issue #2596, opened 2018-07-12
Refactor menu-bar component


chrisgarrity:
### Expected Behavior

_Please describe what should happen_
Menubar has grown from a simple component to one connected to redux with handlers. It should be refactored to use the container/component pattern we use for most elements.

I can probably fix it this month, but didn't want to hold up smoke-testing of localization.

PR #4676, opened 2019-03-21
Move projectChanged into HOC wrapping MenuBar, to avoid unnecessary renders

mzgoddard:
### Resolves

Less time rendering on project change and unchange.

### Proposed Changes

Move projectChanged out of MenuBar into a parent component and pass MenuBar a function to test what the projectChanged logic was doing.

### Reason for Changes

MenuBar uses projectChanged in some event handling behaviour. So when projectChanged changes, MenuBar renders leading to its tree being reconsidered when the nothing has changed.

Another possible implementation would be use a shouldComponentUpdate lifecycle handle. I figured the approach in this PR would be more robust (create less bugs).

### Benchmark Data

These values are the amount of time spent deserializing scratch blocks from an sb3 project (173918262). The change (%) doesn't represent this very well since its over deserializing a projects blocks, when the change specifically applies to the first block being deserialized.

|  device | develop (ms) | menu-bar (ms) | difference (ms) | change (%) |
| --- | --- | --- | --- | --- |
|  chrome | 268 | 254 | 14 | 5.15% |
|  firefox | 444 | 392 | 51 | 11.54% |
|  safari | 531 | 502 | 29 | 5.50% |
|  chromebook | 881 | 790 | 91 | 10.37% |

### DIFF
--- a/src/components/menu-bar/confirm-replace-hoc.jsx
+++ b/src/components/menu-bar/confirm-replace-hoc.jsx
@@ -1,0 +1,50 @@
+import {connect} from 'react-redux';
+import PropTypes from 'prop-types';
+import bindAll from 'lodash.bindall';
+import React from 'react';
+
+const ConfirmReplaceHOC = function (WrappedComponent) {
+    class ConfirmReplaceProject extends React.PureComponent {
+        constructor (props) {
+            super(props);
+
+            bindAll(this, [
+                'confirmReadyToReplaceProject'
+            ]);
+        }
+
+        confirmReadyToReplaceProject (message) {
+            let readyToReplaceProject = true;
+            if (this.props.projectChanged && !this.props.canCreateNew) {
+                readyToReplaceProject = confirm(message); // eslint-disable-line no-alert
+            }
+            return readyToReplaceProject;
+        }
+
+        render () {
+            const {
+                /* eslint-disable no-unused-vars */
+                projectChanged,
+                /* eslint-enable no-unused-vars */
+                ...props
+            } = this.props;
+            return (<WrappedComponent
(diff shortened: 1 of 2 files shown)


thisandagain:
@mzgoddard Can you take a look at the failing menu bar tests?

mzgoddard:
I think they were all linting issues. I'll see how the tests are tomorrow if its also failing units/integrations.

paulkaplan:
@benjiwheeler could you review this since you are working in a very similar area of code, and make sure that this approach could work with your change to the confirmation that you are working on? 

For context, this PR extracts the confirmation logic into an HOC (something I know you were considering) but for a different reason–adding the `projectChanged` prop to the menubar only to use it in an event callback means that it will trigger unneeded re-renders (any render it triggers is unneeded, since it is not used in the rendering). 

Additionally, it would also have the side benefit of making the logic for when to trigger a confirm more testable, since it extracts it from the menubar, like you were looking for.

benjiwheeler:
@paulkaplan Yes, I'll review. This is one of the approaches I've looked at for where to put `sb-file-uploader` -- the holdup has been my concern about having the file picker loaded for everyone, all the time, when it's only used on rare occasions.

benjiwheeler:
Tested this, it does indeed seem to avoid unnecessary renders. It's not a huge difference in number of renders, because the boundary between `projectChanged` being true and false isn't crossed often.

It seems like this is a subset of the container-esque functions and state that should be pulled out of the `menu-bar` component, eventually. See this issue: https://github.com/LLK/scratch-gui/issues/2596 . So maybe this is the first step towards splitting `menu-bar` into container and presentation components.

PR #4677, opened 2019-03-21
render only new toolbox xmls

mzgoddard:
### Resolves

Faster loading.

### Proposed Changes

Check if toolbox needs to be updated with new XML during updateToolbox.

### Reason for Changes

src/containers/Blocks.jsx:updateToolbox may be called when the xml hasn't changed. Calling this.ScratchBlocks.updateToolbox with the unchanged XML does a lot of work for no change.

### Benchmark Data

|  device | before (ms) | after (ms) | difference (ms) | change (%) |
| --- | --- | --- | --- | --- |
|  chrome | 117 | 1 | 116 | 99.15% |
|  firefox | 143 | 1 | 142 | 99.30% |
|  safari | 99 | 1 | 98 | 98.79% |
|  chromebook | 425 | 2 | 422 | 99.43% |

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -30,6 +30,10 @@ import {
     SOUNDS_TAB_INDEX
 } from '../reducers/editor-tab';
 
+const UNINITIALIZED_TOOLBOX_XML = `<xml style="display: none">
+    <category name="%{BKY_CATEGORY_MOTION}" id="motion" colour="#4C97FF" secondaryColour="#3373CC"></category>
+</xml>`;
+
 const addFunctionListener = (object, property, callback) => {
     const oldFn = object[property];
     object[property] = function () {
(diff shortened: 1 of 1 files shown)


paulkaplan:
The integration test that is failing is relevant here. I believe the toolbox not being updated is causing the variables not to appear.

mzgoddard:
The second failing test is caused by the first failing test. On its own it doesn't fail.

The first is failing I think because the toolbox rendering is not deterministically built from the string passed to updateToolbox. The data category determines its variable list from the variables in the main workspace's variable map instead of the variables in the xml passed to it. The failing tests is failing at the last assertion. The first check for the variable passes as its populated from a updateToolbox call after the default project loads and that includes the default workspace. The second check fails because the toolbox rendering at inject time does not correctly render the toolbox. It is missing the variables that have not been loaded into the workspace. A updateToolbox fires later after the workspace is updated but the correct toolbox xml has already been rendered. I think there is a fix here we can do but I'll have to figure it out tomorrow.

mzgoddard:
I added some work that tracks when the workspace has first rendered. Until that first render we can't authoritatively or deterministically render the toolbox. If we render only toolbox changes after the workspace has rendered it'll be all up to date.

-----

Given the increased work for the toolbox render after the screen transition I wonder if we might switch the timeout in blocks to promises or another same-microtask-queue behaviour. That'll move the render to before the new screen is displayed and there won't be the larger flash caused by this change. (There is currently a flash. Just I supposed less noticeable.)

I could add something like ...

```diff
diff --git a/src/containers/blocks.jsx b/src/containers/blocks.jsx
index c366da45..f9cc3ba1 100644
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -180,13 +180,19 @@ class Blocks extends React.Component {
     componentWillUnmount () {
         this.detachVM();
         this.workspace.dispose();
-        clearTimeout(this.toolboxUpdateTimeout);
+        this.toolboxUpdateTimeout && this.toolboxUpdateTimeout.cancel();
     }
     requestToolboxUpdate () {
-        clearTimeout(this.toolboxUpdateTimeout);
-        this.toolboxUpdateTimeout = setTimeout(() => {
-            this.updateToolbox();
-        }, 0);
+        this.toolboxUpdateTimeout && this.toolboxUpdateTimeout.cancel();
+        let running = true;
+        this.toolboxUpdateTimeout = Promise.resolve().then(() => {
+            if (running) {
+                this.updateToolbox();
+            }
+        });
+        this.toolboxUpdateTimeout.cancel = () => {
+            running = false;
+        };
     }
     setLocale () {
         this.ScratchBlocks.ScratchMsgs.setLocale(this.props.locale);
```

paulkaplan:
@mzgoddard I think that the large flash from empty -> full of blocks doesn't seem like an acceptable change in user experience. The alternative you suggested seems worth a try.

Also I don't know what the benchmark numbers are referring to. Could you give a specific start/end that you are measuring from so I can replicate?

mzgoddard:
The benchmark numbers refer to the first toolbox render when blocks is injected into the dom in the componentDidMount handle. I measured with a `start = Date.now(); /* inject blocks with uninitialized toolbox xml */ console.log('time', Date.now() - start);` kind of thing.

mzgoddard:
I thought I check for comments on Monday and should have seen your response @paulkaplan. Sorry for the delay. I'll have the mentioned patch in, in a bit.

mzgoddard:
I switch the label to review since I added the promise logic and waited for the tests to pass.

Issue #4604, opened 2019-03-01
'See project page' button should not save unless project has changed


chrisgarrity:
### Expected Behavior

_Please describe what should happen_
Clicking the 'See project page' button should not save the project unless there have been changes since it was last saved.

### Actual Behavior

_Describe what actually happens_
The 'See project page' button always forces a save.

PR #4710, opened 2019-04-01
Share, See-project-page buttons will only save if project has changed

benjiwheeler:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/4604

### Proposed Changes

Instead of auto-saving whenever a logged in user clicks the `Share` or `See project page` buttons, before sharing or navigating to the project page, only auto-save if the project has changed.

### Reason for Changes

It's slow to auto-save unnecessarily. This is part of the larger effort to make saving happen at the right times, and only at the right times.

### Test Coverage

No tests written yet; should write some

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -194,7 +194,7 @@ class MenuBar extends React.Component {
         this.props.onRequestCloseFile();
     }
     handleClickSeeCommunity (waitForUpdate) {
-        if (this.props.canSave) { // save before transitioning to project page
+        if (this.props.canSave && this.props.projectChanged) { // save before transitioning to project page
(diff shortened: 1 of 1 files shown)


towerofnix:
Thank you! This is definitely an appreciated change. It's a little frustrating to always have to make sure the screen is exactly as you want the project thumbnail to be whenever you see-outside the project. (Especially if you'd just seen-inside and hadn't made any changes.)

PR #4712, opened 2019-04-02
Register callback keys for buttons used in future extensions.

kchadha:
### Proposed Changes

Adds support for buttons used in future extensions to be able to call into the scratch blocks functions for creating variables, lists and custom procedures.

Related to LLK/scratch-vm#2083

### Test Coverage

- [x] Tested locally in with changes in LLK/scratch-vm#2083 with the core extension.
- [x] Tested locally embedded in www

### Browser Coverage

- [x] Mac OS Chrome
- [x] Mac OS Safari


kchadha:
Removed the newlines and tested in www!

PR #4720, opened 2019-04-08
Add Vernier Force and Accel extension

ericrosenbaum:
Add the Vernier Force and Acceleration extension to the library:

Screen Shot 2019-04-08 at 2 17 35 PM


src/lib/libraries/extensions/gdxfor/gdxfor-illustration.svg
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
thisandagain: Can you compress these SVGs (SVGO)?

ericrosenbaum: all set

PR #4724, opened 2019-04-09
Changing tutorial video width to be smaller

### LINKED ISSUE #4668: Update Video Size in Tutorial Window
## Current
Currently any step that contains a video is much larger than all other tutorial steps. Most tutorials start with a video so the tutorials start by taking up a lot of editor room.

**Tutorial Video** 
current tutorial video step

**Tutorial Step** 
current tutorial step

## Update
In efforts to reduce the frequency of people closing the tutorial window immediately or during the video, we'd like to reduce the dimensions of all video steps to be the same size all other steps. This will also help with the jarring transition between a video step and all other steps.

**Maintain the current tutorial step width** 
The height of tutorial steps vary based on the content. To maintain the video aspect ratio, we should not force a height, but we should force the width to be the same size as all other tutorial steps.

evhan55:
### Resolves

- Resolves #4668

### Proposed Changes

Change tutorial video in iframe to be 466x265 to more closely match other tutorial steps.

### DIFF
--- a/src/components/cards/card.css
+++ b/src/components/cards/card.css
@@ -137,7 +137,7 @@
 }
 
 .step-video {
-    height: 337px;
+    height: 256px;
 }
 
 .step-image {
--- a/src/components/cards/cards.jsx
+++ b/src/components/cards/cards.jsx
@@ -69,11 +69,11 @@ const VideoStep = ({video, dragging}) => (
             allowFullScreen
             allowTransparency="true"
             frameBorder="0"
-            height="338"
+            height="265"
             scrolling="no"
             src={`https://fast.wistia.net/embed/iframe/${video}?seo=false&videoFoam=true`}
             title="📹"
-            width="600"
+            width="466"
         />
         <script
             async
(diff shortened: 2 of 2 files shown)


evhan55:
@carljbowman @ericrosenbaum 
Here is a new video width to match other tutorial steps.  It turns out the aspect ratio of the video naturally matched the other steps, so there is no change in size at all.  I tried to tweak it to get it pretty perfect.  Here are some snapshots for now, let me know what you think:

Screen Shot 2019-04-09 at 1 57 51 PM
Screen Shot 2019-04-09 at 1 57 55 PM
Screen Shot 2019-04-09 at 1 57 58 PM

evhan55:
@carljbowman Here is a non-video-step image set to height 500px with this proposed code:

Screen Shot 2019-04-09 at 3 21 09 PM

carljbowman:
@evhan55 - Wonderful. Thanks for testing that. Great to confirm that the fixed height is scoped only to videos and that the other steps can grow vertically.

PR #4729, opened 2019-04-10
Add stand-alone player to playground

spectranaut:
### Proposed Changes

This PR adds a experimental "standalone player" to the GUI's playground. The `standalone-player.html` is example of how to create minimalist scratch project player without the editor view. I hoping to get feedback and continue to work to make the example more robust -- this is just a first draft! I'll also be looking to contribute back to the various repos to make including a player for scratch projects easier to embed in websites.

### Reason for Changes

There is no standard or recommended way to use the scratch player alone.

### Test Coverage

No tests, this is just a draft.

### DIFF
--- a/src/lib/storage.js
+++ b/src/lib/storage.js
@@ -2,6 +2,9 @@ import ScratchStorage from 'scratch-storage';
 
 import defaultProject from './default-project';
 
+const DEFAULT_ASSET_SERVER = 'https://assets.scratch.mit.edu';
+const DEFAULT_PROJECT_SERVER = 'https://projects.scratch.mit.edu';
+
 /**
  * Wrapper for ScratchStorage which adds default web sources.
  * @todo make this more configurable
@@ -10,6 +13,8 @@ class Storage extends ScratchStorage {
     constructor () {
         super();
         this.cacheDefaultProject();
+        this.projectHost = DEFAULT_PROJECT_SERVER;
+        this.assetHost = DEFAULT_ASSET_SERVER;
     }
     addOfficialScratchWebStores () {
         this.addWebStore(
--- a/src/playground/standalone-player.jsx
+++ b/src/playground/standalone-player.jsx
@@ -55,7 +55,7 @@ window.onload = function () {
     let attemptFullscreen = Boolean(document.body.requestFullscreen);
     document.getElementById('green-flag').addEventListener('click', () => {
         document.getElementById('overlay').classList.add('hidden');
-        vm.greenFlag()
+        vm.greenFlag();
         if (attemptFullscreen) {
             document.body.requestFullscreen();
             attemptFullscreen = false;
--- a/src/playground/standalone-player.jsx
+++ b/src/playground/standalone-player.jsx
@@ -41,6 +41,24 @@ window.onload = function () {
     const audioEngine = new AudioEngine();
     vm.attachAudioEngine(audioEngine);
 
+    // Resets size of canvas directly for proper image calcuations
+    // when the window is resized
+    const resize = () => {
+      renderer.resize(canvas.clientWidth, canvas.clientHeight);
+    };
+    window.addEventListener('resize', resize);
+
+    resize();
+
+    // Browser fullscreen mode
+    let attemptFullscreen = Boolean(document.body.requestFullscreen);
+    document.body.addEventListener('click', e => {
+        if (attemptFullscreen) {
+            document.body.requestFullscreen();
+            attemptFullscreen = false;
+        }
+    });
+
     // Feed mouse events as VM I/O events.
     document.body.addEventListener('mousemove', e => {
         const rect = canvas.getBoundingClientRect();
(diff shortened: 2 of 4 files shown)


src/lib/storage.js
```diff
@@ -10,6 +13,8 @@ class Storage extends ScratchStorage {
     constructor () {
         super();
         this.cacheDefaultProject();
+        this.projectHost = DEFAULT_PROJECT_SERVER;
+        this.assetHost = DEFAULT_ASSET_SERVER;
```
spectranaut: Is there a reason not to do this? In scratch-gui code, `setAssetHost` and `setProjectHost` will be called and overwrite this value (with the same value, in the scratch-gui case).

The standalone player could explicitly set these values as well, but having a default lowers the start up cost/boiler plate code for the standalone player scenario.

towerofnix:
:tada:! This will definitely be a handy reference for the many developers who frequently ask how to embed Scratch in their program or website.

src/playground/standalone-player.jsx
```diff
@@ -0,0 +1,105 @@
...
+    vm.attachStorage(storage);
+
+    // Compatibility mode will set the frame rate to 30 TPS,
+    // which is the standard for the scratch player.
+    vm.setCompatibilityMode(true);
+
+    vm.downloadProjectId(EXAMPLE_PROJECT);
+
+    vm.on('workspaceUpdate', () => {
+        setTimeout(() => vm.greenFlag(), 1000);
```
mzgoddard: I don't think this should autoplay. Some user input should be required.

src/playground/standalone-player.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+    // Initialize storage
+    storage.addOfficialScratchWebStores();
+    vm.attachStorage(storage);
+
+    // Compatibility mode will set the frame rate to 30 TPS,
+    // which is the standard for the scratch player.
+    vm.setCompatibilityMode(true);
+
+    vm.downloadProjectId(EXAMPLE_PROJECT);
```
mzgoddard: Why is this hard coded?

src/playground/standalone-player.jsx
```diff
@@ -0,0 +1,105 @@
...
+        // E.g., prevent scroll.
+        if (e.target !== document && e.target !== document.body) {
+            e.preventDefault();
+        }
+    });
+
+    // Run threads
+    vm.start();
+
+    vm.greenFlag();
```
mzgoddard: Do we need to green flag here?

src/playground/standalone-player.jsx
```diff
@@ -0,0 +1,105 @@
...
+    vm.attachV2BitmapAdapter(new BitmapAdapter());
+    vm.attachV2SVGAdapter(new SVGRenderer());
+
+    // Initialize storage
+    storage.addOfficialScratchWebStores();
+    vm.attachStorage(storage);
+
+    // Compatibility mode will set the frame rate to 30 TPS,
+    // which is the standard for the scratch player.
+    vm.setCompatibilityMode(true);
```
spectranaut: Why isn't this the default? Took me and @z a minute to figure out why everything was sped up (before adding this).

src/playground/standalone-player.jsx
```diff
@@ -0,0 +1,105 @@
+import VirtualMachine from 'scratch-vm';
+import RenderWebGL from 'scratch-render';
+import AudioEngine from 'scratch-audio';
+import {SVGRenderer, BitmapAdapter} from 'scratch-svg-renderer';
+import storage from '../lib/storage';
+
+// This file is an example of how to create a standalone, full screen
+// minimal scratch player without the editor view.
+
+const EXAMPLE_PROJECT = 10128067; // 'Make it Dance' Project
```
mzgoddard: I think we want the example project to follow the default the other scratch-gui pages use. The Scratch Cat default project id 0.

src/lib/storage.js
mzgoddard: This makes it easier. I think though its one of those slippery slopes kind of deals. We probably don't want a default configuration for scratch that is distributed over Scratch's implementation?

src/playground/standalone-player.jsx
spectranaut: nope

src/playground/standalone-player.jsx
spectranaut: I added the green flag overlay -- click to start :)

benjiwheeler:
Thanks for this, @spectranaut ! We're also starting work on a related feature for the website's project pages, that will give you the code you need to embed a given project. 

See: https://github.com/LLK/scratch-www/issues/2905

src/playground/standalone-player.jsx
spectranaut: Made an issue to discuss this: https://github.com/LLK/scratch-vm/issues/2137

spectranaut:
> Code LG. I think it would probably be good to isolate the storage config even if it means duplicating it, just to keep this a truly standalone demo.

Done!

spectranaut:
closing this in favor of landing in the VM playground! Thanks for everyones help and support here :)

new PR: https://github.com/LLK/scratch-vm/pull/2156

alexwenbj:
@spectranaut Hi,This is useful for me.Just one problem, it doesn't support the "`ask ... and wait`"  and "`monitor the variable`".Any idea for these?

Issue #4649, opened 2019-03-13
Some projects only load in editor mode but give Asset ID of Null error in player mode


BryceLTaylor:
### Expected Behavior

All projects should load on the project page and in the editor.

### Actual Behavior

Some projects don't load, showing the blue "Oops something went wrong" message.  If you check the console, they have an error that says "TypeError: Cannot read property 'assetId' of null"

![Screen Shot 2019-03-13 at 4 40 48 PM](https://user-images.githubusercontent.com/1088291/54313252-4920f200-45af-11e9-9573-9c79152e938c.png)

Here are some example:
https://scratch.mit.edu/projects/245878906/
https://scratch.mit.edu/projects/292083579/
https://scratch.mit.edu/projects/281999647
https://scratch.mit.edu/projects/291931151/

For 245878906 I downloaded the json from s3 and tried uploading it, but it wouldn't load saying "The project file that was selected failed to load"

![Screen Shot 2019-03-13 at 4 49 57 PM](https://user-images.githubusercontent.com/1088291/54313617-1aefe200-45b0-11e9-9b6e-7359f5458e27.png)

The json named as an .sb3 and then zipped:
[245878906.sb3.zip](https://github.com/LLK/scratch-gui/files/2963622/245878906.sb3.zip)

### Steps to Reproduce

Open one of the projects above
Note: The project doesn't load

### Operating System and Browser

Mac Chrome

BryceLTaylor:
More projects:
https://scratch.mit.edu/projects/292874938/

BryceLTaylor:
It seams that for at least https://scratch.mit.edu/projects/292874938/editor/ if you go straight to the editor (not to the project page and then see inside) you can see the project.  When you open the javascript console you see two errors:

`Failed to load the resource: the server responded with a status of 404 ()`
![Screen Shot 2019-03-26 at 11 51 26 AM](https://user-images.githubusercontent.com/1088291/55012250-8167fd80-4fbd-11e9-9ff3-dd2663342fce.png)

They both have to do with the sound file e4d85d5100e1eff4dcd1d152d217f0d2.wav which in the project is called "Song 1 20130127" on the stage.  I downloaded the project, unzipped it, and found that the file had downloaded.  

The file is 5.6MB and 2:06 long

BryceLTaylor:
There is an interesting report from the community that the projects can be loaded from "my stuff": https://scratch.mit.edu/discuss/topic/345883/

[...]

PR #4733, opened 2019-04-11
Make sure to save assets before sending them to the backpack

paulkaplan:
This prevents assets from existing in the backpack server without also existing in the asset server. Credit to @kchadha for coming up with this fix.

### Resolves

A proposed change to prevent https://github.com/LLK/scratch-gui/issues/4649 from continuing in the future. 

### Proposed Changes

Makes sure to save any unsaved assets to the asset server prior to saving them to the backpack.

### Reason for Changes

This makes it impossible to have an asset saved to the backpack that is not also saved to the asset server (the canonical asset storage location)

This only needs to be done for costumes and sounds, since the other types of backpack assets do not exist in the asset server.

---
/cc @kchadha this is one method for enforcing the implicit assumption about where assets live. Seems like a pretty good solution!

### DIFF
--- a/src/containers/backpack.jsx
+++ b/src/containers/backpack.jsx
@@ -83,12 +83,15 @@ class Backpack extends React.Component {
     }
     handleDrop (dragInfo) {
         let payloader = null;
+        let presaveAsset = null;
         switch (dragInfo.dragType) {
         case DragConstants.COSTUME:
             payloader = costumePayload;
+            presaveAsset = dragInfo.asset;
             break;
         case DragConstants.SOUND:
             payloader = soundPayload;
+            presaveAsset = dragInfo.asset;
             break;
         case DragConstants.SPRITE:
             payloader = spritePayload;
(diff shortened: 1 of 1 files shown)


src/containers/backpack.jsx
```diff
@@ -83,12 +83,15 @@ class Backpack extends React.Component {
     }
     handleDrop (dragInfo) {
         let payloader = null;
+        let presaveAsset = null;
```
kchadha:
```suggestion
        // Adding a test comment
        let presaveAsset = null;
```

src/containers/backpack.jsx
```diff
@@ -83,12 +83,15 @@ class Backpack extends React.Component {
...
         switch (dragInfo.dragType) {
         case DragConstants.COSTUME:
             payloader = costumePayload;
+            presaveAsset = dragInfo.asset;
```
kchadha:
```suggestion
            presaveAsset = dragInfo.payload.asset;
```

src/containers/backpack.jsx
```diff
@@ -83,12 +83,15 @@ class Backpack extends React.Component {
...
             break;
         case DragConstants.SOUND:
             payloader = soundPayload;
+            presaveAsset = dragInfo.asset;
```
kchadha:
```suggestion
            presaveAsset = dragInfo.payload.asset;
```

src/containers/backpack.jsx
```diff
@@ -102,6 +105,14 @@ class Backpack extends React.Component {
         // Creating the payload is async, so set loading before starting
         this.setState({loading: true}, () => {
             payloader(dragInfo.payload, this.props.vm)
+                .then(payload => {
+                    // Force the asset to save to the asset server before storing in backpack
+                    // Ensures any asset present in the backpack is also on the asset server
+                    if (presaveAsset && !presaveAsset.clean) {
+                        return storage.store(presaveAsset).then(() => payload);
```
kchadha: Found another bug... `storage.store` needs more arguments than just the asset or the request fails and throws an error.
```suggestion
                        return storage.store(
                            presaveAsset.assetType,
                            presaveAsset.dataFormat,
                            presaveAsset.data,
                            presaveAsset.assetId
                        ).then(() => payload);
```

PR #4757, opened 2019-04-18
Remove unneeded feedback button in menubar

paulkaplan:
Remove unneeded feedback button in menubar. It is no longer needed, as the playground is no longer the public face of scratch gui


paulkaplan:
@chrisgarrity good call. done

PR #4761, opened 2019-04-19
Add Boost extension assets

ericrosenbaum:
Update the Boost extension tile to include all the assets used by the connection modal, and the rest of the necessary metadata. Also, re-order extension tiles.

(pretty much copied over from @knandersen's version so I could make a PR).

To test, you can just changed `disabled` to false.


evhan55:
This looks great!

- tile is in new place in the library
- all icons during connecting looked great

Issue #4783, opened 2019-04-30
See Project Page button never auto-saves


benjiwheeler:
### Expected Behavior

1. edit a program you own while logged in
2. make a change
3. click project page
4. it auto-saves before transitioning

### Actual Behavior

... 4. it does NOT auto-save before transitioning :(

### Why??

...because two recent PRs made changes to the same area of code, and weren't tested thoroughly enough in staging after the second one went in.

One change added a place in `menu-bar.jsx` where the `projectChanged` prop was used; but the other change removed the `projectChanged` prop from `menu-bar.jsx` entirely! So `projectChanged` was undefined.

PR #4784, opened 2019-04-30
Fix auto-save when clicking See Project Page

benjiwheeler:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/4783

### Proposed Changes

* moves remaining `projectChanged` checking from `menu-bar` to new `confirm-replace-hoc` function, called `shouldSaveBeforeTransition`
* renames `confirm-replace-hoc` to `menu-bar-hoc`, since its purpose has expanded

### Test Coverage

Added tests to cover both this change, and the previous change that introduced the `confirm-replace-hoc`, https://github.com/LLK/scratch-gui/pull/4676

### DIFF
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -192,8 +192,8 @@ class MenuBar extends React.Component {
         this.props.onRequestCloseFile();
     }
     handleClickSeeCommunity (waitForUpdate) {
-        if (this.props.shouldSaveBeforeTransition()) { // save before transitioning to project page
-            this.props.autoUpdateProject();
+        if (this.props.shouldSaveBeforeTransition()) {
+            this.props.autoUpdateProject(); // save before transitioning to project page
             waitForUpdate(true); // queue the transition to project page
         } else {
             waitForUpdate(false); // immediately transition to project page
--- a/src/components/menu-bar/menu-bar.jsx
+++ b/src/components/menu-bar/menu-bar.jsx
@@ -28,7 +28,7 @@ import LoginDropdown from './login-dropdown.jsx';
 import SB3Downloader from '../../containers/sb3-downloader.jsx';
 import DeletionRestorer from '../../containers/deletion-restorer.jsx';
 import TurboMode from '../../containers/turbo-mode.jsx';
-import ConfirmReplaceHOC from './confirm-replace-hoc.jsx';
+import MenuBarHOC from '../../containers/menu-bar-hoc.jsx';
 
 import {openTipsLibrary} from '../../reducers/modals';
 import {setPlayer} from '../../reducers/mode';
@@ -192,7 +192,7 @@ class MenuBar extends React.Component {
         this.props.onRequestCloseFile();
     }
     handleClickSeeCommunity (waitForUpdate) {
-        if (this.props.canSave && this.props.projectChanged) { // save before transitioning to project page
+        if (this.props.shouldSaveBeforeTransition()) { // save before transitioning to project page
             this.props.autoUpdateProject();
             waitForUpdate(true); // queue the transition to project page
         } else {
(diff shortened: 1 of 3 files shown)


src/components/menu-bar/menu-bar.jsx
```diff
@@ -28,7 +28,7 @@ import LoginDropdown from './login-dropdown.jsx';
 import SB3Downloader from '../../containers/sb3-downloader.jsx';
 import DeletionRestorer from '../../containers/deletion-restorer.jsx';
 import TurboMode from '../../containers/turbo-mode.jsx';
-import ConfirmReplaceHOC from './confirm-replace-hoc.jsx';
+import MenuBarHOC from '../../containers/menu-bar-hoc.jsx';
```
benjiwheeler: I know that putting `MenuBarHOC` in the `containers/` directory doesn't match where other HOCs are -- but I think this should probably just be a standard container, rather than a general HOC. I didn't want to fully transform it into a container now, because I wanted to keep the scope of the changes narrow.

src/containers/menu-bar-hoc.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+            bindAll(this, [
+                'confirmReadyToReplaceProject',
+                'shouldSaveBeforeTransition'
+            ]);
+        }
+        confirmReadyToReplaceProject (message) {
+            let readyToReplaceProject = true;
+            if (this.props.projectChanged && !this.props.canCreateNew) {
+                readyToReplaceProject = this.props.confirmWithMessage(message);
```
benjiwheeler: changed this line to use a props function, to make unit testing work without actually causing a js confirm to appear

src/containers/menu-bar-hoc.jsx
```diff
@@ -0,0 +1,66 @@
...
+            }
+            return readyToReplaceProject;
+        }
+        shouldSaveBeforeTransition () {
```
benjiwheeler: This is the main change I made to this file: adding the `shouldSaveBeforeTransition` function and passing it to the wrapped component as a prop

src/containers/menu-bar-hoc.jsx
```diff
@@ -0,0 +1,66 @@
...
+        render () {
+            const {
+                /* eslint-disable no-unused-vars */
+                projectChanged,
+                /* eslint-enable no-unused-vars */
+                ...props
+            } = this.props;
+            return (<WrappedComponent
+                confirmReadyToReplaceProject={this.confirmReadyToReplaceProject}
+                shouldSaveBeforeTransition={this.shouldSaveBeforeTransition}
```
benjiwheeler: Now passing `shouldSaveBeforeTransition`

src/containers/menu-bar-hoc.jsx
```diff
@@ -0,0 +1,66 @@
...
+                {...props}
+            />);
+        }
+    }
+
+    MenuBarContainer.propTypes = {
+        canCreateNew: PropTypes.bool,
+        canSave: PropTypes.bool,
```
benjiwheeler: added this prop -- it was already being passed through to `menu-bar`, but now we use it here too

src/containers/menu-bar-hoc.jsx
```diff
@@ -0,0 +1,66 @@
...
+    MenuBarContainer.defaultProps = {
+        // default to using standard js confirm
+        confirmWithMessage: message => (confirm(message)) // eslint-disable-line no-alert
+    };
+    const mapStateToProps = state => ({
+        projectChanged: state.scratchGui.projectChanged
+    });
+    const mapDispatchToProps = () => ({});
+    // Allow incoming props to override redux-provided props. Used to mock in tests.
+    const mergeProps = (stateProps, dispatchProps, ownProps) => Object.assign(
```
benjiwheeler: added `mergeProps` to facilitate prop overriding in testing

benjiwheeler:
> Also this seems like it got through a lot of layers of our testing without being noticed, so I think we should talk through (offline) what happened, figure out what could have caught this, etc.

Agreed. I think the two changes were being reviewed separately, each passed tests, time went by, and I merged each without realizing that one should really be rebased on top of the other before merging because of the overlap in what they changed.

Travis caught this after merging, and it looks like merges to scratch-gui have been failing for the past 13 days :[

PR #4786, opened 2019-05-01
Update eslint-plugin-react and fix newly discovered lint errors.

paulkaplan:
We had pinned this dependency because it once started causing build failures (https://github.com/LLK/scratch-gui/pull/4215) but then by not automatically getting updates, and not merging the new greenkeeper updates (https://github.com/LLK/scratch-gui/pull/4221) we didn't see errors that should have been caught by the linter. The old version of the linter didn't catch undefined prop issues completely, which is why [this issue ](https://github.com/LLK/scratch-gui/pull/4784) wasn't caught by the linter. :sigh:

Talking with @benjiwheeler we decided to unpin it to get new updates, leaning towards possibly breaking the build instead of possibly missing important updates that can catch bugs.

---
The lint fixes were mostly just missing proptypes, not real bugs. But this includes a small bug where the target pane wouldn't automatically update if changes to the sprite occurred while in fullscreen/player mode (in vm-listener-hoc)

Closes #4221


src/lib/vm-listener-hoc.jsx
```diff
@@ -60,9 +60,9 @@ const vmListenerHOC = function (WrappedComponent) {
                 this.props.vm.postIOData('userData', {username: this.props.username});
             }
 
-            // Re-request a targets update when the shouldEmitTargetsUpdate state changes to true
+            // Re-request a targets update when the shouldEmitUpdate state changes to true
```
benjiwheeler: This one seems like it uncovered an error; I think it was introduced here: https://github.com/LLK/scratch-gui/commit/c799541d8ec5a80765392638f235ec52f811adc1

Does this need to be tested before we make this change?

benjiwheeler:
Blocked on tests failing (rerunning now), and the one possible `vm-listener-hoc` bug this uncovered.

src/lib/vm-listener-hoc.jsx
paulkaplan: Yup, mentioned that in the PR
> The lint fixes were mostly just missing proptypes, not real bugs. But this includes a small bug where the target pane wouldn't automatically update if changes to the sprite occurred while in fullscreen/player mode (in vm-listener-hoc)
I did check manually that that issue was fixed.

benjiwheeler:
Third time restarting identical travis build was the charm! :(

PR #4790, opened 2019-05-01
Supporting GUI changes for extensionification

cwillisf:
### Dependency

This PR and LLK/scratch-vm#2161 should be merged at about the same time, as they make a small-but-important change to the way extension block information is communicated between the VM and the GUI.

In particular, integration tests are expected to fail until LLK/scratch-vm#2161 is merged.

### Proposed Changes

This PR includes several supporting changes for extensionification.

Primarily this PR is about support for interpreting extension block info through a Blockly mutation (late / "dynamic") instead of in the VM (immediate / "static"). This makes it possible to have two different blocks with the same opcode but different attributes in some other way, such as two variable blocks or custom procedure call blocks. This also opens the door to removing some of the `scratch-blocks` / Blockly knowledge from `scratch-vm` if we decide to remove the "static" path.

This dynamic path needs more work before it can support all the features of the current static path.

Other supporting changes include:
* Loading a "core" extension will now place the extension blocks in the correct location in the toolbox instead of at the bottom. Non-"core" extensions will still load at the bottom.

### Reason for Changes

It might make sense to review and/or merge this as one PR, or it might make sense to break it up into more than one. Either way, this PR can serve as a way to keep tabs on the current state of the `e16n` branch.

### Test Coverage

Added `test/unit/util/define-dynamic-block.test.js` to cover the new "dynamic" block info code path.

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


rschamp:
Try it out: [e16n](https://llk.github.io/scratch-gui/e16n)

src/containers/blocks.jsx
```diff
@@ -414,20 +415,50 @@ class Blocks extends React.Component {
...
+                    // Anything else will be picked up from the XML attached to the block instance.
+                    const extendedOpcode = `${categoryInfo.id}_${blockInfo.info.opcode}`;
+                    const blockDefinition =
+                        defineDynamicBlock(this.ScratchBlocks, categoryInfo, blockInfo, extendedOpcode);
+                    this.ScratchBlocks.Blocks[extendedOpcode] = blockDefinition;
+                });
+            }
+        };
+
+        // scratch-blocks implements a menu or custom field as a special kind of block
```
kchadha: I would prefer that the comment just be specific here, otherwise it can be confusing e.g.
```suggestion
        // scratch-blocks implements a menu or custom field as a special kind of block ("shadow" block)
```

src/lib/define-dynamic-block.js
```diff
@@ -0,0 +1,90 @@
...
+            colourTertiary: categoryInfo.color3
+        };
+        // There is a scratch-blocks / Blockly extension called "scratch_extension" which adjusts the styling of
+        // blocks to allow for an icon, a feature of Scratch extension blocks. However, Scratch "core" extension
+        // blocks don't have icons and so they should not use 'scratch_extension'. Adding a scratch-blocks / Blockly
+        // extension after `jsonInit` isn't fully supported (?), so we decide now whether there will be an icon.
+        if (staticBlockInfo.blockIconURI || categoryInfo.blockIconURI) {
+            blockJson.extensions = ['scratch_extension'];
+        }
+        this.jsonInit(blockJson);
```
kchadha: Maybe add a comment here that we're basically initing an empty block, which will later have its layout constructed using the `domToMutation` function.

src/lib/define-dynamic-block.js
```diff
@@ -0,0 +1,90 @@
+// TODO: access `BlockType` and `ArgumentType` without reaching into VM
+// Should we move these into a new extension support module or something?
+import ArgumentType from 'scratch-vm/src/extension-support/argument-type';
+import BlockType from 'scratch-vm/src/extension-support/block-type';
+
+// TODO: grow this until it can fully replace `_convertForScratchBlocks` in the VM runtime
```
kchadha:
```suggestion
/**
 * Define a block using extension info which has the ability to dynamically determine (and update) its layout.
 * This functionality is used for extension blocks which can change its shape based on different state information.
 * (E.g. The control_stop block changes its shape based on which menu item is selected.)
 */
// TODO: grow this until it can fully replace `_convertForScratchBlocks` in the VM runtime
```

src/lib/make-toolbox-xml.js
```diff
@@ -716,35 +716,57 @@ const xmlClose = '</xml>';
 /**
  * @param {!boolean} isStage - Whether the toolbox is for a stage-type target.
  * @param {?string} targetId - The current editing target
- * @param {?string} categoriesXML - null for default toolbox, or an XML string with <category> elements.
+ * @param {?Array.<object>} categoriesXML - optional array of `{id,xml}` for categories.
```
kchadha: Hmmm maybe say more about why one would want to provide this?

kchadha:
```suggestion
 * @param {?Array.<object>} categoriesXML - optional array of `{id,xml}` for additional ("extension") categories in the order that they should appear in the toolbox. If a category in this list shares an ID with a "core" block category, the XML from this list will be used in place of the default xml defined in this file.
```

cwillisf: heh - I wrote another version of this comment at more or less the same time. Let me know if you think mine covers it well enough or if you think I should use yours or merge the two. Thanks :)

test/unit/util/define-dynamic-block.test.js
```diff
@@ -0,0 +1,202 @@
...
+const categoryInfo = {
+    name: 'test category',
+    color1: '#111',
+    color2: '#222',
+    color3: '#333'
+};
+
+const penIconURI = 'data:image/svg+xml;base64,fake_pen_icon_svg_base64_data';
+
+const testBlockInfo = {
```
kchadha: hmmm... probably we should have a test block with arguments...?

kchadha: Ah I see what we're doing below. I guess this is a todo along w/ the todo for `interpolate_`

test/unit/util/define-dynamic-block.test.js
```diff
@@ -0,0 +1,202 @@
...
+        block.domToMutation(xmlElement);
+
+        return block;
+    }
+
+    jsonInit (json) {
+        this.result = Object.assign({}, json);
+    }
+    interpolate_ () {
+        // TODO: add tests for this?
```
kchadha: We should probably create an issue for this and put it in the e16n github project. =D

test/unit/util/define-dynamic-block.test.js
```diff
@@ -0,0 +1,202 @@
...
+    });
+    test('can define a command block without an icon', () => {
+        const extendedOpcode = 'test.commandWithoutIcon';
+        const block = MockBlock.create(testBlockInfo.commandWithoutIcon, extendedOpcode);
+        expect(block.result).toEqual({
+            category: categoryInfo.name,
+            colour: categoryInfo.color1,
+            colourSecondary: categoryInfo.color2,
+            colourTertiary: categoryInfo.color3,
+            // extensions: undefined, // no icon means no extension
```
kchadha: Why is this commented out?

kchadha: Or maybe we can add a separate test for the fact that `block.result.extensions === undefined`?

kchadha: Same questions below for undefined properties.

cwillisf: Our linting rules forbid the use of `undefined`. Since these tests use `expect(...).toEqual({...})` they would fail if the property is defined. The comments listing these properties as undefined are just meant to emphasize that they're expected to be undefined instead of being don't-care values.

test/unit/util/define-dynamic-block.test.js
cwillisf: LLK/scratch-gui#4929

test/unit/util/define-dynamic-block.test.js
cwillisf: LLK/scratch-gui#4929

test/unit/util/define-dynamic-block.test.js
kchadha: Ah, makes sense!

kchadha:
Marking as blocked until we merge the corresponding VM PR and update the package json version here.

PR #4794, opened 2019-05-02
Add support for custom context menu items for dynamic extension blocks.

kchadha:
### Resolves

Towards LLK/scratch-vm#2120.

Depends on LLK/scratch-vm#2163.

### Proposed Changes

Add support for creating custom context menu items for dynamic extension blocks. This includes support for built in functionality like opening a custom procedure edit modal as well as a callback completely controlled by the extension.

Currently the custom context menu is registered as a ScratchBlocks extension when initializing the block. This throws an error which we are currently catching and ignoring, but in the future, we need to handle registering the context menu callbacks only once.

### Reason for Changes

Extensionification project!

### Test Coverage

Together with LLK/scratch-vm#2163, adds a custom context menu item to the core example extension.

### DIFF
--- a/src/lib/define-dynamic-block.js
+++ b/src/lib/define-dynamic-block.js
@@ -24,7 +24,7 @@ const setupCustomContextMenu = (ScratchBlocks, contextMenuInfo, extendedOpcode) 
                                 break;
                             }
                         } else if (contextOption.callback) {
-                            contextOption.callback();
+                            contextOption.callback({blockInfo: JSON.parse(this.blockInfoText)});
                         }
                     }
                 };
--- a/src/lib/define-dynamic-block.js
+++ b/src/lib/define-dynamic-block.js
@@ -2,6 +2,7 @@
 // Should we move these into a new extension support module or something?
 import ArgumentType from 'scratch-vm/src/extension-support/argument-type';
 import BlockType from 'scratch-vm/src/extension-support/block-type';
+import log from './log.js';
 
 /**
  * Define a block using extension info which has the ability to dynamically determine (and update) its layout.
@@ -31,6 +32,46 @@ const defineDynamicBlock = (ScratchBlocks, categoryInfo, staticBlockInfo, extend
         if (staticBlockInfo.blockIconURI || categoryInfo.blockIconURI) {
             blockJson.extensions = ['scratch_extension'];
         }
+
+        // Handle custom context menu options
+        // TODO this should probably not live here in the future, or at least
+        // we need some way of registering a context menu option only once for
+        // each block (see try catch below)
+        if (staticBlockInfo.info.customContextMenu) {
+            const customContextMenuForBlock = {
+                customContextMenu: function (options) {
+                    staticBlockInfo.info.customContextMenu.forEach(contextOption => {
+                        options.push({
+                            enabled: true,
+                            text: contextOption.name,
+                            callback: () => {
+                                if (contextOption.builtInCallback) {
+                                    switch (contextOption.builtInCallback) {
+                                    case 'EDIT_A_PROCEDURE':
+                                        // TODO FILL THIS IN
+                                        break;
+                                    case 'RENAME_A_VARIABLE':
+                                        // TODO FILL THIS IN
+                                        break;
+                                    }
+                                } else if (contextOption.callback) {
+                                    contextOption.callback();
+                                }
+                            }
+                        });
+                    });
+                }
+            };
+            const contextMenuName = `${blockJson.type}_context_menu`;
+            try {
+                ScratchBlocks.Extensions.registerMixin(contextMenuName, customContextMenuForBlock);
+            } catch (e) {
+                log.warn("Context menu callback was already registered, but we're going to ignore this for now");
+            }
+            blockJson.extensions = blockJson.extensions || [];
+            blockJson.extensions.push(contextMenuName);
+        }
+
         // initialize the basics of the block, to be overridden & extended later by `domToMutation`
         this.jsonInit(blockJson);
         // initialize the cached block info used to carry block info from `domToMutation` to `mutationToDom`
--- a/src/lib/define-dynamic-block.js
+++ b/src/lib/define-dynamic-block.js
@@ -8,6 +8,13 @@ import log from './log.js';
 const setupCustomContextMenu = (ScratchBlocks, contextMenuInfo, extendedOpcode) => {
     // Handle custom context menu options
     const customContextMenuForBlock = {
+        /**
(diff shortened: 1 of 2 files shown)


src/lib/define-dynamic-block.js
```diff
@@ -21,6 +22,46 @@ const defineDynamicBlock = (ScratchBlocks, categoryInfo, staticBlockInfo, extend
...
+                                    }
+                                } else if (contextOption.callback) {
+                                    contextOption.callback();
+                                }
+                            }
+                        });
+                    });
+                }
+            };
+            const contextMenuName = `${blockJson.type}_context_menu`;
```
cwillisf: I think this would be slightly more readable but I suppose it depends on whether the reader is more familiar with Scratch Blocks or extensions :)
```suggestion
            const contextMenuName = `${extendedOpcode}_context_menu`;
```

src/lib/define-dynamic-block.js
```diff
@@ -21,6 +22,46 @@ const defineDynamicBlock = (ScratchBlocks, categoryInfo, staticBlockInfo, extend
         if (staticBlockInfo.blockIconURI || categoryInfo.blockIconURI) {
             blockJson.extensions = ['scratch_extension'];
         }
+
+        // Handle custom context menu options
+        // TODO this should probably not live here in the future, or at least
```
cwillisf: I'd say this should probably be a separate function but it's fine to be in this file, called from this location. In my mind the problem of re-registering context menus is basically the same as the existing problem we have with re-registering blocks. I think we might just have to accept the `delete ScratchBlocks.somewhere[blockType]` / `delete ScratchBlocks.somewhereElse[contextMenuName]` kind of approach. Would that be sufficient in the menu case, or is there more we need to do in order to "unregister" a context menu?

kchadha: I haven't tested this out, but I think we might be able to do something smarter with it (with a little bit of effort that I did not think was worth it for the time being)... 

Looking at the data category as an example from scratch blocks (https://github.com/LLK/scratch-blocks/blob/develop/blocks_vertical/data.js#L51), we can have a map of contextMenu options per block to do something similar to what we're doing for the `scratch_extension` ScratchBlocks extension above.

kchadha: Not sure if that comment was totally clear. I was thinking we'd do something similar to how we handle block menus where we define the shadow blocks for menus separately from the specific blocks specified by the extensions. These blocks are defined only once during `handleExtensionAdded` in the blocks container. We could probably similarly have a blockContextMenus section of the _extension info_ and register the mixins only once when the extension is loaded.

kchadha: @cwillisf, I just reorganized the code so that there is a separate helper function for the custom context menu setup. I've also changed the structure of define dynamic block a bit so that it calls the setup function before returning the block definition object (containing the `init`, `domToMutation` and `mutationToDom` functions). This means that the context menu registration attempt is happening way less often now (previously it was running every time you move the block on the workspace), but it's still not completely de-duplicated for the same reason as the blocks getting redefined issue.

Tracking in #TBD

kchadha: I forgot to mention, I recommend viewing these changes with the `hide whitespace changes` option checked because of the reorganized code.

src/lib/define-dynamic-block.js
```diff
@@ -2,6 +2,62 @@
...
+        customContextMenu: function (options) {
+            contextMenuInfo.forEach(contextOption => {
+                const option = {
+                    enabled: true,
+                    text: contextOption.text,
+                    callback: () => {
+                        if (contextOption.builtInCallback) {
+                            switch (contextOption.builtInCallback) {
+                            case 'EDIT_A_PROCEDURE':
+                                // TODO FILL THIS IN
```
cwillisf: What kind of thing do you imagine going here? Firing an event maybe? Or will this be a direct call into `ScratchBlocks`?

kchadha: In this block of code (and the associated `RENAME_A_VARIABLE` block) you use the gui context to open the appropriate modal (e.g. edit modal or rename modal) and then handle the data submitted from the modal and eventually make a call to the contextOption.callback so that the extension callback also has access to the data submitted from the modal.

test/unit/util/define-dynamic-block.test.js
```diff
@@ -19,50 +28,73 @@ const penIconURI = 'data:image/svg+xml;base64,fake_pen_icon_svg_base64_data';
...
 };
 
 class MockBlock {
-    constructor (blockInfo, extendedOpcode) {
+    constructor (staticBlockInfo, extendedOpcode) {
         // mimic Closure-style inheritance by mixing in `defineDynamicBlock` output as this instance's prototype
         // see also the `Blockly.Block` constructor
-        const prototype = defineDynamicBlock(MockScratchBlocks, categoryInfo, blockInfo, extendedOpcode);
+        const ddb = defineDynamicBlock.bind(this);
+        const prototype = ddb(MockScratchBlocks, categoryInfo, staticBlockInfo, extendedOpcode);
```
cwillisf: Seems like these two lines could be replaced with:
```js
const prototype = defineDynamicBlock.call(this, MockScratchBlocks, categoryInfo, staticBlockInfo, extendedOpcode);
```

kchadha: Oh no, that was an accidental remnant of me trying to make the tests work! I'm going to change it back to the one prototype line that it was before.

kchadha:
```suggestion
        const prototype = defineDynamicBlock(MockScratchBlocks, categoryInfo, staticBlockInfo, extendedOpcode);
```

test/unit/util/define-dynamic-block.test.js
```diff
@@ -19,50 +28,73 @@ const penIconURI = 'data:image/svg+xml;base64,fake_pen_icon_svg_base64_data';
...
     }
 };
 
 class MockBlock {
-    constructor (blockInfo, extendedOpcode) {
+    constructor (staticBlockInfo, extendedOpcode) {
         // mimic Closure-style inheritance by mixing in `defineDynamicBlock` output as this instance's prototype
         // see also the `Blockly.Block` constructor
-        const prototype = defineDynamicBlock(MockScratchBlocks, categoryInfo, blockInfo, extendedOpcode);
+        const ddb = defineDynamicBlock.bind(this);
```
kchadha:
```suggestion
```

test/unit/util/define-dynamic-block.test.js
kchadha: Fixed!

PR #4795, opened 2019-05-04
Make GUI integration tests reliable

### LINKED ISSUE #4653: Some integration tests are flaky WRT "loader background"
### Expected Behavior

Integration test results should only depend on the code: re-running a failed test without any code changes should never lead to success.

### Actual Behavior

Occasionally, some of the integration tests fail with an error like this:

```
 FAIL  test/integration/blocks.test.js (57.199s)
  ● Working with the blocks › Blocks report when clicked in the toolbox
    WebDriverError: unknown error: Element <span>...</span> is not clickable at point (59, 74). Other element would receive the click: <div class="loader_background_2DPrW">...</div>
```

I have personally seen this issue with `blocks.test.js` and `backpack.test.js` but I suspect it's possible with other tests as well. These tests seem to fail for me locally more often than on Travis CI, but I have seen both.

Example: https://travis-ci.org/LLK/scratch-gui/builds/506044876#L2046

I believe this is a timing issue related to the loading / project creation screen sometimes being in the way of the test's simulated click.

### Steps to Reproduce

1. Build with `npm run build`
2. Run integration tests with `npm run test:integration`

Alternatively:
1. Build with `npm run build`
2. Run an individual integration test with `./node_modules/.bin/jest test/integration/backpack.test.js'

cwillisf:
### Resolves

Resolves #4653

### Proposed Changes

The most important change here is that `SeleniumHelper`'s `getDriver()` and `getSauceDriver()` no longer replace an existing driver if present. Also, `SeleniumHelper`'s `loadUri()` now checks if `loader_background` element(s) is/are present, and if so it waits for it/them to disappear.

### Reason for Changes

The previous implementation of `getDriver()` was sometimes causing trouble with tests running in parallel. The trouble shows up in cases like this pseudocode (note that I've made `helper` calls explicit for illustration purposes):
```js
let driver;
beforeAll(() => {
  driver = helper.getDriver();
}

test('test A', async () => {
  await helper.loadUri(somwhere);
  await helper.clickText('Fancy Button');
});

test('test B', async () => {
  await driver.quit();
  driver = helper.getDriver();
});
```

The above code may sometimes run without issue, but if `test B` runs `driver.quit()` after `test A` starts `loadUri()` but before `clickText()` finishes there will be trouble. The exact nature of the trouble depends on when exactly `driver = helper.getDriver()` finishes; since both `loadUri` and `clickText` contain multiple asynchronous steps which internally check `this.driver` the error can take several forms.

Now, tests which have a specific need for multiple sessions are responsible for creating a separate `SeleniumHelper` instance for each session. It's a bit of a pain, but it's reliable. I'm open to proposals for other ways to handle this.

After implementing the `getDriver()` change, I found that some tests were now less reliable than before -- or even were reliably failing. To fix this, I adjusted `SeleniumHelper`'s `loadUri()` method to wait until React has rendered at least once, then check if the `loader_background` element is present and if so wait for it to go away. A few tests included similar logic already, which I removed now that it's handled centrally.

### DIFF
--- a/test/helpers/selenium-helper.js
+++ b/test/helpers/selenium-helper.js
@@ -44,38 +44,46 @@ class SeleniumHelper {
     }
 
     getDriver () {
-        const chromeCapabilities = webdriver.Capabilities.chrome();
-        const args = [];
-        if (USE_HEADLESS) {
-            args.push('--headless');
+        // It's not safe to replace an existing driver since there may be outstanding async chains already in progress.
+        // Alternative solution: remove/replace all async code in this class which mentions `this.driver`
+        if (!this.driver) {
+            const chromeCapabilities = webdriver.Capabilities.chrome();
+            const args = [];
+            if (USE_HEADLESS) {
+                args.push('--headless');
+            }
+
+            // Stub getUserMedia to always not allow access
+            args.push('--use-fake-ui-for-media-stream=deny');
+
+            chromeCapabilities.set('chromeOptions', {args});
+            chromeCapabilities.setLoggingPrefs({
+                performance: 'ALL'
+            });
+            this.driver = new webdriver.Builder()
+                .forBrowser('chrome')
+                .withCapabilities(chromeCapabilities)
+                .build();
         }
-
-        // Stub getUserMedia to always not allow access
-        args.push('--use-fake-ui-for-media-stream=deny');
-
-        chromeCapabilities.set('chromeOptions', {args});
-        chromeCapabilities.setLoggingPrefs({
-            performance: 'ALL'
-        });
-        this.driver = new webdriver.Builder()
-            .forBrowser('chrome')
-            .withCapabilities(chromeCapabilities)
-            .build();
         return this.driver;
     }
 
     getSauceDriver (username, accessKey, configs) {
-        this.driver = new webdriver.Builder()
-            .withCapabilities({
-                browserName: configs.browserName,
-                platform: configs.platform,
-                version: configs.version,
-                username: username,
-                accessKey: accessKey
-            })
-            .usingServer(`http://${username}:${accessKey
-            }@ondemand.saucelabs.com:80/wd/hub`)
-            .build();
+        // It's not safe to replace an existing driver since there may be outstanding async chains already in progress.
+        // Alternative solution: remove/replace all async code in this class which mentions `this.driver`
+        if (!this.driver) {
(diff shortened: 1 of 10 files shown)


benjiwheeler:
Wow, seems like awesome work, CWF!

paulkaplan:
@cwillisf i'm a bit surprised about things happening in parallel. We explicitly remove the parallel test running in jest (`--runInBand` in the package.json script for test:integration) and i thought within a test that it would wait until the test finishes to start the next test. Did you see this behavior explicitly (using logging or something?).

w/r/t the tests needing to explicitly restart the session, i think your solution is fine, it is definitely not worse than calling driver.quit -> new driver...

test/helpers/selenium-helper.js
```diff
@@ -87,19 +95,31 @@ class SeleniumHelper {
         return this.findByXpath(`//body//${scope || '*'}//*[contains(text(), '${text}')]`);
     }
 
-    loadUri (uri) {
+    async loadUri (uri) {
```
paulkaplan: What is the significance of this change?

cwillisf: This just allows me to use `await` in the body... IMO that makes it easier to read, but I could switch back to a `.then(...)` chain if you prefer.

cwillisf:
> We explicitly remove the parallel test running in jest (--runInBand in the package.json script for test:integration) and i thought within a test that it would wait until the test finishes to start the next test.

True, but I believe that only prevents separate test files from running together -- it prevents running, say, `backdrops.test.js` and `backpack.test.js` in parallel. The trouble I'm trying to fix here is when two tests within the same file are interleaving their tasks with each other -- say, `Nonexistent projects show error screen` and `Load a project by ID directly through url` from `project-loading.test.js` -- which AFAIK isn't controlled by Jest at all.

paulkaplan:
@cwillisf Hm yup you are right about runInBand only keeping files separate, but I'm still confused about how tests in the same file could be being interleaved. I can't seem to reproduce this locally (on develop, I put in logging at the top/bottom of each test, but they seem to log in order with no overlap).

test/helpers/selenium-helper.js
paulkaplan: Ah right sounds good!

cwillisf:
out of date - closing

Issue #4762, opened 2019-04-19
When confirming project upload with user, use custom modal


benjiwheeler:
Follow-on to https://github.com/LLK/scratch-gui/pull/4700, further resolving https://github.com/LLK/scratch-gui/issues/3509

When confirming project upload with user, user should see the nifty custom modal @carljbowman designed,  instead of the default browser js modal.

Current:

![image](https://user-images.githubusercontent.com/3431616/56439802-700ccb00-62b5-11e9-85b6-c06d83838f95.png)

Desired:

![55671093-1f7b8380-585a-11e9-9b84-d63ed9369054](https://user-images.githubusercontent.com/3431616/56439817-78650600-62b5-11e9-9ec0-4da0cea2039f.png)

apple502j:
Why don't you use Alerts?

alexwenbj:
@apple502j I think because the built-in Alert's style is not unified in different browsers.

benjiwheeler:
@apple502j Do you mean a specific js/css library called Alerts? Or do you mean the built-in js alert()/confirm() functions?

If you mean the built-in ones, then I echo what @alexwenbj said -- we are more confident about what we're communicating to the user if we're using our own design.

apple502j:
@benjiwheeler The alert system - like the one that pops up when you save projects

benjiwheeler:
@apple502j That's a great idea. We discussed this possibility early on. I believe there's no technical reason why we couldn't do it, it's just a design choice.

Ultimately we designed this custom prompt, which I think serves a distinct role from the alerts. (It could also be useful for the sprite deletion confirm.)

What turned out to be tricky is that this custom prompt is a modal -- which means we automatically close it on any mouse click outside its bounds. But the upload menu selection itself is _also_ listening for clicks, and closing on any clicks outside _its_ bounts. So attempting to click "OK" closes both of them, and the button never knows it was clicked!

We have a fix worked out, but it'll take some work -- the first step is to examine and reconcile this click-to-close behavior. 

Just been trying to get through other stuff, and get back to this!

apple502j:
@benjiwheeler FYI: #5318 fixes it a bit

PR #4809, opened 2019-05-10
refactor SB* file uploader into top-level HOC

benjiwheeler:
### Resolves

Big step towards resolving https://github.com/LLK/scratch-gui/issues/4762

### Proposed Changes

* move `sb-file-uploader` to a top-level HOC that wraps `gui`
* in situation where user has selected to upload project, but file content is not ready/available, dispatch `project-state` action `RETURN_TO_SHOWING`

### Reason for Changes

1. we made design decision to use custom confirm modal, to have greater confidence in and control of the user experience when the user is at risk of overwriting their work. But this is not doable yet because clicking on the modal is interpreted as a click outside the open file menu, which closes the file menu. This is a problem because...
2. previously, the `sb-file-uploader` would be unmounted any time the file menu closed; its existence/mounting should not depend on a menu. 
3. in testing, when i would clear the contents of the file to upload before it could be loaded into the vm, `sb-file-uploader-hoc` would correctly avoid loading the empty content; but it would be stuck in the `LOADING_VM_FILE_UPLOAD` state. Submitting `RETURN_TO_SHOWING` in this case makes sure we will stay in the right state in that case. (We shouldn't actually be missing the contents of the file to upload in practice, but just in case!)

### Test Coverage

Added a test and moved over tests to new HOC, but ideally we would test even more.


src/reducers/project-state.js
```diff
@@ -435,10 +435,21 @@ const onLoadedProject = (loadingState, canSave, success) => {
...
+    } else {
+        switch (loadingState) {
+        case LoadingState.LOADING_VM_WITH_ID:
+        case LoadingState.LOADING_VM_FILE_UPLOAD:
+            return {
+                type: RETURN_TO_SHOWING
+            };
+        case LoadingState.LOADING_VM_NEW_DEFAULT:
+            return {
+                type: START_ERROR
```
benjiwheeler: If loading fails (success === false), recover by going back to just showing project as-is. But if we were loading the default project, a failure in loading probably indicates a serious error.

test/unit/util/project-saver-hoc.test.jsx
```diff
@@ -56,7 +56,7 @@ describe('projectSaverHOC', () => {
         expect(mockedUpdateProject).toHaveBeenCalled();
     });
 
-    test('if canSave is alreatdy true and we show a project with an id, project will NOT be saved', () => {
+    test('if canSave is already true and we show a project with an id, project will NOT be saved', () => {
```
benjiwheeler: fixed a typo

src/lib/sb-file-uploader-hoc.jsx
```diff
@@ -0,0 +1,241 @@
+import bindAll from 'lodash.bindall';
```
benjiwheeler: I tried to keep as much as possible in this file the same as before. There are very few differences, aside from what was necessary to refactor it into a top-level HOC wrapping GUI.

benjiwheeler:
Ready for review; just blocking until after this week's deploy.

benjiwheeler:
@paulkaplan and I talked about this. Some takeaways:

Concerns:
* Z is currently working to start loading project before gui fully renders. Adding a top level HOC, with an additionally rendered input component, can slow this down.
* need to assess/measure/understand how much this increases the overall performance burden on gui.

Possible alternative approaches:
* in `containers/menu.jsx`, `mouseup` events cause the menu to be closed. This is the original problem leading to this PR's refactor. What if instead, we just checked if the mouse target was a modal, and didn't close the menu then?

Next step:
* pick this back up when we have more time to discuss. Possibly with @rschamp .

Labeling as blocked, per need for discussion.

src/reducers/project-state.js
```diff
@@ -435,10 +435,21 @@ const onLoadedProject = (loadingState, canSave, success) => {
         default:
             return;
         }
+    } else {
```
rschamp: Can this be extracted into a different PR?

src/lib/sb-file-uploader-hoc.jsx
```diff
@@ -0,0 +1,241 @@
...
+                /* eslint-enable no-unused-vars */
+                ...componentProps
+            } = this.props;
+            return (
+                <React.Fragment>
+                    <WrappedComponent
+                        onStartSelectingFileUpload={this.handleStartSelectingFileUpload}
+                        {...componentProps}
+                    />
+                    {fileInput}
```
rschamp: Let's try to create this element dynamically in the handler, rather than mounting it here, to avoid issues with components unmounting.

benjiwheeler:
Discussed with @rschamp and @paulkaplan. Current plan: 

1. The menu click issue that prompted this should be fixed in and of itself
2 I should pull out my project state tests and state fix, and introduce those separately
3. The custom modal for project upload can then be introduced
4. Redo this PR as only a refactor, which doesn't change behavior at all
    * change placement of the file picker to be a createelement that’s not even on the dom, per Ray's comment: "Let's try to create this element dynamically in the handler, rather than mounting it here, to avoid issues with components unmounting."

src/reducers/project-state.js
benjiwheeler: Did so -- see https://github.com/LLK/scratch-gui/pull/5291

benjiwheeler:
Updated plan:

1. Redo this PR as only a refactor, which doesn't change behavior at all.
    * change placement of the file picker to be a createelement that’s not even on the dom, per Ray's comment: "Let's try to create this element dynamically in the handler, rather than mounting it here, to avoid issues with components unmounting."
2. Re-introduce custom modal; discuss with participants on https://github.com/LLK/scratch-gui/issues/5386

src/lib/sb-file-uploader-hoc.jsx
benjiwheeler: > I tried to keep as much as possible in this file the same as before. There are very few differences, aside from what was necessary to refactor it into a top-level HOC wrapping GUI.

Now there are more differences -- as part of making the <input> component dynamic, I reordered functions and added comments.

benjiwheeler:
This PR is ready for review, but tests aren't passing; it's difficult to diagnose because the same tests pass locally.

benjiwheeler:
OK @rschamp, take a look at the dynamic way sb-file-uploader-hoc is now creating the <input> element.

src/lib/sb-file-uploader-hoc.jsx
```diff
@@ -0,0 +1,269 @@
...
+                    .then(() => {
+                        this.props.onLoadingFinished(this.props.loadingState, true);
+                        if (filename) {
+                            const uploadedProjectTitle = this.getProjectTitleFromFilename(filename);
+                            this.props.onUpdateProjectTitle(uploadedProjectTitle);
+                        }
+                    })
+                    .catch(error => {
+                        log.warn(error);
+                        this.props.intl.formatMessage(messages.loadError); // eslint-disable-line no-alert
```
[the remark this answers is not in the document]
benjiwheeler: Doing it

src/lib/sb-file-uploader-hoc.jsx
```diff
@@ -0,0 +1,269 @@
...
+        const loadingState = state.scratchGui.projectState.loadingState;
+        const user = state.session && state.session.session && state.session.session.user;
+        return {
+            isLoadingUpload: getIsLoadingUpload(loadingState),
+            isShowingWithoutId: getIsShowingWithoutId(loadingState),
+            loadingState: loadingState,
+            projectChanged: state.scratchGui.projectChanged,
+            userOwnsProject: ownProps.authorUsername && user &&
+                (ownProps.authorUsername === user.username),
+            vm: state.scratchGui.vm // NOTE: double check this belongs here
```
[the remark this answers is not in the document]
benjiwheeler: Doing it

PR #4835, opened 2019-05-21
Pause video on tutorial window shrink

### LINKED ISSUE #4766: Clicking "shrink" on tutorial window should pause the video
If you play a video in a tutorial window, then click the "shrink" button, it should pause the video.

evhan55:
### Resolves

- Resolves #4766: Clicking "shrink" on tutorial window should pause the video

### Proposed Changes

To access the `Wistia` API object on the `window`, we need to switch the video embed from an iframe (which prevents access to its window for security reasons), to the inline-embedding style that Wistia suggests. (https://wistia.com/support/embed-and-share/video-on-your-website). Wistia video player API docs for reference: https://wistia.com/support/developers/player-api.

This also required upgrading the VideoStep React component from a functional one to a normal one, to handle the video-pausing in `componentDidUpdate`.

### DIFF
--- a/src/components/cards/cards.jsx
+++ b/src/components/cards/cards.jsx
@@ -86,10 +86,6 @@ const CardHeader = ({onCloseCards, onShrinkExpandCards, onShowAll, totalSteps, s
 );
 
 class VideoStep extends React.Component {
-    // eslint-disable-next-line no-useless-constructor
-    constructor (props) {
-        super(props);
-    }
     componentDidMount () {
         const script = document.createElement('script');
         script.src = `https://fast.wistia.com/embed/medias/${this.props.video}.jsonp`;
--- a/src/components/cards/cards.jsx
+++ b/src/components/cards/cards.jsx
@@ -86,33 +86,35 @@ const CardHeader = ({onCloseCards, onShrinkExpandCards, onShowAll, totalSteps, s
     </div>
 );
 
-// Video step needs to know if the card is being dragged to cover the video
-// so that the mouseup is not swallowed by the iframe.
 class VideoStep extends React.Component {
     // eslint-disable-next-line no-useless-constructor
     constructor (props) {
         super(props);
     }
     componentDidMount () {
-        console.log('Video1 step did mount.');
         const script = document.createElement('script');
-
         script.src = `https://fast.wistia.com/embed/medias/${this.props.video}.jsonp`;
         script.async = true;
-    
         document.body.appendChild(script);
 
         const script2 = document.createElement('script');
-
+        script2.onload = () => {
+            window._wq = window._wq || [];
+            window._wq.push({
+                id: `${this.props.video}`,
+                onReady: function (video) {
+                    console.log('I got a handle to the video!', video);
+                }
+            });
+        };
         script2.src = 'https://fast.wistia.com/assets/external/E-v1.js';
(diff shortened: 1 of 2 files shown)


benjiwheeler:
I'm not sure how much this matters, but this change affects the way that dragging on the video timeline (called "scrobbling", IIRC) behaves. It wasn't perfect before -- on Mac Chrome, I could get into a weird state where the video would track my mouse even when I'm not clicking. But now, dragging in the timeline will move the whole window.

src/components/cards/cards.jsx
```diff
@@ -85,32 +85,46 @@ const CardHeader = ({onCloseCards, onShrinkExpandCards, onShowAll, totalSteps, s
...
-        />
-        <script
-            async
-            src="https://fast.wistia.net/assets/external/E-v1.js"
-        />
-    </div>
-);
+class VideoStep extends React.Component {
+    componentDidMount () {
+        const script = document.createElement('script');
```
benjiwheeler: One concern here: this adds script elements to the page, but never removes them.

Maybe you could do something like:

Call `script.setAttribute('id', 'wistia-video-content')` here, and then in `componentWillUnmount` you could call

```
var wistiaScript = document.getElementById('wistia-video-content');
wistiaScript.parentNode.removeChild(wistiaScript);
```

src/components/cards/cards.jsx
```diff
@@ -85,32 +85,46 @@ const CardHeader = ({onCloseCards, onShrinkExpandCards, onShowAll, totalSteps, s
...
+                </div>
+            </div>
+        );
+    }
+}
 
 VideoStep.propTypes = {
     dragging: PropTypes.bool.isRequired,
+    expanded: PropTypes.bool.isRequired,
     video: PropTypes.string.isRequired
```
benjiwheeler: This idea is beyond the scope of this change, but I'd love to see the `video` prop renamed. What is it? A video Id? A video URL?

src/components/cards/cards.jsx
evhan55: @benjiwheeler Ah great idea, thanks!  I will try that.

src/components/cards/cards.jsx
evhan55: I believe `video` is an id for Wistia embed URLs, like the value for `video` of `ir0j8ljsgm` is one I have stored for testing purposes (it's the id for the 'Make Music' tutorial video).

evhan55:
@benjiwheeler 

Thanks for the review!
- I will check to see about the scrobbling behavior
- I will also try the dynamic script removal suggestion

evhan55:
@ericrosenbaum The change from 'iframe' to 'div' here (to make dynamic video pausing possible) is introducing a new bug: you can't scroll on the video timeline now because the mouse click to do that is triggering a drag operation instead.  I am investigating how to prevent this dragging but it's hairy.  Should we keep this PR open while we figure out how to prevent the dragging, or should we land this PR and open a new issue to track the new bug?

evhan55:
@benjiwheeler 

I think I addressed your two points:
- scripts now dynamically removed
- video div is disabled for dragging now

Please let me know if it's working any better for you!
Also I can rename the `video-div` id to something better if you have any ideas to match id naming conventions elsewhere.

Issue #2744, opened 2018-07-31
List monitors don't wrap to the next line


thisandagain:
### Expected Behavior

Long items should wrap to the next line:
![image](https://user-images.githubusercontent.com/31634240/43318730-eb819ddc-916f-11e8-856f-5a6a1c0f0bb1.png)

### Actual Behavior

They don't:
![image](https://user-images.githubusercontent.com/31634240/43318742-fa6528be-916f-11e8-8fd1-08e40659bd41.png)

### Steps to Reproduce

Create a list with a long item.
### Operating System and Browser

Windows 10 Chrome 67

---
Moved from LLK/scratch-render#324
/cc @kyleplo

towerofnix:
This is a much-requested feature so I'd like to work on implementing it, but I'm not too sure where to start. The trouble is that list monitors make use of react-virtualized's List component - they aren't just ordinary CSS-styled elements. In order for react-virtualized to do what it does, it needs to know the height of the elements. At the moment, we set it to the hard-coded value 24; but in the case of this suggestion, it will be a dynamic value. Luckily, you can pass a function to react-virtualized to calculate the height, but we still need to figure out *how* to calculate the height. And I'm not sure how to do that. In order to know the height, I'd guess that we need to know how many lines of text there will be - and I'm not sure how to calculate that without, say, rendering the element in memory or doing some similar black magic. Anyone have any thoughts..?

FWIW the code to modify is mostly in `components/monitor/list-monitor-scroller.jsx` - particularly change `rowHeight={24}` to, for example, `rowHeight={this.calculateRowHeight}` (and define the code in there). Various CSS in `monitor.css` needs to be adjusted too.

towerofnix:
I'd quite like to focus on fixing this as it remains an issue that many people notice and are frustrated by: [List Items Got Shortened](https://scratch.mit.edu/discuss/topic/347102/)

>I was actually in the process of making a text based game- called “Potato Adventure”- and it gets very annoying to use abbreviations and shorter words.

>But the fact that list items are getting cut off instead of wrapping to the next line is also hurting some projects I liked to play. Support.

>Support. I'm making an OS (VFDan OS), and one of the features is notes. Long notes get cut off. It would be great if it wrapped instead of cutting off.

Plus plenty of other posts here.

For a help-wanted issue, I can't go further until the discussion above is figured out. I suppose that simply removing react-virtualized is not an option..? :)

adroitwhiz:
[...]

PR #4854, opened 2019-05-26
Always display whitespace in list monitors

### LINKED ISSUE #4823: Contiguous space characters collapse to one in var/list monitors
Not sure if this was a conscious decision, but it's been something that's niggled me about Scratch 3 since it first came out, so thought I'd bring it up and see how people respond...

### Expected Behavior

Can use spaces in variable/list values to 'pad' how it appears.

### Actual Behavior

As title.

### Steps to Reproduce

Create a variable and give it a value `"a     b"` (note multiple spaces between letters).
See that the var's monitor shows it as "a b" (only a single space between letters).

See https://scratch.mit.edu/projects/20460790/ for an example of how multiple spaces could have some degree of use. (Compare against project running within Scratch 2. Note that Scratch 2 did drop trailing spaces – not sure if that would remain a desired behaviour?)

N/A

towerofnix:
### Resolves

Fixes #4823.

### Proposed Changes

Changes the whitespace CSS on the element from `nowrap` to `pre`.

### Reason for Changes

We want to display the whitespace in lists exactly as it was added!

### Test Coverage

Tested manually:

Four items: "a b", "c  d" (2 spaces), "e   f" (3 spaces), "very long item" (cut off with an ellipsis after "it")

Note that `white-space: pre` prevents the text from wrapping, just as we want ([until this is fixed](https://github.com/LLK/scratch-gui/issues/2744)). (We'd want `white-space: pre-wrap` to wrap it.)

### Browser Coverage

Linux Firefox, Chromium.

### DIFF
--- a/src/components/monitor/monitor.css
+++ b/src/components/monitor/monitor.css
@@ -149,7 +149,7 @@
     overflow: hidden; /* Don't let long values escape container */
     text-overflow: ellipsis;
     user-select: text; /* Allow selecting list values for 2.0 compatibility, only relevant in player */
-    white-space: nowrap;
+    white-space: pre;
 }
 
 .list-input {
--- a/src/components/monitor/monitor.css
+++ b/src/components/monitor/monitor.css
@@ -43,6 +43,7 @@
     margin: 0 5px;
     border-radius: calc($space / 2);
     padding: 0 2px;
+    white-space: pre-wrap;
     transform: translateZ(0); /* Fixes flickering in Safari */
 }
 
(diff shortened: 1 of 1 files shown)


thisandagain:
@rschamp Would you mind taking a look at this?

towerofnix:
@rschamp Oh, good catch! I've pushed a commit that makes ordinary monitors show whitespace too:

A list, normal, and large-readout monitor, each presenting multiple whitespace characters in a row

I use the style `white-space: pre-wrap` for these, instead of `pre`, because these classes wrap by default (the standardized browser default style is `white-space: normal`, which wraps text).

PR #4871, opened 2019-05-29
Stabilize integration tests

BryceLTaylor:
### Resolves

This should make the integration tests more stable.  This makes sure elements are visible on the page during integration tests before moving on in the tests.

### Proposed Changes

Makes the FindByXpath function in Selenium helpers wait until element is visible before returning.  This replaces the WaitUntilGone method that was used in some tests. 

### Reason for Changes

FindByXpath is ultimately used by all of the other ways of finding elements in Selenium helpers so this should effect most of the tests.  This should specifically make it so elements that tare hidden by the loading screen are revealed before they are clicked in tests.

After the changes the WaitUntilGone function was causing tests to fail.  Since it was supposed to solve the same problem, I removed that function from the rest of the tests.  All the tests passed locally after that.

### Browser Coverage
Mac Chrome

### DIFF
--- a/test/helpers/selenium-helper.js
+++ b/test/helpers/selenium-helper.js
@@ -85,7 +85,11 @@ class SeleniumHelper {
     }
 
     findByXpath (xpath, timeoutMessage = `findByXpath timed out for path: ${xpath}`) {
-        return this.driver.wait(until.elementLocated(By.xpath(xpath)), DEFAULT_TIMEOUT_MILLISECONDS, timeoutMessage);
+        return this.driver.wait(until.elementLocated(By.xpath(xpath)), DEFAULT_TIMEOUT_MILLISECONDS, timeoutMessage)
+            .then(el => (
+                this.driver.wait(el.isDisplayed(), DEFAULT_TIMEOUT_MILLISECONDS, `${xpath} is not visible`)
+                    .then(() => el)
+            ));
     }
 
     findByText (text, scope) {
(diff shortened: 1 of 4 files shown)


test/helpers/selenium-helper.js
```diff
@@ -85,7 +84,11 @@ class SeleniumHelper {
     }
 
     findByXpath (xpath, timeoutMessage = `findByXpath timed out for path: ${xpath}`) {
-        return this.driver.wait(until.elementLocated(By.xpath(xpath)), DEFAULT_TIMEOUT_MILLISECONDS, timeoutMessage);
+        return this.driver.wait(until.elementLocated(By.xpath(xpath)), DEFAULT_TIMEOUT_MILLISECONDS, timeoutMessage)
+            .then(el => (
+                this.driver.wait(el.isDisplayed(), DEFAULT_TIMEOUT_MILLISECONDS, `${xpath} is not visible`)
```
benjiwheeler: This line may be a cause of the testing errors seen at https://github.com/LLK/scratch-gui/issues/4902 and https://github.com/LLK/scratch-gui/issues/4903

benjiwheeler:
DId a bunch of research on why the "Adding costumes DOES update the default costume name in the toolbox" test in integration/blocks.test.js test fails. 
(See https://github.com/LLK/scratch-gui/issues/5429 ) 

I identified the `this.driver.wait(el.isDisplayed()` line in selenium-helper.js as the culprit. (See my previous note above.)

PR #4874, opened 2019-05-30
Revert "render only new toolbox xmls"

paulkaplan:
Reverts LLK/scratch-gui#4677

This may be causing a bug where after switching languages, the "Say" block contains an updated default message but when you run the block it still says "hello".


rschamp:
Try it out: [revert-4677-update-toolbox](https://llk.github.io/scratch-gui/revert-4677-update-toolbox)

PR #4881, opened 2019-05-30
Revert "Revert "Pause video on tutorial window shrink""

evhan55:
Reverts LLK/scratch-gui#4877

### DIFF
--- a/src/components/cards/card.css
+++ b/src/components/cards/card.css
@@ -275,14 +275,6 @@
     margin-right: 0.5rem;
 }
 
-.video-cover {
-    width: 100%;
-    height: 100%;
-    position: absolute;
-    top: 0;
-    left: 0;
-}
-
 .steps-list {
     display: flex;
     flex-direction: row;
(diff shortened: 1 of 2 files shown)


rschamp:
Try it out: [revert-4877-revert-4835-bugfix/4766](https://llk.github.io/scratch-gui/revert-4877-revert-4835-bugfix/4766)

evhan55:
@benjiwheeler 

Thanks for the bug hunt testing of this today!

I have fixed it so that the video updates if a new one is selected from the library.
One thing I notice is that if a video has started playing, and then select the **same** tutorial from the library again, it doesn't update to restart the video, it just stays where it was (paused or playing).  Is that okay?

benjiwheeler:
> if a video has started playing, and then you select the same tutorial from the library again, it doesn't update to restart the video, it just stays where it was (paused or playing). Is that okay?

Yes, that seems to be expected behavior to me.

Issue #4895, opened 2019-06-04
Replace tutorial videos with thumbnails on "offline" platforms


thisandagain:
When loading tutorials on an "offline" platform (e.g. Scratch Desktop, Scratch for Android) we would like to replace the initial video with the tutorial thumbnail (as shown in the tutorial library) as to avoid situations where the video fails to load when internet access is either unreliable or unavailable. 

We feel that this is an important enhancement to have in place prior to the start of Africa Code Week (https://africacodeweek.org/). While the event is not until October, the Africa Code Week team has requested that our materials be delivered by the end of June in time for the start of teacher training which means that we would want to have this landed, tested, and released before the end of this iteration.

/cc @ntlrsk @kosiecki @carljbowman

cwillisf:
Can we do some fancy JavaScript stuff to make it so that we initially load the thumbnail and then if possible replace the thumbnail with the video? If we can, should we?

For that matter, if it's reasonable to do the above we might want to do that even for the normal, online version of Scratch since it make the tutorials more robust against (for example) the video provider being down. On the other hand, maybe we don't want this feature in "offline" versions: I could see someone objecting to it if they're trying to reduce data charges.

thisandagain:
@cwillisf I most certainly feel like this is something we may want to consider long-term. @paulkaplan and I also talked about perhaps including the videos as an option in a preferences pane somewhere. I think we should file something to track these potential paths for enabling users of the Desktop, Android, and iOS editors to access the videos when they have access to the internet.

PR #4901, opened 2019-06-05
Use a "title card" instead of video if using Scratch Desktop

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/4895

### Proposed Changes

Pass in a `showVideos` flag that controls whether to show videos steps or instead show a "title card" consisting of the deck image and name. This is controlled by the `notScratchDesktop()` utility function from the container.

### Test Coverage

Added a `cards` unit test to test whether Video or Image steps are shown based on varying that flag.

Tested the scratchDesktop flag compatibility manually, by using the `isScratchDesktop=true` url param.

### DIFF
--- a/src/containers/cards.jsx
+++ b/src/containers/cards.jsx
@@ -19,6 +19,7 @@ import {
 
 import CardsComponent from '../components/cards/cards.jsx';
 import {loadImageData} from '../lib/libraries/decks/translate-image.js';
+import {notScratchDesktop} from '../lib/isScratchDesktop';
 
 class Cards extends React.Component {
     componentDidMount () {
@@ -52,7 +53,8 @@ const mapStateToProps = state => ({
     y: state.scratchGui.cards.y,
     isRtl: state.locales.isRtl,
     locale: state.locales.locale,
-    dragging: state.scratchGui.cards.dragging
+    dragging: state.scratchGui.cards.dragging,
+    showVideos: notScratchDesktop()
 });
 
 const mapDispatchToProps = dispatch => ({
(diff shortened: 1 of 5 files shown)


paulkaplan:
@thisandagain just a note, I added the title of the tutorial as well as the image because the cover image doesn't say the name and felt kind of strange without any text. That seem ok?

image

paulkaplan:
@thisandagain I talked with @kosiecki and found two other issues. One, we wanted to shrink the title image (which has a different aspect ratio) to fit so the first card matches the height of the other cards. We did that 
image

We also found that some tutorials _only_ have videos, and brainstormed what to do about that. The most straightforward is to filter those items out of the library for offline mode, because if you loaded them they would contain only the title card, with no steps. I did that, using similar logic to how we filter out the tutorial that requires a specific project to be loaded.

There is still an issue of being able to be linked from a tutorial to one with no steps in the offline editor, which I did not fix, since it felt like it was getting to be a bigger deal than initially expected. I'd like to get confirmation about these workarounds/ this implementation.

src/containers/cards.jsx
```diff
@@ -19,6 +19,7 @@ import {
 
 import CardsComponent from '../components/cards/cards.jsx';
 import {loadImageData} from '../lib/libraries/decks/translate-image.js';
+import {notScratchDesktop} from '../lib/isScratchDesktop';
```
cwillisf: Not for this PR, but do you think this should change from indicating Scratch Desktop specifically to some generic `isThis[Not]ABuildWhichShouldSupportRunningWithoutInternetAccess`? If so, any thoughts on a name that's fewer than 60 characters?

Kenny2github: `noNeedInternet` maybe? or `shouldHaveInternet` for the reverse?

Issue #4903, opened 2019-06-05
Inconsistently failing test in blocks.test.js


benjiwheeler:
In https://travis-ci.org/LLK/scratch-gui/builds/541871720 , ran into this test failure:

```
FAIL test/integration/blocks.test.js (57.708s)
  ● Working with the blocks › Adding a sound DOES NOT update the default sound name in the toolbox
    StaleElementReferenceError: stale element reference: element is not attached to the page document
      (Session info: headless chrome=75.0.3770.80)
      (Driver info: chromedriver=74.0.3729.6 (255758eccf3d244491b8a1317aa76e1ce10d57e9-refs/branch-heads/3729@{#29}),platform=Linux 4.4.0-101-generic x86_64)
      
      at Object.checkLegacyResponse (node_modules/selenium-webdriver/lib/error.js:546:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:509:13)
      at doSend.then.response (node_modules/selenium-webdriver/lib/http.js:441:30)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)
      From: Task: WebElement.isDisplayed()
      at thenableWebDriverProxy.schedule (node_modules/selenium-webdriver/lib/webdriver.js:807:17)
      at WebElement.schedule_ (node_modules/selenium-webdriver/lib/webdriver.js:2010:25)
      at WebElement.isDisplayed (node_modules/selenium-webdriver/lib/webdriver.js:2362:17)
      at test/helpers/selenium-helper.js:89:32
      at ManagedPromise.invokeCallback_ (node_modules/selenium-webdriver/lib/promise.js:1376:14)
      at TaskQueue.execute_ (node_modules/selenium-webdriver/lib/promise.js:3084:14)
      at TaskQueue.executeNext_ (node_modules/selenium-webdriver/lib/promise.js:3067:27)
      at asyncRun (node_modules/selenium-webdriver/lib/promise.js:2927:27)
      at node_modules/selenium-webdriver/lib/promise.js:668:7
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)
```

Retrying the build, this same test passed: https://travis-ci.org/LLK/scratch-gui/builds/541876228

benjiwheeler:
We are currently hoping that #4871 has improved things, and that this problem will not reoccur. But if necessary, we will disable this test.

Issue #4902, opened 2019-06-05
"Change categories" test in exampltes.test.js failing inconsistently


benjiwheeler:
In https://travis-ci.org/LLK/scratch-gui/builds/541819019 , we saw this failure:

```
FAIL test/integration/examples.test.js
  ● blocks example › Change categories
    StaleElementReferenceError: stale element reference: element is not attached to the page document
      (Session info: headless chrome=75.0.3770.80)
      (Driver info: chromedriver=74.0.3729.6 (255758eccf3d244491b8a1317aa76e1ce10d57e9-refs/branch-heads/3729@{#29}),platform=Linux 4.4.0-101-generic x86_64)
      
      at Object.checkLegacyResponse (node_modules/selenium-webdriver/lib/error.js:546:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:509:13)
      at doSend.then.response (node_modules/selenium-webdriver/lib/http.js:441:30)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)
      From: Task: WebElement.isDisplayed()
      at thenableWebDriverProxy.schedule (node_modules/selenium-webdriver/lib/webdriver.js:807:17)
      at WebElement.schedule_ (node_modules/selenium-webdriver/lib/webdriver.js:2010:25)
      at WebElement.isDisplayed (node_modules/selenium-webdriver/lib/webdriver.js:2362:17)
      at test/helpers/selenium-helper.js:89:32
      at ManagedPromise.invokeCallback_ (node_modules/selenium-webdriver/lib/promise.js:1376:14)
      at TaskQueue.execute_ (node_modules/selenium-webdriver/lib/promise.js:3084:14)
      at TaskQueue.executeNext_ (node_modules/selenium-webdriver/lib/promise.js:3067:27)
      at asyncRun (node_modules/selenium-webdriver/lib/promise.js:2927:27)
      at node_modules/selenium-webdriver/lib/promise.js:668:7
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)
```

Retrying the build, this same test passed: https://travis-ci.org/LLK/scratch-gui/builds/541876228

kchadha:
Seen again three builds in a row last night on gui develop:
https://travis-ci.org/LLK/scratch-gui/jobs/547545197#L4703
https://travis-ci.org/LLK/scratch-gui/builds/547555384#L4703
[...]

PR #4905, opened 2019-06-05
simpler findByXpath helper

benjiwheeler:
Resolves https://github.com/LLK/scratch-gui/issues/4906

The line https://github.com/LLK/scratch-gui/pull/4871/files#r290948638 may have been the cause of testing errors seen in #4902 and #4903.

Here, we remove part of the fix introduced in https://github.com/LLK/scratch-gui/pull/4871 , and use the simpler, older version of `findByXpath`.

### DIFF
--- a/test/helpers/selenium-helper.js
+++ b/test/helpers/selenium-helper.js
@@ -84,11 +84,7 @@ class SeleniumHelper {
     }
 
     findByXpath (xpath, timeoutMessage = `findByXpath timed out for path: ${xpath}`) {
-        return this.driver.wait(until.elementLocated(By.xpath(xpath)), DEFAULT_TIMEOUT_MILLISECONDS, timeoutMessage)
-            .then(el => (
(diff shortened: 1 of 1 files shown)


benjiwheeler:
Tried running Travis builds on the `simpler-find-xpath` branch that has this same code.

Try # 1: pass https://travis-ci.org/LLK/scratch-gui/builds/547793490?utm_source=github_status&utm_medium=notification

benjiwheeler:
Closing this for now because it's not clear that this improves things, though I'd still argue that we should merge it because we know #4871 was not strictly better than what we had before.

Note that I tested locally with this change side by side with develop (with only this change different) and my network artificially slowed down, and I got stale element errors reliably with develop and never saw them with this change.

benjiwheeler:
I revised the branch to split the change categories test up, and to try repeatedly getting the element so as to try to keep it from becoming stale.

Try #1: pass https://travis-ci.org/LLK/scratch-gui/builds/548735607
Try #2: building... https://travis-ci.org/LLK/scratch-gui/builds/548771986

PR #4909, opened 2019-06-07
Show libraries before loading their content

paulkaplan:
Resolves an issue where there was no feedback when clicking the "add a sprite/costume/sound/backdrop" buttons because the libraries are waiting to render all their content before showing up. Instead of that, show a spinner and then load the data. 

This is particularly noticeable on the costumes library, and particularly particularly noticeable on low-power devices. 

This also sets up future improvements where the library content (json, etc.) could be loaded async instead of bundled with the app, saving on package size.

library-spinner


rschamp:
Try it out: [progressive-libraries](https://llk.github.io/scratch-gui/progressive-libraries)

PR #4915, opened 2019-06-11
Add larger delete icon to make it easier to do on touch devices

### LINKED ISSUE #4911: Difficult to delete a sprite, costume, or sound on touch devices
### Expected Behavior

It should not take multiple tries to delete a sprite on a touch device.

### Actual Behavior

The 'delete' icon is so small that it is difficult to tap it with a finger on a touch device.

chrisgarrity:
### Resolves

- Resolves #4911

### Proposed Changes

Since it's using a new icon, don't repurpose the `CloseButton` and instead add a new `DeleteButton` with the new asset and styles.

### Reason for Changes
When using the gui on ChromeOS or Android touch devices we observed that it would take multiple tries to delete a sprite.

### Test Coverage

Updated the previous Delete tests to test the new component.

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


src/components/delete-button/delete-button.css
```diff
@@ -0,0 +1,48 @@
...
+    user-select: none;
+    cursor: pointer;
+    transition: all 0.15s ease-out;
+
+}
+
+/* .delete-button.large:hover {
+    transform: scale(1.1, 1.1);
+    box-shadow: 0 0 0 4px $ui-black-transparent;
+} */
```
rschamp: Cruft?

src/components/delete-button/delete-button.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+const DeleteButton = props => (
+    <div
+        aria-label="Delete"
+        className={classNames(
+            styles.deleteButton,
+            props.className
+        )}
+        role="button"
+        tabIndex="0"
```
rschamp: Eventually this should probably be a prop with a default value of `0` so it can be changed.  But for now I think if they're all the same, the browser does the right thing anyway.

chrisgarrity:
@rschamp I did the cleanup, but now there's a test failing on travis - times out converting to vector in the costumes test. However, it doesn't fail when I run it locally. Thoughts?

rschamp:
Let's try out the "merged into develop" version of the code locally and see if we can reproduce. And also rerun that build. It doesn't look completely unrelated to these changes even though it's not obvious why this particular action would fail.

Secret-chest:
However, what does that button do? Delete the trash bin? The icon should be either a trash bin or a X, not both.

PR #4921, opened 2019-06-17
Add VM init callback

paulkaplan:
Add a callback for when the VM is available so it can be configured by a VM consumer, similar to the way `storage` can be configured.

### DIFF
--- a/src/containers/gui.jsx
+++ b/src/containers/gui.jsx
@@ -52,6 +52,7 @@ class GUI extends React.Component {
         setIsScratchDesktop(this.props.isScratchDesktop);
         this.setReduxTitle(this.props.projectTitle);
         this.props.onStorageInit(storage);
+        this.props.onVmInit(this.props.vm);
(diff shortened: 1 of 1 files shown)


src/containers/gui.jsx
```diff
@@ -142,7 +145,8 @@ GUI.defaultProps = {
     isScratchDesktop: false,
     onStorageInit: storageInstance => storageInstance.addOfficialScratchWebStores(),
     onProjectLoaded: () => {},
-    onUpdateProjectId: () => {}
+    onUpdateProjectId: () => {},
+    onVmInit: (/* vm */) => {}
```
chrisgarrity: Is the parameter commented out to avoid unused variable errors? Or something else?

paulkaplan: Oh just as a kind of light form of documentation, to show that a param does get passed to that function, but it isn't used in that default function.

Issue #4927, opened 2019-06-18
On slow network, tutorial gifs lag behind the tutorial step you are on


paulkaplan:
### Expected Behavior

_Please describe what should happen_

The gif should always match the step title in tutorials.

### Actual Behavior

_Describe what actually happens_

With a slow network (throttled to "fast 3g" in the inspector) you see the next step title but the old gif.

PR #4932, opened 2019-06-19
Prevent previous tutorial images from hanging around on next/prev

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/4927

### Proposed Changes

Use the react `key` property on the image to make sure it does not recycle the image DOM node, instead creating a new one. 

### Reason for Changes

If we allow react to recycle the img node and just change its `src` property, the previous image will show until the new image is loaded. This makes it look like the image "lags behind" the tutorial step.

I needed to ensure a consistent min-height in order to prevent the card from collapsing each time the image is replaced.

### Test Coverage

Tested manually by using a slow connection and stepping through the cards.

### DIFF
--- a/src/components/cards/card.css
+++ b/src/components/cards/card.css
@@ -170,6 +170,9 @@
     align-items: center;
     position: relative;
     text-align: center;
+
+    /* Min height prevents layout changing when images change */
+    min-height: 256px;
 }
 
 .step-video {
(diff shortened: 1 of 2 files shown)


chrisgarrity:
Test failure was flaky change categories test. It should be fine to merge (or rebase and retest)

PR #4935, opened 2019-06-19
Allow vertical scrolling of sprite tile pane on touch devices

paulkaplan:
Resolves a long standing issue where you could not scroll vertically in the sprites box (or costumes/sounds lists) on touch devices. This fixes that problem by deciding whether to trigger a scroll based on both distance and angle. That is the same strategy used by the blocks to decide whether to scroll or pop a block out of the blocks palette. 

The first commit shows this work dropped into the existing mouse/touch handlers. 

The next commit refactored that into a `DragRecognizer`, so that I could add tests for it. 

The final commit fixes a related issue where the context menu would come up after trying to drag the tile. The drag recognizer makes it easy to prevent this, so I figured I should include that fix here.

/cc @carljbowman This actually feels pretty great! I'm surprised at how well it seems to work. Let's test it in person before merging this PR. It'll be up on a branch.

### DIFF
--- a/src/components/sprite-selector-item/sprite-selector-item.css
+++ b/src/components/sprite-selector-item/sprite-selector-item.css
@@ -20,7 +20,6 @@
     cursor: pointer;
 
     user-select: none;
-    touch-action: none;
 }
 
 .sprite-selector-item.is-selected {
--- a/src/components/sprite-selector/sprite-selector.css
+++ b/src/components/sprite-selector/sprite-selector.css
@@ -68,6 +68,8 @@
     padding-left: calc($space / 2);
     padding-right: calc($space / 2);
     padding-bottom: $space;
+
+    overflow-x: hidden;
 }
 
 .add-button {
--- a/src/lib/drag-recognizer.js
+++ b/src/lib/drag-recognizer.js
@@ -1,0 +1,125 @@
+import bindAll from 'lodash.bindall';
+import {getEventXY} from '../lib/touch-utils';
+
+class DragRecognizer {
+    /* Gesture states */
+    static get STATE_UNIDENTIFIED () {
+        return 'unidentified';
+    }
+    static get STATE_SCROLL () {
+        return 'scroll';
+    }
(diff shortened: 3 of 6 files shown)


src/components/sprite-selector-item/sprite-selector-item.css
```diff
@@ -20,7 +20,6 @@
     cursor: pointer;
 
     user-select: none;
-    touch-action: none;
```
paulkaplan: Removing this line because it was added to prevent scrolling by touching the tiles, which is the opposite of what we want now. Now the scroll prevention is handled dynamically by selectively preventing default on certain touch moves.

src/components/sprite-selector/sprite-selector.css
```diff
@@ -68,6 +68,8 @@
     padding-left: calc($space / 2);
     padding-right: calc($space / 2);
     padding-bottom: $space;
+
+    overflow-x: hidden;
```
paulkaplan: This is needed because on iOS you get a little wiggle of the sprite list when scrolling without it.

rschamp:
Try it out: [drag-angle](https://llk.github.io/scratch-gui/drag-angle)

src/lib/drag-recognizer.js
```diff
@@ -0,0 +1,125 @@
...
+    _unbindListeners () {
+        window.removeEventListener('mouseup', this._handleEnd);
+        window.removeEventListener('mousemove', this._handleMove);
+        window.removeEventListener('touchend', this._handleEnd);
+        window.removeEventListener('touchmove', this._handleMove, {passive: false});
+    }
+
+    _handleMove (event) {
+        // For gestures identified as vertical scrolls, do not process movement events
+        if (this._isScroll()) return;
```
chrisgarrity: defining `_isScroll` instead of just using `this._gestureState === DragRecognizer.STATE_SCROLL` seems like overkill. Is there a javascript 'convention' that recommends it?

PR #4948, opened 2019-06-24
Fix PropType on Loupe component

paulkaplan:
Updated linter caught this missing prop type! Hooray!

This was the goal of unpinning the linter deps, so yay it worked! See https://github.com/LLK/scratch-gui/pull/4786 for the history of that issue.


rschamp:
Try it out: [paulkaplan-patch-1](https://llk.github.io/scratch-gui/paulkaplan-patch-1)

PR #4960, opened 2019-06-28
Customize menubar

chrisgarrity:
### Resolves

In some uses of gui (e.g. ChromeOS/Android), the app will handle changing language and file management. So the menu should be customizable to show or hide the language and file menus along with allowing a gui consumer to provide an alternate logo.

### Proposed Changes

Adds `canChangeLanguage`, `canManageFiles` and `logo` props to the gui and menubar. The menus default to `true`, and the default scratch-logo is shown if no logo is provided.

### Test Coverage

Current tests pass. The current menu bar integration test will use the default config. Is there a good way to add a test for the options to hide the menus?

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -87,6 +87,7 @@ const GUIComponent = props => {
         isRtl,
         isShared,
         loading,
+        logo,
         renderLogin,
         onClickAccountNav,
         onCloseAccountNav,
@@ -215,6 +216,7 @@ const GUIComponent = props => {
                     className={styles.menuBarPosition}
                     enableCommunity={enableCommunity}
                     isShared={isShared}
+                    logo={logo}
                     renderLogin={renderLogin}
                     showComingSoon={showComingSoon}
                     onClickAccountNav={onClickAccountNav}
@@ -384,6 +386,7 @@ GUIComponent.propTypes = {
     isRtl: PropTypes.bool,
     isShared: PropTypes.bool,
     loading: PropTypes.bool,
+    logo: PropTypes.string,
     onActivateCostumesTab: PropTypes.func,
     onActivateSoundsTab: PropTypes.func,
     onActivateTab: PropTypes.func,
--- a/src/components/gui/gui.jsx
+++ b/src/components/gui/gui.jsx
@@ -66,8 +66,10 @@ const GUIComponent = props => {
         backpackVisible,
         blocksTabVisible,
         cardsVisible,
+        canChangeLanguage,
         canCreateNew,
         canEditTitle,
+        canManageFiles,
         canRemix,
         canSave,
         canCreateCopy,
@@ -202,9 +204,11 @@ const GUIComponent = props => {
                     authorId={authorId}
                     authorThumbnailUrl={authorThumbnailUrl}
                     authorUsername={authorUsername}
+                    canChangeLanguage={canChangeLanguage}
                     canCreateCopy={canCreateCopy}
                     canCreateNew={canCreateNew}
                     canEditTitle={canEditTitle}
+                    canManageFiles={canManageFiles}
                     canRemix={canRemix}
                     canSave={canSave}
                     canShare={canShare}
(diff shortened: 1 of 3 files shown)


src/components/gui/gui.jsx
```diff
@@ -66,8 +66,10 @@ const GUIComponent = props => {
         backpackVisible,
         blocksTabVisible,
         cardsVisible,
+        canChangeLanguage,
```
chrisgarrity: There are a lot of props that are all related to the menu bar. Would it make sense to refactor into a menuOptions object? E.g.:
```
menuOptions = {
    canChangeLanguage: true,
    canEditTitle: false,
    canManageFiles: true,
    canRemix: false,
    canShare: false,
    enableCommunity: false
}
```
Perhaps other options should also be in there (e.g. canCreateNew)

rschamp: I personally prefer the flat structure since it's less to keep in your head, and easier to define, override, etc.

src/components/menu-bar/menu-bar.jsx
```diff
@@ -327,11 +327,11 @@ class MenuBar extends React.Component {
                                     [styles.clickable]: typeof this.props.onClickLogo !== 'undefined'
                                 })}
                                 draggable={false}
-                                src={scratchLogo}
+                                src={this.props.logo ? this.props.logo : scratchLogo}
```
rschamp: This looks like `scratchLogo` should be the default of the `logo` prop.

src/components/menu-bar/menu-bar.jsx
```diff
@@ -345,82 +345,86 @@ class MenuBar extends React.Component {
...
-                                <MenuSection>
-                                    <SBFileUploader
-                                        canSave={this.props.canSave}
-                                        userOwnsProject={this.props.userOwnsProject}
-                                        onUpdateProjectTitle={this.props.onUpdateProjectTitle}
-                                    >
-                                        {(className, renderFileInput, loadProject) => (
+                                    {(this.props.canSave || this.props.canCreateCopy || this.props.canRemix) && (
+                                        <MenuSection>
+                                            {this.props.canSave ? (
```
rschamp: We're mixing the ternary style and the `&&` style for these optional components. It seems like the `&&` version is more straightforward, so maybe we should use that everywhere?

src/components/menu-bar/menu-bar.jsx
```diff
@@ -345,82 +345,86 @@ class MenuBar extends React.Component {
...
+                                                    {/* eslint-enable max-len */}
+                                                    {renderFileInput()}
+                                                </MenuItem>
+                                            )}
+                                        </SBFileUploader>
+                                        <SB3Downloader>{(className, downloadProjectCallback) => (
                                             <MenuItem
                                                 className={className}
-                                                onClick={loadProject}
+                                                onClick={this.handleSaveToComputer(downloadProjectCallback)}
```
rschamp: Just a note — I know you didn't name this. This method name is misleading since it looks like an error. I would expect `onClick="{this.handleSaveToComputer}`, i.e. set the `onClick` handler to that method. But instead this is a function that returns a handler, so the name should probably reflect that.

src/components/menu-bar/menu-bar.jsx
chrisgarrity: I switched to `&&` style for these optional components. There are other places in the code where we use the ternary operator, but some of them are nested within other ternary operators where we are deciding between two options (not optional). I decided not to mess with those.

paulkaplan:
@chrisgarrity looks like this PR didn't allow these options to be passed in from the container, so this can't be used externally

chrisgarrity:
@paulkaplan aargh - yes, forgot to add the props to the gui container and pass them through.

chrisgarrity:
@paulkaplan on second look, the GUI container passes everything not explicitly called out to the component as `{...componentProps}`, so just adding the props to `guiProps` in the android `index.jsx` will pass them through. (I have a version that does it) However, it adds a new svg file to the android webpack that requires the file-loader.

PR #4962, opened 2019-07-02
Correctly fail on non-200 responses when saving

paulkaplan:
This patches an issue where we are not correctly handling non-200 (i.e. failures) when saving assets and saving projects.

The assets part of this is also https://github.com/LLK/scratch-storage/issues/100, but we are able to detect it here by inspecting the JSON response, which should have `status: 'ok'` but instead has a different JSON body. I think it is an open question how storage should handle network responses like this, but we can at least prevent further damage this way.

Incidentally, we were doing the same thing when trying to save projects – ignoring non-200 responses. It is important to note that, per the `xhr` docs (and it is the same with fetch) that error responses do not trigger an `error` in the promise chain or an `error` param in the callback, that only happens when something prevents the request from being made. All "error-y" responses are still valid responses, and you need to check the response status code to be sure the request succeeded.

I built this as a hotfix branch in case we want to hotfix into www. But because hotfix/ builds to travis, nothing special needs to be done here to support that.

/cc @kchadha @BryceLTaylor

### DIFF
--- a/src/lib/save-project-to-server.js
+++ b/src/lib/save-project-to-server.js
@@ -45,6 +45,7 @@ export default function (projectId, vmState, params) {
     return new Promise((resolve, reject) => {
         xhr(opts, (err, response) => {
             if (err) return reject(err);
+            if (response.statusCode !== 200) return reject(response.statusCode);
             let body;
             try {
                 // Since we didn't set json: true, we have to parse manually
--- a/src/lib/project-saver-hoc.jsx
+++ b/src/lib/project-saver-hoc.jsx
@@ -234,9 +234,14 @@ const ProjectSaverHOC = function (WrappedComponent) {
                         asset.dataFormat,
                         asset.data,
                         asset.assetId
-                    ).then(
-                        () => (asset.clean = true)
-                    )
+                    ).then(response => {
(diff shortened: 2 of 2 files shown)


rschamp:
Try it out: [hotfix/2019-07-02](https://llk.github.io/scratch-gui/hotfix/2019-07-02)

benjiwheeler:
Reminder to set the milestone to July, if you want this reviewed this month! :)

src/lib/save-project-to-server.js
```diff
@@ -45,6 +45,7 @@ export default function (projectId, vmState, params) {
     return new Promise((resolve, reject) => {
         xhr(opts, (err, response) => {
             if (err) return reject(err);
+            if (response.statusCode !== 200) return reject(response.statusCode);
```
benjiwheeler: My only note is that I find `return reject()` confusing, since you're not actually doing anything (as I understand it) with the return value -- this is just a shorthand for
```
reject(response.statusCode);
return;
```

(see https://stackoverflow.com/questions/33445415/javascript-promises-reject-vs-throw#comment71729644_38181613 )

PR #4973, opened 2019-07-09
Add callback to update dynamic block on workspace.

kchadha:
### Resolves

Towards LLK/scratch-vm#2120

### Proposed Changes

Add a callback to update a dynamic extension block on the workspace. This callback gets passed into the vm so that the extension definition can control when a block gets updated (e.g. renaming a variable, editing a custom procedure, changing the shape of the control_stop block).

### Reason for Changes

This is a feature needed for #2120, especially to support the blocks mentioned above.

### Test Coverage

Added a unit test for the function which mocks scratch-blocks specific functionality.

### Related PRs
- Depends on LLK/scratch-vm#TBD

### DIFF
--- a/src/lib/define-dynamic-block.js
+++ b/src/lib/define-dynamic-block.js
@@ -177,6 +177,12 @@ const defineDynamicBlock = (ScratchBlocks, categoryInfo, staticBlockInfo, extend
                 return `%${++argCount}`;
             });
             this.interpolate_(scratchBlocksStyleText, args);
+        },
+        setBlockInfo: function (blockInfo) {
+            this.needsBlockInfoUpdate = true;
+            this.blockInfoText = JSON.stringify(blockInfo);
+            const mutation = this.mutationToDom();
+            this.domToMutation(mutation);
         }
     });
 };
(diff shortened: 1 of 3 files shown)


src/lib/define-dynamic-block.js
```diff
@@ -5,6 +5,15 @@ import BlockType from 'scratch-vm/src/extension-support/block-type';
 import ContextMenuContext from 'scratch-vm/src/extension-support/context-menu-context';
 import log from './log.js';
 
+const updateBlockOnWorkspace = (ScratchBlocks, workspace, blockId, blockInfo) => {
+    const block = workspace.getBlockById(blockId);
+    // TODO this seems hacky but is needed for how defineDynamicBlock currently works.
+    block.needsBlockInfoUpdate = true;
+    block.blockInfoText = JSON.stringify(blockInfo);
+    const mutation = block.mutationToDom();
+    block.domToMutation(mutation);
```
cwillisf: I think I'd prefer these 4 lines to be in a method on the block so that we can treat `blockInfoText` as somewhat private to the block. I believe it would work to just add another method to the object returned by `defineDynamicBlock`.

PR #4982, opened 2019-07-11
Only call Renderer.pick() on draggable targets in Stage.onStartDrag

### LINKED ISSUE #5781: Clicking and dragging on the stage could be expensive when the pen is used
### Expected Behavior

Clicking and dragging should not cause any particularly expensive function calls.

### Actual Behavior

Every time you click and drag on the stage, it ends up calling `RenderWebGL.pick`, which can in some cases be expensive.

When `RenderWebGL.pick` is called and the pen layer has been updated, the renderer must read all pixels from the pen layer so that it can test whether they are being touched.

In projects with both click-and-drag and pen usage, for instance projects which let you draw with the pen by clicking and dragging (updating the pen layer each frame when you drag), this could negatively impact performance.

In my rough testing, I discovered that reading pen layer pixels takes 1ms on my PC, out of the 33ms per frame we have, but this was on relatively high-end hardware and it will likely take longer on something like a Chromebook.

### Steps to Reproduce

A quick demo of the type of project this affects is as such:
1. Add the pen extension
2. Create the following script:
image
    - This will allow you to draw pen lines on the stage.
3. Using the profiler in your browser's dev tools, monitor performance of the project when drawing pen lines by dragging
4. Look for the amount of time used by calls to `updateSilhouette` in `onStartDrag`:
image

All

adroitwhiz:
Resolves https://github.com/LLK/scratch-gui/issues/5776
Resolves #5781

### Proposed Changes

This PR changes the `Stage` component's `onStartDrag` handler to only `Renderer.pick` drawables that are currently draggable, instead of `pick`ing all drawables then checking if each is draggable as it currently does.

### Reason for Changes

Currently, `Renderer.pick()` is called every time you click and drag the mouse on the stage. Projects which make use of a "dragging" action do this a lot.

`Renderer.pick()` can be expensive, especially if they trigger a silhouette update on a pen skin (which calls `gl.readPixels` and massively impacts performance), and `onStartDrag` seems to be called every frame the mouse is down. By only `pick`ing draggable targets, performance can be improved, especially for mouse-controlled projects which make use of the pen.

The last time I benchmarked this, it saved 5ms out of the 33ms frame budget, but since https://github.com/LLK/scratch-render/pull/475 went in, this may have changed. I'll try to get some better numbers now.

Note: this may *appear* to cause a performance regression on Edge (large amounts of lag appear when the mouse is being dragged), but that seems to happen in dev-server builds (and only dev-server builds) with and without this patch.

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [x] Edge

Linux
* [x] Chrome
* [x] Firefox
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/stage.jsx
+++ b/src/containers/stage.jsx
@@ -342,15 +342,23 @@ class Stage extends React.Component {
     }
     onStartDrag (x, y) {
         if (this.state.dragId) return;
-        const drawableId = this.renderer.pick(x, y);
+
+        // Because pick queries are expensive, only perform them for drawables that are currently draggable.
+        let draggableTargets = this.props.vm.runtime.targets;
+        if (!this.props.useEditorDragStyle) {
+            draggableTargets = draggableTargets.filter(
+                target => Number.isFinite(target.drawableID) && target.draggable
+            );
+        }
+        if (draggableTargets.length === 0) return;
+
+        const draggableIDs = draggableTargets.map(target => target.drawableID);
+        const drawableId = this.renderer.pick(x, y, 1, 1, draggableIDs);
         if (drawableId === null) return;
         const targetId = this.props.vm.getTargetIdForDrawableId(drawableId);
         if (targetId === null) return;
 
         const target = this.props.vm.runtime.getTargetById(targetId);
-
-        // Do not start drag unless in editor drag mode or target is draggable
-        if (!(this.props.useEditorDragStyle || target.draggable)) return;
 
         // Dragging always brings the target to the front
         target.goToFront();
--- a/src/containers/stage.jsx
+++ b/src/containers/stage.jsx
@@ -343,11 +343,16 @@ class Stage extends React.Component {
     onStartDrag (x, y) {
         if (this.state.dragId) return;
 
-        // Because pick queries are expensive, only perform them for drawables that are currently draggable.
-        let draggableTargets = this.props.vm.runtime.targets;
+        // Targets with no attached drawable cannot be dragged.
+        let draggableTargets = this.props.vm.runtime.targets.filter(
+            target => Number.isFinite(target.drawableID)
+        );
+
+        // Because pick queries can be expensive, only perform them for drawables that are currently draggable.
+        // If we're in the editor, we can drag all targets. Otherwise, filter.
         if (!this.props.useEditorDragStyle) {
             draggableTargets = draggableTargets.filter(
-                target => Number.isFinite(target.drawableID) && target.draggable
+                target => target.draggable
             );
         }
         if (draggableTargets.length === 0) return;


paulkaplan:
@adroitwhiz can you give a project ID of a project where the difference here would be especially noticeable? I'm having trouble creating a repro.

adroitwhiz:
> @adroitwhiz can you give a project ID of a project where the difference here would be especially noticeable? I'm having trouble creating a repro.

In [11489772](https://scratch.mit.edu/projects/11489772/), which uses the pen to draw a "blade swipe" effect when the mouse is held down, `updateSilhouette` calls triggered by `pick` take up ~5ms of the 33ms frame budget:

Firefox | Chrome
-|-
image|image

benjiwheeler:
@paulkaplan @adroitwhiz does this one need attention?

benjiwheeler:
@adroitwhiz, @chrisgarrity and I were checking this out. (I also rebased it on develop.) 

We found it difficult to detect a performance difference to the naked eye using that Pen Benchmark project, but we did see that essentially all the silhouette update time vanished when we used this code.

What are your current thoughts on this? We want to see it go in if it's still relevant -- and, we weren't sure if you think it still is.

adroitwhiz:
The performance difference isn't apparent on my test project, but in real projects where more of the "frame budget" is spent, and especially on lower-end hardware, this will likely have a tangible effect. If you have a Chromebook at your disposal, maybe see how long the silhouette updates take there?

cwillisf:
Hmm... will this let you pick up a draggable sprite that's hidden on the bottom of a stack of non-draggable sprites? That might not be what we want...

benjiwheeler:
This PR is in a bit of limbo -- it's proven difficult to see the difference it makes in practice. However, if @cwillisf's question can be addressed, it seems like this would probably be an improvement overall. It's unclear what the assignment and milestone should be.

fsih:
@benjiwheeler I tested and it doesn't seem to allow that, it's working as it should

benjiwheeler:
After discussing with @ray and @chrisgarrity, I'm deciding to merge this. I know it isn’t a strictly better improvement, but it seems likely to be a tiny performance improvement, and we need those! I think we checked off the major concerns.

benjiwheeler:
In the bug hunt testing, we realized that this does indeed let you pick up a draggable sprite that's hidden on the bottom of a stack of non-draggable sprites; contrast the "Purple cats are a drag" in these studios: 

* https://scratch.mit.edu/studios/28480703/
* https://scratch.ly/studios/10003668/

Reverting.

PR #4983, opened 2019-07-11
Move browsers options to .browserslistrc

rschamp:
Apparently this is the recommendation for loaders that look at the `browsers` option.

I saw this warning while trying to re-enable Greenkeeper.

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,3 @@
+last 3 versions
+Safari >= 8
+iOS >= 8
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -55,10 +55,7 @@ const base = {
                     ['react-intl', {
                         messagesDir: './translations/messages/'
                     }]],
-                presets: [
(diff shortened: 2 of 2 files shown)


paulkaplan:
Bizarre, but ok... It might be worth checking if there is any change to the build output when you make this change? It might be the easiest way to test this change too.

rschamp:
Good point — I just diffed the output and there's no change.  I didn't put this in the description, but this resolves the warnings that look like this when you build:
```
  Replace Autoprefixer browsers option to Browserslist config.
  Use browserslist key in package.json or .browserslistrc file.

  Using browsers option cause some error. Browserslist config
  can be used for Babel, Autoprefixer, postcss-normalize and other tools.

  If you really need to use option, rename it to overrideBrowserslist.

  Learn more at:
  https://github.com/browserslist/browserslist#readme
  https://twitter.com/browserslist
```

rschamp:
Also please remember to actually use the "approve" feature — non-admins don't have the ability to merge unless you use that.

PR #5021, opened 2019-07-24
Support for variables extension - Variable Renaming Context Menu

kchadha:
### Resolves

Towards LLK/scratch-vm#2120

### Proposed Changes

Adds support for the variable renaming context menu on a variable reporter block.
- fills in `RENAME_A_VARIABLE` built-in context menu callback, which opens the appropriate rename variable modal and passes back information to the VM extension callback for the context menu item.
- Updates the code in domToMutation which dynamically updates the layout of the block. Ensures that that shadow blocks (and any other blocks plugged into inputs) on updated dynamic blocks stay connected after layout update.
- Adds support for number arguments on dynamic extension blocks (this is for list blocks which use a `math_number` shadow block.
- Adds support for non-droppable menus in dynamic extension blocks

### Related PRs
- This PR depends on LLK/scratch-vm#2242

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -417,7 +417,7 @@ class Blocks extends React.Component {
                     // Anything else will be picked up from the XML attached to the block instance.
                     const extendedOpcode = `${categoryInfo.id}_${blockInfo.info.opcode}`;
                     const blockDefinition =
-                        defineDynamicBlock(this.ScratchBlocks, categoryInfo, blockInfo, extendedOpcode);
+                        defineDynamicBlock(this, this.ScratchBlocks, categoryInfo, blockInfo, extendedOpcode);
                     this.ScratchBlocks.Blocks[extendedOpcode] = blockDefinition;
                 });
             }
--- a/src/lib/define-dynamic-block.js
+++ b/src/lib/define-dynamic-block.js
@@ -116,6 +116,8 @@ const defineDynamicBlock = (guiContext, ScratchBlocks, categoryInfo, staticBlock
             this.blockInfoText = '{}';
             // we need a block info update (through `domToMutation`) before we have a completely initialized block
             this.needsBlockInfoUpdate = true;
+            // Keep track of menu info for use in laying out the block and its menus
+            this.menus = categoryInfo.convertedMenuInfo;
         },
         mutationToDom: function () {
             const container = document.createElement('mutation');
@@ -172,7 +174,12 @@ const defineDynamicBlock = (guiContext, ScratchBlocks, categoryInfo, staticBlock
                 const arg = blockInfo.arguments[argName];
                 switch (arg.type) {
                 case ArgumentType.STRING:
-                    args.push({type: 'input_value', name: argName});
+                    if (arg.menu && !this.menus[arg.menu].acceptReporters) {
+                        const options = this.menus[arg.menu].items;
+                        args.push({type: 'field_dropdown', name: argName, options: options});
+                    } else {
+                        args.push({type: 'input_value', name: argName});
+                    }
                     break;
                 case ArgumentType.BOOLEAN:
                     args.push({type: 'input_value', name: argName, check: 'Boolean'});
--- a/src/lib/define-dynamic-block.js
+++ b/src/lib/define-dynamic-block.js
@@ -22,6 +22,10 @@ const setupCustomContextMenu = (guiContext, ScratchBlocks, contextMenuInfo, exte
                     text: contextOption.text,
                     callback: () => {
                         const blockInfo = JSON.parse(this.blockInfoText);
+                        const contextOptionCallbackData = {
+                            blockInfo: blockInfo,
+                            blockId: this.id
+                        };
                         if (contextOption.builtInCallback) {
                             switch (contextOption.builtInCallback) {
                             case 'EDIT_A_PROCEDURE':
@@ -40,9 +44,13 @@ const setupCustomContextMenu = (guiContext, ScratchBlocks, contextMenuInfo, exte
                                 // strings to come from scratch-vm instead
                                 const varType = blockInfo.opcode === 'variable' ? '' : 'list';
 
-                                const renameCallback = newName => contextOption.callback({
-                                    blockInfo: blockInfo, newName: newName, varType: varType
-                                });
+                                const renameCallback = newName => {
+                                    // Pass in additional information about the variable renaming to the
+                                    // extension callback
+                                    contextOptionCallbackData.newName = newName;
+                                    contextOptionCallbackData.varType = varType;
+                                    contextOption.callback(contextOptionCallbackData);
+                                };
 
                                 // Use scratch-blocks to open the variable rename modal
                                 // TODO change this to open the variable prompt directly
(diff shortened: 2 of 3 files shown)


src/containers/blocks.jsx
```diff
@@ -417,7 +417,7 @@ class Blocks extends React.Component {
                     // Anything else will be picked up from the XML attached to the block instance.
                     const extendedOpcode = `${categoryInfo.id}_${blockInfo.info.opcode}`;
                     const blockDefinition =
-                        defineDynamicBlock(this.ScratchBlocks, categoryInfo, blockInfo, extendedOpcode);
+                        defineDynamicBlock(this, this.ScratchBlocks, categoryInfo, blockInfo, extendedOpcode);
```
kchadha: Need to pass in the whole gui container so that the built in `RENAME_A_VARIABLE` callback (and when we add support for editing custom procedures, the built in `EDIT_A_PROCEDURE` callback) can open the appropriate modal.

src/lib/define-dynamic-block.js
```diff
@@ -168,6 +203,14 @@ const defineDynamicBlock = (ScratchBlocks, categoryInfo, staticBlockInfo, extend
                 const arg = blockInfo.arguments[argName];
                 switch (arg.type) {
                 case ArgumentType.STRING:
+                    if (arg.menu && !this.menus[arg.menu].acceptReporters) {
+                        const options = this.menus[arg.menu].items;
+                        args.push({type: 'field_dropdown', name: argName, options: options});
```
kchadha: Add support for non-droppable menus in dynamic extension blocks. All variable blocks/list blocks are now dynamic, so that we can dynamically update the block instance when a variable is renamed.

src/lib/define-dynamic-block.js
```diff
@@ -168,6 +203,14 @@ const defineDynamicBlock = (ScratchBlocks, categoryInfo, staticBlockInfo, extend
...
+                    } else {
+                        args.push({type: 'input_value', name: argName});
+                    }
+                    break;
+                case ArgumentType.NUMBER:
```
kchadha: Add support for `ArgumentType.NUMBER` which is used in some list blocks.

src/lib/define-dynamic-block.js
```diff
@@ -176,14 +219,64 @@ const defineDynamicBlock = (ScratchBlocks, categoryInfo, staticBlockInfo, extend
                 }
                 return `%${++argCount}`;
             });
+            const wasRendered = this.rendered;
+            this.rendered = false;
+            const connectionMap = this.disconnectOldBlocks_();
```
kchadha: Keep track of blocks (and shadows) that were previously connected to this block before we start dynamically updating its layout, so that we can reconnect them after we're done.

src/lib/define-dynamic-block.js
```diff
@@ -176,14 +219,64 @@ const defineDynamicBlock = (ScratchBlocks, categoryInfo, staticBlockInfo, extend
...
         setBlockInfo: function (blockInfo) {
             this.needsBlockInfoUpdate = true;
             this.blockInfoText = JSON.stringify(blockInfo);
             const mutation = this.mutationToDom();
             this.domToMutation(mutation);
-        }
+        },
+        // Extra functions for dynamically updating the layout of a block and reconnecting
+        // inputs and shadow blocks.
+        removeAllInputs_: ScratchBlocks.ScratchBlocks.ProcedureUtils.removeAllInputs_,
```
kchadha: It's not great that this and the following functions are being borrowed from ScratchBlocks.ProcedureUtils, but maybe we can move those scratch-blocks functions into a more generic util post-extensionification.

src/lib/define-dynamic-block.js
```diff
@@ -21,17 +21,49 @@ const setupCustomContextMenu = (ScratchBlocks, contextMenuInfo, extendedOpcode)
...
+                            case 'RENAME_A_VARIABLE': {
+                                // Extract variable info from this block
+                                const varName = blockInfo.text;
+                                // TODO it's not great that the following is hard coded...
+                                // We could use ScratchBlocks.SCALAR_VARIABLE_TYPE and
+                                // ScratchBlocks.LIST_VARIABLE_TYPE here instead, but
+                                // given that we eventually want to decouple this code from
+                                // using ScratchBlocks variables, we ultimately want these
+                                // strings to come from scratch-vm instead
+                                const varType = blockInfo.opcode === 'variable' ? '' : 'list';
```
cwillisf: Yep - I agree with your comment. I have yet to hear about a good way to do this in JS though it could be added to `scratch-vm`'s main export list. That's what I've been doing to access `scratch-gui` internals from `scratch-desktop`, though I've been trying to minimize that.

src/lib/define-dynamic-block.js
cwillisf: 🎉

To celebrate I'ma make a block where the menu flips between droppable and non-droppable based on which item is selected in the menu! ;)

src/lib/define-dynamic-block.js
cwillisf: Does `connectionMap` also contain the information destroyed by `removeAllInputs_()`? Or is that what the stuff after `interpolate_()` is doing?

src/lib/define-dynamic-block.js
```diff
@@ -176,14 +219,64 @@ const defineDynamicBlock = (ScratchBlocks, categoryInfo, staticBlockInfo, extend
...
+                        type = 'b';
+                    } else {
+                        type = 's';
+                    }
+                    if (inputIndex < this.inputList.length) {
+                        const input = this.inputList[inputIndex];
+                        this.populateArgument_(type, inputIndex, connectionMap, input.name, input);
+                    }
+                }
+            });
```
cwillisf: This seems tied to `scratch-blocks` internals even tighter than the rest of the code in this file. If it's necessary to do it this way then c'est la vie but if there's a way to use a scratch-blocks utility for it that would make me sleep better. I imagine you looked, though...

src/lib/define-dynamic-block.js
cwillisf: I agree that it looks a bit... not-great... but if all these methods expect to be called with the block as `this` then your current approach might be the best option. If that's the case then I'm inclined to just call this a consequence of the scratch-blocks API.

src/lib/define-dynamic-block.js
kchadha: `removeAllInputs_` actually just clears all the input holes from the block. It has nothing to do with previously connected shadows etc. (`disconnectOldBlocks_` takes care of that)

src/lib/define-dynamic-block.js
kchadha: The scratch-blocks utilities in question are from the custom procedure code. All of these functions are custom to the custom procedure code, but the good news is that we can make them general utilities in scratch-blocks post-extensionification maybe?

kchadha: This is sort of why we temporarily put extensionifying custom procedures on the back burner...

src/lib/define-dynamic-block.js
kchadha: > If that's the case then I'm inclined to just call this a consequence of the scratch-blocks API.

I think I'm resigned to this statement. These are the woes of dynamically updating a block...

src/lib/define-dynamic-block.js
cwillisf: Ah, that makes sense.

PR #5057, opened 2019-08-01
Add Jest-specific linting rules

rschamp:
This is follow up from https://github.com/LLK/scratch-gui/pull/5049, so that kind of mistake will cause linting to fail.

Adding this caught a couple of important things:
- The sound integration tests had a `.only` in them on the version of the code I had at the time
- The project-saver-hoc tests had two tests with the same description, causing possible confusion if that test broke

### DIFF
--- a/test/helpers/selenium-helper.js
+++ b/test/helpers/selenium-helper.js
@@ -1,4 +1,4 @@
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 30000; // eslint-disable-line no-undef
+jest.setTimeout(30000); // eslint-disable-line no-undef
 
 import bindAll from 'lodash.bindall';
 import 'chromedriver'; // register path
--- a/test/unit/containers/slider-prompt.test.jsx
+++ b/test/unit/containers/slider-prompt.test.jsx
@@ -55,7 +55,7 @@ describe('Slider Prompt Container', () => {
         const componentProps = wrapper.find(SliderPromptComponent).props();
         componentProps.onChangeMin({target: {value: '1.0'}});
         componentProps.onOk();
-        expect(onOk).toBeCalledWith(1, 100, false);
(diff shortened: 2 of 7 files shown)


test/helpers/selenium-helper.js
```diff
@@ -1,4 +1,4 @@
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 30000; // eslint-disable-line no-undef
+jest.setTimeout(30000); // eslint-disable-line no-undef
```
rschamp: Confirmed that these are equivalent via https://jestjs.io/docs/en/jest-object.html#jestsettimeouttimeout and https://github.com/facebook/jest/issues/652

test/unit/containers/slider-prompt.test.jsx
```diff
@@ -55,7 +55,7 @@ describe('Slider Prompt Container', () => {
         const componentProps = wrapper.find(SliderPromptComponent).props();
         componentProps.onChangeMin({target: {value: '1.0'}});
         componentProps.onOk();
-        expect(onOk).toBeCalledWith(1, 100, false);
+        expect(onOk).toHaveBeenCalledWith(1, 100, false);
```
rschamp: Would not be opposed to disabling this rule if we think it's annoying, it's just part of the recommended set. These changes were made just by using `--fix`, so pretty easy to deal with.

test/unit/util/project-saver-hoc.test.jsx
```diff
@@ -423,7 +423,7 @@ describe('projectSaverHOC', () => {
         expect(mockedOnRemixing).toHaveBeenCalledWith(true);
     });
 
-    test('when starting to remix, onRemixing should be called with param true', () => {
+    test('when starting to remix, onRemixing should be called with param false', () => {
```
rschamp: The test above this one has the same name, so the linter caught this copy-paste error

PR #5071, opened 2019-08-06
Add webpack resolver to eslintrc so linter doesn't give false positive noUnresolved errors

fsih:
Developer ergonomics fix


src/.eslintrc.js
```diff
@@ -20,6 +20,18 @@ module.exports = {
     settings: {
         react: {
             version: '16.2' // Prevent 16.3 lifecycle method errors
+        },
+        'import/resolver': {
```
rschamp: I think we want this to use the default webpack config, (`'import/resolver': 'webpack',`) so when we update it, the linter will notice those updates. Also makes this config less scary looking :)

fsih: The syntax `'import/resolver': 'webpack'` is failing for me. I get `"TypeError: Cannot read property 'externals' of undefined"` when linting.
The syntax
```
    'import/resolver': {
        'webpack': {
            'config': './webpack.config.js'
        }
    }
```
also isn't working for me, I'm getting `Error resolving webpackConfig { Error: Cannot find module '/Users/dd/Git/scratch-gui/node_modules/react-intl/webpack.config.js'`

rschamp: It looks like the default behavior when you just use `'import/resolver': 'webpack'` is to look for the closest webpack config to the file doing the requiring. This is causing issues because the scratch-paint config is returning an array rather than an object from its webpack config, which is an issue that was supposedly fixed (https://github.com/benmosher/eslint-plugin-import/issues/1219), although apparently not as you can see here.

In any case, the "use the closest webpack config to the file" behavior isn't how webpack works, so I think it would be incorrect to use that for linting. So the second syntax you have is how to go.  Again, unfortunately this plugin seems to be bizarrely literal about relative paths, giving you the second error. To fix, use this to get an absolute path, like we do occasionally in `webpack.config.js` itself!
```javascript
const path = require('path');
// ...
        'import/resolver': {
            webpack: {
                config: path.resolve(__dirname, '../webpack.config.js')
            }
        }
```

fsih: Thank you for doing this digging for me! I updated eslint-plugin-import to double check, and the 'import/resolver': 'webpack' syntax still didn't work. I filed https://github.com/benmosher/eslint-plugin-import/issues/1448

Issue #2755, opened 2018-08-01
Drag cursor (and delete cursor) disappears from block after starting drag


kchadha:
### Expected Behavior

Drag cursor should remain on block while dragging.
Delete cursor should appear on block when dragging over the toolbox.

### Actual Behavior

Neither of the two happen. This regression was introduced by #2584.

#### Behavior before regression:

![block_cursor](https://user-images.githubusercontent.com/1786240/43530046-9dd54c5a-957a-11e8-8bec-c13bac82ac01.gif)

#### Current Behavior:

![block_cursor_regression](https://user-images.githubusercontent.com/1786240/43530053-a12a70d8-957a-11e8-9f8b-fab0a6ef759e.gif)

epicfaace:
@kchadha Undoing the change in #2584 fixes this; what problem, exactly, was #2584 supposed to fix?

kchadha:
@epicfaace, #2584 fixed an issue where you couldn't share blocks to a sprite twice.

To Repro (without the changes in #2584):
1. Open a new 3.0 project
2. Create another sprite
3. Drag out a block from the toolbox into Sprite 1
4. Drag that block from Sprite1 into Sprite2
6. Without scrolling the workspace at all, try to drag that same block again from Sprite1 to Sprite2 again
7. Observe that Sprite 2 highlights, but its sprite tile does not wiggle (indicating that blocks can be shared to it)
8. Switch to Sprite 2, observe that it only has one instance of that shared block

Here is a gif:

![share-blocks-bug](https://user-images.githubusercontent.com/1786240/50351613-c1c46500-0510-11e9-9de4-248297707d66.gif)

epicfaace:
@kchadha , #4184 should fix both issues #2584 and this one.

adroitwhiz:
This is happening again because #5078 was reverted

Issue #1307, opened 2018-01-18
Connecting duplicated blocks can result in the stack being executed


thisandagain:
![untitled3](https://user-images.githubusercontent.com/747641/35123568-d2730176-fc70-11e7-8575-e20375b97982.gif)

paulkaplan:
This appears to be due to the pointer-events: none; that GUI applies to the `.blocklyBlockDragSurface`, but I'm not exactly sure why, still investigating. Removing that line fixes the problem, but there is a comment about why that line is needed. https://github.com/LLK/scratch-gui/blob/develop/src/components/blocks/blocks.css#L83

/ht @picklesrus

ErikMejerHansen:
Possible duplicate of: https://github.com/LLK/scratch-blocks/issues/1935

paulkaplan:
@ErikMejerHansen yup can confirm this is the same issue. The issue does not appear in the blocks playground, it is actually a problem introduced by the GUI CSS, so I'm going to close that one as a duplicate of this.

adroitwhiz:
This is happening again because #5078 was reverted

griffpatch:
As you probably are all aware, but I just wanted to state it. The execution
occurs because you click to drop the duplicated block stack. If you are
attaching to the bottom of another stack then the scripts are not executed,
but if you drop between other blocks in a block stack then the stack is
executed. This is because your cursor is 'over' the block stack when you
click to place the duplicated stack, and are therefore clicking the stack
behind too. If we could just cancel the event bubbling it might all go away?

PR #5078, opened 2019-08-08
Fix duplicate running blocks by removing pointer-events: none

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/1307
- Fixes https://github.com/LLK/scratch-gui/issues/2755

### Proposed Changes

Instead of using `pointer-events: none` on the blocks drag layer, set its width/height to 1 (setting to zero makes it disappear) and allow overflow. This allows hover events to fall through (used for sharing blocks to other sprites) but keep the click behavior where the drag layer handles the click on duplicate and drag. 

### Reason for Changes

This is hard to explain, but here goes:
- After duplicating blocks, you then need to click to drop the blocks.
- If you click on a stack to insert the duplicated blocks into a stack, it should attach the blocks but not cause the blocks to run.
- Usually this is handled by the drag surface handling the block click, which cancels the gesture and allows the blocks to be dropped.
- By setting `pointer-events: none` on the drag surface (to fix issues with hovering over sprite tiles to share them), we inadvertently changed the drop behavior so that the workspace handles the click instead of the drag surface.
- A few things are possible: your click could land on the workspace, or on a single block, in both cases nothing interesting happens, the stack is dropped/attached without running any blocks
- If you click in the middle/end of a stack, due to a strange quirk of the way block mousedown events are bound, a mousedown is triggered on _every_ block in that stack, instead of just the one you are directly over.
- Multiple block mousedowns cause a weird issue where the gesture is cancelled correctly, but then successive block mousedowns cause a new gesture to start.
- If a gesture is started, it will eventually be ended by mouseup, at which point the stackclick event is triggered and the blocks are run.

/ht to @kchadha and @picklesrus for helping to find this issue.

I'd like to follow up on a couple issues on the scratch-blocks side. I think that we may want to change the way mousedowns are bound so that the targets are not nested in a stack, this causes way to many mousedown events in large stacks that is probably a big performance problem. Additionally, we could remove the event binding from the drag layer and officially set pointer-events to none on the scratch-blocks side, which would allow us to remove this kludge.

### DIFF
--- a/src/components/blocks/blocks.css
+++ b/src/components/blocks/blocks.css
@@ -79,8 +79,13 @@
     /*
         Fix an issue where the drag surface was preventing hover events for sharing blocks.
         This does not prevent user interaction on the blocks themselves.
+        This was previously pointer-events: none, but that allowed clicks to fall through
+        causing incorrect stack clicks on duplicate, and prevented the correct
+        cursor from showing when dragging over the toolbox to delete.
     */
-    pointer-events: none;
+    width: 1px;
+    height: 1px;
+    overflow: visible;
     z-index: $z-index-drag-layer; /* make blocks match gui drag layer */
 }
 


rschamp:
Try it out: [fix-duplicate-run](https://llk.github.io/scratch-gui/fix-duplicate-run)

src/components/blocks/blocks.css
```diff
@@ -79,8 +79,13 @@
     /*
         Fix an issue where the drag surface was preventing hover events for sharing blocks.
         This does not prevent user interaction on the blocks themselves.
+        This was previously pointer-events: none, but that allowed clicks to fall through
+        causing incorrect stack clicks on duplicate, and prevented the correct
+        cursor from showing when dragging over the toolbox to delete.
     */
-    pointer-events: none;
+    width: 1px;
```
kchadha: 🤣

picklesrus:
LGTM 👍😰

picklesrus:
So, sadly, this seems to break Chrome on windows. 😢 I think we should revert.

paulkaplan:
@picklesrus thanks for spotting this early. Welp, back to the drawing board...

Issue #5089, opened 2019-08-13
GUI hangs when uploading a .sb3/.zip instead of .sprite3 file for sprite upload


paulkaplan:
### Expected Behavior

_Please describe what should happen_

When you upload a file that is not a sprite3 after clicking the sprite upload button, you should see a message about it failing, and you should be able to go back to what you were doing.

### Actual Behavior

_Describe what actually happens_

It leaves the "importing" spinner open forever.

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_

In Mac/Chrome, after clicking the sprite file upload button, click the "Options" button at the bottom left of the file menu and change "Custom Files" to "All Files". Then choose a different zip file (e.g. .zip or .sb3). The "importing" spinner never goes away. 

Note if I try to import a non-zip file (e.g. a .csv) I don't get the spinner, and nothing bad happens, but I also do not get any kind of alert about it failing. Better than infinite spinner I guess.

### Operating System and Browser

_e.g. Mac OS 10.11.6 Safari 10.0_

Mac/Chrome

---

This may seem like the user is doing a strange thing, but on some platforms you can't stop people (or even forcefully suggest like with Chrome) that you should use only certain file extensions.

paulkaplan:
@carljbowman we do not have an error state for when you import a file that could not be parsed, like here. Right now we are just closing the "importing..." alert and doing nothing. Should we show a message like "Could not import that file"?

PR #5090, opened 2019-08-13
Show an error when trying to import an invalid sprite file

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/5089

### Proposed Changes

1. Add a specific "failed to import" error alert, and use it instead of just closing the importing alert.
2. Add a `.catch` to the sprite import promise. We were only catching errors from the file reading process itself before, but this way we can also catch issues from parsing the file as a sprite.

### Test Coverage

Tested manually using the upload steps from https://github.com/LLK/scratch-gui/issues/5089

If there is copy/useablility concerns with the new alert error, I can also resubmit this just fixing (2) and making it just close the import without any other feedback, the existing behavior for when the file reading fails.


paulkaplan:
Going to close this pending discussion. This PR should be updated to show errors for all types of uploads, not just sprites.

PR #5092, opened 2019-08-13
Close importing after failing to import

### LINKED ISSUE #4639: Importing unsupported files result in forever "Importing..."
### Expected Behavior
It should reject unsupported files.

### Actual Behavior
On Edge, the file chooser ignores ".mp3, .wav" rule so anything can be chosen.

Note from @ericrosenbaum : we have also found that you can get the bug on windows chrome by changing the file type in the file type selector to "all files" and then selecting a file that is not a sound. update! there's also a way to get it to happen on mac chrome.

### Steps to Reproduce
Scratcher mii-10 reported that importing flac on Edge showed Importing... popup.
https://youtu.be/vBbvmG_ylMw

Edge/Windows 10

paulkaplan:
### Resolves

This is a partial fix for the issue where importing non-sprite zips as sprites causes the import spinner to show up forever. I broke this out from https://github.com/LLK/scratch-gui/pull/5090 since I wanted to fix the functionality in the current system (where there is no error alert) before introducing new functionality (a dismissable error alert). 

### Proposed Changes

Like in other situations where file parsing fails, just close the importing alert.

Resolves #4639

### DIFF
--- a/src/containers/target-pane.jsx
+++ b/src/containers/target-pane.jsx
@@ -141,11 +141,13 @@ class TargetPane extends React.Component {
         this.props.onShowImporting();
         handleFileUpload(e.target, (buffer, fileType, fileName, fileIndex, fileCount) => {
             spriteUpload(buffer, fileType, fileName, storage, newSprite => {
-                this.handleNewSprite(newSprite).then(() => {
-                    if (fileIndex === fileCount - 1) {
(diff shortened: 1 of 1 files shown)


apple502j:
Resolves #4639

paulkaplan:
@apple502j oh thanks! I was looking for a related issue but didn't see one.

PR #5101, opened 2019-08-15
Allow some kind of downloading to work on iOS Safari

paulkaplan:
This allows exporting of projects/sprites/costumes/sounds in iOS Safari.

iOS safari does not support the `download` attribute on `<a>` elements, so we need to read the blob as a data-uri and assign that to a popup window. This prevents the situation where trying to export closes the project, possibly losing work. It is not a great experience on iOS, though. It opens a new tab, but the file name is always "unknown". It does allow the user to download sprite/project files to the local "Files" app. For costumes/sprites, it just shows the asset in a new tab, using the menu at the top/right allows the user to download the image/sound.

Apparently iOS13 will have support for downloading files, so we should keep an eye on this in the fall, but it looks like it will just automatically switch over to using the current download method. 

### Testing plan:
* try exporting costume, sound, sprite
* try downloading project
* make sure doing this doesn't close the project or make you lose work!

### DIFF
--- a/src/lib/download-blob.js
+++ b/src/lib/download-blob.js
@@ -8,11 +8,23 @@ export default (filename, blob) => {
         return;
     }
 
-    const url = window.URL.createObjectURL(blob);
-    downloadLink.href = url;
-    downloadLink.download = filename;
-    downloadLink.type = blob.type;
-    downloadLink.click();
-    window.URL.revokeObjectURL(url);
-    document.body.removeChild(downloadLink);
+    if ('download' in HTMLAnchorElement.prototype) {
+        const url = window.URL.createObjectURL(blob);
+        downloadLink.href = url;
+        downloadLink.download = filename;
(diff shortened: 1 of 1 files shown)


rschamp:
Try it out: [file-download-safari](https://llk.github.io/scratch-gui/file-download-safari)

paulkaplan:
@ericrosenbaum just a heads up i'm reassigning this from chris g to you

benjiwheeler:
Todo: 
1. prepare an ipad to access scratchly
1. write up testing steps for what to do
1. merge this and test on an ipad on scratchly using these steps
1. add them to the bug hunt doc

benjiwheeler:
@paulkaplan @ericrosenbaum I've merged this and it's up on scratch.ly, but it does not seem to have any effect. I'm using an iPad Pro with iOs 13.3.1 .

Issue #2444, opened 2018-06-22
"Change color" option in variable/list menu


towerofnix:
Simple suggestion/request - the ability to change the color of variable and list monitors by right-clicking the monitor and picking a "change color" option. This would do exactly what one would expect - it'd change the orange/red color of the monitor to something else. Handy for designing any sort of game or UI where you want the variables to seamlessly fit in with the rest of the stage, or if you want to assign particular emphasis to the meanings of variable monitors by giving them unique colors (like making hit points a green monitor and magic points blue).

wenxy:
https://github.com/source-tutorial/scratch-gui/tree/dev

PR #5102, opened 2019-08-15
Variables Extension Monitors

kchadha:
### Proposed Changes

Add support for monitors in variables extension. Monitor info now uses color directly instead of using a category string to look up the appropriate color. This also means that the VM can then use the block instance color for the monitor color. Additionally add support for enabling a monitor for a variable/list that was just created using the modal.

### Test Coverage

Update tests that were previously broken. Update monitor related tests to account for new changes.

### Related PRs
- This PR depends on LLK/scratch-vm#2262

### DIFF
--- a/src/components/monitor/monitor.jsx
+++ b/src/components/monitor/monitor.jsx
@@ -13,15 +13,8 @@ import ListMonitor from '../../containers/list-monitor.jsx';
 
 import styles from './monitor.css';
 
-const categories = {
-    data: '#FF8C1A',
-    sensing: '#5CB1D6',
-    sound: '#CF63CF',
-    looks: '#9966FF',
-    motion: '#4C97FF',
-    list: '#FC662C',
-    extension: '#0FBD8C'
-};
+
+const extensionColor = '#0FBD8C';
 
 const modes = {
     default: DefaultMonitor,
@@ -49,7 +42,7 @@ const MonitorComponent = props => (
                 onDoubleClick={props.mode === 'list' || !props.draggable ? null : props.onNextMode}
             >
                 {React.createElement(modes[props.mode], {
-                    categoryColor: categories[props.category],
+                    categoryColor: props.color,
                     ...props
                 })}
             </Box>
@@ -114,12 +107,11 @@ const MonitorComponent = props => (
 
 );
 
-MonitorComponent.categories = categories;
 
 const monitorModes = Object.keys(modes);
 
 MonitorComponent.propTypes = {
-    category: PropTypes.oneOf(Object.keys(categories)),
+    color: PropTypes.string,
     componentRef: PropTypes.func.isRequired,
     draggable: PropTypes.bool.isRequired,
     label: PropTypes.string.isRequired,
@@ -135,7 +127,7 @@ MonitorComponent.propTypes = {
 };
 
 MonitorComponent.defaultProps = {
-    category: 'extension',
+    color: extensionColor,
     mode: 'default'
 };
 
--- a/src/lib/define-dynamic-block.js
+++ b/src/lib/define-dynamic-block.js
@@ -246,14 +246,45 @@ const defineDynamicBlock = (guiContext, ScratchBlocks, categoryInfo, staticBlock
                 }
             });
 
+            // Add a validator function to each field_dropdown which is responsible
+            // for keeping the mutation and the field value in sync. Emit a change
+            // event for the mutation so that the VM can also stay in sync with
+            // these changes.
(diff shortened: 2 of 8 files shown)


src/components/monitor/monitor.jsx
```diff
@@ -49,7 +42,7 @@ const MonitorComponent = props => (
                 onDoubleClick={props.mode === 'list' || !props.draggable ? null : props.onNextMode}
             >
                 {React.createElement(modes[props.mode], {
-                    categoryColor: categories[props.category],
+                    categoryColor: props.color,
```
towerofnix: Ooh! Does this hypothetically make implementing #2444 easier? Exciting..!

src/lib/define-dynamic-block.js
```diff
@@ -246,14 +246,45 @@ const defineDynamicBlock = (guiContext, ScratchBlocks, categoryInfo, staticBlock
...
+                const currBlockInfo = JSON.parse(this.sourceBlock_.blockInfoText);
+                currBlockInfo.arguments[argName].selectedValue = selectedItem;
+                this.sourceBlock_.blockInfoText = JSON.stringify(currBlockInfo);
+                const newMutation = ScratchBlocks.Xml.domToText(this.sourceBlock_.mutationToDom());
+                ScratchBlocks.Events.fire(new ScratchBlocks.Events.BlockChange(this.sourceBlock_,
+                    'mutation', null, oldMutation, newMutation));
+                this.setValue(selectedItem);
+                return null;
+                /* eslint-enable no-invalid-this */
+            });
```
cwillisf: It might be more readable to pull this function out: in particular, it might avoid confusion about `this`. Also, `defineDynamicBlock` is kind of too big already... :/

I'm imagining something like:
```js
const makeFieldValidator = function (argName) {
    return function (selectedItem) {
        const oldMutation = // ...
    };
};

const defineDynamicBlock = // ...
```

I suspect something like this might make the `eslint-disable no-invalid-this` unnecessary, too, but I'm not certain.

src/lib/define-dynamic-block.js
```diff
@@ -246,14 +246,45 @@ const defineDynamicBlock = (guiContext, ScratchBlocks, categoryInfo, staticBlock
...
+                    } else {
+                        // Update the block info to keep track of the selected value for the
+                        // next time this block gets rendered without any external forces
+                        // changing it
+                        // See generic blockInfoText update below.
+                        // This prevents block arguments accidentally getting updated because
+                        // they don't have a selectedValue in their block info, and the
+                        // defaultValue is also dynamic and changed (e.g. a new variable is
+                        // added that is sorted first alphabetically)
+                        blockInfo.arguments[arg.name].selectedValue = field.getValue();
```
cwillisf: I'm not sure I understand what's going on here, even with the comment. What I'm reading is (pseudocode):

```
if (the block thinks something should be selected) {
    set the selection to(whatever the block thinks should be selected) // for checkboxes this will always be true if we get here
} else {
    // the block doesn't think anything should be selected, or it doesn't know yet, or it's a non-checked checkbox
    set the block's idea of the selection to(whatever is selected in the UI)
}
```

Is that accurate?

src/lib/opcode-labels.js
```diff
@@ -115,6 +115,19 @@ const messages = defineMessages({
...
+const categoryColors = {
+    data: '#FF8C1A',
+    sensing: '#5CB1D6',
+    sound: '#CF63CF',
+    looks: '#9966FF',
+    motion: '#4C97FF',
+    list: '#FC662C',
+    extension: '#0FBD8C'
+};
+
```
cwillisf: 🎉

src/lib/define-dynamic-block.js
kchadha: Hmmm I agree that `defineDynamicBlock` is getting much too bloated. When pulling this function out though, we need to pass in ScratchBlocks as an argument, and unfortunately the `this` issue does not go away (still need the eslint-disables).

kchadha: Moved the `makeFieldValidator` functionality into its own function as you suggested. I think even with passing in `ScratchBlocks` as an argument, it helps break up the complexity of `domToMutation` a little bit.

src/lib/define-dynamic-block.js
cwillisf: I'm not sure where checkboxes came from in what I was talking about... sorry about that :D

Anyway, thanks for the thorough explanation! I get it now -- that makes a lot of sense.

PR #5103, opened 2019-08-15
Added the Imagine If video to Imagine If tutorial, and changed last step

zoebentley:
Changes 
- Added a video as the first step of the Imagine If tutorial
- Changed the image for the last step of the Imagine If tutorial


paulkaplan:
@zoebentley looks good!

Issue #5105, opened 2019-08-16
On Edge, some vector images turn into grey rect & ?


apple502j:
### Expected Behavior
Image should be kept as-is!

### Actual Behavior
![ball-e](https://user-images.githubusercontent.com/33279053/63168003-e2c46200-c06d-11e9-89a6-1c5176181f96.png)
Some of the **vector** images (which are usually in the library) turn into grey rect & ? image on Edge. (PNG version above. For SVGs see the projects, MD5 checksum of SVG `8E768A5A5A01891B05C01C9CA15EB6AA`)

### Bug Reports
- https://scratch.mit.edu/projects/190478909/ `Cat1`, `Cat2` and `Cat6` (all Scratch Cat)
- https://scratch.mit.edu/projects/241515902/ `Sprite1` (Scratch Cat)
- https://scratch.mit.edu/projects/323776983/ `Ball1` (sprite library name Ball) - `Ball10`

### Steps to Reproduce
On Edge, I tried adding some sprites using the Surprise and found "that":
- Starfish
- Winter Hat
- Cat
- Ball
- Story letters
and more.

The console showed me something "interesting": `XML5662 "The "xml" prefix is reserved and must have the URI "http://www.w3.org/XML/1998/namespace/"."` This can be read as "asset problems" but I thought it will be nice to post here.

See also https://docs.microsoft.com/en-us/microsoft-edge/devtools-guide/console/error-and-status-codes#xml-codes

### Operating System and Browser
Edge on Windows 10

*Note: This **is** data loss. Make sure to put **critical**!*

lktornado360:
also https://scratch.mit.edu/projects/323718694/ and https://scratch.mit.edu/projects/10016238/

PR #5106, opened 2019-08-16
Revert render to last good version

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-gui/issues/5105

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -998,9 +998,9 @@
       "dev": true
     },
     "@types/node": {
-      "version": "12.7.1",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-12.7.1.tgz",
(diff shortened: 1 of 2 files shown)


towerofnix:
(do see https://github.com/LLK/scratch-svg-renderer/pull/104 before deciding to merge this!)

thisandagain:
Marking as "critical' for hotfix on Monday.

/cc @fsih @paulkaplan @BryceLTaylor

paulkaplan:
@fsih I built this up to a hotfix branch on www so that it is ready to go. We need to also make sure to revert whatever brought this change into scratch-render, so that we do not block future scratch-render updates.

paulkaplan:
@towerofnix we are going to revert it for now to make sure we go back to a known state, then move forward from there to fix t correctly

Issue #4244, opened 2019-01-03
Reports that iPad hardware keyboards are not triggering Scratch keypresses


benjiwheeler:
See discussion: https://scratch.mit.edu/discuss/topic/329379

![image](https://user-images.githubusercontent.com/3431616/50651712-3906e980-0f52-11e9-9af7-1b123045daff.png)

paulkaplan:
I was able to repro some issues with an apple bluetooth keyboard connected to an iPad. specifically:
- The keypressed reporter reports false until i hit the key, then it stays "true" forever afterwards (i.e. it never sees keyup). this is true for all keys
-  I was not able to confirm the second comment, the "when down arrow key pressed" seems to work correctly for me.

paulkaplan:
Wow this is very new to me. The bluetooth keyboard `keyup` event does not contain the `key` property, instead it sets the `key="Dead"`... but it does send the correct keyCode. We are already sending both `key` and `keyCode` to the VM, but the VM only uses `key`. We should be able to use the `keyCode` if the `key==="Dead"`. Bizarre...

lihongtao2012:
Did the blogger solve it? I downloaded your project. It's not easy to use

PR #5115, opened 2019-08-21
Fix bluetooth keyboards with iOS keypress blocks

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/4244

### Proposed Changes

Instead of sending both `key` and `keyCode` (which the VM ignores), send `keyCode` as `key` when the `key` is the string `"Dead"`. That is what happens when you keyup on a bluetooth keyboard in safari. The keycode information is still valid, but for some reason the `key` property is not.

### Reason for Changes

Keypressed block would remain true forever when using iOS Safari and a bluetooth keyboard.

### Test Coverage

Added unit tests for the keyup/keydown listeners of the vm-listener-hoc, including this new special behavior

Note: needs to be tested on iOS Safari + bluetooth keyboard, @benjiwheeler I've got one on on my desk if you need to use

### DIFF
--- a/src/lib/vm-listener-hoc.jsx
+++ b/src/lib/vm-listener-hoc.jsx
@@ -87,9 +87,9 @@ const vmListenerHOC = function (WrappedComponent) {
             // Don't capture keys intended for Blockly inputs.
             if (e.target !== document && e.target !== document.body) return;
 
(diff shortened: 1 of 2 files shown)


paulkaplan:
(adding @rschamp to review this since I know @benjiwheeler has a lot of work to do. @rschamp if you want me to find a different reviewer i can)

Issue #3278, opened 2018-10-01
Sound/costume scroller "fade" effect should not cover scrollbar


towerofnix:
### Expected Behavior

The fade effect should be adjacent to the scrollbar; it should not actually cover it.

### Actual Behavior

The scrollbar covers it. This does not look good and causes difficulty as described below.

### Steps to Reproduce

Simply open the "Costumes" or "Sounds" tab while the scrollbar is visible. Note that the fade effect covers the scrollbar.

If there are a lot of items in the sidebar, the scrollbar handle can become very short and actually be *covered* by the fade effect. This makes it impossible to grab the scroll handle, and some other scrolling method must be used (e.g. keyboard, mouse wheel, or clicking elsewhere in the scroll bar - but the important part is that this is *not* how a user might usually scroll, and in that case is annoying).

Here is a screenshot of the scrollbar handle covered by the fade effect:

![The scrollbar handle, relatively short because the scroller has 110 sounds in it; entirely covered by the fade effect](https://user-images.githubusercontent.com/9948030/46321470-d18b7780-c5b9-11e8-8b3d-62f3b03e45ea.png)

### Operating System and Browser

Firefox 63.0b10, Linux. (The scrollbar is blackish because I'm using the Arc Dark theme.)

benjiwheeler:
Thanks, I was able to reproduce this on Mac Chrome. We should definitely fix it.

benjiwheeler:
Ran into this again... impossible to click on scrollbar. Should fix!

towerofnix:
So, design question - do we want to have the scrollbar be positioned above the space the button takes up, ala this...

![a costume at the bottom of a scrolled-down costume list. the scrollbar's visual bottom edge is above the space behind the add-costume button](https://user-images.githubusercontent.com/9948030/64048836-47d98680-cb49-11e9-9c9f-b56c75aa08ff.png)

Or would we rather do something more like the scratch-blocks toolbox, adding padding to the end of the scrollable area?

The question is actually more a matter of this:

1) Separate scrollbar area and button area: this introduces a new visual border which the tiles are cut off at. The add button is never visually above any element (so the tile at the bottom of the scroll area is always easily selectable (provided at least a few pixels of it are visible from scrolling)).

2) Add padding to the bottom of (inside) the scrollable area: no new visual border, but the add button still always covers one tile—except, key point here, when you've scrolled to the very bottom. This allows you to access the last tile (hence resolves the issue).

[...]

PR #5143, opened 2019-09-03
Use list-area:after to pad scroll area

towerofnix:
### Resolves

Resolves LLK/scratch-gui#3278.

### Proposed Changes

Uses the `.list-area:after` pseudoelement to add padding to the end of the scroll area. See [this StackOverflow post](https://stackoverflow.com/a/22040425/4633828) for my reference; I had to add `flex-shrink: 0` so that the element wouldn't just get squished when it's meant to grow the scroll area.

### Reason for Changes

This makes it easy to drag and select the bottom tile in asset pickers.

I considered reusing the `$fade-out-distance` variable but decided against it because I saw there was already a `padding-bottom: 70px` property, which was meant to fix this issue itself but apparently no longer works (or never did in the first place). I used that value instead.

I tried putting a `margin-bottom` on `.list-area > :last-item` as well, but that didn't seem to have any effect. In the end, using `:after` worked on the browsers I tested.

### Test Coverage

No new tests; tested manually.

### Browser Coverage

Linux Firefox, Chromium.

---

/cc @benjiwheeler

### DIFF
--- a/src/components/asset-panel/selector.css
+++ b/src/components/asset-panel/selector.css
@@ -50,8 +50,16 @@ $fade-out-distance: 100px;
     overflow-y: scroll;
     display: flex;
     flex-direction: column;
+}
+
+.list-area:after {
     /* Make sure there is room to scroll beyond the last tile */
-    padding-bottom: 70px;
+    content: '';
+    display: block;
+    height: 70px;
+    width: 100%;
+    flex-shrink: 0;
+    order: 99999999;
 }
 
 .list-item {


paulkaplan:
Do you mind adding some screenshots?

towerofnix:
@paulkaplan sure. Here's what it looks like now, not scrolled (left) and scrolled (right):

not scrolled scrolled

When not scrolled it looks the same as before, but there is now always space so you can scroll down and fully uncover the bottom tile. See https://github.com/LLK/scratch-gui/issues/3278 for what the issue looks like (w/o this PR).

paulkaplan:
Interesting, seems to be Firefox specific? On Mac/Chrome it looks like your image on the right
image

towerofnix:
Oh curious - @benjiwheeler said they could repro on Mac Chrome back in October: https://github.com/LLK/scratch-gui/issues/3278#issuecomment-426386836. Though like you, I can't repro in Chromium today (but I've confirmed the issue doesn't come *back* in Chromium from merging this PR).

Issue #5142, opened 2019-09-02
Make the sound editor 'louder" button cause less distortion


ericrosenbaum:
Pressing the sound editor's "louder" button repeatedly causes the sound to become badly distorted (primarily because of "clipping," where the sound samples reach their limiting values at -1 or 1, causing loud crackling). We'd like to find a way that reduces or prevents this distortion. 

Scratch 2.0 had a "louder" effect that was strictly limited in such a way that the maximum gain allowed would cause the largest sample value to reach 1 or -1 (this is basically "normalization"). This completely prevents clipping, but it also causes it the effect to stop working once this limit is reached. This approach is not desirable for our new editor, in which it the louder effect is easier to get to and press repeatedly. Also, many of the sounds in the sound library are at or near normalization, so the effect would work only a tiny bit or not at all.

I've investigated using web audio's built in compression (DynamicsCompressorNode), but it adds a short delay (about 6ms) to the beginning of the sound. Repeatedly applying the effect with this node in the chain causes the delay to accumulate, so there's a growing silence at the beginning of the sound. We could try to trim this silence off automatically, but the pre-delay amount is not specified by the web audio spec so it could vary between browsers. I suppose we could measure the pre-delay when the effect initializes, but that starts to seem like a complicated solution. It's worth noting that the robot effect uses the DynamicsCompressorNode, so it already has this problem of the growing silence at the start of the sound, but it matters less because you are less likely to apply that effect repeatedly.

I think the next idea to try is to measure the RMS of the sound, as an estimate of its loudness (as we do in several other places: the recording level meter, the waveform renderer, and the loudness reporter block), and then apply the louder effect only if the maximum RMS in the sound is below some threshold. This may allow some clipping to occur, but will stop it before the distortion gets too bad.

PR #5149, opened 2019-09-03
Limit louder effect with a max RMS

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/5142

### Proposed Changes

Before applying the "louder" effect in the sound editor, measure the RMS of the sound. If any part of it is above a threshold, do not apply any gain.

You can try out this change here:
https://ericrosenbaum.github.io/scratch-gui/limit-louder-rms/

**Questions**
- How does the threshold feel? Should it be higher or lower?
- ~~How does it feel for the effect to simply stop working above the threshold? Should we disable the button in that case (not clear exactly how this would work)?~~ **UPDATE** I'm now disabling the button when the sound or selected region of the sound is too loud. This also prevents the issue with the undo stack when you were able to press the button with no result.
- The RMS limit is checked *before* applying the effect, instead of after. Perhaps it would be better to apply the effect, and then check the RMS, and "undo" if the limit has been crossed? This might make the results more consistent. This would require a somewhat more complex setup in the code.
- ~~There's a strange subtlety: When you apply gain that causes samples to exceed 1.0 or -1.0, then save and load, and play the sound again, the sound actually becomes slightly **more** distorted. I'm guessing that web audio treats those out-of-range sample values differently than if they were hard-clipped (i.e. to 1.0 or -1.0), but the wav encoding hard clips them. Also, after loading You can then generally apply louder one more time. I'm investigating, but it appears we can prevent these issues by applying a "hard limiter" inside the volume effect chain.~~ **UPDATE** I'm now using a waveshaper to hard limit the samples. The behavior is now consistent: the distortion gets worse faster, but does not change on save/load. 

### Reason for Changes

It's easy to press the louder button repeatedly and create badly distorted sounds. This limitation does not completely prevent the distortion, but reduces how bad you can quickly make it.

### DIFF
--- a/src/components/sound-editor/sound-editor.jsx
+++ b/src/components/sound-editor/sound-editor.jsx
@@ -265,6 +265,7 @@ const SoundEditor = props => (
                 onClick={props.onSlower}
             />
             <IconButton
+                disabled={props.tooLoud}
                 className={classNames(styles.effectButton, styles.flipInRtl)}
                 img={louderIcon}
                 title={<FormattedMessage {...messages.louder} />}
@@ -340,6 +341,7 @@ SoundEditor.propTypes = {
     onUndo: PropTypes.func.isRequired,
     playhead: PropTypes.number,
     setRef: PropTypes.func,
+    tooLoud: PropTypes.bool.isRequired,
     trimEnd: PropTypes.number,
     trimStart: PropTypes.number
 };
--- a/src/containers/sound-editor.jsx
+++ b/src/containers/sound-editor.jsx
@@ -14,6 +14,8 @@ import log from '../lib/log.js';
 
 const UNDO_STACK_SIZE = 99;
 
+const MAX_RMS = 1.2;
+
 class SoundEditor extends React.Component {
     constructor (props) {
         super(props);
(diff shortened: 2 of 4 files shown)


paulkaplan:
One thing i noticed is that the "no-effect" effect is being added to the undo stack, so even though we aren't changing the sound there are new copies of the sound on the undo stack. this makes undoing a bit strange because you have to click undo a bunch of times until something starts happening.

thisandagain:
This feels really great @ericrosenbaum. Thank you! A couple minor notes:

#### Threshold
Overall I think the threshold feels pretty good. For sounds the have a really wide dynamic range you can feel a little too constrained, but in most cases (sound library, quiet voice recordings, loud voice recordings, foley recording w/ my laptop) the balance is seems ok.

In most scenarios it feels that the limit generally lets you get right to the point of "clipping" and then no further. This still permits you to make pretty loud (i.e., annoying) sounds by selecting quieter subsections of the waveform and then selectively applying the effect. This could be an argument to push the threshold down a little bit, but I worry about overly constraining users in scenarios where they have a particularly "hot" recording.

#### Feedback
In general I think it would be good to provide some feedback to the user when "louder" reaches the limit. There were a few states (particularly with very short sounds) where the waveform changes were very subtle and it might not be clear when you have hit the limit visually. My initial thought is that simply applying a "disabled" state to the louder button could be enough, but I also don't feel like this is necessarily a blocker.

#### Pre-Check vs. Post-Check
The approach right now feels pretty good. I would be inclined to proceed with the strategy you have implemented here.

#### Undo
+1 to @paulkaplan's observation that this causes "undo" to behave in unexpected ways. This might be something to consider "blocking".

Issue #3433, opened 2018-10-19
in standalone, project title persists after upload; isn't set to default when File->New


benjiwheeler:
to repro:
1. open beta, do try it
2. upload file; title changes
3. File->New
4. observe that title is still uploaded version

benjiwheeler:
#5168 fixed many things, but we actually removed the part that reset the title on file->new! So we still need to do that...

Issue #4759, opened 2019-04-19
Loading project from computer doesn't name the project


BryceLTaylor:
### Expected Behavior

If you choose "File>Load from your computer" and upload a Scratch project it should import the name of the project.

### Actual Behavior

If you are looking at someone else's project when you choose "File>Load from your computer" it will create a new project but and populate it with the uploaded project but the title will be Untitled-## rather than the title of the project.

If you upload a project while looking at one of your projects it DOES rename the project.  I think the process of creating a new project is interfering with the naming in some way.

### Steps to Reproduce

Sign in
Look at a project you don't own
See Inside
Click "File>Load from your computer"
Choose a scratch file to upload
Note: when the project loads it has a title along the lines of "Untitled - 13"
Click "File>New"
Click "File>Load from your computer"
Choose a scratch file
Note: the project has the title of the uploaded project

### Operating System and Browser

Mac Chrome (Also seen on Windows Edge)

PR #5168, opened 2019-09-12
Refactor title handling

benjiwheeler:
### Resolves

Should resolve https://github.com/LLK/scratch-gui/issues/3433 and https://github.com/LLK/scratch-desktop/issues/71 ; 

Note that it does not seem to resolve the related https://github.com/LLK/scratch-gui/issues/4759 .

### Proposed Changes

Repurpose TitledHOC so that it:
* no longer maintains projectTitle state
* no longer passes projectTitle down to children
* no longer sets `canEditTitle` to true on children
* has responsibility for setting redux projectTitle when parent passes a changed projectTitle prop (previously this functionality was in containers/gui.jsx)
* notices when the project-state changes to `isShowingWithoutId`, uses the translated default project name to set the redux projectTitle, and also calls the parent's`onUpdateProjectTitle` prop function

GUI's playground/standalone versions previously used TitledHOC, so they need to explicitly pass `canEditTitle` to children.

scratch-desktop will need to no longer wrap itself with TitledHOC, and instead maintain its own projectTitle state.

### Related

If this is merged, we will need to merge https://github.com/LLK/scratch-desktop/pull/74 as well so that scratch-desktop is in line with these changes.

### Reason for Changes

1) https://github.com/LLK/scratch-gui/issues/3433 and https://github.com/LLK/scratch-desktop/issues/71 were bugs
2) we want to localize the default project name


src/containers/gui.jsx
```diff
@@ -126,17 +103,16 @@ GUI.propTypes = {
     isLoading: PropTypes.bool,
     isScratchDesktop: PropTypes.bool,
     isShowingProject: PropTypes.bool,
+    isShowingWithoutId: PropTypes.bool,
```
rschamp: Is this added intentionally? It looks unrelated.

src/lib/titled-hoc.jsx
```diff
@@ -12,26 +26,89 @@ const TitledHOC = function (WrappedComponent) {
...
+            if (this.props.projectTitle !== prevProps.projectTitle) {
+                this.setReduxTitle(this.props.projectTitle);
+            }
+            if (this.props.isShowingWithoutId && !prevProps.isShowingWithoutId) {
+                const defaultProjectTitle = this.titleWithDefault();
+                this.setReduxTitle(defaultProjectTitle);
+                this.props.onUpdateProjectTitle(defaultProjectTitle);
+            }
+        }
+        titleWithDefault (title) {
```
rschamp: Why is this a function that takes a `title` argument? It looks like it's only called in one place, without an argument.

src/lib/titled-hoc.jsx
```diff
@@ -12,26 +26,89 @@ const TitledHOC = function (WrappedComponent) {
...
+                // redux value
+                projectTitle,
+                /* eslint-enable no-unused-vars */
+                ...componentProps
+            } = this.props;
             return (
                 <WrappedComponent
-                    canEditTitle
-                    projectTitle={this.state.projectTitle}
                     onUpdateProjectTitle={this.handleUpdateProjectTitle}
```
rschamp: If we expect children to receive the title from redux, I think we should we also expect them to set it through redux.

src/containers/gui.jsx
benjiwheeler: Removing...

src/lib/titled-hoc.jsx
benjiwheeler: I made this change, now testing it...

src/lib/titled-hoc.jsx
benjiwheeler: I think I meant to refactor setReduxTitle further... doing this now

src/containers/gui.jsx
benjiwheeler: Done

src/lib/titled-hoc.jsx
benjiwheeler: Done

src/lib/titled-hoc.jsx
benjiwheeler: Tested well

src/lib/titled-hoc.jsx
```diff
@@ -1,37 +1,97 @@
...
-                projectTitle: null
-            };
+        componentDidMount () {
+            this.props.updateReduxProjectTitle(this.titleWithDefault(this.props.projectTitle));
+        }
+        componentDidUpdate (prevProps) {
+            if (this.props.projectTitle !== prevProps.projectTitle) {
+                this.props.updateReduxProjectTitle(this.titleWithDefault(this.props.projectTitle));
+            }
+            // if the projectTitle hasn't changed, but the reduxProjectTitle
```
benjiwheeler: Note this addition... here is where the change to redux title gets echoed up to the enclosing context (i.e., www)

src/components/menu-bar/project-title-input.jsx
```diff
@@ -19,47 +19,37 @@ const messages = defineMessages({
     }
 });
 
-class ProjectTitleInput extends React.Component {
```
benjiwheeler: I should change this back to how it was before

src/lib/titled-hoc.jsx
```diff
@@ -1,37 +1,97 @@
...
-        constructor (props) {
-            super(props);
-            bindAll(this, [
-                'handleUpdateProjectTitle'
-            ]);
-            this.state = {
-                projectTitle: null
-            };
+        componentDidMount () {
+            this.props.updateReduxProjectTitle(this.titleWithDefault(this.props.projectTitle));
```
benjiwheeler: can combine these calls

src/components/menu-bar/project-title-input.jsx
rschamp: I think this is fine to do since your handler is coming from Redux now. I just think the name of the prop should change.

src/playground/player.jsx
```diff
@@ -24,6 +23,7 @@ const Player = ({isPlayerOnly, onSeeInside, projectId}) => (
     <Box className={classNames(isPlayerOnly ? styles.stageOnly : styles.editor)}>
         {isPlayerOnly && <button onClick={onSeeInside}>{'See inside'}</button>}
         <GUI
+            canEditTitle
```
rschamp: Curious why this is necessary to add to the playgrounds.

rschamp: Nevermind I see now!

src/components/menu-bar/project-title-input.jsx
benjiwheeler: Doing this...

src/components/menu-bar/project-title-input.jsx
```diff
@@ -19,47 +19,37 @@ const messages = defineMessages({
...
+        tabIndex="0"
+        type="text"
+        value={projectTitle}
+        onSubmit={handleUpdateReduxProjectTitle}
+    />
+);
 
 ProjectTitleInput.propTypes = {
     className: PropTypes.string,
+    handleUpdateReduxProjectTitle: PropTypes.func,
```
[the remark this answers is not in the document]
benjiwheeler: What I'm used to the linter insisting on is that this form be followed:

```
onXXXX={this.handleXXXX}
```

In my experience it doesn't like
```
onXXXX={this.onXXXX}
```

but it's ok with
```
onXXXX={this.props.whateverNameIsOK}
```

src/containers/sb-file-uploader.jsx
```diff
@@ -179,9 +180,9 @@ SBFileUploader.propTypes = {
     loadingState: PropTypes.oneOf(LoadingStates),
     onLoadingFinished: PropTypes.func,
     onLoadingStarted: PropTypes.func,
-    onUpdateProjectTitle: PropTypes.func,
     projectChanged: PropTypes.bool,
     requestProjectUpload: PropTypes.func,
+    updateReduxProjectTitle: PropTypes.func,
```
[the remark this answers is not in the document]
benjiwheeler: Changing...

src/lib/titled-hoc.jsx
benjiwheeler: Changing...

src/components/menu-bar/project-title-input.jsx
benjiwheeler: Done

src/lib/titled-hoc.jsx
benjiwheeler: Done

src/containers/sb-file-uploader.jsx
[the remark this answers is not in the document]
benjiwheeler: Done

Issue #1735, opened 2018-04-06
Sounds recorded in certain situations are glitchy


ericrosenbaum:
### Expected Behavior

The sound recorder should be able to record audio smoothly and continuously no matter what else Scratch is doing.

### Actual Behavior

In some cases, recorded sounds have loud crackles and other glitches in them.

I have observed this:

- On a Chrome Book, while the video motion extension is running
- On my Macbook Pro, in Chrome, while the video motion extension is running 
- On my Macbook Pro, in Safari, in a project that is not currently running but has 14 bitmap assets that are 480x360
- @kathymakes reports a similar problem when recording sound on an Android tablet

### Steps to Reproduce

- Enable the "video sensing" extension
- Record a sound and play it back

OR

- Use Mac Safari
- Add a dozen full stage size bitmap assets
- Record a sound and play it back

### Operating System and Browser

Mac OS 10.13.2 Chrome 65
Samsung Chrome Book
Android tablet

ktrvs:
I can definitely repro this. macOS 10.13.3, Chrome Latest.

paulkaplan:
After some investigation, it looks like the react update to the svg waveform is eating a ton of CPU time. A quick prototype of a version that directly draws the waveform to canvas instead of going through the react update lifecycle with SVG is considerably faster, and I think is the first easy win for making this better.

PR #5169, opened 2019-09-13
Use a larger buffer size for recording to prevent glitching

paulkaplan:
### Resolves

Helps with https://github.com/LLK/scratch-gui/issues/1735

### Proposed Changes

Use an 8x larger buffer size for script processor node, which allows the browser much more leeway in terms of how quickly it needs to run the script processor node before buffers start getting dropped. This does not impact the smoothness of the visualization, which is run at RAF framerate using a separate analyzer path. 

I had to change the `stop` function, which previously was relying on the buffer size being 1024 implicitly (it would submit RMS chunks depending on the script processor buffer length). Instead use the other RMS utility by switching around the order of operations.

### Reason for Changes

By using a small buffer size, we were causing glitchiness in exchange for latency, which it turns out we do not need because we are recording!

### DIFF
--- a/src/components/record-modal/record-modal.css
+++ b/src/components/record-modal/record-modal.css
@@ -20,8 +20,10 @@
     border: 1px solid $ui-black-transparent;
     border-radius: 5px;
     padding: 3px;
-
+    /* Force these to be on their own render layer because they update often */
+    transform: translateZ(0);
 }
+
 .meter-container {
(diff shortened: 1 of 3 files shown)


rschamp:
Try it out: [recording-glitch](https://llk.github.io/scratch-gui/recording-glitch)

PR #5171, opened 2019-09-16
explicitly track whether we are in embed mode; block cloud vars on embed

benjiwheeler:
### Resolves

With https://github.com/LLK/scratch-www/pull/3356, resolves https://github.com/LLK/scratch-www/issues/2471

Must be merged before https://github.com/LLK/scratch-www/pull/3356

### Proposed Changes

* adds `isEmbed` property to `mode` reducer
* sets `isEmbed` to true if GUI is inited as embed
* stops any cloud connection from being made if project is an embed

### Reason for Changes

Embeds make too many requests of our servers! 

### Test Coverage

None

### DIFF
--- a/src/lib/cloud-manager-hoc.jsx
+++ b/src/lib/cloud-manager-hoc.jsx
@@ -52,7 +52,10 @@ const cloudManagerHOC = function (WrappedComponent) {
             this.disconnectFromCloud();
         }
         canUseCloud (props) {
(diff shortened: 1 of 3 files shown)


benjiwheeler:
We're using a different approach in https://github.com/LLK/scratch-www/pull/3356 that makes this PR not necessary.

PR #5194, opened 2019-09-20
Don’t push tags to gh-pages

rschamp:
These builds make our gh-pages branch too large for GitHub to handle. Thanks to @fsih for deleting all those builds. This will keep them from re-populating.

@fsih I'm never sure who to assign for this type of change, so I assigned you since we talked about it. Here's the relevant part of Travis' docs. https://docs.travis-ci.com/user/deployment/#conditional-releases-with-on

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,7 @@ deploy:
   on:
     all_branches: true
     condition: $TRAVIS_EVENT_TYPE != cron && ! $TRAVIS_BRANCH =~ ^greenkeeper/
+    tags: false # Don't push tags to gh-pages
   skip_cleanup: true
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -74,6 +74,7 @@ deploy:
   on:
     all_branches: true
     condition: $TRAVIS_EVENT_TYPE != cron && ! $TRAVIS_BRANCH =~ ^greenkeeper/
+    tags: false # Don't push tags to gh-pages
```
fsih: It says setting tags "causes the branch condition to be ignored", so I think that means you don't need the all_branches line

rschamp: From what I can tell, without `all_branches`, it uses the default for `branch` which is whichever primary branch is configured in GitHub (develop on most repos), which would cause this step to only trigger on one branch. So I _think_ it's necessary to keep it.

PR #5198, opened 2019-09-20
simplify props of cloud-manager-hoc

benjiwheeler:
### Proposed Changes

* removes `shouldNotModifyCloudData` function and replaces it with prop `canModifyCloudData` (with opposite boolean value)
* Sets prop defaults so that code which renders GUI (such as www's embed view) can safely omit most props, and trust GUI to still render as expected

### DIFF
--- a/src/lib/cloud-manager-hoc.jsx
+++ b/src/lib/cloud-manager-hoc.jsx
@@ -54,13 +54,10 @@ const cloudManagerHOC = function (WrappedComponent) {
         canUseCloud (props) {
             return !!(props.cloudHost && props.username && props.vm && props.projectId && props.hasCloudPermission);
         }
-        shouldNotModifyCloudData (props) {
-            return (props.hasEverEnteredEditor && !props.canSave);
-        }
         shouldConnect (props) {
             return !this.isConnected() && this.canUseCloud(props) &&
                 props.isShowingWithId && props.vm.runtime.hasCloudData() &&
-                !this.shouldNotModifyCloudData(props);
+                props.canModifyCloudData;
         }
         shouldDisconnect (props, prevProps) {
             return this.isConnected() &&
@@ -70,7 +67,7 @@ const cloudManagerHOC = function (WrappedComponent) {
                     (props.projectId !== prevProps.projectId) ||
                     (props.username !== prevProps.username) ||
                     // Editing someone else's project
-                    this.shouldNotModifyCloudData(props)
+                    !props.canModifyCloudData
                 );
         }
         isConnected () {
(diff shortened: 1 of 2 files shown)


benjiwheeler:
> I think adding CloudManager.defaultProps is a good idea. However I think we should leave canSave as .isRequired, since previously we had decided to try to keep as many things as required as possible, unless they are explicitly not required. The strictness helps things from being overlooked as props change.

Take a look now. I added `isRequired` to each propType where it seems appropriate.

src/components/gui/gui.jsx
```diff
@@ -425,7 +425,6 @@ GUIComponent.defaultProps = {
     canEditTitle: false,
     canManageFiles: true,
     canRemix: false,
-    canSave: false,
```
rschamp: Confused why this is still moving out of GUI and into CloudManagerHOC. It's a prop for GUI, not anything to do with cloud variables, so it would make sense to me for it to live here.

src/lib/cloud-manager-hoc.jsx
```diff
@@ -126,15 +126,23 @@ const cloudManagerHOC = function (WrappedComponent) {
     CloudManager.propTypes = {
         canSave: PropTypes.bool.isRequired,
```
rschamp: Also confused why this is a prop for CloudManager, even though you haven't touched that.

src/components/gui/gui.jsx
benjiwheeler: It does have to do with cloud variables, because it's used in the `shouldNotModifyCloudData` function.

In short, the idea there is that users who are not the project's owner should not be able to modify the code and then still be able to set cloud data (prevents the "high score hack"); whereas the project's owner should be able to still set cloud data.

src/lib/cloud-manager-hoc.jsx
benjiwheeler: I don't know what you mean about having touched that. For why `canSave` is a prop here, see my comment above, at https://github.com/LLK/scratch-gui/pull/5198#discussion_r362549403

src/components/gui/gui.jsx
rschamp: Do you understand why I think it's confusing for CloudManager to be setting the default for `canSave`?

src/lib/cloud-manager-hoc.jsx
rschamp: By "you haven't touched that", I was just acknowledging that this line isn't necessarily in scope for this PR since it's unchanged.

benjiwheeler:
Next steps:
* make shouldNotModifyCloudData into a prop, using ownProps

src/components/gui/gui.jsx
benjiwheeler: Resolving this per offline discussion

src/lib/cloud-manager-hoc.jsx
```diff
@@ -124,23 +121,30 @@ const cloudManagerHOC = function (WrappedComponent) {
     }
 
     CloudManager.propTypes = {
-        canSave: PropTypes.bool.isRequired,
+        canModifyCloudData: PropTypes.bool.isRequired,
```
benjiwheeler: note that I switched the boolean value of shouldNotModifyCloudData canModifyCloudData.

There's potential confusion with `hasCloudPermission`, which in www is set to `isScratcher`; I have ideas for further simplification here, but I think we should take it one step at a time.

PR #5229, opened 2019-10-07
Depend on isShowingWithId for updating thumbnail

chrisgarrity:
In addition to being loaded, make sure the project is showing with id. I think this will prevent trying to save a thumbnail when projectId is `“0”`. Moved it out of the telemetry event condition, but it could go back if that seems like the right place for it.

### DIFF
--- a/src/components/scratch-image/scratch-image.jsx
+++ b/src/components/scratch-image/scratch-image.jsx
@@ -136,6 +136,7 @@ ScratchImage.ImageSourcePropType = PropTypes.oneOfType([
 ]);
 
 ScratchImage.propTypes = {
+    src: PropTypes.string,
     imageSource: ScratchImage.ImageSourcePropType.isRequired
 };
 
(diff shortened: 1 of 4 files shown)


src/components/scratch-image/scratch-image.jsx
```diff
@@ -136,6 +136,7 @@ ScratchImage.ImageSourcePropType = PropTypes.oneOfType([
 ]);
 
 ScratchImage.propTypes = {
+    src: PropTypes.string,
```
chrisgarrity: @paulkaplan lint was complaining about `src` being missing from the props validation, so I added it. However, I'm not sure if src is actually used/needed (defined above on line 97). 
/cc @cwillisf

chrisgarrity:
While I made the 'defensive' change on the Android side to make sure that the project exists when SaveThumbnail is called, this PR should make sure that we don't try to save a thumbnail for the default project before it has an id. Checking for `projectId > 0` isn't going to work because Android ids are not numeric (UUIDs). I wonder if there are other places where that could be a problem.

chrisgarrity:
Closing - no longer needed.

PR #5237, opened 2019-10-10
Update paint to add swahili to overflow languages

fsih:
Screen Shot 2019-10-10 at 12 21 45


fsih:
@BryceLTaylor @chrisgarrity how does this look?

fsih:
If I narrow the browser a bit, I get this
Screen Shot 2019-10-10 at 12 22 53

PR #5291, opened 2019-11-12
reset title on transitioning to isShowingWithoutId

benjiwheeler:
https://github.com/LLK/scratch-gui/pull/5168 was originally intended to resolve https://github.com/LLK/scratch-gui/issues/3433 and https://github.com/LLK/scratch-desktop/issues/71 , but then it was narrowed in scope to a general refactor of project title handling.

This change should resolve https://github.com/LLK/scratch-gui/issues/3433 and https://github.com/LLK/scratch-desktop/issues/71 .

### Proposed Changes

When project state transitions into `isShowingWithoutId` being true, set the project title to the default title, and call GUI's owner's `onUpdateProjectTitle` function.

### Reason for Changes

In standalone and desktop modes, File->New was not resetting title.

### Test Coverage

We don't have good testing for this functionality yet. What would it look like?

### DIFF
--- a/src/lib/titled-hoc.jsx
+++ b/src/lib/titled-hoc.jsx
@@ -27,6 +27,10 @@ const TitledHOC = function (WrappedComponent) {
             if (this.props.projectTitle !== prevProps.projectTitle) {
                 this.handleReceivedProjectTitle(this.props.projectTitle);
             }
+            if (this.props.isShowingWithoutId && !prevProps.isShowingWithoutId) {
+                const defaultProjectTitle = this.handleReceivedProjectTitle();
+                this.props.onUpdateProjectTitle(defaultProjectTitle);
+            }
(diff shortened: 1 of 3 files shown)


paulkaplan:
Code LG but it is tough to think about all the implications. could you test with selenium integration test? Like load app, change project title, file -> new, check project title?

Could you put together a list of actions we should make sure to test that might intersect with this, like importing a file when showing with ID and when showi without Id?

benjiwheeler:
Things we'd test, ideally:
* in standalone mode:
  * showing default project
     * change title, then do file->new; title should reset
     * change title, then do file->load; title should be from file
        * then do file->new; title should reset

* embedded in www:
  * logged in
     * showing default project
        * change title, then do file->new; title should be some version of "Untitled-NUMBER"
        * change title, then do file->load; title should be from file
           * then do file->new; title should reset

PR #5328, opened 2019-12-05
Revert "Update scratch-render to the latest version 🚀"

paulkaplan:
Reverts LLK/scratch-gui#5319

Revert the renderer update because it includes the large mipmap change. We want to be conservative about deploys before HOC. I'll re-revert this change after we have a built version that can be released.


fsih:
Oops sorry

PR #5338, opened 2019-12-12
Implement sound downsampler to try to allow more sounds to be edited

paulkaplan:
### Proposed Changes

Add a step to the sound editor to downsample the sound if needed in order to be saved. This augments the previous behavior where it just refused to edit sounds that were too long.

In addition to the downsampler, which just reuses some code that @ericrosenbaum had already made for matching sample rates, this required changing the main entry point function for updating sounds from synchronous to asynchronous, since the resampling process is async.

### Reason for Changes

This change allows for 2x longer sounds to be edited with the sound editor. This includes things like making background music quieter (thanks @zoebentley for example projects with this), muting bad words in songs and otherwise trimming and editing imported songs.

Due to the 10MB asset limit (10,000,000 bytes), the length limit for editable sounds is currently:
10000000 / (44100 * 2) = 113 seconds (1min 53sec)
This change makes it twice as long:
10000000 / (22050 * 2) = 226 seconds (3min 46sec) 

### Test Coverage

Added a unit test for the downsampler logic and updated tests that depended on a synchronous sound update

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [x] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/sound-editor.jsx
+++ b/src/containers/sound-editor.jsx
@@ -6,7 +6,11 @@ import VM from 'scratch-vm';
 
 import {connect} from 'react-redux';
 
-import {computeChunkedRMS, encodeAndAddSoundToVM, SOUND_BYTE_LIMIT} from '../lib/audio/audio-util.js';
+import {
+    computeChunkedRMS,
+    encodeAndAddSoundToVM,
+    downsampleIfNeeded
+} from '../lib/audio/audio-util.js';
 import AudioEffects from '../lib/audio/audio-effects.js';
 import SoundEditorComponent from '../components/sound-editor/sound-editor.jsx';
 import AudioBufferPlayer from '../lib/audio/audio-buffer-player.js';
@@ -132,29 +136,27 @@ class SoundEditor extends React.Component {
         });
     }
     submitNewSamples (samples, sampleRate, skipUndo) {
-        return WavEncoder.encode({
-            sampleRate: sampleRate,
-            channelData: [samples]
-        })
-            .then(wavBuffer => {
-                if (wavBuffer.byteLength > SOUND_BYTE_LIMIT) {
-                    log.error(`Refusing to encode sound larger than ${SOUND_BYTE_LIMIT} bytes`);
-                    return Promise.reject();
-                }
-                if (!skipUndo) {
-                    this.redoStack = [];
-                    if (this.undoStack.length >= UNDO_STACK_SIZE) {
-                        this.undoStack.shift(); // Drop the first element off the array
+        return downsampleIfNeeded(samples, sampleRate, this.resampleBufferToRate)
+            .then(({samples: newSamples, sampleRate: newSampleRate}) =>
+                WavEncoder.encode({
+                    sampleRate: newSampleRate,
+                    channelData: [newSamples]
+                }).then(wavBuffer => {
+                    if (!skipUndo) {
+                        this.redoStack = [];
+                        if (this.undoStack.length >= UNDO_STACK_SIZE) {
+                            this.undoStack.shift(); // Drop the first element off the array
+                        }
+                        this.undoStack.push(this.getUndoItem());
                     }
-                    this.undoStack.push(this.getUndoItem());
-                }
-                this.resetState(samples, sampleRate);
-                this.props.vm.updateSoundBuffer(
-                    this.props.soundIndex,
-                    this.audioBufferPlayer.buffer,
-                    new Uint8Array(wavBuffer));
-                return true; // Edit was successful
-            })
+                    this.resetState(newSamples, newSampleRate);
+                    this.props.vm.updateSoundBuffer(
+                        this.props.soundIndex,
+                        this.audioBufferPlayer.buffer,
+                        new Uint8Array(wavBuffer));
+                    return true; // Edit was successful
+                })
+            )
             .catch(e => {
                 // Encoding failed, or the sound was too large to save so edit is rejected
                 log.error(`Encountered error while trying to encode sound update: ${e}`);
--- a/src/containers/sound-editor.jsx
+++ b/src/containers/sound-editor.jsx
@@ -362,9 +362,11 @@ class SoundEditor extends React.Component {
     }
     backupDownSampler (buffer, newRate) {
         log.warn(`Using backup down sampler for conversion from ${buffer.sampleRate} to ${newRate}`);
-        const newSamples = buffer.samples.filter((element, index) =>
-            index % 2 === 0
-        );
+        const newLength = Math.floor(buffer.samples.length / 2);
+        const newSamples = new Float32Array(newLength);
+        for (let i = 0; i < newLength; i++) {
+            newSamples[i] = buffer.samples[i * 2];
+        }
         return {
             samples: newSamples,
             sampleRate: newRate
--- a/src/containers/sound-editor.jsx
+++ b/src/containers/sound-editor.jsx
@@ -325,25 +325,30 @@ class SoundEditor extends React.Component {
         });
     }
     resampleBufferToRate (buffer, newRate) {
-        return new Promise(resolve => {
+        return new Promise((resolve, reject) => {
+            const sampleRateRatio = newRate / buffer.sampleRate;
+            const newLength = sampleRateRatio * buffer.samples.length;
+            let offlineContext;
             if (window.OfflineAudioContext) {
-                const sampleRateRatio = newRate / buffer.sampleRate;
-                const newLength = sampleRateRatio * buffer.samples.length;
-                const offlineContext = new window.OfflineAudioContext(1, newLength, newRate);
-                const source = offlineContext.createBufferSource();
-                const audioBuffer = offlineContext.createBuffer(1, buffer.samples.length, buffer.sampleRate);
-                audioBuffer.getChannelData(0).set(buffer.samples);
-                source.buffer = audioBuffer;
-                source.connect(offlineContext.destination);
-                source.start();
-                offlineContext.startRendering();
-                offlineContext.oncomplete = ({renderedBuffer}) => {
-                    resolve({
-                        samples: renderedBuffer.getChannelData(0),
-                        sampleRate: newRate
-                    });
-                };
+                offlineContext = new window.OfflineAudioContext(1, newLength, newRate);
+            } else if (window.webkitOfflineAudioContext) {
+                offlineContext = new window.webkitOfflineAudioContext(1, newLength, newRate);
+            } else {
+                return reject('No offline audio context');
             }
+            const source = offlineContext.createBufferSource();
+            const audioBuffer = offlineContext.createBuffer(1, buffer.samples.length, buffer.sampleRate);
+            audioBuffer.getChannelData(0).set(buffer.samples);
+            source.buffer = audioBuffer;
+            source.connect(offlineContext.destination);
+            source.start();
+            offlineContext.startRendering();
+            offlineContext.oncomplete = ({renderedBuffer}) => {
+                resolve({
+                    samples: renderedBuffer.getChannelData(0),
+                    sampleRate: newRate
+                });
+            };
         });
     }
     paste () {
--- a/src/containers/sound-editor.jsx
+++ b/src/containers/sound-editor.jsx
@@ -132,45 +132,34 @@ class SoundEditor extends React.Component {
         });
     }
     submitNewSamples (samples, sampleRate, skipUndo) {
-        // Encode the new sound into a wav so that it can be stored
-        let wavBuffer = null;
-        try {
-            wavBuffer = WavEncoder.encode.sync({
-                sampleRate: sampleRate,
-                channelData: [samples]
-            });
-
-            if (wavBuffer.byteLength > SOUND_BYTE_LIMIT) {
-                // Cancel the sound update by setting to null
-                wavBuffer = null;
-                log.error(`Refusing to encode sound larger than ${SOUND_BYTE_LIMIT} bytes`);
-            }
-        } catch (e) {
-            // This error state is mostly for the mock sounds used during testing.
-            // Any incorrect sound buffer trying to get interpretd as a Wav file
-            // should yield this error.
-            // This can also happen if the sound is too be allocated in memory.
-            log.error(`Encountered error while trying to encode sound update: ${e}`);
-        }
-
-        // Do not submit sound if it could not be encoded (i.e. if too large)
-        if (wavBuffer) {
-            if (!skipUndo) {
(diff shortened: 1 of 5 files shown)


ericrosenbaum:
@paulkaplan I added the emergency-backup-low-fi-downsampler system, so that we can get down to 22kHz on Safari. Using the `for` loop is way faster than `filter`!

test/__mocks__/audio-effects.js
```diff
@@ -14,7 +14,10 @@ export default class MockAudioEffects {
         this.buffer = buffer;
         this.name = name;
         this.process = jest.fn(done => {
-            this._finishProcessing = renderedBuffer => done(renderedBuffer, 0, 1);
+            this._finishProcessing = renderedBuffer => {
+                done(renderedBuffer, 0, 1);
+                return new Promise(resolve => setTimeout(resolve));
```
paulkaplan: Now that the rest submitNewSamples pipeline is async, this was needed to allow the tests to be run using async/await

src/containers/sound-editor.jsx
```diff
@@ -322,27 +327,51 @@ class SoundEditor extends React.Component {
...
+            offlineContext.startRendering();
+            offlineContext.oncomplete = ({renderedBuffer}) => {
+                resolve({
+                    samples: renderedBuffer.getChannelData(0),
+                    sampleRate: newRate
+                });
+            };
         });
     }
+    backupDownSampler (buffer, newRate) {
```
paulkaplan: @ericrosenbaum I think this should be moved to audio-util so that it can be easily tested.

ericrosenbaum: Good point- I'll give that a try. Re-assigning myself.

src/containers/sound-editor.jsx
```diff
@@ -209,11 +202,13 @@ class SoundEditor extends React.Component {
...
-            trimEnd: null
+        this.submitNewSamples(newSamples, sampleRate).then(() => {
+            this.setState({
+                trimStart: null,
+                trimEnd: null
+            });
         });
+
     }
     handleDeleteInverse () {
```
benjiwheeler: I know this is outside the scope of your change, but I'm confused about what "handleDeleteInverse" means

src/containers/sound-editor.jsx
```diff
@@ -6,7 +6,12 @@ import VM from 'scratch-vm';
 
 import {connect} from 'react-redux';
 
-import {computeChunkedRMS, encodeAndAddSoundToVM, SOUND_BYTE_LIMIT} from '../lib/audio/audio-util.js';
+import {
+    computeChunkedRMS,
+    encodeAndAddSoundToVM,
+    downsampleIfNeeded,
+    backupDownSampler
```
benjiwheeler: This is outside the scope of your change, but what does "backup" mean in this context?

src/containers/sound-editor.jsx
```diff
@@ -322,25 +327,37 @@ class SoundEditor extends React.Component {
...
+                return reject('No offline audio context');
             }
+            const source = offlineContext.createBufferSource();
+            const audioBuffer = offlineContext.createBuffer(1, buffer.samples.length, buffer.sampleRate);
+            audioBuffer.getChannelData(0).set(buffer.samples);
+            source.buffer = audioBuffer;
+            source.connect(offlineContext.destination);
+            source.start();
+            offlineContext.startRendering();
+            offlineContext.oncomplete = ({renderedBuffer}) => {
```
benjiwheeler: can oncomplete fail? Do you need to handle a situation where it should reject()?

src/containers/sound-editor.jsx
```diff
@@ -322,25 +327,37 @@ class SoundEditor extends React.Component {
...
-                        samples: renderedBuffer.getChannelData(0),
-                        sampleRate: newRate
-                    });
-                };
+                offlineContext = new window.OfflineAudioContext(1, newLength, newRate);
+            } else if (window.webkitOfflineAudioContext) {
+                try {
+                    offlineContext = new window.webkitOfflineAudioContext(1, newLength, newRate);
+                } catch {
+                    if (newRate === (buffer.sampleRate / 2)) {
```
benjiwheeler: another way of writing this would be that sampleRateRatio === 0.5, right? though maybe testing a floating point number for equality is unreliable?

src/lib/audio/audio-util.js
```diff
@@ -59,9 +60,39 @@ const encodeAndAddSoundToVM = function (vm, samples, sampleRate, name, callback)
...
+    // If encodeable at 22khz, resample and call submitNewSamples again
+    if (duration * 22050 * 2 < SOUND_BYTE_LIMIT) {
+        return resampler({samples, sampleRate}, 22050);
+    }
+    // Cannot save this sound at 22khz, refuse to edit
+    // In the future we could introduce further compression here
+    return Promise.reject('Sound too large to save, refusing to edit');
+};
+
+const backupDownSampler = (buffer, newRate) => {
```
benjiwheeler: I'm not sure I understand what "backup" means in this context. I know from your perspective this downsampling is abnormal, but to an outsider like me it doesn't seem that way -- this just seems like more downsampling to make the data smaller and more editable.

src/containers/sound-editor.jsx
```diff
@@ -322,25 +327,37 @@ class SoundEditor extends React.Component {
...
+                        return resolve(backupDownSampler(buffer, newRate));
```
benjiwheeler: I don't understand why this is happening only if we failed to construct a webkitOfflineAudioContext. What's the relationship?

src/containers/sound-editor.jsx
```diff
@@ -322,25 +327,37 @@ class SoundEditor extends React.Component {
...
-                source.connect(offlineContext.destination);
-                source.start();
-                offlineContext.startRendering();
-                offlineContext.oncomplete = ({renderedBuffer}) => {
-                    resolve({
-                        samples: renderedBuffer.getChannelData(0),
-                        sampleRate: newRate
-                    });
-                };
+                offlineContext = new window.OfflineAudioContext(1, newLength, newRate);
```
benjiwheeler: I don't understand what cases are being covered by these two different ways of getting offline audio context.

src/lib/audio/audio-util.js
```diff
@@ -59,9 +60,39 @@ const encodeAndAddSoundToVM = function (vm, samples, sampleRate, name, callback)
...
+const downsampleIfNeeded = (samples, sampleRate, resampler) => {
+    const duration = samples.length / sampleRate;
+    const encodedByteLength = samples.length * 2; /* bitDepth 16 bit */
+    // Resolve immediately if already within byte limit
+    if (encodedByteLength < SOUND_BYTE_LIMIT) {
+        return Promise.resolve({samples, sampleRate});
+    }
+    // If encodeable at 22khz, resample and call submitNewSamples again
+    if (duration * 22050 * 2 < SOUND_BYTE_LIMIT) {
+        return resampler({samples, sampleRate}, 22050);
```
benjiwheeler: I do follow how the sampler is being passed to downsampleIfNeeded and then returned here, but that seems hard to follow. Is there any way to simplify it, given that there's only one resampler actually used in the code?

benjiwheeler:
Still reviewing, but I have some questions in the meantime!

src/containers/sound-editor.jsx
paulkaplan: I asked the exact same question while writing this code. No, surprisingly, once it starts the only thing it can do is complete. Check out the `events` section in the sidebar https://developer.mozilla.org/en-US/docs/Web/API/OfflineAudioContext

src/containers/sound-editor.jsx
paulkaplan: deleteInverse is one of the operations you can take on a sound selection. It deletes everything you do not have selected (the inverse of the selection). You do this by shift+delete, so it is a bit of a hidden feature.

src/lib/audio/audio-util.js
paulkaplan: Ah yeah. So it is the "backup" method because the ideal method is to use the audio context to resample the audio, which involves some interpolation instead of just dropping every other sample, which introduces audio artifacts. But if the audio context does not support resampling at the desired frequency (like safari with anything other than 44khz), this is the "backup" option. @ericrosenbaum I think adding this as jsdoc (and generally jsdoc-ing the arguments to all these utility functions) would be a good idea

src/containers/sound-editor.jsx
paulkaplan: On safari you will throw trying to create an offline audio context using anything other than 44khz. The backup resampler is used in those situations. Leaving that this as a try/catch is to future proof if safari is brought closer to the spec in the future.

src/lib/audio/audio-util.js
paulkaplan: The dependency injection here is mostly to make unit testing the logic of `downsampleIfNeeded` separate from the resampling process itself.

src/containers/sound-editor.jsx
paulkaplan: `OfflineAudioContext` is not supported on all browsers we support. Specifically, safari still has the offline audio context under vendor prefix.

src/containers/sound-editor.jsx
benjiwheeler: I guess what I mean is, it would be clearer to just write `if (sampleRateRatio === 0.5)`

src/lib/audio/audio-util.js
benjiwheeler: I see -- there are two types of downsampling -- you might think of them as:
1)"downsampleUsingInterpolation" or maybe "downsampleSmartly" or "smartDownsample", and
2) "downsampleByDroppingSamples" or "dropEveryOtherSample" or "bruteForceDownsample".

I think drawing this distinction using a pair of functions with linked names might make their respective roles clearer. For me, the "backup" nature of 2) seems like it would be clearer if it came in the context of the branching logic, with the function name being more specific to what that function does to its data (rather than the name referring to the situation in which it is called). 

But, my function naming preferences aren't necessarily better -- I'm fine with what you prefer, as long as you consider a change like i'm suggesting.

src/containers/sound-editor.jsx
benjiwheeler: some comments explaining some of that would be helpful -- maybe something like "if browser doesn't support our desired sample rate, downsample using a brute force method"

src/containers/sound-editor.jsx
benjiwheeler: maybe add a comment before this test explaining that -- something like "there are two ways major browsers support resampling: window.OfflineAudioContext and window.webkitOfflineAudioContext"

and maybe before the line `offlineContext = new window.OfflineAudioContext(1, newLength, newRate);` you could comment "if window.OfflineAudioContext exists, we're confident we can use it to resample"

and before the `offlineContext = new window.webkitOfflineAudioContext(1, newLength, newRate);` line below you might comment "some webkit browsers are currently known to crash on some sample rates"

src/lib/audio/audio-util.js
benjiwheeler: that makes sense

benjiwheeler: Here's another way you might handle it:

change the function declaration line from
```
const downsampleIfNeeded = (samples, sampleRate, resampler) => {
```
to
```
const downsampleIfNeeded = (samples, sampleRate, resampler = this.resampleBufferToRate) => {
```

...then, when you call `downsampleIfNeeded()` in your actual code, the context will be simpler for an outsider to understand, and you can still test `downsampleIfNeeded` the same way.

Do you like that idea?

test/__mocks__/audio-effects.js
benjiwheeler: and it doesn't work without the setTimeout, like 

```
                return new Promise(resolve => { resolve(); }));
```
? Interesting/surprising!

src/containers/sound-editor.jsx
benjiwheeler: Ah, that makes sense. Could it make sense to add a comment above this function saying something like "deletes everything you do not have selected (the inverse of the selection)" ?

test/__mocks__/audio-effects.js
paulkaplan: Yeah unfortunately because there are several promises chained

src/containers/sound-editor.jsx
benjiwheeler: FWIW I think it's better to put a comment that describes the whole function above the function (or on the same line as its declaration)-- below the declaration line, it sorta looks like it just refers to the variable declarations.

src/containers/sound-editor.jsx
```diff
@@ -322,25 +327,39 @@ class SoundEditor extends React.Component {
...
-                    });
-                };
+        return new Promise((resolve, reject) => {
+            const sampleRateRatio = newRate / buffer.sampleRate;
+            const newLength = sampleRateRatio * buffer.samples.length;
+            let offlineContext;
+            // Try to use either OfflineAudioContext or webkitOfflineAudioContext to resample
+            // The constructors will throw if trying to resample at an unsupported rate
+            // (e.g. Safari/webkitOAC does not support lower than 44khz).
+            try {
```
benjiwheeler: I like this refactor

src/containers/sound-editor.jsx
```diff
@@ -322,25 +327,39 @@ class SoundEditor extends React.Component {
...
+                if (window.OfflineAudioContext) {
+                    offlineContext = new window.OfflineAudioContext(1, newLength, newRate);
+                } else if (window.webkitOfflineAudioContext) {
+                    offlineContext = new window.webkitOfflineAudioContext(1, newLength, newRate);
+                }
+            } catch {
+                // If no OAC available and downsampling by 2, downsample by dropping every other sample.
+                if (newRate === buffer.sampleRate / 2) {
```
benjiwheeler: do you like the idea of instead saying `if (sampleRateRatio === 0.5)` ?

benjiwheeler:
LGTM -- I just had one note you didn't address about sound-editor.jsx line 345, feel free to change or leave as-is.

benjiwheeler:
Tested with a track around 4 mins long, difference between build with your changes and without matches my expectations!

apple502j:
This would be a great Christmas gift!

benjiwheeler:
More of a new year's gift, calendar-wise ;)

PR #5343, opened 2019-12-17
Fix `doneUpdatingProject` call signature

taybenlor:
ProjectSaverHOC has been using the wrong call signature for `doneUpdatingProject`.

### Resolves

I didn't look for an issue, since this doesn't fix any behaviour.

### Proposed Changes

The call signature for `doneUpdatingProject` expects a single argument, the `loadingState`. However in the `mapDispatchToProps` it was being called with two arguments, but in a way that was actually logically correct. This change reduces it to one argument to match the function definition and usage.

Relevant code sections:

- `project-saver-hoc.jsx:updateProjectToStorage` uses `this.props.onUpdatedProject` with just a loading state
- `project-saver-hoc.jsx:mapDispatchToProps` defines `onUpdatedProject` as a function with two arguments and passes both to `doneUpdatingProject`
- `project-state.js:doneUpdatingProject` defines `doneUpdatingProject` as a function that takes a single argument (loading state)

### Reason for Changes

This looks like a mistake.

### Test Coverage

There is no change in behaviour from this code change.

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/lib/project-saver-hoc.jsx
+++ b/src/lib/project-saver-hoc.jsx
@@ -433,7 +433,7 @@ const ProjectSaverHOC = function (WrappedComponent) {
         onShowCreatingRemixAlert: () => showAlertWithTimeout(dispatch, 'creatingRemix'),
         onShowSaveSuccessAlert: () => showAlertWithTimeout(dispatch, 'saveSuccess'),
         onShowSavingAlert: () => showAlertWithTimeout(dispatch, 'saving'),
-        onUpdatedProject: (projectId, loadingState) => dispatch(doneUpdatingProject(projectId, loadingState)),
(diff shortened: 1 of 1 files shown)


benjiwheeler:
Thanks for bringing this to our attention, we'll take a look!

benjiwheeler:
This tests fine, and it's a small change. I see how we didn't catch it before -- there is no actual bug, just a remnant of another version of the code that adds an undefined param that isn't used.

Thanks for catching and fixing this!

taybenlor:
No worries! Happy holidays!

Issue #5361, opened 2020-01-06
when default project can't load, project state should transition to error


benjiwheeler:
### Expected Behavior

* GUI attempts to load default project
* something goes wrong, and it cannot load
* GUI shows error screen

### Actual Behavior

* whatever is already in vm, title just stays there, even if both are blank

PR #5362, opened 2020-01-06
refactor onLoadedProject() creator to handle success==false consistently; add tests of project state action creators

benjiwheeler:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/5361

### Proposed Changes

Adds unit tests for many project state reducer action creator functions.

Also refactors the project state reducer's `onLoadedProject()` function to be clearer, and to specifically handle failures to load the default project by switching to an error state.

(This relates to the project state reducer's handling of success and failure in loading the default project. Since we get that project data locally, and not from the server, loading it into the VM should always succeed. If it fails, we should switch to an error state.)

This change doesn't make a change to actual behavior, only to the project state reducer's API for clarity and consistency.

### DIFF
--- a/src/reducers/project-state.js
+++ b/src/reducers/project-state.js
@@ -17,7 +17,7 @@ const START_AUTO_UPDATING = 'scratch-gui/project-state/START_AUTO_UPDATING';
 const START_CREATING_NEW = 'scratch-gui/project-state/START_CREATING_NEW';
 const START_ERROR = 'scratch-gui/project-state/START_ERROR';
 const START_FETCHING_NEW = 'scratch-gui/project-state/START_FETCHING_NEW';
-const START_LOADING_VM_FILE_UPLOAD = 'scratch-gui/project-state/START_LOADING_FILE_UPLOAD';
+const START_LOADING_VM_FILE_UPLOAD = 'scratch-gui/project-state/START_LOADING_VM_FILE_UPLOAD';
 const START_MANUAL_UPDATING = 'scratch-gui/project-state/START_MANUAL_UPDATING';
 const START_REMIXING = 'scratch-gui/project-state/START_REMIXING';
 const START_UPDATING_BEFORE_CREATING_COPY = 'scratch-gui/project-state/START_UPDATING_BEFORE_CREATING_COPY';
@@ -435,10 +435,21 @@ const onLoadedProject = (loadingState, canSave, success) => {
         default:
             return;
         }
+    } else {
+        switch (loadingState) {
+        case LoadingState.LOADING_VM_WITH_ID:
+        case LoadingState.LOADING_VM_FILE_UPLOAD:
+            return {
+                type: RETURN_TO_SHOWING
+            };
+        case LoadingState.LOADING_VM_NEW_DEFAULT:
+            return {
+                type: START_ERROR
+            };
+        default:
+            return;
+        }
     }
-    return {
-        type: RETURN_TO_SHOWING
-    };
 };
 
 const doneUpdatingProject = loadingState => {
--- a/src/reducers/project-state.js
+++ b/src/reducers/project-state.js
@@ -413,42 +413,30 @@ const onFetchedProjectData = (projectData, loadingState) => {
 };
 
 const onLoadedProject = (loadingState, canSave, success) => {
-    if (success) {
-        switch (loadingState) {
-        case LoadingState.LOADING_VM_WITH_ID:
-            return {
-                type: DONE_LOADING_VM_WITH_ID
-            };
-        case LoadingState.LOADING_VM_FILE_UPLOAD:
+    switch (loadingState) {
+    case LoadingState.LOADING_VM_WITH_ID:
+        if (success) {
+            return {type: DONE_LOADING_VM_WITH_ID};
+        }
+        // failed to load project; just keep showing current project
+        return {type: RETURN_TO_SHOWING};
+    case LoadingState.LOADING_VM_FILE_UPLOAD:
+        if (success) {
             if (canSave) {
-                return {
-                    type: DONE_LOADING_VM_TO_SAVE
-                };
+                return {type: DONE_LOADING_VM_TO_SAVE};
             }
-            return {
-                type: DONE_LOADING_VM_WITHOUT_ID
-            };
-        case LoadingState.LOADING_VM_NEW_DEFAULT:
-            return {
-                type: DONE_LOADING_VM_WITHOUT_ID
-            };
-        default:
-            return;
+            return {type: DONE_LOADING_VM_WITHOUT_ID};
         }
-    } else {
-        switch (loadingState) {
-        case LoadingState.LOADING_VM_WITH_ID:
-        case LoadingState.LOADING_VM_FILE_UPLOAD:
-            return {
-                type: RETURN_TO_SHOWING
-            };
-        case LoadingState.LOADING_VM_NEW_DEFAULT:
-            return {
-                type: START_ERROR
-            };
-        default:
-            return;
+        // failed to load project; just keep showing current project
+        return {type: RETURN_TO_SHOWING};
+    case LoadingState.LOADING_VM_NEW_DEFAULT:
+        if (success) {
+            return {type: DONE_LOADING_VM_WITHOUT_ID};
         }
+        // failed to load default project; show error
+        return {type: START_ERROR};
+    default:
+        return;
     }
 };
 
--- a/test/unit/reducers/project-state-reducer.test.js
+++ b/test/unit/reducers/project-state-reducer.test.js
@@ -96,142 +96,173 @@ test('onFetchedProjectData new loads project data into vm', () => {
 
 test('onLoadedProject(LOADING_VM_WITH_ID, true, true) results in state SHOWING_WITH_ID', () => {
     const initialState = {
+        projectId: '100',
         loadingState: LoadingState.LOADING_VM_WITH_ID
     };
     const action = onLoadedProject(initialState.loadingState, true, true);
     const resultState = projectStateReducer(initialState, action);
     expect(resultState.loadingState).toBe(LoadingState.SHOWING_WITH_ID);
+    expect(resultState.projectId).toBe('100');
 });
 
 test('onLoadedProject(LOADING_VM_WITH_ID, false, true) results in state SHOWING_WITH_ID', () => {
     const initialState = {
+        projectId: '100',
         loadingState: LoadingState.LOADING_VM_WITH_ID
     };
-    const action = onLoadedProject(initialState.loadingState, true, true);
+    const action = onLoadedProject(initialState.loadingState, false, true);
     const resultState = projectStateReducer(initialState, action);
     expect(resultState.loadingState).toBe(LoadingState.SHOWING_WITH_ID);
+    expect(resultState.projectId).toBe('100');
 });
 
+// case where we started out viewing a project with a projectId, then
+// started to load another project; but loading fails. We go back to
+// showing the original project.
 test('onLoadedProject(LOADING_VM_WITH_ID, false, false), with project id, ' +
     'results in state SHOWING_WITH_ID', () => {
     const initialState = {
-        loadingState: LoadingState.LOADING_VM_WITH_ID,
-        projectId: '12345'
+        projectId: '100',
+        loadingState: LoadingState.LOADING_VM_WITH_ID
     };
     const action = onLoadedProject(initialState.loadingState, false, false);
     const resultState = projectStateReducer(initialState, action);
     expect(resultState.loadingState).toBe(LoadingState.SHOWING_WITH_ID);
+    expect(resultState.projectId).toBe('100');
 });
 
+// case where we started out viewing a project with default projectId, then
+// started to load one with an id, such as in standalone mode when user adds
+// '#PROJECT_ID_NUMBER' to the URI; but loading fails. We go back to
+// showing the original project.
 test('onLoadedProject(LOADING_VM_WITH_ID, false, false), with no project id, ' +
     'results in state SHOWING_WITHOUT_ID', () => {
     const initialState = {
-        loadingState: LoadingState.LOADING_VM_WITH_ID,
-        projectId: null
+        projectId: '0',
+        loadingState: LoadingState.LOADING_VM_WITH_ID
     };
     const action = onLoadedProject(initialState.loadingState, false, false);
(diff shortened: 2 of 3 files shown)


src/reducers/project-state.js
```diff
@@ -435,10 +435,21 @@ const onLoadedProject = (loadingState, canSave, success) => {
         default:
             return;
         }
+    } else {
```
kchadha: This code seems to be getting more confusing because of the combination of `if` and `switch`. I realize that's the way the code was already written, but for example, I think this `else` is not needed since there's a `default: return` above. However, that fact is not immediately obvious/easy to reason about. 

Do you have any thoughts on how we could simplify this?

src/reducers/project-state.js
```diff
@@ -435,10 +435,21 @@ const onLoadedProject = (loadingState, canSave, success) => {
...
+        switch (loadingState) {
+        case LoadingState.LOADING_VM_WITH_ID:
+        case LoadingState.LOADING_VM_FILE_UPLOAD:
+            return {
+                type: RETURN_TO_SHOWING
```
kchadha: What is the difference between RETURN_TO_SHOWING and START_ERROR? Since I am not familiar w/the project state machine, I think it would be helpful to have comments describing the state changes that are happening (this does not necessarily need to be added to this PR but would be useful to have in the future).

test/unit/reducers/project-state-reducer.test.js
```diff
@@ -92,16 +92,49 @@ test('onFetchedProjectData new loads project data into vm', () => {
     expect(resultState.projectData).toBe('1010101');
 });
 
-test('onLoadedProject upload, with canSave false, shows without id', () => {
+// onLoadedProject: LOADING_VM_WITH_ID
```
kchadha: Cruft?

test/unit/reducers/project-state-reducer.test.js
```diff
@@ -110,25 +143,38 @@ test('onLoadedProject upload, with canSave true, prepares to save', () => {
...
-        loadingState: LoadingState.LOADING_VM_WITH_ID
+        loadingState: LoadingState.LOADING_VM_FILE_UPLOAD
     };
-    const action = onLoadedProject(initialState.loadingState, true, true);
+    const action = onLoadedProject(initialState.loadingState, false, true);
+    const resultState = projectStateReducer(initialState, action);
+    expect(resultState.loadingState).toBe(LoadingState.SHOWING_WITHOUT_ID);
+});
+
+test('onLoadedProject (LOADING_VM_FILE_UPLOAD, false, false), with project id, shows with id', () => {
```
kchadha: Not sure I understand what's going on in this test...

src/reducers/project-state.js
benjiwheeler: Thanks for pointing this out. I tried using one switch statement instead, and I think it's better:

```
    switch (loadingState) {
    case LoadingState.LOADING_VM_WITH_ID:
        if (success) {
            return {type: DONE_LOADING_VM_WITH_ID};
        }
        // failed to load project; just keep showing current project
        return {type: RETURN_TO_SHOWING};
    case LoadingState.LOADING_VM_FILE_UPLOAD:
        if (success) {
            if (canSave) {
                return {type: DONE_LOADING_VM_TO_SAVE};
            }
            return {type: DONE_LOADING_VM_WITHOUT_ID};
        }
        // failed to load project; just keep showing current project
        return {type: RETURN_TO_SHOWING};
    case LoadingState.LOADING_VM_NEW_DEFAULT:
        if (success) {
            return {type: DONE_LOADING_VM_WITHOUT_ID};
        }
        // failed to load default project; show error
        return {type: START_ERROR};
    default:
        return;
    }
```

src/reducers/project-state.js
benjiwheeler: Thanks -- I'll put in a bit of commenting here, and I have done some experiments on how to add documentation to this file that I'd like to show you

test/unit/reducers/project-state-reducer.test.js
benjiwheeler: the test names are not exactly very clear! Let me try revising them.

benjiwheeler:
Please take a look -- I revised a few things, and I think made it a bit clearer!

test/unit/reducers/project-state-reducer.test.js
benjiwheeler: Is the revised name more helpful?

src/reducers/project-state.js
```diff
@@ -413,32 +413,31 @@ const onFetchedProjectData = (projectData, loadingState) => {
...
+            return {type: DONE_LOADING_VM_WITHOUT_ID};
         }
+        // failed to load project; just keep showing current project
+        return {type: RETURN_TO_SHOWING};
+    case LoadingState.LOADING_VM_NEW_DEFAULT:
+        if (success) {
+            return {type: DONE_LOADING_VM_WITHOUT_ID};
+        }
+        // failed to load default project; show error
+        return {type: START_ERROR};
```
kchadha: Just to be clear, this is the piece of logic that is new to this PR, right? 
The rest of this is chunk of code is a refactor?

src/reducers/project-state.js
kchadha: This feels much better/easier to reason about!

test/unit/reducers/project-state-reducer.test.js
```diff
@@ -92,16 +92,51 @@ test('onFetchedProjectData new loads project data into vm', () => {
...
+    const action = onLoadedProject(initialState.loadingState, true, true);
+    const resultState = projectStateReducer(initialState, action);
+    expect(resultState.loadingState).toBe(LoadingState.SHOWING_WITH_ID);
+});
+
+test('onLoadedProject(LOADING_VM_WITH_ID, false, true) results in state SHOWING_WITH_ID', () => {
+    const initialState = {
+        loadingState: LoadingState.LOADING_VM_WITH_ID
+    };
+    const action = onLoadedProject(initialState.loadingState, true, true);
```
kchadha: Is this one supposed to be `false, true`? It looks like this test is exactly the same as the one above...

benjiwheeler: Will fix!

test/unit/reducers/project-state-reducer.test.js
```diff
@@ -92,16 +92,51 @@ test('onFetchedProjectData new loads project data into vm', () => {
...
+    const initialState = {
+        loadingState: LoadingState.LOADING_VM_WITH_ID,
+        projectId: '12345'
+    };
+    const action = onLoadedProject(initialState.loadingState, false, false);
+    const resultState = projectStateReducer(initialState, action);
+    expect(resultState.loadingState).toBe(LoadingState.SHOWING_WITH_ID);
+});
+
+test('onLoadedProject(LOADING_VM_WITH_ID, false, false), with no project id, ' +
```
benjiwheeler: What should happen here? State is LOADING_VM_WITH_ID, but projectId is null -- is that even possible?

test/unit/reducers/project-state-reducer.test.js
```diff
@@ -92,16 +92,51 @@ test('onFetchedProjectData new loads project data into vm', () => {
...
+    const initialState = {
+        loadingState: LoadingState.LOADING_VM_WITH_ID
+    };
+    const action = onLoadedProject(initialState.loadingState, true, true);
+    const resultState = projectStateReducer(initialState, action);
+    expect(resultState.loadingState).toBe(LoadingState.SHOWING_WITH_ID);
+});
+
+test('onLoadedProject(LOADING_VM_WITH_ID, false, false), with project id, ' +
+    'results in state SHOWING_WITH_ID', () => {
```
kchadha: Could you add a comment here that explains why it results in state "SHOWING_WITH_ID"? That we're showing the original project w/ID (not the one that failed to load)

test/unit/reducers/project-state-reducer.test.js
kchadha: Would be good to add a comment here explaining that this is for the case where we started viewing a project that doesn't have an ID (e.g. in standalone mode where we go from viewing a project w/o an ID and then add an id using the `#....` method in the URL

test/unit/reducers/project-state-reducer.test.js
```diff
@@ -110,25 +145,40 @@ test('onLoadedProject upload, with canSave true, prepares to save', () => {
...
+
+test('onLoadedProject(LOADING_VM_FILE_UPLOAD, false, false), when we know project id, ' +
+    'results in state SHOWING_WITH_ID', () => {
+    const initialState = {
+        loadingState: LoadingState.LOADING_VM_FILE_UPLOAD,
+        projectId: '12345'
+    };
+    const action = onLoadedProject(initialState.loadingState, false, false);
     const resultState = projectStateReducer(initialState, action);
     expect(resultState.loadingState).toBe(LoadingState.SHOWING_WITH_ID);
```
benjiwheeler: We should be also checking that the projectid is what we expect -- in each expect line

test/unit/reducers/project-state-reducer.test.js
benjiwheeler: Done

test/unit/reducers/project-state-reducer.test.js
benjiwheeler: Done

test/unit/reducers/project-state-reducer.test.js
benjiwheeler: Done

src/reducers/project-state.js
benjiwheeler: Yes.

test/unit/reducers/project-state-reducer.test.js
benjiwheeler: Done

benjiwheeler:
@kchadha Made the changes we discussed!

test/unit/reducers/project-state-reducer.test.js
```diff
@@ -92,99 +92,181 @@ test('onFetchedProjectData new loads project data into vm', () => {
...
+});
+
+// case where we started out viewing a project with default projectId, then
+// started to load one with an id, such as in standalone mode when user adds
+// '#PROJECT_ID_NUMBER' to the URI; but loading fails. We go back to
+// showing the original project.
+test('onLoadedProject(LOADING_VM_WITH_ID, false, false), with no project id, ' +
+    'results in state SHOWING_WITHOUT_ID', () => {
+    const initialState = {
+        projectId: '0',
```
kchadha: Hmmm did the nature of this test change here by adding the projectId field? E.g. is it the same test as the one above or does the '0' here have special meaning that exercises a different code path?

benjiwheeler: That's a good question. '0' is the specific projectId value of the default project. It's defined at the top of the project-state reducer file.

So this test's expecting to reach the state SHOWING_WITHOUT_ID, whereas the test above is expecting to reach the state SHOWING_WITH_ID.

benjiwheeler:
@kchadha I'm realizing I named this PR poorly at the start; this was intended as a change in project state reducer to support further changes in vm-manager-hoc. 

So this change only changes the project state reducer's onLoadedProject() action creator to be more consistent about how it will handle the cases when the `success` parameter is false. It doesn't actually change functionality yet -- it just sets us up to do that next.

The change it was saying it made -- switching to an error state if loading the default project fails -- is actually satisfied by gui's current functionality! It's other cases that don't work as intended: if loading fails for a different fetched project, currently gui will enter an error state without even calling onLoadedProject(). 

After this change, we would/will be able to change vm-manager-hoc to call onLoadedProject() with the `success` parameter `false`, so that it can go into an error state for the default project, but *avoid* going into one for other projects.

benjiwheeler:
Also, I replied to your outstanding question on that test!

benjiwheeler:
Here's a patch that you can apply locally, which will make vm loading fail 50% of the time, and call `onLoadedProject()` with the resulting success or failure:

```
diff --git a/src/containers/sb-file-uploader.jsx b/src/containers/sb-file-uploader.jsx
index 2019f719e..14e2a2ba9 100644
--- a/src/containers/sb-file-uploader.jsx
+++ b/src/containers/sb-file-uploader.jsx
@@ -127,6 +127,8 @@ class SBFileUploader extends React.Component {
             const filename = this.fileToUpload && this.fileToUpload.name;
             this.props.vm.loadProject(this.reader.result)
                 .then(() => {
+                    const success = (Math.random() > .5);
+                    if (!success) throw "random error";
                     this.props.onLoadingFinished(this.props.loadingState, true);
                     // Reset the file input after project is loaded
                     // This is necessary in case the user wants to reload a project
diff --git a/src/lib/vm-manager-hoc.jsx b/src/lib/vm-manager-hoc.jsx
index 7947dfd64..42170df2b 100644
--- a/src/lib/vm-manager-hoc.jsx
+++ b/src/lib/vm-manager-hoc.jsx
@@ -54,7 +54,9 @@ const vmManagerHOC = function (WrappedComponent) {
         loadProject () {
             return this.props.vm.loadProject(this.props.projectData)
                 .then(() => {
-                    this.props.onLoadedProject(this.props.loadingState, this.props.canSave);
+                    const success = (Math.random() > .5);
+                    if (!success) throw "random error";
+                    this.props.onLoadedProject(this.props.loadingState, this.props.canSave, true);
                     // Wrap in a setTimeout because skin loading in
                     // the renderer can be async.
                     setTimeout(() => this.props.onSetProjectUnchanged());
@@ -71,7 +73,7 @@ const vmManagerHOC = function (WrappedComponent) {
                     }
                 })
                 .catch(e => {
-                    this.props.onError(e);
+                    this.props.onLoadedProject(this.props.loadingState, this.props.canSave, false);
                 });
         }
         render () {
@@ -137,8 +139,8 @@ const vmManagerHOC = function (WrappedComponent) {

         onError: error => dispatch(projectError(error)),
-        onLoadedProject: (loadingState, canSave) =>
-            dispatch(onLoadedProject(loadingState, canSave, true)),
+        onLoadedProject: (loadingState, canSave, success) =>
+            dispatch(onLoadedProject(loadingState, canSave, success)),
         onSetProjectUnchanged: () => dispatch(setProjectUnchanged())
     });

```

When I run gui locally with this, what I see is
* errors for loading the default project result in an error screen
* errors for loading any other project result in still showing the editor normally

...which is what's desired.

Issue #4907, opened 2019-06-06
Modals should handle clicks on them, and not have other components see those clicks


benjiwheeler:
### Expected Behavior

If a menu is open and a modal is open too, and you click on the modal, the click should be handled by the modal, without causing the menu to close.

### Actual Behavior

The mouseup causes the menu to close before the modal handles the mouseup.

This can pose a problem if the modal has been presented by the menu. We ran into a situation with `sb-file-uploader` where clicking on its modal caused the menu to close -- which caused `sb-file-uploader` to unmount. 

### Can't reproduce

You can't run into this using current code, just with code we've tried to introduce that hasn't worked because of this problem

PR #5364, opened 2020-01-06
don't close menus on mouse click, if element clicked is modal

benjiwheeler:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/4907

### Proposed Changes

When a menu bar menu handles mouseup, it should check if the mouse was in a modal; if so, the menu should not close itself.

### Reason for Changes

It became clear in working on https://github.com/LLK/scratch-gui/issues/4762 that this issue was a blocker for introducing a custom confirm modal for project upload.

### Test Coverage

None -- we don't yet have a modal that would be opened by a menu, so it's hard to test this with existing code.

### DIFF
--- a/src/containers/menu.jsx
+++ b/src/containers/menu.jsx
@@ -30,8 +30,18 @@ class Menu extends React.Component {
     removeListeners () {
         document.removeEventListener('mouseup', this.handleClick);
     }
+    elementIsModal (element) {
+        const modalPortals = document.getElementsByClassName('ReactModalPortal');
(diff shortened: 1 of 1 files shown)


benjiwheeler:
blocked on finding someone to review

benjiwheeler:
update: talked to @rschamp, he felt this was a hacky direction and it might make sense to not try this. Instead, maybe we should do what MacOS does, and close the menu when a selection causes a dialog/modal to be opened.

PR #5377, opened 2020-01-09
[fixup] Implement sound downsampler to try to allow more sounds to be edited

paulkaplan:
This is a fixed version of https://github.com/LLK/scratch-gui/pull/5338 which was reverted due to a bug that is fixed in 76f5cfe

### DIFF
--- a/test/unit/util/audio-util.test.js
+++ b/test/unit/util/audio-util.test.js
@@ -1,4 +1,9 @@
-import {computeRMS, computeChunkedRMS} from '../../../src/lib/audio/audio-util';
+import {
+    computeRMS,
+    computeChunkedRMS,
+    downsampleIfNeeded,
+    dropEveryOtherSample
+} from '../../../src/lib/audio/audio-util';
 
 describe('computeRMS', () => {
     test('returns 0 when given no samples', () => {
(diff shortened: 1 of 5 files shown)


test/unit/util/audio-util.test.js
```diff
@@ -87,11 +87,15 @@ describe('dropEveryOtherSample', () => {
         sampleRate: 2
     };
     test('result is half the length', () => {
-        const {samples} = dropEveryOtherSample(buffer, 1);
+        const {samples} = dropEveryOtherSample(buffer);
         expect(samples.length).toEqual(Math.floor(buffer.samples.length / 2));
     });
     test('result contains only even-index items', () => {
-        const {samples} = dropEveryOtherSample(buffer, 1);
```
paulkaplan: These extra args were leftover from a previous incarnation of the function, another thing unit testing didn't show.

test/unit/util/audio-util.test.js
```diff
@@ -87,11 +87,15 @@ describe('dropEveryOtherSample', () => {
...
+        const {samples} = dropEveryOtherSample(buffer);
         expect(samples).toEqual(new Float32Array([1, 2, 3]));
     });
+    test('result sampleRate is given sampleRate / 2', () => {
+        const {sampleRate} = dropEveryOtherSample(buffer);
+        expect(sampleRate).toEqual(buffer.sampleRate / 2);
```
paulkaplan: Confirmed this test failed without the fix, it was giving NaN's which is what we saw in the bughunt

Issue #5084, opened 2019-08-09
Library should expect 3.0 sprite format


ericrosenbaum:
Currently, the library json files contain sprite and asset info generated from 2.0 projects. Once we've updated the library generator to use 3.0 projects, the GUI will need to handle the changes to a few properties. Here are the changes I've found so far:

- The library item container should use the `md5ext` instead of `baseLayerMD5` to animate the sprite's costumes on hover
- The randomize sprite position util should set the `x` and `y` properties instead of `scratchX` and `scratchY`  (Paul: this one is already handled)
- lib/empty-assets.js may need to be changed to use the 3.0 format as well, depending on how it is used.

Another note: to eliminate dependency on library generation using 3.0 syntax, it would be best to implement this in a way that supports both formats, that way we can release without waiting for the actual library to change, which may be blocked by other issues.

paulkaplan:
@ericrosenbaum looked into this, I think the second one you mention is not needed because the code already handles 3.0 sprite objects

PR #5385, opened 2020-01-14
Update library-item to handle 3.0 or 2.0 library formats

paulkaplan:
### Resolves

- Resolves https://github.com/LLK/scratch-gui/issues/5084

### Proposed Changes

Updates the `library-item` which was expecting library items to have the `baseLayerMD5` property to now use either `baseLayerMD5` or `md5ext`, which is what is what the libraries json will have once we start using the new library generator that works on 3.0 projects. 

This _does not_ update the library JSON files, as there are other blockers for that.

### Reason for Changes

See https://github.com/LLK/scratch-gui/issues/5084 for more

### Test Coverage

Manually tested that costumes rotate correctly using the existing format, so this can be merged right away. Also tested using the new library format which is in progress, and it works for that as well. There is an automated test around costume rotation that should continue to work both here and (maybe more importantly) when the libraries themselves are updated

### DIFF
--- a/src/containers/library-item.jsx
+++ b/src/containers/library-item.jsx
@@ -92,14 +92,16 @@ class LibraryItem extends React.PureComponent {
         this.setState({iconIndex: nextIconIndex});
     }
     curIconMd5 () {
+        const iconMd5Prop = this.props.iconMd5;
         if (this.props.icons &&
             this.state.isRotatingIcon &&
-            this.state.iconIndex < this.props.icons.length &&
-            this.props.icons[this.state.iconIndex] &&
-            this.props.icons[this.state.iconIndex].baseLayerMD5) {
-            return this.props.icons[this.state.iconIndex].baseLayerMD5;
+            this.state.iconIndex < this.props.icons.length) {
+            const icon = this.props.icons[this.state.iconIndex] || {};
+            return icon.md5ext || // 3.0 library format
+                icon.baseLayerMD5 || // 2.0 library format, TODO GH-5084
+                iconMd5Prop;
         }
-        return this.props.iconMd5;
+        return iconMd5Prop;
     }
     render () {
         const iconMd5 = this.curIconMd5();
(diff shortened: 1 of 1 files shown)


src/containers/library-item.jsx
```diff
@@ -92,14 +92,16 @@ class LibraryItem extends React.PureComponent {
...
         if (this.props.icons &&
             this.state.isRotatingIcon &&
-            this.state.iconIndex < this.props.icons.length &&
-            this.props.icons[this.state.iconIndex] &&
-            this.props.icons[this.state.iconIndex].baseLayerMD5) {
-            return this.props.icons[this.state.iconIndex].baseLayerMD5;
+            this.state.iconIndex < this.props.icons.length) {
+            const icon = this.props.icons[this.state.iconIndex] || {};
+            return icon.md5ext || // 3.0 library format
+                icon.baseLayerMD5 || // 2.0 library format, TODO GH-5084
```
paulkaplan: This line and the corresponding prop type can be removed after we update the library JSON files

src/containers/library-item.jsx
```diff
@@ -92,14 +92,16 @@ class LibraryItem extends React.PureComponent {
...
     curIconMd5 () {
+        const iconMd5Prop = this.props.iconMd5;
         if (this.props.icons &&
             this.state.isRotatingIcon &&
-            this.state.iconIndex < this.props.icons.length &&
-            this.props.icons[this.state.iconIndex] &&
-            this.props.icons[this.state.iconIndex].baseLayerMD5) {
-            return this.props.icons[this.state.iconIndex].baseLayerMD5;
+            this.state.iconIndex < this.props.icons.length) {
+            const icon = this.props.icons[this.state.iconIndex] || {};
```
paulkaplan: Tried to make the logic here a bit more readable. Using || {} to make it possible to access properties on a possibly undefined object, and using md5ext || baseLayer || iconMd5Prop to get to the same result without so many if conditions.

PR #5395, opened 2020-01-23
Asset library update

ericrosenbaum:
Now that we have fixed the library generator to work with 3.0 projects, we can do a full update of the libraries. 

Blocking issues:
- [x] Fix missing tags on watermelon costumes


adroitwhiz:
Any reason why the tags have all been removed from certain costumes (e.g. the watermelons)?

paulkaplan:
@adroitwhiz thanks for noticing that. This is a work in progress, we are having some issues with the process of updating those libraries. @ericrosenbaum I've added a checklist section the the PR body for tracking what things are blocking. I know you had generated a list,  could your list to the PR body?

paulkaplan:
@adroitwhiz if you've noticed others feel free to comment with more, i'll add them to the checklist

src/lib/libraries/costumes.json
```diff
@@ -6811,9 +6815,7 @@
         "name": "Marian-a",
         "md5": "e9577a1eb098905dd386135bb38c0398.png",
         "type": "costume",
-        "tags": [
-            "people"
-        ],
+        "tags": [],
```
adroitwhiz: Tags disappeared here too

src/lib/libraries/sounds.json
```diff
@@ -676,7 +676,7 @@
     {
         "name": "C2 Sax",
         "md5": "ea8d34b18c3d8fe328cea201666458bf.wav",
-        "sampleCount": 7349,
+        "sampleCount": 8129,
         "rate": 22050,
         "format": "adpcm",
```
adroitwhiz: Is it worth converting these to regular .wavs so they can be decoded by the browser?

src/lib/libraries/sounds.json
```diff
@@ -816,19 +816,16 @@
...
-        "sampleCount": 2912,
-        "rate": 11025,
+        "sampleCount": 11648,
+        "rate": 44100,
         "format": "",
-        "tags": [
-            "effects",
-            "human"
-        ]
+        "tags": []
```
adroitwhiz: Tags gone here as well

src/lib/libraries/sounds.json
```diff
@@ -926,26 +923,42 @@
...
+    },
+    {
+        "name": "Computer Beeps2",
+        "md5": "28c76b6bebd04be1383fe9ba4933d263.wav",
+        "sampleCount": 38144,
+        "rate": 44100,
+        "format": "",
+        "tags": []
+    },
```
adroitwhiz: These have the same md5 as "Computer Beep2" and "Computer Beep"

src/lib/libraries/sounds.json
```diff
@@ -2271,18 +2284,15 @@
...
-        "sampleCount": 2464,
-        "rate": 22050,
+        "sampleCount": 4928,
+        "rate": 44100,
         "format": "",
-        "tags": [
-            "human",
-            "percussion"
-        ]
+        "tags": []
```
adroitwhiz: Tags are gone here

src/lib/libraries/costumes.json
ericrosenbaum: fixed!

src/lib/libraries/sounds.json
ericrosenbaum: this is a reasonable question - it's a tradeoff of file size vs js decoding time I guess- ADPCM is smaller but takes a little time to decode.  the other issue is that we don't generally save at 22kHz (i.e. sounds are always saved at 44kHz, except very long sounds are now saved at 22kHz, thanks to a recent update), and we no longer save in adcpm format. so any updates we make to library asset sounds will make them increase in size by 2x or more.

src/lib/libraries/sounds.json
ericrosenbaum: fixed

src/lib/libraries/sounds.json
ericrosenbaum: fixed

ericrosenbaum:
@BryceLTaylor once this is merged lets talk about how to properly QA it. basically all the contents of all the libraries are affected, so maybe we can schedule some time to test it manually before the bug hunt to make sure we don't miss anything.

ericrosenbaum:
I wrote a little script to check this over and generated a list of the 40 sounds that are currently missing tags (i.e. pre-existing issue). I'll file an issue to fix that separately.

ericrosenbaum:
We found a blocking issue: newly added sprites do not appear in randomized positions. For example if you add the Abby sprite twice, the second one is in the same place as the first (on production they are in different places).

ericrosenbaum:
blocking issue: first costume of the sprite should always be selected when the sprite loads (we can update libgen to make sure this is true).

ericrosenbaum:
Another issue (not sure if it's a blocker): some sprites show up in the sprite info area with lowercase names (they should always be title case). e.g the Dragon sprite loads as "dragon" and "Giga Walking" loads as "Giga walking".

apple502j:
> We found a blocking issue: newly added sprites do not appear in randomized positions. For example if you add the Abby sprite twice, the second one is in the same place as the first (on production they are in different places).

This is because library sprites have `json` key. In https://github.com/LLK/scratch-gui/blob/develop/src/lib/randomize-sprite-position.js#L5, library item with `json` key is considered "2.0 format". Maybe you can swap those conditions, so that it checks X and Y first?

paulkaplan:
Sounds like a good idea, Eric do you want that together in this pr or
separate?

ericrosenbaum:
@paulkaplan what @apple502j  is pointing out is part of the larger issue that the library generator is still generating 2.0-style sprite data. at this point it's sort of a hybrid of the 2.0 and 3.0 style formats, which is the cause of the bugs we're seeing (using both `currentCostume` and `currentCostumeIndex`, having the sprite name in three (!) places instead of one, etc). The library sprites now have both `draggable` and `isDraggable` properties.  I'm investigating...

paulkaplan:
Ohh I see. Hm let’s fix libgen then?

ericrosenbaum:
I did some work on changing libgen to generate 3.0-schema compliant library data, and on fixing the GUI to work with that, but it started to feel complicated and a bit risky. So I went back and fixed the known issues in this PR by making small fixes in libgen, and one in GUI.

src/lib/libraries/costumes.json
```diff
@@ -2453,7 +2453,7 @@
     },
     {
         "name": "Dee-d",
-        "md5": "2159a6be8f7ff450d0b5e938ca34a3f4.svg",
+        "md5": "320a892c86e9b039ba9d6d50a4897276.svg",
```
fsih: I like this one

ericrosenbaum: hahahah nice

src/lib/libraries/costumes.json
```diff
@@ -3003,17 +3003,6 @@
             2
         ]
     },
-    {
```
fsih: Just to double check, is this gone intentionally?

ericrosenbaum: yup- it's a duplicate

PR #5397, opened 2020-01-28
initialize Google Analytics only if GA_ID is set

cwillisf:
Resolves https://github.com/LLK/scratch-gui/issues/5775

### Proposed Changes

Call `GoogleAnalytics.initialize(...)` only if either `process.env.GA_ID` or `window.GA_ID` is set.

### Reason for Changes

It turns out that `GoogleAnalytics.initialize(...)` causes the download of `https://www.google-analytics.com/analytics.js` even if passed a null or undefined GA_ID. Skipping the call means we don't download that file in cases where we don't need it. In particular, we shouldn't try to download that file in Scratch Desktop since a) we should assume that the Internet is not available and b) we don't intend to use Google Analytics in Scratch Desktop.

### Test Coverage

We don't have automated tests for our usage of GA, but I've tested this locally in the context of Scratch Desktop. I'm unsure of how best to test this in the context of scratch-www or scratch-gui standalone. Any advice is welcome.

### Notes for Code Review

This diff is easier to read with "Hide whitespace changes" turned on.

### DIFF
--- a/src/lib/analytics.js
+++ b/src/lib/analytics.js
@@ -1,10 +1,13 @@
 import GoogleAnalytics from 'react-ga';
 
-GoogleAnalytics.initialize(process.env.GA_ID || window.GA_ID, {
-    debug: (process.env.NODE_ENV !== 'production'),
-    titleCase: true,
-    sampleRate: (process.env.NODE_ENV === 'production') ? 100 : 0,
-    forceSSL: true
-});
+const GA_ID = (process.env.GA_ID || window.GA_ID);
+if (GA_ID) {
+    GoogleAnalytics.initialize(GA_ID, {
+        debug: (process.env.NODE_ENV !== 'production'),
+        titleCase: true,
+        sampleRate: (process.env.NODE_ENV === 'production') ? 100 : 0,
+        forceSSL: true
+    });
+}
 
 export default GoogleAnalytics;
--- a/src/lib/analytics.js
+++ b/src/lib/analytics.js
@@ -8,6 +8,13 @@ if (GA_ID) {
         sampleRate: (process.env.NODE_ENV === 'production') ? 100 : 0,
         forceSSL: true
     });
+} else {
+    window.ga = () => {
+        // The `react-ga` module calls this function to implement all Google Analytics calls. Providing an empty
+        // function effectively disables `react-ga`. This is similar to the `testModeAPI` feature of `react-ga` except
+        // that `testModeAPI` logs the arguments of every call into an array. That's nice for testing purposes but
+        // would look like a memory leak in a live program.
+    };
 }
 
 export default GoogleAnalytics;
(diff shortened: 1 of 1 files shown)


cwillisf:
Well... that's disappointing. Every call to `analytics.event(...)` will result in a `console.warn(...)` in the console, like this:
image

I don't see a good way around this unless we wrap `react-ga`'s methods (or at least `event`) ourselves...

cwillisf:
It turns out that all calls to `react-ga` methods eventually end up as calls to `window.ga(...)`. Specifically, everything goes through this function in `react-ga/src/index.js`:

```js
const internalGa = (...args) => {
  if (_testMode) return TestModeAPI.ga(...args);
  if (!window.ga) return warn('ReactGA.initialize must be called first or GoogleAnalytics should be loaded manually');
  return window.ga(...args);
};
```

Providing an empty function for `window.ga` effectively means that these calls will be silently dropped instead of triggering the warning about calling `ReactGA.initialize`.

cwillisf:
Also, I looked into using `_testMode` but it's implemented like this:

```js
  ga: (...args) => {
    gaCalls.push([...args]);
  }
```

...which means that the `gaCalls` array would grow unbounded unless we explicitly clear it, which in turn means we'd be assuming more about `react-ga`'s than we already are and slightly harming performance besides.

paulkaplan:
@cwillisf it is strange especially because one of the useful things about the `debug` flag in init is that if debug=true you get the event with all the data you passed logged to the console so you can tell you've done it right.

cwillisf:
@paulkaplan do you think it's worth adding a notice if `debug` is true but `GA_ID` is missing, maybe `skipping GA initialization because GA_ID is not set` or something like that? That would be unobtrusive in cases where you want GA to be disabled, but still give you a hint if you're trying to debug a situation where you want GA but it isn't working.

benjiwheeler:
@cwillisf What is the state of this PR? Do you need help with it?

cwillisf:
@benjiwheeler I think it just needs re-review, now that I have (I think?) handled Paul's concerns. I see that it failed on Travis, though, and I don't have a good explanation for the errors I see there. I'll dig into it...

cwillisf:
@benjiwheeler looks like it might have been an ancient flaky test issue; rebasing onto current `develop` seems to have resolved the problem and tests are passing. So, I guess I just need a re-review, especially regarding whether or not I've resolved @paulkaplan's concerns.

benjiwheeler:
Testing plan:

* Locally, trying this with the line that sets the env var in webpack removed, to confirm no errors appear in console
* after merging and deploying to staging, will confirm that GA events are being sent
* CWF will bring this into desktop branch eventually and confirm that it no longer attempts to download GA js

benjiwheeler:
LGTM!

Issue #5140, opened 2019-09-01
File too large in the GIT history


garrapato:
I have noticed that the repository contains a **very large file in the GIT history**. Is it possible that someone made a commit with some file or directory by mistake? It can be for example nodes_modules or some other like that, unless that really is the correct size of the repository.

I just made a clone of the repository and it took a long time to download it.

The file I refer to in a fresh clone of the repository is:
```
scratch-gui/.git/objects/pack/pack-c81e535f2cf1cd650ef7a6e69553ee444473a465.pack
```
### Expected Behavior

Less time to clone (download) the repo

### Actual Behavior

I just made a clone of the repository and it took a very long time to download it.

### Steps to Reproduce

```
$ git clone https://github.com/LLK/scratch-gui.git
Cloning into 'scratch-gui'...
remote: Enumerating objects: 48, done.
remote: Counting objects: 100% (48/48), done.
remote: Compressing objects: 100% (47/47), done.
Receiving objects:  100% (67966/67966), 11.68 MiB | 815.00 KiB

$ cd scratch-gui
$ du -ch . | grep "G\t"
1.0G	./.git/objects/pack
1.1G	./.git/objects
1.1G	./.git
1.1G	.
1.1G	total

$ cd ./.git/objects/pack
$ ll
total 2199344
-r--r--r--  1 garrapato  staff     1904120 Sep  1 06:09 pack-c81e535f2cf1cd650ef7a6e69553ee444473a465.idx
-r--r--r--  1 garrapato  staff  1111074667 Sep  1 06:24 pack-c81e535f2cf1cd650ef7a6e69553ee444473a465.pack
 ```
The file already measures more than 1 GB!

### Possible solution
If a file or directory was uploaded (committed) by mistake, it must be deleted from the story and the following article shows how to do it:

[Removing sensitive data from a repository](https://help.github.com/en/articles/removing-sensitive-data-from-a-repository)

### Operating System and Browser

Mac OS 10.11.14
Chrome Versión 76.0.3809.132 (Build oficial) (64 bits)

I hope this information will be useful

Regards

apple502j:
Maybe this is because of tutorial animated GIFs. We recommend the use of `--depth 10` when cloning this repo.

Kenny2github:
In general, this repo's length commit history will make a full clone take an incredible amount of time. I recommend `--depth 1` rather than 10 because it really can get to be too much.

rschamp:
[...]

PR #5411, opened 2020-02-03
Add note about cloning to README

glenpike:
Because there are some very large files in the repository history

### Resolves

Workaround for #5140

### Proposed Changes

Add a note to the README about cloning with minimum depth

### Reason for Changes

So people don't have to wait such a long time to clone the repository

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ cd scratch-gui
 npm install
 ```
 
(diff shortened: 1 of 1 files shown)


benjiwheeler:
@rschamp @kchadha This note makes sense to me, OK to merge?

benjiwheeler:
Thanks, @glenpike! I merged this.

If you'd like to go even further, you might submit an additional PR that puts the depth argument into the first suggested clone command, then explains below that you can also run it without the depth argument.

I can't promise we'd merge that change, but I'd advocate for it!

PR #5412, opened 2020-02-03
show clickable 'About' button if handler is present

cwillisf:
### Resolves

This is a step toward resolving LLK/scratch-desktop#97

### Proposed Changes

This PR adds an optional icon to the top right of the main menu bar. This optional icon looks like a question mark in a circle and only appears if the new `onClickAbout` handler property is set.

### Reason for Changes

Scratch Desktop will use this new button to display its "About" window, which will have a link to the in-app privacy policy display.

I don't expect that we'll want to show the "About" button at the same time as the account menu or the "My Stuff" button, but if we do then the "About" button is currently set to display to the right of those other items.

### Test Coverage

TBD

### Browser Coverage

TBD

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


cwillisf:
Looks like I accidentally opened two PRs for the same work. Closing this one in favor of 5434.

PR #5418, opened 2020-02-04
don’t update project title if it was undefined

benjiwheeler:
### Resolves

- Resolves https://github.com/LLK/scratch-www/issues/3673

### Proposed Changes

Instead of calling `onUpdateProjectTitle()` any time the title has changed in scratch-gui, only call `onUpdateProjectTitle()` if the title has changed *and the title prop passed in to scratch-gui is set.* 

That is, if the title passed in is undefined or null, then it's not really "stale" -- we trust the code which renders scratch-gui to determine the title and pass it in to us, so we should only correct it if it really is being passed in.

### Reason for Changes

We were seeing an invalid request to api, using an undefined project id; that's because www was trying to set the title when the project data hadn't fully loaded yet.

### Test Coverage

Added tests

### DIFF
--- a/src/lib/titled-hoc.jsx
+++ b/src/lib/titled-hoc.jsx
@@ -36,10 +36,14 @@ const TitledHOC = function (WrappedComponent) {
                 const defaultProjectTitle = this.handleReceivedProjectTitle();
                 this.props.onUpdateProjectTitle(defaultProjectTitle);
             }
-            // if the projectTitle hasn't changed, but the reduxProjectTitle
-            // HAS changed, we need to report that change to the projectTitle's owner
-            if (this.props.reduxProjectTitle !== prevProps.reduxProjectTitle &&
-                this.props.reduxProjectTitle !== this.props.projectTitle) {
+            // if the user has changed the project Title, and the projectTitle prop
+            // passed in to us is stale (defined, but old), we need to notify
+            // the source of the projectTitle that it should be updated
+            const reduxTitleHasChanged =
+                (this.props.reduxProjectTitle !== prevProps.reduxProjectTitle &&
+                this.props.reduxProjectTitle !== this.props.projectTitle);
+            const titlePropIsSet = !!this.props.projectTitle;
+            if (titlePropIsSet && reduxTitleHasChanged) {
                 this.props.onUpdateProjectTitle(this.props.reduxProjectTitle);
             }
         }
--- a/src/lib/titled-hoc.jsx
+++ b/src/lib/titled-hoc.jsx
@@ -42,7 +42,8 @@ const TitledHOC = function (WrappedComponent) {
             const reduxTitleHasChanged =
                 (this.props.reduxProjectTitle !== prevProps.reduxProjectTitle &&
                 this.props.reduxProjectTitle !== this.props.projectTitle);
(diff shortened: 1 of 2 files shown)


src/lib/titled-hoc.jsx
```diff
@@ -36,10 +36,14 @@ const TitledHOC = function (WrappedComponent) {
...
-            // HAS changed, we need to report that change to the projectTitle's owner
-            if (this.props.reduxProjectTitle !== prevProps.reduxProjectTitle &&
-                this.props.reduxProjectTitle !== this.props.projectTitle) {
+            // if the user has changed the project Title, and the projectTitle prop
+            // passed in to us is stale (defined, but old), we need to notify
+            // the source of the projectTitle that it should be updated
+            const reduxTitleHasChanged =
+                (this.props.reduxProjectTitle !== prevProps.reduxProjectTitle &&
+                this.props.reduxProjectTitle !== this.props.projectTitle);
+            const titlePropIsSet = !!this.props.projectTitle;
```
rschamp, line 45: Currently we allow you to set your title to an empty string — would this break that? I think it would be safer to test for `null` or `undefined`.  As an aside, I thought our eslint rules disallowed `!!`.

test/unit/util/titled-hoc.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+        );
+        expect(mockOnUpdateProjectTitle).not.toHaveBeenCalled();
+        mounted.setProps({
+            reduxProjectTitle: 'new title'
+        });
+        expect(mockOnUpdateProjectTitle).toHaveBeenCalled();
+        expect(mockOnUpdateProjectTitle.mock.calls[0][0]).toBe('new title');
+    });
+
+    test('when new title is set in GUI but external title is blank, it does not call onUpdateProjectTitle', () => {
```
rschamp, line 40: Since we allow blank titles to be set, then I think this behavior would make it impossible to set your title to anything after you did set it to be blank.

src/lib/titled-hoc.jsx
benjiwheeler: Good point. Changed to test for null or undefined.

test/unit/util/titled-hoc.test.jsx
benjiwheeler: I agree, thanks for pointing this out. Take a look now.

rschamp:
This looks good to me as far as the code goes. It still seems like it could have some unintended consequences, given that it changes the way all instances of GUI handle titles, but is only intended to resolve an issue with www. So I'm still looking for an answer to this:

> Did you consider making the `onUpdateProjectTitle` function in scratch-www test for `undefined` before making the API request? If we do this here, can the standalone GUI still manage its own titles? Honestly asking — this could be the right place to make the change, I'm just not sure without knowing why it shouldn't be in [www](http://www). Another way of thinking about it — is this the right behavior for _every_ instance of GUI (i.e. all the native apps, and the standalone GUI)? Or just what we need for www? If it's just for www I lean to fixing it in www rather than making it universal behavior.

benjiwheeler:
Closing this as too old to be worked on more; but this would be useful to pick back up one day!

PR #5434, opened 2020-02-06
Add about menu button

cwillisf:
### Resolves

This supports work toward LLK/scratch-desktop#97

### Proposed Changes

If the GUI element receives a value for the new `onClickAbout` property then it will display a new "(?)" button in the top menu bar. If the button is clicked, the `onClickAbout` function will be called.

These changes have been discussed with and approved by the design team; the "About" button itself was supplied by @carljbowman. Here's a preview of the button as it looks in Scratch Desktop on Windows 10:

image

### Reason for Changes

We need to display a privacy policy within the Scratch Desktop application, and we've chosen to do so by adding a link from the "About" window. Currently the "About" window is displayed when the user clicks on the Scratch logo, which was easy to code but is a bit unusual so it might be hard for users to find. To make it easier to find the "About" window & privacy policy, this PR moves ties the window to this new button.

### Test Coverage

The existing tests in `sratch-gui` should help verify that this change doesn't impact other menus or buttons. I also added new tests in a new file called `test/unit/components/menu-bar.test.jsx` which verify that the "About" button does not appear when there is no handler, and that when a handler is installed the button appears and calls the handler when clicked.


benjiwheeler:
@cwillisf Before I review, any thoughts on why this isn't passing tests?

cwillisf:
Sorry about that! It turns out that `scratch-l10n` build output uses `export default` and our Jest settings don't activate Babel for those files. One option would be to change `scratch-l10n` to always ES5-ify its output but it was easy enough to just mock the relevant file and have a fix here in GUI.

I'm not 100% sure why I wasn't consistently seeing the problem locally but I believe it had something to do with recent changes in `develop`, or maybe some weird state that I had on my local machine which changed when I merged in the latest changes from `develop`. At any rate, this change works both on my local machine and on Travis CI.

benjiwheeler:
Currently blocked on two things:
1) the gui branch "scratch-desktop" needs to be updated with recent changes to develop; currently when i try to rebase it on develop, I can't run scratch-desktop (loading screen appears forever)
2) a review of the policy the about button links to is underway

cwillisf:
@benjiwheeler I don't think this should be blocked on either of those things:
1. The `scratch-desktop` branch of `scratch-gui` is meant only for use in the Scratch Desktop app, and specifically in Electron. We do intend to merge the two development lines at some point in the future but right now there's no guarantee that the `scratch-desktop` branch will run correctly in a browser.
   - That's part of the reason this PR is targeted to merge into `develop`: this change should be available to, and compatible with, all branches of `scratch-gui`, and in particular should be testable in `develop`. It's hard to test manually, that's true, but I was intending the new tests in `menu-bar.test.jsx` to cover the testing requirement.
   - Also: it's likely that the Android build will want to use the "About" button for similar reasons, but they don't use the `scratch-desktop` branch so they'd probably rather merge this feature from `develop` than cherry-pick it from `scratch-desktop`.
2. The button itself doesn't depend on the wording of the privacy policy. In fact, the current design has this "About" button opening the existing "About" dialog, currently attached to clicking the Scratch logo in Scratch Desktop. I'm working on a separate change to add a privacy policy link to that dialog, but I don't think this PR should be blocked on that change.

If you'd like to see how this functions in a Scratch Desktop context, there are several more steps to take, but IMO these steps aren't strictly required as part of reviewing this PR:
1. Rebase a local clone of this branch onto `scratch-desktop` branch instead of `develop`.
2. Make sure that version of `scratch-gui` is registered using `npm link`.
3. Get the `privacy-policy` branch of `scratch-desktop`, which is in active development based on some feedback we received from our lawyers.
4. In `scratch-desktop`, run `npm i` or `npm ci` then `npm link scratch-gui`
5. Still in `scratch-desktop`, run `npm run build-gui`
5. Finally, run `npm start` to try it out

benjiwheeler:
@cwillisf All that makes sense, thanks! I don't totally remember the context of this proposed change, or what might hold it up. Why don't we:

1. test it out in just normal scratch-gui
2. put together a description of this change to show to the design and Android teams that they're ok with this change going in

cwillisf:
Sorry @benjiwheeler, I realized that the issue description was a little out of date in that it had a "TODO" about tests even though I had already added some, and that might have caused some confusion. I've updated the description to include the tests that I've added.

Also, this actually was discussed with the design team a while back, and they approved this approach. I added that to the issue description as well since it wasn't obvious. I've only talked briefly & informally to the Android team but they seem to like it so far, too.

Is there more "test it out in just normal scratch-gui" and "test it out in scratch-desktop" that you have in mind? IMO the tests I've added cover the normal scratch-gui case, and my day-to-day work in the scratch-desktop repository is effectively testing this in that context constantly.

.gitattributes
```diff
@@ -12,6 +12,7 @@
 *.svg binary
 
 # Prefer LF for most file types
+*.css text eol=lf
```
benjiwheeler, line 15: why is this here?

benjiwheeler:
@cwillisf OK, it sounds like you have been testing it in desktop? Then I won't worry about testing it myself.

The code LGTM, but please take a look at the .gitattributes change -- is that intentional?

.gitattributes
cwillisf: I added this because a lot of Windows-based editors tend to use CRLF unless you explicitly say to use LF, and CSS files weren't already covered here. I noticed this when I went to edit `src/components/menu-bar/menu-bar.css` as part of the code in this PR. With this line in place, `git` will automatically ensure that LF-only line endings are committed to the repository  in any CSS file, regardless of whether the editor used CRLF, LF, CR, or some mix of them in that file.

benjiwheeler:
OK! LGTM

PR #5444, opened 2020-02-09
fix: initialize audio context only with user interaction

### LINKED ISSUE #4514: The AudioContext was not allowed to start.
### Expected Behavior

No warnings in console for production build

### Actual Behavior

Keep raising warning (The AudioContext was not allowed to start. It must be resumed (or created) after a user gesture on the page. https://goo.gl/7K7WLu) before user click the page. 

And this problem is exists in the current online version https://scratch.mit.edu/projects/editor

screen shot 2019-02-07 at 5 37 19 pm

### Steps to Reproduce

* open Chrome
* Go to https://scratch.mit.edu/projects/editor
* open developer tool
* open console
* refresh page and **do not click page**
* you will see the warnings

* Mac OS 10.14.2 Chrome 71
* Chrome OS 71

related issue #2017

watilde:
### Resolves

- Resolves #4514

### Proposed Changes

Add a one-dispatch event listener to trigger the shared audio context.

### Reason for Changes

To help developers to debug by removing a warning.

### Test Coverage

WIP

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/lib/audio/shared-audio-context.js
+++ b/src/lib/audio/shared-audio-context.js
@@ -2,10 +2,20 @@ import StartAudioContext from 'startaudiocontext';
 import bowser from 'bowser';
 
 let AUDIO_CONTEXT;
-if (!bowser.msie) {
-    AUDIO_CONTEXT = new (window.AudioContext || window.webkitAudioContext)();
 
-    StartAudioContext(AUDIO_CONTEXT);
+if (!bowser.msie) {
+    /**
+     * AudioContext can be initialized only when user interaction event happens
+     */
+    const event =
+        typeof document.ontouchend === 'undefined' ? 'mouseup' : 'touchend';
+    const initAudioContext = () => {
+        document.removeEventListener(event, initAudioContext);
+        AUDIO_CONTEXT = new (window.AudioContext ||
+            window.webkitAudioContext)();
+        StartAudioContext(AUDIO_CONTEXT);
+    };
+    document.addEventListener(event, initAudioContext);
 }
 
 /**


benjiwheeler:
This seems like a simple fix that should not be hard to test

apple502j:
Hmm, I have a question... isn't that `startaudiocontext`'s purpose?

watilde:
@apple502j That's right. We can leverage https://github.com/tambien/StartAudioContext instead as well. @benjiwheeler What do you think?

Let me also add a test for this.

apple502j:
Hmm, I don't think we need `startaudiocontext` because the thing is already implemented. Also, https://github.com/LLK/scratch-audio/blob/develop/src/StartAudioContext.js may also need this fix

watilde:
@apple502j That's good to know, thank you. As the original issue still exists, we still need to have new audio context initialization. Turned out the audio context initialization issue exists in three places.

Here is a screenshot on Chrome from the latest develop branch([hash](https://github.com/LLK/scratch-gui/commit/cc22bfc636f86b56b43e8422084a3b64a63e875a)).
Screenshot

### Where the issue happens

**1. src/lib/audio/shared-audio-context.js#6**
This PR fixes this issue. Line: https://github.com/LLK/scratch-gui/blob/cc22bfc636f86b56b43e8422084a3b64a63e875a/src/lib/audio/shared-audio-context.js

**2. node_modules/audio-context/index.js#36**
We need to have another patch to fix it. This call is from [`node_modules/scratch-audio/src/AudioEngine.js`](https://github.com/LLK/scratch-audio/blob/develop/src/AudioEngine.js#L44) that calls [`StartAudioContext`](https://github.com/LLK/scratch-audio/blob/develop/src/StartAudioContext.js) you mentioned.

**3. node_modules/startaudiocontext/StartAudioContext.js#114**
This is just the place where `scratch-gui` calls AudioContext. If we fix the above two, this will not be an issue anymore.

watilde:
@benjiwheeler Thank you for your comment. I just added tests.

paulkaplan:
This PR causes a bug where the sound tab will crash due to undefined audio context in the situation where you have a device with `ontouchstart` defined but not used, for example on certain devices with a touchscreen and a trackpad mouse. Also 

> Hmm, I have a question... isn't that `startaudiocontext`'s purpose?

Yeah @apple502j it definitely is, so I'm pretty confused about why it wasn't working. Ideally we can  remove any synchronous uses of AudioContext and move everything to a promise, although that might be hard because we may rely on synchronous access to the audio context in many places. 

A simple fix for right now will be to just bind to both those events and unbind both to cover the bases, although that still leaves open the 2 questions: (1) why startaudiocontext wasnt doing its job, (2) is there a way to not rely on audiocontext being available right away.

PR #5437, opened 2020-02-06
Revert "refactor SB* file uploader into top-level HOC"

benjiwheeler:
Reverts LLK/scratch-gui#4809

PR #5450, opened 2020-02-11
Revert "Revert "refactor SB* file uploader into top-level HOC""

### LINKED ISSUE #3365: Can't upload scratch project by double clicking the file
### Expected Behavior
Double clicking in the file selector selects the file

### Actual Behavior
Double clicking in the file selector closes the file selector with no action

### Steps to Reproduce
1. Go to develop
2. File > Upload from your computer an sb3
 - If you double click to select the file, the file dialog closes with no action
 - If you click the file and then press the 'Open' button, it works fine

upload2

Attached the file I'm using in case it's relevant (rename to sb3)
[you totally ca.zip](https://github.com/LLK/scratch-gui/files/2479185/you.totally.ca.zip)
Windows 10 Firefox

benjiwheeler:
Reverts LLK/scratch-gui#5437; redo of original PR https://github.com/LLK/scratch-gui/pull/4809

### Resolves

Big step towards resolving #4762
Resolves #3365

### Changes from previous version

Adds a fix to a saving error we saw, where after upload, the first attempt to save the project would appear to fail.

The actual problem was twofold:
1) in a promise `then` block, we were calling `onLoadingFinished()` with the `success` parameter set to `true` before other calls which could throw errors. Then in the corresponding `catch` block, we were calling `onLoadingFinished()` with the `success` parameter set to `false`. This could result in `onLoadingFinished()` being called twice, with the second call failing (as it should) and causing a visible error state. 
2) why was the actual error being thrown, though? Because the title-setting function we were calling, `onUpdateProjectTitle()`, was unset by separate changes to titled-hoc.js (see https://github.com/LLK/scratch-gui/pull/5168 ). Those changes should have included changing sb-file-uploader.js to no longer call `onUpdateProjectTitle()`.

The fixes:
1) move the `onLoadingFinished()` calls to the end of their respective blocks, to avoid their being called twice in case of an error
2) instead of calling `onUpdateProjectTitle()`, instead call the redux function `setProjectTitle()` per the refactor of titled-hoc.js .


rschamp:
Try it out: [revert-5437-revert-4809-refactor-file-uploader](https://llk.github.io/scratch-gui/revert-5437-revert-4809-refactor-file-uploader)

src/lib/sb-file-uploader-hoc.jsx
```diff
@@ -0,0 +1,272 @@
...
+            if (this.fileReader) {
+                this.props.onLoadingStarted();
+                const filename = this.fileToUpload && this.fileToUpload.name;
+                this.props.vm.loadProject(this.fileReader.result)
+                    .then(() => {
+                        if (filename) {
+                            const uploadedProjectTitle = this.getProjectTitleFromFilename(filename);
+                            this.props.onSetProjectTitle(uploadedProjectTitle);
+                        }
+                        this.props.onLoadingFinished(this.props.loadingState, true);
```
rschamp: I think it might be a better design to move this into the final `.then()`, since you want to ensure that either it will be called in the `catch` OR it will be called when it's determined the promise chain was successful. Then it's safer for that pattern to be repeated in other places if it gets copied around and the details of the error-throwing code changes.

src/lib/sb-file-uploader-hoc.jsx
```diff
@@ -0,0 +1,272 @@
...
+                    })
+                    .catch(error => {
+                        log.warn(error);
+                        this.props.intl.formatMessage(messages.loadError);
+                        this.props.onLoadingFinished(this.props.loadingState, false);
+                    })
+                    .then(() => {
```
rschamp, line 164: This is the `.then()` I'm referring to.

src/lib/sb-file-uploader-hoc.jsx
benjiwheeler: That makes sense, thanks!

benjiwheeler:
added tests; want to manually test more thoroughly now that code has changed

benjiwheeler:
Tested lots of UX sequences manually, I'm satisfied!

rschamp (approved):
We should import previous testing steps from the first time we tried to put this in, and make sure that importing a project twice in a row works as expected.

PR #5460, opened 2020-02-14
Use sb3 schema-compliant asset library data  

ericrosenbaum:
This is a follow-up to https://github.com/LLK/scratch-gui/pull/5395, updating all the asset libraries. This change makes the asset library data compliant with our schema for sb3 files: https://github.com/LLK/scratch-parser/blob/master/lib/sb3_definitions.json

The corresponding update to the library generator is https://github.com/LLK/scratch-libgen/pull/13

This change required modifying several files where the GUI expected data from the libraries in our previous format.


towerofnix:
I think I might've asked about this a while ago, but - if there's nothing in the way and there's permission from whoever is in charge of which repos are open source (used to be some division of MIT IIRC, but that may have changed recently?)—would it make sense to publish the scratch-libgen repo? It sounds like a fairly handy utility for folx in the community, and I'm sure people would be happy to assist with any minor issues laying around there!

ericrosenbaum:
@towerofnix good question. I don't think there's a single specific thing preventing us from making the scratch-libgen repo public. We just haven't prioritized it. There's a sort of general sense that the way it works right now is not ideal (i.e. the way we generate the asset libraries by extracting the assets from a set of unshared projects) and we'd like to re-think it, and perhaps build a completely different system, but that will not likely happen soon. The library data itself is not set up in a way that can be localized (i.e. our transifex-based system does not work with this type of json file, if I understand correctly), so we may need to re-engineer that part. And also we may not want to encourage the creation of utilities like libgen (i.e. node programs that directly request projects from the site etc). All that said I'll raise the question in a meeting soon to see what others think.

kchadha (approved):
LGTM! Thanks for all the follow up work on this @ericrosenbaum!

PR #5475, opened 2020-02-26
(demo) Remove bitmap resolution from asset library json

ericrosenbaum:
NOT TO BE MERGED

this is just an illustrative PR for @fsih 

This is a variation of https://github.com/LLK/scratch-gui/pull/5460, but with the added change of removing the `bitmapResolution` field from all assets. This change should theoretically be safe (because the information in that field is redundant or unnecessary), but I noticed that it caused a bug: 

- Add an SVG sprite from the library containing 2 or more costumes (e.g. "ball")
- click the "next costume" block
- note that the position of the sprite changes

My best guess is that this is caused by a line in the VM:  https://github.com/LLK/scratch-vm/blob/develop/src/sprites/rendered-target.js#L712

basically, if the bitmapResolution is missing when you add a costume, it sets it to 2. A couple lines later, it divides rotation center by bitmapResolution. So for SVGs with a missing bitmapResolution, the rotation center gets incorrectly changed


adroitwhiz:
I think it may be worth mentioning that the VM doesn't actually need to ever set the rotation centers after loading them initially, so the buggy VM code you mentioned [can be completely removed](https://github.com/LLK/scratch-vm/pull/2314/).

fsih:
Blocked on https://github.com/LLK/scratch-vm/pull/2329

fsih:
@ericrosenbaum I tested this out and I think it can be merged now!

ericrosenbaum:
@fsih cool! this is actually on top of https://github.com/LLK/scratch-gui/pull/5460, which is still in review, though, so it will have to come after that one

chrisgarrity:
@ericrosenbaum should this just get closed now?

PR #5496, opened 2020-03-06
Check list limit in list monitor

### LINKED ISSUE #4303: Project with large imported list cannot be saved
### Expected Behavior

Importing a list should not be able to get the editor into a state in which saving fails.

### Actual Behavior

When I import a large list of words (4.9MB), the import succeeds, and I have a list with over 460,000 items (note that this exceeds the normal list length limit). When I click "save now", and it appears to succeed, but there is a console error (a 413 error, which seems reasonable!), but when I reload the project my changes have not been saved.

### Steps to Reproduce

- download a large word list in a text file (e.g. from https://github.com/dwyl/english-words/raw/master/words.txt)
- create a list in scratch
- right-click the list monitor and import the file
- click 'save now'
- reload the project and the list (and other changes) are gone

### LINKED ISSUE #5495: List monitor can be used to bypass list limit
### Expected Behavior
List item limit cannot be bypassed

### Actual Behavior
Using list monitor, you can bypass the 200000 item limitation.

### Steps to Reproduce
1. Add 200000 item
2. Click + button or select one and press Enter
3. The list now contains 200000 item

any

apple502j:
### Resolves
Resolves #4303 (partially)
Resolves #5495

### Proposed Changes
Adds a new function, `canSetVariableValue`. This returns true when the variable is not a list, or the limit is not reached. This is called inside `list-monitor.jsx` to check if the new item can be added. (if not, Enter or + button simply fail)

Also, this changes `monitor.jsx` so that only the first 200000 item is imported from text file.

### Reason for Changes
The list limit is not supposed to be bypassed.

### Test Coverage
Added a unit test for the new function. Manual test for other fixes.

### DIFF
--- a/src/containers/monitor.jsx
+++ b/src/containers/monitor.jsx
@@ -6,7 +6,7 @@ import {injectIntl, intlShape, defineMessages} from 'react-intl';
 import monitorAdapter from '../lib/monitor-adapter.js';
 import MonitorComponent, {monitorModes} from '../components/monitor/monitor.jsx';
 import {addMonitorRect, getInitialPosition, resizeMonitorRect, removeMonitorRect} from '../reducers/monitor-layout';
-import {getVariable, setVariableValue} from '../lib/variable-utils';
+import {getVariable, setVariableValue, LIST_ITEM_LIMIT} from '../lib/variable-utils';
 import importCSV from '../lib/import-csv';
 import downloadBlob from '../lib/download-blob';
 import SliderPrompt from './slider-prompt.jsx';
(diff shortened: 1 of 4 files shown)


src/containers/monitor.jsx
```diff
@@ -171,7 +171,8 @@ class Monitor extends React.Component {
                 columnNumber = parseInt(prompt(msg), 10); // eslint-disable-line no-alert
             }
             const newListValue = rows.map(row => row[columnNumber - 1])
-                .filter(item => typeof item === 'string'); // CSV importer can leave undefineds
+                .filter(item => typeof item === 'string') // CSV importer can leave undefineds
+                .splice(0, LIST_ITEM_LIMIT);
```
adroitwhiz: `splice` modifies the existing array, I believe. Is this okay/intentional?

apple502j: `filter` should deepcopy the array anyway, so it's okay.

fsih:
Thanks for your PR! However, we think there are some design questions with this issue that we want to address before accepting PRs, such as whether a warning should show, or if the buttons should be disabled.

PR #5515, opened 2020-03-13
findByXpath: catch StaleElementReferenceError

### LINKED ISSUE #5429: Flaky integration test: blocks.test.js
Passed when I reran it

FAIL test/integration/blocks.test.js (46.004s)
  ● Working with the blocks › Adding costumes DOES update the default costume name in the toolbox
    StaleElementReferenceError: stale element reference: element is not attached to the page document
      (Session info: headless chrome=80.0.3987.87)
      (Driver info: chromedriver=80.0.3987.16 (320f6526c1632ad4f205ebce69b99a062ed78647-refs/branch-heads/3987@{#185}),platform=Linux 4.15.0-1028-gcp x86_64)
      
      at Object.checkLegacyResponse (node_modules/selenium-webdriver/lib/error.js:546:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:509:13)
      at doSend.then.response (node_modules/selenium-webdriver/lib/http.js:441:30)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)
      From: Task: WebElement.click()
      at thenableWebDriverProxy.schedule (node_modules/selenium-webdriver/lib/webdriver.js:807:17)
      at WebElement.schedule_ (node_modules/selenium-webdriver/lib/webdriver.js:2010:25)
      at WebElement.click (node_modules/selenium-webdriver/lib/webdriver.js:2092:17)
      at test/helpers/selenium-helper.js:124:73
      at ManagedPromise.invokeCallback_ (node_modules/selenium-webdriver/lib/promise.js:1376:14)
      at TaskQueue.execute_ (node_modules/selenium-webdriver/lib/promise.js:3084:14)
      at TaskQueue.executeNext_ (node_modules/selenium-webdriver/lib/promise.js:3067:27)
      at asyncRun (node_modules/selenium-webdriver/lib/promise.js:2927:27)
      at node_modules/selenium-webdriver/lib/promise.js:668:7
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)

### LINKED ISSUE #5514: Stabilize integration tests by removing use of stale element?
As discussed in https://github.com/LLK/scratch-gui/issues/5429#issuecomment-598543870, the change in https://github.com/LLK/scratch-gui/pull/4871 introduced an experiment: each time we found an element by XPath, we would then try repeatedly calling `isDisplayed()` on the element until it became visible.

This usually works fine, but sometimes -- such as when first finding sections of the blocks menu/palette, which rapidly replaces itself twice on initialization -- the element has been removed by the time we're calling `isDisplayed()` on it.

We should try removing this `isDisplayed()` loop, and seeing if the tests seem more stable or less stable afterwards. Since this was always an experiment, and we've identified at least one case where it specifically keeps a test from working, I think we should err on the side of removing it.

adroitwhiz:
### Resolves

- Resolves #5429
- Resolves #5514

### Proposed Changes

This PR replaces the `findByXpath` wait introduced in #4871 with a `catch` block that retries the `findByXpath` call if a `StaleElementReferenceError` is found.

### Reason for Changes

This is a complete shot in the dark, but the `StaleElementReferencError`s seem to originate from the following line (which is unusual because it seems to point to an issue in WebDriver itself):
```js
this.driver.wait(until.elementLocated(By.xpath(xpath)), DEFAULT_TIMEOUT_MILLISECONDS, timeoutMessage)
```

Using the `cpulimit` command to slow down Node + Chrome and alter the timing in subtle ways, I was able to reproduce the test failure locally about half the time. After adding the `catch`, the tests began passing 100% of the time.

### DIFF
--- a/test/helpers/selenium-helper.js
+++ b/test/helpers/selenium-helper.js
@@ -4,7 +4,7 @@ import bindAll from 'lodash.bindall';
 import 'chromedriver'; // register path
 import webdriver from 'selenium-webdriver';
 
-const {By, until, Button} = webdriver;
+const {By, until, Button, error} = webdriver;
 
 const USE_HEADLESS = process.env.USE_HEADLESS !== 'no';
 
@@ -91,10 +91,12 @@ class SeleniumHelper {
 
     findByXpath (xpath, timeoutMessage = `findByXpath timed out for path: ${xpath}`) {
         return this.driver.wait(until.elementLocated(By.xpath(xpath)), DEFAULT_TIMEOUT_MILLISECONDS, timeoutMessage)
-            .then(el => (
-                this.driver.wait(el.isDisplayed(), DEFAULT_TIMEOUT_MILLISECONDS, `${xpath} is not visible`)
-                    .then(() => el)
-            ));
+            .catch(err => {
+                // driver.wait(until.elementLocated(...)) seems to throw a StaleElementReferenceError if the element is
+                // removed while waiting. Not sure why this happens. If it does, just try again.
+                if (err instanceof error.StaleElementReferenceError) return this.findByXpath(xpath, timeoutMessage);
+                throw err;
+            });
     }
 
     findByText (text, scope) {


benjiwheeler:
Thanks, I think this is a good response to those two issues. cc @BryceLTaylor

apple502j:
There's one strange error - `WebDriverError` caused by Chrome crashing.

benjiwheeler:
Hm, I'm still running into the same error, even with this change.

To repro:
* comment out all integration/blocks.test.js tests except "'Adding costumes DOES update the default costume name in the toolbox'"
* within that test, comment out all lines after `await clickText('Looks', scope.blocksTab);`
* run `$(npm bin)/jest --runInBand test/integration/blocks.test.js`

adroitwhiz:
Managed to get it to occur as well.

I tried adding a `driver.sleep(500)` above the problematic `clickText`, and it didn't help either.

It looks like there's currently no system for telling when the GUI has loaded, and hence no way to sleep for a certain amount of time *before* clicking the first element. I assume that's why my `sleep` call failed-- it was sleeping for 500ms at the beginning of the page load, then going on to wait for the block palette to appear before the loader had even gone away.

PR #5539, opened 2020-03-24
prevent duplicate selections in sensing_of

### LINKED ISSUE #5488: Some variable names break sensing_of
### Expected Behavior
Variables named `volume` should not interfere with the `volume` option in sensing_of. Same goes for these: global variable named `backdrop #` `backdrop name` local variable named `x position` `y position` `size` `costume #` `costume name` `direction`

### Actual Behavior
It is currently impossible to tell the difference between "sprite's volume" option and "sprite's variable named volume" from the code. 
image
This is `volume of stage`, and the `音量` (`volume` in Japanese) and `volume` variable is handled as two options with the same name. If you change the language to English, it will be impossible for the user as well.

### Steps to Reproduce
Make a variable named `volume` (or other names listed above) and see sensing_of options.

Windows 10 Firefox 73.0.1, though I believe this issue occurs in all environment.

lholliger:
### Resolves

- Resolves #5488 

### Proposed Changes

This PR changes `sensing_of` so that instead of building a list of variables as `[variable, variable]` (which leads to there being a duplicate between a variable name and one of the pre-defined options (ex: volume) it uses `[variable.name, variable.id]` so there is no conflict. This allowed for the pre-defined value (ex: `volume`) and a user-made variable of the same name to not be checked at the same time.

### Reason for Changes

These changes allow for `sensing_of` to use variables that keep the same names as the pre-defined ones, such as `direction` and `volume`.

### Test Coverage

Manually tested, changes shown:
fixed:
image
original:
image
The pre-defined `volume` returns `100` as expected, and the user-defined `volume` returns `100`, as defined.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 

iPad
* [x] Safari

### DIFF
--- a/src/lib/blocks.js
+++ b/src/lib/blocks.js
@@ -245,13 +245,13 @@ export default function (vm) {
                     // The block was in the flyout so look up future block info there.
                     lookupBlocks = vm.runtime.flyoutBlocks;
                 }
-                const sort = function (options) {
-                    options.sort(ScratchBlocks.scratchBlocksUtils.compareStrings);
-                };
+                const sort = options => {
+                    options.sort((var1, var2) => ScratchBlocks.scratchBlocksUtils.compareStrings(var1.name, var2.name));
+                }
                 // Get all the stage variables (no lists) so we can add them to menu when the stage is selected.
-                const stageVariableOptions = vm.runtime.getTargetForStage().getAllVariableNamesInScopeByType('');
+                const stageVariableOptions = Object.values(vm.runtime.getTargetForStage().variables);
                 sort(stageVariableOptions);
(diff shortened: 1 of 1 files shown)


src/lib/blocks.js
```diff
@@ -245,13 +245,13 @@ export default function (vm) {
                     // The block was in the flyout so look up future block info there.
                     lookupBlocks = vm.runtime.flyoutBlocks;
                 }
-                const sort = function (options) {
-                    options.sort(ScratchBlocks.scratchBlocksUtils.compareStrings);
-                };
+                const sort = options => {
+                    options.sort((var1, var2) => ScratchBlocks.scratchBlocksUtils.compareStrings(var1.name, var2.name));
+                }
```
apple502j: *Linter angers*

PR #5543, opened 2020-03-27
Meow meow meow purr meow

### LINKED ISSUE #518: Support for Edible Scratch and Meow
Do we want to support Edible Scratch (yum) and Meow (cat) in Scratch 3.0

apple502j:
### Resolves
Resolves #518

### Proposed Changes
Makes it accessible for all cats.

### Reason for Changes
I tried to [teach Scratch to my cat](https://scratch.wistia.com/medias/u1aibu2mfq), but it failed because Meow language was not supported.

### Test Coverage
Tested using my cat

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [x] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


chrisgarrity:
@apple502j unfortunately we made the hard decision not to support meow or edible scratch in 3.0. If one of these languages returns, even for a day, people will expect it to stay. The problem in general with fake languages is that we cannot maintain them in transifex, and they create bugs in Scratch when any strings change later. For the sake of not creating drama, I don't think we should add this right now. Thanks for understanding.

PR #5552, opened 2020-03-30
Redraw stage on fullscreen

apple502j:
### Resolves
Resolves LLK/scratch-www#2721

### Proposed Changes
Re-draws the stage when fullscreen state is changed

### Reason for Changes
Without redrawing, it seemed like the canvas was still in the old size, and other area was transparent.

### Test Coverage
Manual testing: https://scratch.mit.edu/projects/357736772/fullscreen/
image

### DIFF
--- a/src/containers/stage.jsx
+++ b/src/containers/stage.jsx
@@ -95,9 +95,7 @@ class Stage extends React.Component {
         }
         this.updateRect();
         this.renderer.resize(this.rect.width, this.rect.height);
-        if (this.props.isFullScreen !== prevProps.isFullScreen ||
-            this.props.stageSize !== prevProps.stageSize ||
-            this.props.isPlayerOnly !== prevProps.isPlayerOnly) {
+        if (this.props.isFullScreen !== prevProps.isFullScreen) {
             // Re-draw the stage so that it fits the screen LLK/scratch-www#2721
             this.renderer.draw();
         }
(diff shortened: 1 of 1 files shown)


src/containers/stage.jsx
```diff
@@ -95,6 +95,10 @@ class Stage extends React.Component {
         }
         this.updateRect();
         this.renderer.resize(this.rect.width, this.rect.height);
+        if (this.props.isFullScreen !== prevProps.isFullScreen) {
+            // Re-draw the stage so that it fits the screen LLK/scratch-www#2721
+            this.props.vm.renderer.draw();
```
adroitwhiz: nit: we already have `this.renderer`
```suggestion
            this.renderer.draw();
```

src/containers/stage.jsx
```diff
@@ -95,6 +95,10 @@ class Stage extends React.Component {
         }
         this.updateRect();
         this.renderer.resize(this.rect.width, this.rect.height);
+        if (this.props.isFullScreen !== prevProps.isFullScreen) {
```
adroitwhiz: Would this be better as `this.props.stageSize !== nextProps.stageSize`?

There's another bug where:
1. You switch to small stage mode
2. You switch from the editor to the project page
3. The stage appears low-resolution

adroitwhiz:
Closing because https://github.com/LLK/scratch-render/pull/588 fixes this-- switching to fullscreen calls `resize` which will now properly redraw the stage within the renderer itself

PR #5553, opened 2020-03-30
Fix wobbly dragging in Edge

adroitwhiz:
Let me know if you don't want to merge this or enable wobbles by default; I understand if this "feature" has become a bit of a chore.

Edge's `getBoundingClientRect` doesn't return a rectangle with `x` or `y`, but it does have `left` and `top`, so use those when positioning the drag canvas.


rschamp:
Unfortunately we ran out of time to look at this before we needed to freeze for the deploy. Thanks for your diligence though!

PR #5558, opened 2020-04-02
fix costume animation in sprite library

cwillisf:
### Resolves

Resolves a problem specific to the scratch-desktop branch in which the library icons for sprites would not correctly animate even in the presence of multiple costumes.

### Proposed Changes

Two changes:
- if a library item (or sub-item) contains both `assetId` and `dataFormat` properties, use those to determine the icon. Prefer that over options which require splitting on `.`
- if a library item (or sub-item) has a property called `md5ext` use that (higher priority than existing code which used `md5` or `baseLayerMD5`).

### Reason for Changes

An animated library item may now have `assetId` and `dataFormat` properties. The new library generation code provides these properties and they align better to the way `scratch-storage` works.

Also, a library item may now contain a property called `md5ext`, which is just a renamed version of the old `md5` property. We are trying to encourage `md5ext` rather than just `md5` to describe a property or variable which contains both the MD5 hash and the file extension, just so that the code is more self-documenting.

### Test Coverage

Tested by hand:
1. Open Scratch Desktop
2. Click "Choose a Sprite"
3. Hover over "Bear-walking"

Expected: the "Bear-walking" library item shows the bear's walking animation.

Also tested:
- Loading a sprite, costume, or sound still works as intended.
- Playing sounds in the sound library works as intended.

### DIFF
--- a/src/components/library/library.jsx
+++ b/src/components/library/library.jsx
@@ -60,8 +60,15 @@ const getItemImageSource = function (item) {
         };
     }
 
-    // TODO: adjust libraries to be more storage-friendly; don't use split() here.
-    const md5ext = item.md5 || item.baseLayerMD5;
+    if (item.assetId && item.dataFormat) {
+        return {
+            assetId: item.assetId,
+            assetType: getAssetTypeForFileExtension(item.dataFormat)
+        };
+    }
+
+    // legacy
+    const md5ext = item.md5ext || item.md5 || item.baseLayerMD5;
     if (md5ext) {
         const [assetId, fileExtension] = md5ext.split('.');
         return {
--- a/src/components/library/library.jsx
+++ b/src/components/library/library.jsx
@@ -49,11 +49,19 @@ const getAssetTypeForFileExtension = function (fileExtension) {
 };
 
 /**
- * Figure out an `imageSource` (URI or asset ID & type) for a library item's icon.
+ * Figure out one or more icon(s) for a library item.
+ * If it's an animated thumbnail, this will return an array of `imageSource`.
+ * Otherwise it'll return just one `imageSource`.
  * @param {object} item - either a library item or one of a library item's costumes.
- * @returns {object} - an `imageSource` ready to be passed to a `ScratchImage`.
+ *   The latter is used internally as part of processing an animated thumbnail.
+ * @returns {LibraryItem.PropTypes.icons} - an `imageSource` or array of them, ready for `LibraryItem` & `ScratchImage`
  */
-const getItemImageSource = function (item) {
+const getItemIcons = function (item) {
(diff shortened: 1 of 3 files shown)


src/components/library/library.jsx
```diff
@@ -60,8 +60,15 @@ const getItemImageSource = function (item) {
...
-    const md5ext = item.md5 || item.baseLayerMD5;
+    if (item.assetId && item.dataFormat) {
+        return {
+            assetId: item.assetId,
+            assetType: getAssetTypeForFileExtension(item.dataFormat)
+        };
+    }
+
+    // legacy
+    const md5ext = item.md5ext || item.md5 || item.baseLayerMD5;
```
cwillisf, line 71: are these all still possible?

ericrosenbaum:
Looks like travis failed due to props validation:
```
/home/travis/build/LLK/scratch-gui/src/components/scratch-image/scratch-image.jsx
  97:13  error  'src' is missing in props validation  react/prop-types
```

ericrosenbaum:
also @cwillisf I'm curious how this will interact with the possible-future-changes (still in review) in https://github.com/LLK/scratch-gui/pull/5460, which changes the shape of the sprite library data

cwillisf:
OK, I think these changes preemptively resolve any trouble that might come from #5460, though it's possible I missed some detail. Also, these changes should make it easier to quickly fix anything broken by #5460 if I did miss anything. :D

src/containers/library-item.jsx
```diff
@@ -21,6 +21,8 @@ class LibraryItem extends React.PureComponent {
             'startRotatingIcons',
             'stopRotatingIcons'
         ]);
+        // TODO: is there a better "is it an array?" check to use here?
+        this.hasIconsArray = props.icons.hasOwnProperty('length');
```
apple502j, line 25:
```suggestion
        this.hasIconsArray = Array.isArray(props.icons);
```

cwillisf: I'm afraid that's just a little too readable! ;)

src/components/library/library.jsx
ericrosenbaum: We don't seem to generally use `baseLayerMD5` anymore, but seems safest to keep it here for now. We can revisit when we do a library data cleanup.

src/containers/library-item.jsx
```diff
@@ -21,6 +21,8 @@ class LibraryItem extends React.PureComponent {
             'startRotatingIcons',
             'stopRotatingIcons'
         ]);
+        // TODO: is there a better "is it an array?" check to use here?
```
adroitwhiz, line 24:
```suggestion

```
nit: this TODO can be removed now

PR #5662, opened 2020-04-21
Close Importing when the sound is invalid

### LINKED ISSUE #4639: Importing unsupported files result in forever "Importing..."

apple502j:
### Resolves
Resolves #4639 

### Proposed Changes
Close importing when sound is invalid

### Reason for Changes
Because sound importing didn't have it

### Test Coverage
manually tested it with random txt file

### DIFF
--- a/src/containers/sound-tab.jsx
+++ b/src/containers/sound-tab.jsx
@@ -140,7 +140,7 @@ class SoundTab extends React.Component {
                         this.props.onCloseImporting();
                     }
                 });
-            });
+            }, this.props.onCloseImporting);
         }, this.props.onCloseImporting);
     }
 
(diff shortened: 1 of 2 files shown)


ericrosenbaum (approved):
This does fix the issue. We noticed that the console warning is lost, but that's okay. Thanks for the fix!

PR #5731, opened 2020-05-08
Add new people sprites

ericrosenbaum:
The Bring Yourself into Scratch (BYIS) project has created several new people sprites, with many costumes, which we are adding to the sprite and costume libraries:

Characters 1
Characters 2
Kia
Luca
Noor
Sasha
Tatiana
Taylor
Trisha

### DIFF
--- a/src/lib/libraries/sprites.json
+++ b/src/lib/libraries/sprites.json
@@ -3426,7 +3426,7 @@
     },
     {
         "name": "Characters 1",
-        "md5": "59099ad030544d62c83da44cfddc315e.svg",
+        "md5": "03bc23a9fa12c1244c83a07a81f20bfd.svg",
         "type": "sprite",
         "tags": [
             "people",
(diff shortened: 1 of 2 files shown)


src/lib/libraries/sprites.json
```diff
@@ -11784,28 +12182,103 @@
...
+                    "dataFormat": "svg",
+                    "rotationCenterX": 50.565256229860694,
+                    "rotationCenterY": 130.02168173555543
                 }
             ],
             "sounds": [
                 {
                     "assetId": "83a9787d4cb6f3b7632b4ddfebf74367",
-                    "name": "pop",
+                    "name": "Pop",
```
kchadha, line 12215: Noticed this weird difference between this "Pop" with capital P and the "pop" sound with lowercase p in the Magic Wand sprite...

ericrosenbaum: Right! we are a bit inconsistent with capitalization across the libraries, unfortunately. There are a couple of pre-existing instances of "Pop" but they are almost all "pop".

Issue #4615, opened 2019-03-04
changing the ____ of ____ sensing block sprite scrolls the workspace to the top


BryceLTaylor:
### Expected Behavior

If you change the sprite/backdrop of the ____ of _____ sensing block should not change the view of the block workspace.

### Actual Behavior

When you change the sprite it scrolls the workspace to the top.  

This issue was brought to our attention from the community by user 34oo3958.

### Steps to Reproduce

Create a new project
Add a second sprite (any method)
On the second sprite drag a block (any) to the top of the workspace
Add a ___ of ___ block to the workspace
Scroll down and bring the ___ of ___ block with you.  
Make sure you can't see the first block you dragged out
Change the second part of that block, the sprite/stage
Note: you have now been scrolled to the top.

### Operating System and Browser

Mac Chrome

towerofnix:
Halfway-tracking in #4400; merging #4613 will fix this.

PR #5774, opened 2020-05-18
On workspace reload, save+load pre-reload metrics

adroitwhiz:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/4615

(See also [this discussion on scratch-gui#4400](https://github.com/LLK/scratch-gui/issues/4400#issuecomment-462994133))

### Proposed Changes

In `onWorkspaceUpdate`, which triggers a workspace clear + reload, save the old workspace metrics (scroll position and zoom) *before* calling `clearWorkspaceAndLoadFromXml`, and restore them afterwards.

### Reason for Changes

The previous code [restored the workspace metrics from `this.state.workspaceMetrics[this.props.vm.editingTarget.id]`](https://github.com/LLK/scratch-gui/blob/dee7462c57e5fe2cd726ef12ef7f338f2ee01868/src/containers/blocks.jsx#L384-L390). This appeared to be correct, because `setState` in React is supposedly asynchronous so `this.state.workspaceMetrics` shouldn't be changed until a tick later.

However, React was apparently setting the workspace metrics state synchronously after `clearWorkspaceAndLoadFromXml` was called, meaning that instead of restoring the *old* workspace metrics, it was restoring the metrics that scratch-blocks had just updated with.

Even if `setState` suddenly turns completely asynchronous in the future, this shouldn't cause a race condition-- restoring the workspace metrics should trigger another `onWorkspaceMetricsChange` call, always setting the proper state afterwards (as long as React guarantees that `setState` calls maintain their order, which I think it does).

### Test Coverage

Tested manually

### Testing Steps
- Add a sprite (so that there are at least 2 total sprites)
- Drag the "(attribute) of (sprite)" block into the block workspace
- Drag that block further down and to the right a lot
- Change the sprite selected in the block's dropdown
- Before: The block workspace will scroll up and to the left
- After: The block workspace should no longer change scroll position

### Browser Coverage

Linux
* [x] Firefox
* [x] Chrome

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -355,7 +355,11 @@ class Blocks extends React.Component {
             this.props.updateToolboxState(toolboxXML);
         }
 
-        if (this.props.vm.editingTarget && !this.state.workspaceMetrics[this.props.vm.editingTarget.id]) {
+        // For some reason, calling clearWorkspaceAndLoadFromXml sets this.state.workspaceMetrics... synchronously?
+        // Ensure that we're retaining the old metrics *before* reloading the workspace.
+        const oldMetrics = this.state.workspaceMetrics[this.props.vm.editingTarget.id];
+
+        if (this.props.vm.editingTarget && !oldMetrics) {
             this.onWorkspaceMetricsChange();
         }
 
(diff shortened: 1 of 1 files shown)


chrisgarrity:
@adroitwhiz could you add testing steps - thanks.

adroitwhiz:
@chrisgarrity Done

adroitwhiz:
#4542 fixed #4615, I believe.

PR #5795, opened 2020-05-20
Refactor monitor dragging

### LINKED ISSUE #2367: Small stage mode causes some problems interacting with monitors
Basically it seems like the monitor area looks only 1/4 of the size of the new stage. 

monitor-small-stage

### LINKED ISSUE #2949: Monitors can't be positioned past halfway right (or halfway down) when stage is in small display mode
...that's as far as you can position it:

screen shot 2018-08-21 at 5 37 58 pm

### LINKED ISSUE #5782: Monitor text can wrap depending on where the monitor loads in
### Expected Behavior

Text in monitors should not wrap, or at least be consistent when it does

### Actual Behavior

Text in monitors will always wrap to avoid overflowing the right edge of the stage, but that's according to where the monitor was *initially* loaded in. If you move the monitors, they will still attempt to wrap to the same width.

### Steps to Reproduce

1. Show the variable monitor for the default variable "my variable"
2. Move it horizontally to around the center of the stage
3. Set the variable value to a long string of text
4. Observe that the variable monitor's text does not wrap
5. Save and reload the project
6. Observe that the variable monitor's text now wraps
7. Drag the variable monitor to the left
8. Observe that the variable monitor's text *still* wraps the same amount, even if you change the variable's value

For instance, compare [this project](https://scratch.mit.edu/projects/397073459/editor/) where the monitor loads in on the left with [this project](https://scratch.mit.edu/projects/397075287/editor) where it loads in in the center.

In the first project, set the variable's value to 0 and drag the variable monitor to the right, then set the variable's value to the long string and observe that it does not wrap.

In the second project, drag the variable monitor to the left and observe that it will never go above a certain width, instead choosing to wrap.

Monitor starts on the left | Monitor starts in the middle
--- | ---
image | image

This is possibly the root cause of #3267.

Firefox + Chrome, probably other browsers as well

adroitwhiz:
### Resolves

- Resolves #5782
- Resolves #2367
- Resolves #2949

### Proposed Changes

This PR reimplements the dragging behavior previously done in react-draggable, but taking stage scale into account.

It also changes the monitors to always position themselves using CSS transforms and never setting the `top` or `left` position--those will affect text wrapping while CSS transforms will not, resulting in wrapping differences between monitors positioned with `top`/`left` and monitors positioned with a CSS transform. This is what led to #5782.

### Reason for Changes

These changes resolve the bugs linked above.

### Test Coverage

Tested manually

### Browser Coverage

Linux
* [x] Chrome
* [x] Firefox

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/monitor/monitor.jsx
+++ b/src/components/monitor/monitor.jsx
@@ -1,7 +1,8 @@
 import React from 'react';
+import classNames from 'classnames';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
-import Draggable from 'react-draggable';
+import {DraggableCore} from 'react-draggable';
 import {FormattedMessage} from 'react-intl';
 import {ContextMenuTrigger} from 'react-contextmenu';
 import {BorderedMenuItem, ContextMenu, MenuItem} from '../context-menu/context-menu.jsx';
@@ -36,15 +37,20 @@ const MonitorComponent = props => (
         holdToDisplay={props.mode === 'slider' ? -1 : 1000}
         id={`monitor-${props.label}`}
     >
-        <Draggable
-            bounds=".monitor-overlay" // Class for monitor container
+        <DraggableCore
             cancel=".no-drag" // Class used for slider input to prevent drag
-            defaultClassNameDragging={styles.dragging}
             disabled={!props.draggable}
+            onStart={props.onDragStart}
+            onDrag={props.onDrag}
             onStop={props.onDragEnd}
         >
             <Box
-                className={styles.monitorContainer}
+                className={classNames(styles.monitorContainer, {
+                    [styles.dragging]: props.dragging
+                })}
+                style={{
+                    transform: `translate(${props.x}px, ${props.y}px)`
+                }}
                 componentRef={props.componentRef}
                 onDoubleClick={props.mode === 'list' || !props.draggable ? null : props.onNextMode}
             >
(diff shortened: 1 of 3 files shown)


src/components/monitor/monitor.jsx
```diff
@@ -36,15 +37,20 @@ const MonitorComponent = props => (
...
+            onDrag={props.onDrag}
             onStop={props.onDragEnd}
         >
             <Box
-                className={styles.monitorContainer}
+                className={classNames(styles.monitorContainer, {
+                    [styles.dragging]: props.dragging
+                })}
+                style={{
+                    transform: `translate(${props.x}px, ${props.y}px)`
```
adroitwhiz, line 52: Will it cause any performance issues to re-render this component for every frame we're dragging?

rschamp:
@adroitwhiz Would you want to keep working on this? We're thinking about closing it unless you think all the issues are resolved.

adroitwhiz:
It's basically done--the only issue is that the monitor component re-renders once per drag event (which I believe is at most once per frame). I'll rebase this, see if I can fix that, and update you on this by next Monday.

adroitwhiz:
The re-rendering once per drag event happens because the monitor's coordinates are now stored as part of the component state, and dragging monitors updates those coordinates immediately. I don't think there's a clean way to stop it from happening--the monitor's coordinates *should* be part of its state.

I think it's perfectly fine to re-render the monitor once per frame whilst dragging--a Scratch project which changes a monitor value once per tick does basically the same thing anyway (albeit at 30 FPS).

PR #5799, opened 2020-05-21
Use DebouncedPropertyHOC instead of ThrottledPropetyHOC

### LINKED ISSUE #4703: The watermark doesn't appear after the first load
### Expected Behavior

The watermark should appear, when the user load (for the first time) a project from his computer.

### Actual Behavior

After the first load, the watermark doesn't appear.

### Steps to Reproduce

1. Create a new project
2. Save this project to your computer
3. Load the saved project from your computer

Peek 2019-04-01 01-42

Ubuntu 18.04, Chrome v72

### Comments

The same problem:
When I change my sprites quickly, the watermark component hasn't got time for rendering.
Peek 2019-04-01 01-43

I think this problem is in the `src/containers/watermark.jsx` file on the `ThrottledPropertyHOC('asset', 500)(Watermark)`
If I change 500 to 300 (or I remove the ThrottledPropertyHOC call), it will solve my problem.

### LINKED ISSUE #5135: Wrong watermark shows up if you click fast
### Expected Behavior

The watermark in the code area should always match the selected sprite

### Actual Behavior

Sometimes it doesn't!

### Steps to Reproduce

1. new project
2. add a sprite
3. click on first sprite in sprite icon pane
4. as quickly as you can, click on second sprite in sprite icon pane
5. wrong watermark is in code pane!

Aug-29-2019 20-31-34

### LINKED ISSUE #5777: Sprite tiles can display out-of-date costumes
### Expected Behavior

Sprite tiles should, given a second or two, display the sprite's latest costume

### Actual Behavior

If you quickly undo/redo in the paint editor, the sprite selector tiles will display the old costume until you hover over them with your mouse:
image

This appears to be a consequence of the `ThrottledPropertyHOC`: it limits a sprite tile's costume to be changed at most once every 500ms, but if you change it more often than that, it "throws away" all later changes.

### Steps to Reproduce

1. With the editor open, switch to the Costumes tab
2. Scribble over the cat's costume with the paintbrush or otherwise make a visible change to it
3. Observe the sprite tile changing
4. Quickly (within half a second) undo and then redo in the paint editor
5. Observe that the sprite tile now shows the original unmodified sprite

This also happens if you rapidly click the undo button in the paint editor.

All

apple502j:
### Resolves
Resolves #5777
Resolves #4703
Resolves #5135

### Proposed Changes
This adds `DebouncedPropertyHOC` and removes `ThrottledPropertyHOC`.

### Reason for Changes
ThrottledPropetyHOC ignored any changes to the props during the throttle, and this made some props not "up-to-date".

### How DebouncedPropetyHOC Works
`DebouncedPropertyHOC("propName", 30000)` will:
1) render the prop using the given propName, and start the timeout
2) propName was changed 10 secs after that, but ignore since we have the timer
3) anotherProp was changed, and render it with the old propName value and new anotherProp
4) After 30secs, the component will be re-rendered using the latest propName
5) After that, propName was updated. Since it's not throttled anymore, update immediately

### Test Coverage
The easy way to see if the bug is fixed is to run code such as
```
repeat 11
next costume
```


adroitwhiz:
Why remove the unit tests instead of modifying them to test the new behavior?

benjiwheeler:
@apple502j Two questions:

1. I have the same question as @adroitwhiz -- would it be possible to make tests for this, or put back in the tests you removed and adapt them?

2. Can you write more in the description about how to test this? This might have unforeseen implications, right? So it'd be helpful if we had several specific sequences of steps to try to confirm that things are working correctly.

apple502j:
@benjiwheeler I am not sure if I can test this as a unittest - unless enzyme actually renders the component, executes setTimeout and re-renders when the props get changed. I don't think overriding global.setTimeout with mocks will work, either. It may be possible to make an integration test but it depends on timing and may create flaky test. 

Manual testing step: the easiest way to test it is by checking the watermark and sprite selector. Use the attached project for testing.
- click the flag. the costume switches from A to B, and B to C
- the watermark changes once; it is because _useInstead is set to the costume B (because we can't access prev props inside renderer) and the component is rendered using _useInstead when it changes from B to C - any way to fix this?  note this will only happen once per `if (!this._timeout)`
- after a bit the watermark should change to C
[test-5777.sb3.zip](https://github.com/LLK/scratch-gui/files/4792881/test-5777.sb3.zip)

benjiwheeler:
I rebased this on develop

benjiwheeler:
@seotts and I took a quick look at this. I tested this and it works and its scope is narrow, and it improves a tiny detail of the editor: it updates the “watermark” sprite image more reliably. 

But it needs tests written (or at least, someone should try!)

Note that we should review this with performance in mind, to make sure the "throttled" approach is as performant as the "debounced" approach.

PR #5855, opened 2020-06-04
Use direction: ltr to prevent RTL languages from mirroring the variable slider

### LINKED ISSUE #5545: Slider monitors should not be mirrored in RTL languages
### Expected Behavior
Cultures that use RTL languages still use the same coordinate system and numbers as other languages. In a number line the numeric values should increase to the right. e.g.
```
0....5....10....15 etc
```

In other words numeric sliders should look identical in LTR and RTL languages.
### Actual Behavior

Slider values have the larger value on the left and smaller value to the right in RTL languages.
slider

### Steps to Reproduce

Create a slider, then switch between LTR and RTL languages.

BoomerScratch:
### Resolves

Resolves #5545

### Proposed Changes

Adds a CSS rule so the RTL languages do not mirror the slider.

### Reason for Changes

The reason is at proposed changes above.

### Test Coverage

Manually tested with a local clone of the GUI.

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/monitor/monitor.css
+++ b/src/components/monitor/monitor.css
@@ -66,6 +66,7 @@
 .slider {
     width: 100%;
     transform: translateZ(0); /* Fixes flickering in Safari */
+    direction: ltr;
 }
 
 .list-monitor {


chrisgarrity:
@BoomerScratch thanks for working on this. It turns out this is not as obvious as it seems. Even though the mathematical coordinate system and number line are still LTR, the best practice for UI elements seems to be that they increase to the left in RTL languages. See https://medium.com/@giladalmosnino/ux-best-practices-for-bi-directional-languages-9bd7b96dc8c2, and https://material.io/design/usability/bidirectionality.html#mirroring-layout.

So I'm not going to close this yet. We need to get more feedback from Arabic, Hebrew, Persian users of Scratch before deciding whether or not to make this change.

chrisgarrity:
@BoomerScratch as I mentioned in closing the issue, we decided not to make this change after getting more feedback from folks in the community. Thanks for working on it.

PR #5861, opened 2020-06-05
Stop building gh pages branches for dependabot 

picklesrus:
Updating the regex since we switched from greenkeeper.


picklesrus:
I think this passes on travis: https://travis-ci.org/github/LLK/scratch-gui/builds/695062870
I'm not sure why github doesn't know so yet.

rschamp:
Thanks @picklesrus!

PR #5877, opened 2020-06-10
Add assets to editingTarget at the time it started uploading

### LINKED ISSUE #5876: Assets will be added to new editingTarget if changed during uploading
### Expected Behavior
Assets should be uploaded to the editingTarget at the time it started uploading, not when it finished uploading

### Actual Behavior
If the editingTarget was changed during uploading, the asset will be added to the new editingTarget.

### Steps to Reproduce
It's easy to do this by uploading GIFs or multiple files:
1. Upload a GIF on Sprite A's costume tab
2. Quickly switch to Sprite B (or stage)
3. Some GIFs are added to Sprite A, and others are added to Sprite B

Using multiple files:
1. Upload a lot of sound files on Sprite A's sound tab
2. Quickly switch to Sprite B (or stage)
3. Some sounds are added to Sprite A, and others are added to Sprite B

any

To fix this, we just need to pass `vm.editingTarget.id` to `addCostume`. However, the value must be obtained before handleFileUpload.

apple502j:
### Resolves
Resolves #5876 

### Proposed Changes
Add assets to editingTarget at the time it started uploading, not when `addCostume`/`addSound` is called. There are still some functions that use the old behavior, but I don't expect adding such assets (like adding blank costume) will be long enough for the user to switch sprites.

### Reason for Changes
Previously if you switch sprites while adding assets it would add some of the assets to the new sprite instead of the previously selected one.

### Test Coverage
Tested on Firefox 77, Windows 10

To test this,
1) Import a GIF file via costumes tab
2) Switch sprites
3) Check that all imported GIFs are still in the previous sprite
4) Import sounds (multiple preferably, to lag it a bit) via sounds tab
5) Switch sprites
6) Check that all imported sounds are still in the previous sprite

### DIFF
--- a/src/containers/costume-tab.jsx
+++ b/src/containers/costume-tab.jsx
@@ -158,14 +158,17 @@ class CostumeTab extends React.Component {
         const blob = new Blob([item.asset.data], {type: item.asset.assetType.contentType});
         downloadBlob(`${item.name}.${item.asset.dataFormat}`, blob);
     }
-    handleNewCostume (costume, fromCostumeLibrary) {
+    handleNewCostume (costume, fromCostumeLibrary, targetId) {
         const costumes = Array.isArray(costume) ? costume : [costume];
 
         return Promise.all(costumes.map(c => {
             if (fromCostumeLibrary) {
                 return this.props.vm.addCostumeFromLibrary(c.md5, c);
             }
-            return this.props.vm.addCostume(c.md5, c);
+            // If targetId is falsy, VM should default it to editingTarget.id
+            // However, targetId should be provided to prevent #5876,
+            // if making new costume takes a while
+            return this.props.vm.addCostume(c.md5, c, targetId);
         }));
     }
     handleNewBlankCostume () {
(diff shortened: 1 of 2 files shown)


src/containers/costume-tab.jsx
```diff
@@ -158,14 +158,17 @@ class CostumeTab extends React.Component {
...
-    handleNewCostume (costume, fromCostumeLibrary) {
+    handleNewCostume (costume, fromCostumeLibrary, targetId) {
         const costumes = Array.isArray(costume) ? costume : [costume];
 
         return Promise.all(costumes.map(c => {
             if (fromCostumeLibrary) {
                 return this.props.vm.addCostumeFromLibrary(c.md5, c);
             }
-            return this.props.vm.addCostume(c.md5, c);
+            // If targetId is falsy, VM should default it to editingTarget.id
```
R4356th, line 168: You probably want it to be "false" instead of "falsy". :P

apple502j: Falsy is a valid word.

Issue #5619, opened 2020-04-09
FAIL test/integration/localization.test.js Localization › Switching languages


fsih:
I have hit this a couple times today. It passes for me locally

FAIL test/integration/localization.test.js (30.112s)
  ● Localization › Switching languages
    Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
      
      at node_modules/jest-jasmine2/build/queue_runner.js:64:21
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:523:19)
      at ontimeout (timers.js:498:11)
      at tryOnTimeout (timers.js:323:5)
      at Timer.listOnTimeout (timers.js:290:5)
  ● Localization › Loading with locale shows correct blocks
    WebDriverError: unknown error: Chrome failed to start: exited abnormally.
      (chrome not reachable)
      (The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
      (Driver info: chromedriver=80.0.3987.106 (f68069574609230cf9b635cd784cfb1bf81bb53a-refs/branch-heads/3987@{#882}),platform=Linux 4.15.0-1028-gcp x86_64)
      
      at Object.checkLegacyResponse (node_modules/selenium-webdriver/lib/error.js:546:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:509:13)
      at doSend.then.response (node_modules/selenium-webdriver/lib/http.js:441:30)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)
      From: Task: WebDriver.createSession()
      at Function.createSession (node_modules/selenium-webdriver/lib/webdriver.js:769:24)
      at Function.createSession (node_modules/selenium-webdriver/chrome.js:761:15)
      at createDriver (node_modules/selenium-webdriver/index.js:170:33)
      at Builder.build (node_modules/selenium-webdriver/index.js:642:16)
      at SeleniumHelper.getDriver (test/helpers/selenium-helper.js:73:7)
      at wrapper (node_modules/lodash.bindall/index.js:550:15)
      at _callee2$ (test/integration/localization.test.js:29:22)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:296:22)
[...]

PR #5885, opened 2020-06-12
Update localization test

BryceLTaylor:
### Resolves

Updates a localization test that was being flaky.  

- Resolves #
https://github.com/LLK/scratch-gui/issues/5619
It may not fully resolve it, but at least it removes unnecessary steps from the test that was timing out.

### Proposed Changes

Removes unnecessary steps from the localization integration test that was failing sometimes from a timeout.
Updates the version of Chromedriver to the most recent version

### Reason for Changes

Hopefully it will stabilize the tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3207,6 +3207,32 @@
         "@types/yargs": "^13.0.0"
       }
     },
+    "@nodelib/fs.scandir": {
+      "version": "2.1.3",
+      "resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.3.tgz",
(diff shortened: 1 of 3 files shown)


fsih:
The tab-switching lines were added here: https://github.com/LLK/scratch-gui/commit/130bd3f05cd737b2d4f782efb067c033cb3aa092
The context doesn't really explain why the tab changes were necessary... I think I agree with you that they aren't.

The await lines were added in https://github.com/LLK/scratch-gui/commit/3d588321ca22398d01f7423cbab469cce4e5f6ab
If we've decided that they're not necessary, we should probably remove them from all the other places they were added in this PR, as well. It looks like the contributor was trying things to figure out how to get the tests to pass, and they were eventually fixed by a different pr: https://github.com/LLK/scratch-gui/pull/4410/commits/9bef38e977982fa3a03fd5004ccc6da6b8dc3017

BryceLTaylor:
@fsih It is probably worth putting this PR in and take a look at how stable the test is for a little while ago before deciding to pull those steps out from other tests.

Issue #1783, opened 2018-04-13
File->Download causes crash on iOS Safari


ericrosenbaum:
### Expected Behavior

Selecting File->Download to your computer on iOS Safari should work, or be grayed out.

### Actual Behavior

It tries to load a URL starting with `blob`, and shows an error page including `WebKitBlobResource error 1` 

### Operating System and Browser

iOS 11.2.2 Safari

hedgehog125:
Maybe it could detect you're on ios and tell the user that the device can't download files? Otherwise it may be confusing for younger users?

tyreun:
Same issue with iOS11.3

paulkaplan:
Looks like using a library that handles the many edge cases of saving files out of the browser like https://github.com/eligrey/FileSaver.js/ would help. It is definitely possible to save those files on mobile safari, just takes a fair bit of trickery.

BryceLTaylor:
https://github.com/LLK/scratch-gui/issues/1030 contained some screenshots that are out of date but might be useful.  These two were duplicates.

exomut:
I took a look around in the “src/containers/project-saver.jsx” file since it seems to be the code causing the problem. I wasn’t able to solve the issue, but if you surround lines 48 and 49 with a setTimeout() you are able to get a downloadable version of your project on iOS.

```javascript
setTimeout(function(){
    window.URL.revokeObjectURL(url);
    document.body.removeChild(saveLink);
}, 1000);
```

The only problem is that the filename can’t be set. It will save as “Unknown.zip”. Using a file manager on iOS you can rename the file to “Your Project.sb3” and upload it back to Scratch 3.0 without error. (Tested on iOS 12)

Running the Scratch site in full screen while canceling a save still crashes Safari though.

I am now able to work on Scratch projects on my iPad and save them. Although, renaming is a hassle. FIlesaver.js looks like it could be an option since it is using the MIT license, but I am not too sure how compatible the MIT and BSD licenses are.

rickwookie:
Disappointing to see no progress on a fix for this a year on, particularly now that iOS/iPadOS 13 fully supports file downloads within mobile Safari.

tyreun:
Hi,

Take a look at 

http://www.ac-grenoble.fr/maths/scratch/

Cheers

Ronan

tyreun:
You can download on iOS with safari

benjiwheeler:
[...]

PR #5927, opened 2020-06-23
Prevent crash when saving locally on iOS 13 safari

ericrosenbaum:
### Resolves

resolves https://github.com/LLK/scratch-gui/issues/1783

### Proposed Changes

When saving to a local file (file -> save to your computer), wait briefly before removing the download link. This prevents the crash on iOS 13, and the file downloads with a correct filename and extension.

### Reason for Changes

Not 100% sure why this works, but it probably prevents a race condition in iOS Safari.

### Test Coverage

No tests

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iOS
* [x] iOS 13 Safari
* [x] iOS 12 Safari (via simulator)

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/lib/download-blob.js
+++ b/src/lib/download-blob.js
@@ -14,10 +14,13 @@ export default (filename, blob) => {
         downloadLink.download = filename;
         downloadLink.type = blob.type;
         downloadLink.click();
-        document.body.removeChild(downloadLink);
-        window.URL.revokeObjectURL(url);
+        // remove the link after a timeout to prevent a crash on iOS 13 Safari
+        window.setTimeout(() => {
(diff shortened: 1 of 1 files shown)


ericrosenbaum:
thanks @chrisgarrity - I was able to save and then load a project on iOS 13 on both the simulator and on my iPhone. We can look into the issue further if we hear reports of people having trouble after we get this change in.

PR #5941, opened 2020-06-26
Don't let user exit full screen when branding shows

BoomerScratch:
### Resolves

Resolves LLK/scratch-www#4102

### Proposed Changes

Adds `&& !this.props.showBranding` (`!this.props.showBranding` will return false when branding shows), so escape won't do anything when the project is embedded. (Because embedded projects are full screen)

### Reason for Changes

As I said above: so escape won't do anything when the project is embedded.
\
\
\
\
@benjiwheeler

### DIFF
--- a/src/containers/stage-header.jsx
+++ b/src/containers/stage-header.jsx
@@ -25,7 +25,7 @@ class StageHeader extends React.Component {
         document.removeEventListener('keydown', this.handleKeyPress);
     }
     handleKeyPress (event) {
-        if (event.key === 'Escape' && this.props.isFullScreen) {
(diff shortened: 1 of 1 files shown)


benjiwheeler:
@BoomerScratch I took the liberty of altering this PR a bit. I introduced a redux prop, `editorModeEnabled`, to track whether the player can shift to showing the editor. I'm hoping this will be clearer than adding meaning to `showBranding`, which might be confusing to developers later. What do you think?

BoomerScratch:
@benjiwheeler Sounds good!

PR #5961, opened 2020-07-02
Handle auto-scrolling and highlighting list monitor index

apple502j:
**LLK/scratch-vm#2507 must be merged first**

### Resolves
GUI part for these issues:
Resolves LLK/scratch-vm#1385
Resolves LLK/scratch-vm#1386

### Proposed Changes
auto-scrolling and highlighting
Adds list index highlighting and list auto-scrolling feature (although the bugtracker says lack of auto-scrolling is a bug)

When a list monitor receives update to its `highlightItem`, it cancels the setTimeout (which i'll mention below), and does two things:
- adds the item to `highlightItems` which is a Map (not immutable) of number to setTimeout which unhighlights it after 500ms
- re-renders list monitor with new `highlightItem` and `highlightItems`, which scrolls to added item (list-monitor-scroller L87)

When highlighting $control-primary is used.

Note that we pass stringified version of highlightItems to List so that it re-renders - it's PureComponent and doesn't re-render when unhighlighting by itself.

### Reason for Changes
compatibility with Scratch 2.0

### Test Coverage
Attached a project sb3 - rename and open. click green flag, go to fullscreen, click the stage (not the stage monitor), try pressing arrow keys!
[scrollable-list.sb3.zip](https://github.com/LLK/scratch-gui/files/4898657/scrollable-list.sb3.zip)

### DIFF
--- a/src/components/monitor-list/monitor-list.jsx
+++ b/src/components/monitor-list/monitor-list.jsx
@@ -26,6 +26,7 @@ const MonitorList = props => (
                     <Monitor
                         draggable={props.draggable}
                         height={monitorData.height}
+                        highlightItem={monitorData.highlightItem}
                         id={monitorData.id}
                         isDiscrete={monitorData.isDiscrete}
                         key={monitorData.id}
--- a/src/components/monitor/list-monitor-scroller.jsx
+++ b/src/components/monitor/list-monitor-scroller.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import bindAll from 'lodash.bindall';
 import {FormattedMessage} from 'react-intl';
-
+import {Set} from 'immutable';
 import styles from './monitor.css';
 import {List} from 'react-virtualized';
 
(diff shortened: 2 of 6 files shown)


ericrosenbaum:
@apple502j thanks for your work on this, and sorry it took a while for us to look at it. We decided that our design team will need to make some decisions before we can proceed. We'll need to decide exactly what behavior we want, e.g. will making this change now break existing 3.0 projects that expect no auto-scrolling, exactly when should auto-scrolling occur vs. not, and should the highlight fade as in 2.0. We will also need to assess performance to make sure this doesn't slow things down, and also ideally add unit and integration tests where possible. We're going to close this for now, and possibly re-open if we decide to return to it.

Wowfunhappy:
I know this is closed, but I've been using it in my own Scratch build because it's great, and more importantly Eric mentioned that LLK might returned to it in the future. So, I wanted to write a note about some odd behavior I noticed with this PR.

If you get the same item from a list repeatedly, the list number will only be highlighted the first time it is accessed. You need to highlight a different number before the first number can be highlighted again.

For example:
Screen Shot 2023-03-26 at 1 58 40 PM
In the above program, if I press the down arrow, `1` and `2` will be highlighted repeatedly, as expected.

Logically, I would think that if I press the up arrow, `1` should be highlighted repeatedly every second. However, `1` is only highlighted the first time the script is run. It will not be highlighted again until after some action causes `2` to be highlighted.

[List Highlighting Bug.sb3.zip](https://github.com/LLK/scratch-gui/files/11072526/List.Highlighting.Bug.sb3.zip)

I made the same program in Scratch 2, and I confirmed that pressing the up arrow makes `1` highlight repeatedly, as expected.

PR #5974, opened 2020-07-09
Fix backpacking zero-height/width costumes

### LINKED ISSUE #5972: (Sprites with) zero-height/width costumes cannot be backpacked
### Expected Behavior
Zero-height/width costumes (and sprites with current costume set to it) can be backpacked

### Actual Behavior
DOMException raised inside https://github.com/LLK/scratch-gui/blob/develop/src/lib/backpack/jpeg-thumbnail.js

### Steps to Reproduce
Try to backpack an vector costume with everything deleted, or with very thin line

any

Code is ready, gonna PR once the team is ready

apple502j:
### Resolves
Resolves #5972

### Proposed Changes
If the sprite height/width is less than one, use 96(maxDimension)x96 white square as the backpack thumbnail

### Reason for Changes
drawImage fails if the image is too small. I did not compare width===0 because it's possible that very-small-but-positive numbers (like 0.01) can cause this as well, and I heard some implementations floor instead of round.

### Test Coverage
Manually tested using http://localhost:8601/?token=X-TOKEN-HERE&backpack_host=https://backpack.scratch.mit.edu&username=apple502j

(by the way, why is it JPEG instead of PNG?)

### DIFF
--- a/src/lib/backpack/jpeg-thumbnail.js
+++ b/src/lib/backpack/jpeg-thumbnail.js
@@ -6,17 +6,24 @@ const jpegThumbnail = dataUrl => new Promise((resolve, reject) => {
 
         const maxDimension = 96; // 3x the maximum displayed size of 32px
 
-        if (image.height > image.width) {
-            canvas.height = maxDimension;
-            canvas.width = (maxDimension / image.height) * image.width;
+        if (image.height < 1 || image.width < 1) {
+            canvas.width = canvas.height = maxDimension;
+            // drawImage can fail if image height/width is less than 1
+            // Use blank image; the costume is too small to render anyway
+            ctx.fillStyle = 'white'; // Create white background, since jpeg doesn't have transparency
+            ctx.fillRect(0, 0, maxDimension, maxDimension);
         } else {
-            canvas.width = maxDimension;
-            canvas.height = (maxDimension / image.width) * image.height;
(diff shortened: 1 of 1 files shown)


adroitwhiz:
According to MDN, an `` element's width and height [must be integers](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img).

src/lib/backpack/jpeg-thumbnail.js
```diff
@@ -6,17 +6,24 @@ const jpegThumbnail = dataUrl => new Promise((resolve, reject) => {
...
 
-        if (image.height > image.width) {
-            canvas.height = maxDimension;
-            canvas.width = (maxDimension / image.height) * image.width;
+        if (image.height < 1 || image.width < 1) {
+            canvas.width = canvas.height = maxDimension;
+            // drawImage can fail if image height/width is less than 1
+            // Use blank image; the costume is too small to render anyway
+            ctx.fillStyle = 'white'; // Create white background, since jpeg doesn't have transparency
+            ctx.fillRect(0, 0, maxDimension, maxDimension);
```
BryceLTaylor: This is slightly more maintainable (in the case that we ever change canvas.width/height).

```suggestion
            ctx.fillRect(0, 0, canvas.width, canvas.height);
```

Issue #5828, opened 2020-05-28
All integration tests fail on CircleCI


rschamp:
### Expected Behavior

_Please describe what should happen_
All the tests should pass when running in CircleCI

### Actual Behavior

_Describe what actually happens_
None of the integration tests ever pass. E.g. https://circleci.com/gh/LLK/scratch-gui/6457

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_
Push a commit to develop and see how it runs in CircleCI

apple502j:
This may be fixed by bumping both node (to 10) and chromedriver (to latest) - it seems like `circleci/node:8.16-browsers` is not updated?

rschamp:
@apple502j That would explain this! Thanks for taking a look. @picklesrus This might be a good first lead. If we do update the CircleCI config it might be a good idea to see if we can lock down the versions better so we can update on our own schedule.

PR #6038, opened 2020-07-27
Get unittests and integration tests (consistently?) passing on circle again. 

picklesrus:
- Update the docker image to node 10 and corresponding browsers (~chrome 83)
- Unit tests: They were running out of memory. Set --maxWorkers to 2 so they don't spawn so many that we run 
out of memory.  This value probably could be optimized.
- Integration tests: costumes test and tutorials test were routinely failing with an error claiming that a tab in chrome had crashed.  I think there is a weird interaction when running some of these tests in parallel. Setting it so each test runs in a separate docker instance also solves the problem, but that's not sustainable. Instead set --runInBand on integration tests (like we do in package.json) so they run one at a time.

Consistency Note: This passed 6 times in a row, which isn't that many... but far better than what I've gotten so far.

https://github.com/LLK/scratch-gui/issues/5828

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ aliases:
         - v1-npm-
   - &defaults
     docker:
-      - image: circleci/node:8.16-browsers
+      - image: circleci/node:10-browsers
     working_directory: ~/repo
 
 jobs:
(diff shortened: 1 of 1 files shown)


rschamp (approved):
🎉

PR #6116, opened 2020-08-21
Force LTR on drag-layer

### LINKED ISSUE #6104: In RTL, asset tiles are not shown while dragging
### Expected Behavior

While dragging an asset tile (sprite, costume, sound or backdrop), it should be visible.

drag-LTR

### Actual Behavior

In RTL, it is invisible, apparently drawn far off screen. This gif from @fsih shows the offset:

dragging rtl

apple502j:
### Resolves
Resolves #6104 

### Proposed Changes
Force LTR on drag-layer, by adding `direction: ltr;`

### Reason for Changes
No matter which direction it displays, values passed to `transform()` is calculated from top-left to bottom-right.

### Test Coverage
Tested on the browser

### DIFF
--- a/src/components/drag-layer/drag-layer.css
+++ b/src/components/drag-layer/drag-layer.css
@@ -9,7 +9,8 @@
     left: 0;
     top: 0;
     width: 100%;
-    height: 100%
+    height: 100%;
+    direction: ltr;
 }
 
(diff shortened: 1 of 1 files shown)


seotts:
This looks good! Thanks for fixing this.

PR #6130, opened 2020-08-27
Move src/* dependencies into `dependencies`

cwillisf:
### Resolves

Progress toward LLK/scratch-desktop#144

### Proposed Changes

The dependencies required to safely `require('scratch-gui/src/index.js')` (or `import` equivalent) are now listed in `dependencies` rather than `devDependencies`.

I also removed a few unused dependencies. Note that several dependencies which initially appear unused are actually used. For example, `babel-core` looks redundant (since we have `@babel/core` as well) but it's used by some of the middleware involved in our tests. Likewise `jest-junit` is not used by our regular tests, but it is used by the CI test scripts. In the end, only `lodash.isequal`, `lodash.pick`, and `svg-to-image` were actually unused.

### Reason for Changes

The `npm install` command installs `dependencies` of dependencies, but not `devDependencies` of dependencies. By moving these into the `dependencies` list, `scratch-desktop` can now `import 'scratch-gui/src/index.js'` without needing to duplicate all of `scratch-gui`'s dependencies in `scratch-desktop`'s `package.json`.

### Test Coverage

Covered by existing `scratch-gui` tests.


package.json
```diff
@@ -115,14 +79,49 @@
...
+    "eslint-config-scratch": "^5.0.0",
+    "eslint-import-resolver-webpack": "^0.11.1",
+    "eslint-plugin-import": "^2.18.2",
+    "eslint-plugin-jest": "^22.14.1",
+    "eslint-plugin-react": "^7.12.4",
+    "gh-pages": "github:rschamp/gh-pages#publish-branch-to-subfolder",
+    "html-webpack-plugin": "^3.2.0",
+    "jest": "^21.0.0",
+    "jest-junit": "^7.0.0",
+    "lodash.throttle": "4.0.1",
```
apple502j: This is used in cloud-provider.js. While not used in Scratch Desktop, it should be moved to dependencies.

package.json
```diff
@@ -115,14 +79,49 @@
...
+    "@babel/core": "^7.1.2",
+    "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
+    "@babel/plugin-syntax-dynamic-import": "^7.0.0",
+    "@babel/plugin-transform-async-to-generator": "^7.1.0",
+    "@babel/preset-env": "^7.1.0",
+    "@babel/preset-react": "^7.0.0",
+    "babel-core": "7.0.0-bridge.0",
+    "babel-eslint": "^10.0.1",
+    "babel-loader": "^8.0.4",
+    "base64-loader": "1.0.0",
```
apple502j: Used in sound-payload.js

package.json
```diff
@@ -115,14 +79,49 @@
...
+    "jest": "^21.0.0",
+    "jest-junit": "^7.0.0",
+    "lodash.throttle": "4.0.1",
+    "mkdirp": "^1.0.3",
+    "raf": "^3.4.0",
+    "react-test-renderer": "16.2.0",
+    "redux-mock-store": "^1.2.3",
+    "rimraf": "^2.6.1",
     "selenium-webdriver": "3.6.0",
     "startaudiocontext": "1.2.1",
```
apple502j: This is definitely dependencies, used in shared-audio-context.js

package.json
```diff
@@ -115,14 +79,49 @@
...
-    "style-loader": "^0.23.0",
-    "svg-to-image": "1.1.3",
     "text-encoding": "0.7.0",
```
apple502j: While not used by Scratch Desktop (I think), default-project/index.js uses text-encoding. Also, why not change to fastestsmallesttextencoderdecoder?

package.json
```diff
@@ -115,14 +79,49 @@
...
-    "to-style": "1.3.3",
     "uglifyjs-webpack-plugin": "^1.2.5",
-    "wav-encoder": "1.3.0",
     "web-audio-test-api": "^0.5.2",
     "webpack": "^4.6.0",
     "webpack-cli": "^3.1.0",
```
apple502j, line 128: `xhr` is used in save-project-to-server.js

apple502j (changes requested):
Some devDependencies should be in dependencies

package.json
cwillisf: It might be a good idea to switch to smallestfastesttextencoderdecoder but I'd rather not make that part of this change.

cwillisf:
Thanks @apple502j! Interestingly, making those changes in package.json did not cause any changes in package-lock.json, maybe because scratch-gui dependencies already listed those packages as non-dev dependencies.

rschamp (approved):
It's possible that the dependencies shared between gui and www could cause problems once this gets to scratch-www. E.g. by making React a dependency here, I believe we are influencing the version of React used upstream.  It sounds like a peer dependency. But we can't leave it solely as a peer dependency or it won't get installed when you're just working on scratch-gui and run npm install.  Peer dependencies are also ignored by tools like dependabot. It's possible it should it be left as a dev-dependency + peer dependency? For now I think we should just try it this way.

I think it's going to be hard to predict any issues with this just by analyzing this list. So I think we should just try merging and paying special attention to what happens when this gets into www. Today is probably the best day to try doing this since then we'll have a week to see/fix any fallout.

PR #6140, opened 2020-08-31
Have circle deploy to the standard npm tags instead of travis.

picklesrus:
Deploy to latest, stable (although we don't use gui's master branch), and hotfix.


rschamp (approved):
(Acknowledging we won't merge this til we're both back from vacation)

PR #6162, opened 2020-09-04
Added Getting Started video in 7 languages

zoebentley:
Adds new Getting Started video for: 
- Spanish
- Latin American Spanish (same video as Spanish)
- French
- isiZulu
- Amharic
- Ukrainian  
- Kiswahili (Swahili)

When you load the Getting Started tutorial with the language set to each of those languages, the video should appear in that language instead of English.


chrisgarrity (approved):
Looks good - I checked the ids on wistia, and they all look like the correct ones.

PR #6195, opened 2020-09-18
Remove Travis config since we're using circle

picklesrus:
Circle should be doing all the work now.


picklesrus:
I think I need to turn Travis off via the UI first. Will do that early next week.

rschamp (approved):
✂️ ✂️ ✂️

PR #6354, opened 2020-11-09
Added Getting Started video in 19 languages for Phase 2

zoebentley:
Modifies translate-video.js file with links to the Getting tutorial video in the following 19 languages: 

1. Arabic
2. Catalan
3. Chinese Simplified
4. Chinese Traditional
5. Danish
6. Dutch
7. German
8. Haitian Creole
9. Hebrew
10. Indonedian
11. Khmer
12. Korean
13. Kurdish
14. Persian
15. Polish
16. Russian
17. Thai
18. Turkish
19. Vietnamese

Note: New tutorial videos were also created in Japanese and Italian, but because those languages had Getting Started videos already, we were able to simply replace the file within our video hosting platform and maintain the same link.


chrisgarrity (approved):
These changes look good. Two things to note about the videos:
* In the Turkish video at about 30 seconds: https://scratch.wistia.com/medias/p3g8ek3cuh, when dragging out a sound block, the name of the sound looks like it's in Thai, not Turkish. Unfortunately the names of assets don't get changed when you switch languages.
* In the original getting started videos the thumbnail is the landscape scene, in the new videos it's a screenshot of blocks (very convenient for reviewing because I knew which ones were new)

Those are both things that can be dealt with on the wistia side, so no need to hold this PR.

PR #6356, opened 2020-11-09
add eslint comments

benjiwheeler:
Adds some eslint comments to make gui ready to pass eslint linting if eslint version is updated.


benjiwheeler:
I'm not positive this is the right approach -- it might be better to try to resolve these violations by fixing them more substantially

benjiwheeler:
Looks good to @seotts and me!

PR #6369, opened 2020-11-10
Revert "build(deps): bump query-string from 5.1.1 to 6.13.7"

picklesrus:
Reverts LLK/scratch-gui#6342

The PR to update gui on www is failing with the same error that the query-string update to www is. I'm thinking this is related so reverting to see if that solves the problem. Here are the builds:

The www version of the PR to bump query string doesn’t build with this error: https://travis-ci.org/github/LLK/scratch-www/builds/742529546#L3016
And there’s a PR to update gui on www that is currently failing with this error: https://travis-ci.com/github/LLK/scratch-www/builds/199659136#L3018 (edited)


apple502j:
This may be due to some issues with uglifyjs.

PR #6463, opened 2020-12-08
About menu

cwillisf:
### Resolves

Work toward LLK/scratch-desktop#97

### Proposed Changes

* Support displaying a menu for the About button, if `onClickAbout` is a list of items instead of just one callback
* Adjust the design of the telemetry modal to better support letting the user change their mind in a later session

### Reason for Changes

Our design for LLK/scratch-desktop#97 involves adding a menu to the Scratch app for Windows & macOS. This menu will contain an entry to open the "About" window, display the privacy policy, and open the telemetry modal.

Viewing the privacy policy in-app is not only a good idea, it's also now required by Microsoft Store policy.

Re-opening the telemetry modal on demand will allow the user to change their mind at any time, which is not required by any policy but was recommended to us and seems like a good idea anyway.

The existing functionality for the "About" button, in which it calls a callback instead of opening a menu, is retained by this set of changes because it's being used by some of the Scratch Lab efforts.

### Test Coverage

Tested in context in scratch-desktop.

### DIFF
--- a/src/components/telemetry-modal/telemetry-modal.css
+++ b/src/components/telemetry-modal/telemetry-modal.css
@@ -51,7 +51,7 @@ $ui-gray: hsla(0, 0%, 95%, 1);
 }
 
 /* stack the radio buttons vertically, not horizontally */
-.radioButtons label {
+.radio-buttons label {
     display: block;
     margin: 0.5rem;
     transition: all .125s ease;
@@ -63,16 +63,16 @@ $ui-gray: hsla(0, 0%, 95%, 1);
     vertical-align: middle;
 }
 
-.radioButtons label:hover {
+.radio-buttons label:hover {
     background-color: $ui-blue-10percent;
 }
 
-.radioButtons label.labelSelected,
-.radioButtons label.labelSelected:hover {
+.radio-buttons label.label-selected,
+.radio-buttons label.label-selected:hover {
     background-color: $ui-blue-25percent;
 }
 
-.radioButtons input[type="radio"] {
+.radio-buttons input[type="radio"] {
     margin: -1px 0.75rem 1px;
     border: 1px solid $active-gray;
     border-radius: 50%;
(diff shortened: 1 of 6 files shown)


rschamp (approved):

src/components/telemetry-modal/telemetry-modal.css
```diff
@@ -107,16 +107,17 @@ $ui-gray: hsla(0, 0%, 95%, 1);
...
     font-weight: bolder;
 }
 
-[dir="ltr"] .button-row {
-    text-align: right;
-}
-
-[dir="rtl"] .button-row {
-    text-align: left;
+.settingWasUpdated {
```
apple502j: Don't this become snakecase?

cwillisf: Apparently it works either way, but I should adjust it to follow our conventions.

cwillisf:
@rschamp your last review was prior to the telemetry settings dialog design being settled. The design is now implemented so I've asked for a new review. If you'd like someone else to review LMK or reassign :)

src/components/telemetry-modal/telemetry-modal.jsx
```diff
@@ -109,20 +158,41 @@ class TelemetryModal extends React.PureComponent {
...
-                        <button
-                            className={styles.optOut}
-                            title={this.props.intl.formatMessage(messages.noTooltip)}
-                            onClick={this.handleOptOut}
-                        >
-                            <FormattedMessage {...messages.noButton} />
-                        </button>
+                        <span
+                            className={styles.settingWasUpdated}
+                            key={this.state.settingWasUpdatedTimer} // restart CSS fade when timer changes
```
cwillisf, line 188: This sets the `key` to the integer returned by `setTimeout` which will be different each time `settingWasUpdated` gets called. Changing the `key` means React will create a brand new element and discard the old one (if it exists), which is the only way I found to reliably reset the CSS fade animation. Creating the new element (and the resulting reflow) could be expensive if we do it a lot, but this `key` changes at most once per user click so I think it's OK.

rschamp (approved):

cwillisf:
Sorry for the noise. I was trying to make it so I could apply this same branch to both `develop` and `scratch-desktop` branches but ran into chromedriver problems in the integration test, so I'll do it the less lazy way ;)

PR #6648, opened 2021-01-26
Revert "set DETECT_CHROMEDRIVER_VERSION for integration tests"

cwillisf:
Reverts LLK/scratch-gui#6647

Never mind: I got confused about which branch I was working with and it turns out that this doesn't fix the problem.


rschamp:
Try it out: [revert-6647-detect-chromedriver-version](https://llk.github.io/scratch-gui/revert-6647-detect-chromedriver-version)

_I'm a bot!_

PR #6649, opened 2021-01-26
About menu (develop)

cwillisf:
This is the same as #6463 but targeting the develop branch.

Since it's not used by the web version of the editor and only activates if a special React property is set, this code is harmless in develop. Merging it into develop will reduce the size of the merge when we someday merge the scratch-desktop branch into develop, and may reduce the likelihood or magnitude of merge conflicts.


rschamp:
Try it out: [about-menu-develop](https://llk.github.io/scratch-gui/about-menu-develop)

_I'm a bot!_

PR #6722, opened 2021-02-10
Update to eslint-config-scratch@6.0.0 and associated fixes

paulkaplan:
This PR updates to eslint-config-scratch@6.0.0 which includes some new lint rules, so I fixed the failing lint errors, namely:

### Must reject promise with Error

This is a good rule @cwillisf I'm glad you added it, it always felt strange that you might `catch` an arbitrary thing)

### No duplicate imports

Overall good, but for this I added a sub-eslintrc to disable this rule in the `lib/libraries` folder because duplicate imports there are used to maintain the parallel structure of the libraries. I thought the code clarity benefitted from the duplication in that situation, but am open to comments)

### Spread instead of arguments
I agree it is more explicit and avoids some common traps, although the existing usage of arguments didn't have the usual problems like treating `arguments` like an array

### Don't return await
There was only one instance of this
```js
    afterEach(async () => await (driver && driver.quit()));	
```
Which actually got me stuck in a loop, since if I took out the `await` then it conflicts with `no async without await`. I definitely need to brush up on how async/await works. If i remove both async and await from that line, is it still the same? I guess because it ultimately would return driver.quit, which is a promise? I sidestepped the issue by changing it to `await driver.quit()` but **not** return it. Since the resolved value of driver.quit is never used, I think this is functionally equivalent (and is actually the way other parts of the code work, like in `integration/stage-size.test.js` for instance). All to say, I gotta brush up on async/await! Maybe @cwillisf you could give a short brown bag on the topic since you noted some of the subtleties in your eslint PR that introduced those rules?

---
Also note the peer dependency on eslint for eslint-config-scratch@6 is only 5.3, but upcoming updates will require 5.7, so bumped straight to that version instead.


src/containers/sound-editor.jsx
```diff
@@ -165,7 +165,7 @@ class SoundEditor extends React.Component {
             )
             .catch(e => {
                 // Encoding failed, or the sound was too large to save so edit is rejected
-                log.error(`Encountered error while trying to encode sound update: ${e}`);
+                log.error(`Encountered error while trying to encode sound update: ${e.message}`);
```
cwillisf, line 168: Wow, did lint catch this? If so, that's great!

cwillisf (approved):
Looks great! Thanks for doing this!

src/containers/sound-editor.jsx
paulkaplan: Alas no, that was me tracing where the rejections were going.

PR #6723, opened 2021-02-10
Mark deprecated lifecycle methods with UNSAFE_

paulkaplan:
NB this depends on https://github.com/LLK/scratch-gui/pull/6722, the only new commit here is 8e7c65e

As preparation for updating to react 17, mark deprecated lifecycle methods as `UNSAFE_` to reduce the console warnings. Once we update to react 17, we can migrate to newer lifecycle methods or otherwise remove the `UNSAFE_` methods.

Include the _correct_ camelcase rule in the eslintrc, since it is only temporary.


apple502j:
Notes on why componentWillReceiveProps is used:
- audio-selector, costume-tab, slider-monitor, sound-editor, sound-tab, : derived state. we can probably use `key` + "default value" props AKA "[Fully uncontrolled component with a key](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key)"
- drop-area-hoc, sortable-hoc: is this more complicated form of derived state? can't we componentDidUpdate?

Also, project-saver-hoc uses componentWillMount - but it could probably be moved to constructor.

paulkaplan:
Hmm also I misunderstood, the UNSAFE_ lifecycle methods were introduced in react 16.3, so we can't just mark these all unsafe (since we are on 16.2).

Since GUI and Paint declare anything in the 16 range as peer dependencies, maybe i can see if we can update www to the latest 16 version (16.14) first and then update lifecycle methods and then move to 17.

paulkaplan:
Closing this in favor of https://github.com/LLK/scratch-gui/pull/6775, but I'll transfer your comment there @apple502j

PR #6775, opened 2021-02-19
[blocked] Update react 16.14.0

paulkaplan:
This PR does 3 related things
1. **Update dev dependency for React to 16.14.0**: This only impacts the playgrounds, since the exported library has react as an external. This brings us far-enough up-to-date to allow for updating sub-dependencies and marking deprecated lifecycle methods, but doesn't go to 17 right now since there are other bugs with that.
2. **Update the peer dependency to 16.8.0**: Since this PR marks the deprecated lifecycle methods unsafe, it needs to have a higher peer dependency than 16.0.0, which doesn't have those methods in it. I chose 16.8.0 because, as the docs say, it is ["The One With Hooks"](https://reactjs.org/blog/2019/02/06/react-v16.8.0.html), so I wanted to future proof a bit.
3. **Mark deprecated lifecycle methods UNSAFE_** Primarily this suppresses a flood of console warnings, there is an associated eslintrc rule for this.

This cannot be merged until https://github.com/LLK/scratch-www/pull/5013 is in. Other repos using scratch-gui, like scratch-desktop, should also update before their next release. I'm going to mark this as "blocked" until I have a PR for scratch-desktop that at updates to a later version of react, just so that update doesn't get blindsided by this or allow some bugs to slip through. /cc @cwillisf 

Also closing https://github.com/LLK/scratch-gui/pull/6723 in favor of this PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3718,21 +3718,163 @@
       }
     },
     "airbnb-prop-types": {
-      "version": "2.15.0",
-      "resolved": "https://registry.npmjs.org/airbnb-prop-types/-/airbnb-prop-types-2.15.0.tgz",
-      "integrity": "sha512-jUh2/hfKsRjNFC4XONQrxo/n/3GG4Tn6Hl0WlFQN5PY9OMC9loSCoAYKnZsWaP8wEfd5xcrPloK0Zg6iS1xwVA==",
+      "version": "2.16.0",
+      "resolved": "https://registry.npmjs.org/airbnb-prop-types/-/airbnb-prop-types-2.16.0.tgz",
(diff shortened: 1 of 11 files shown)


paulkaplan:
From @apple502j 

> Notes on why componentWillReceiveProps is used:
> 
> * audio-selector, costume-tab, slider-monitor, sound-editor, sound-tab, : derived state. we can probably use `key` + "default value" props AKA "[Fully uncontrolled component with a key](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key)"
> * drop-area-hoc, sortable-hoc: is this more complicated form of derived state? can't we componentDidUpdate?
> 
> Also, project-saver-hoc uses componentWillMount - but it could probably be moved to constructor.

PR #6804, opened 2021-02-23
Benchmark deserialize

cwillisf:
### Proposed Changes

This change introduces a benchmark test for loading (deserializing) an SB2 file which contains 1000 stage-size vector costumes, along with a quick-and-dirty benchmark framework to host it and report results.

Depends on LLK/scratch-vm#2903

### Reason for Changes

This is indirectly in support of some SVG measurement work going on in the scratch-svg-renderer repository. My initial attempt hurt performance pretty badly, so I wanted to make sure that my second attempt didn't have a significant performance impact. The aggregate performance impact is on the SB2 deserialization process; see LLK/scratch-vm#2903 for details.

Measuring SVGs requires a DOM, so I added the benchmarking framework to this repository in order to use the Selenium setup that we already have in place.

Note that Jest overrides all console I/O for tests, so it's somewhat difficult to get results out of the tests. For this reason, I record the benchmark "samples" into an external file, then read & parse those samples later to report aggregate results.

I tested this code with some large projects in the community, but I didn't want to add those projects to our repository without permission from both the project authors and the authors of the assets used in those projects, so I generated a test project instead. I also included the script I used to generate that project.

### Test Coverage

Run `npm run test:benchmarks` to see the benchmark tests in action.


kchadha (commented):
Minor changes:
- Let's rename thousand-vector-costumes to many-vector-costumes
- Maybe `test:benchmarks` should be `test:benchmark` ¯\\\_(ツ)\_/¯  (same with the actual `benchmarks` directory name)

PR #6852, opened 2021-03-04
fix telemetry dialog receiving updates

### LINKED ISSUE #183: choice doesn't stick in telemetry dialog
### Expected Behavior

The radio buttons in the telemetry dialog should update when clicked

### Actual Behavior

Clicking the radio buttons doesn't seem to do anything

### Steps to Reproduce

1. Delete your `telemetry.json` file (or remove the "optIn" setting from it)
   * See instructions here: https://github.com/LLK/scratch-desktop#resetting-the-telemetry-system
2. Start Scratch Desktop version 3.20.0
3. Wait for the telemetry dialog to appear
4. Attempt to use the radio buttons

Affects both Mac and Windows builds

cwillisf:
### Resolves

Resolves LLK/scratch-desktop#183, which is a release-blocking issue for the desktop app but does not affect other versions of Scratch.

### Proposed Changes

Remove `isTelemetryEnabled` from the list of properties explicitly NOT passed from `AppStateWrapper` to its wrapped component.

This PR is targeted at `develop` and is just for keeping the branches from diverging further, so it's not urgent. I will go ahead and merge this commit in the `scratch-desktop` branch so that it unblocks the release.

### Reason for Changes

The telemetry dialog relies on this property to determine the current state of the telemetry system. The radio buttons are designed to reflect the current state by inspecting this property. Since it wasn't getting passed through, the telemetry dialog thought the state was always `undefined` -- which corresponds to the user not yet having made a choice.

I'm not sure why this was working before. My gut feeling is that some of the recent React changes are to blame, but I can't be certain.


chrisgarrity (approved):
Looks good - we pair reviewed.

Issue #6898, opened 2021-03-09
Replace the text-encoding package


adroitwhiz:
### Expected Behavior

Dependencies should be actively maintained and preferably not add 500KB to the bundle size

### Actual Behavior

The [`text-encoding` package](https://www.npmjs.com/package/text-encoding), used by `scratch-gui`, `scratch-sb1-converter`, and `scratch-vm`, is deprecated and unmaintained, and adds 500KB worth of encoding indices for text formats which Scratch does not use.

In https://github.com/LLK/scratch-storage/pull/114, I replaced it with [`fastestsmallesttextencoderdecoder`](https://www.npmjs.com/package/fastestsmallesttextencoderdecoder), but that change has not yet been made in other repositories.

PR #6899, opened 2021-03-09
ENA-118 Replace text-encoding polyfill with much smaller library, saving 500KB

adroitwhiz:
### Resolves

- Towards #6898

### Proposed Changes

This PR replaces the `text-encoding` package with `fastestsmallesttextencoderdecoder`.

### Reason for Changes

The `text-encoding` package has been deprecated by its maintainer, and also adds 500KB of encoding data for non-UTF-8 text. Because we're only using it to encode UTF-8, that 500KB is entirely unnecessary.

### Test Coverage

Must be tested manually.

Here, the `TextEncoder` API is used to encode assets for the default project. To test this, create a new default project in Legacy Edge, the only browser without native `TextEncoder` support.

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge

Linux
 * [x] Firefox
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


adroitwhiz:
Test failures seem to be the result of a ChromeDriver version mismatch

benjiwheeler:
To be super clear, this way of providing encoding functions is _only_ needed on legacy Edge, right? Does that include basically all known pre-chromium versions of Edge?

benjiwheeler:
Added "accessibility" label because reducing download size (for both app and website) is meaningful for improving access when data is slow or expensive

apple502j:
@benjiwheeler Yep, this library is used by all Legacy (non-Chromium) Edge - and by ONLY the Legacy Edge (at least for supported environments).

adroitwhiz:
Just a reminder that in order to actually remove this huge dependency from the *bundle*, you'll need to remove it from [all the other packages too](https://github.com/LLK/scratch-gui/issues/6898).

cwillisf (approved):

adroitwhiz:
@cwillisf Since this PR was first opened, I came across https://www.npmjs.com/package/fast-text-encoding, which seems to be far more popular and faster. I can create another PR for that if you want.

You could also consider dropping the polyfill entirely--the only browser which does not support `TextEncoder` and `TextDecoder` is Edge Legacy, which [has reached EOS and doesn't even seem to be installable anymore](https://learn.microsoft.com/en-us/lifecycle/announcements/edge-legacy-eos-details).

cwillisf:
Yeah - I think removing it entirely is the right answer. Honestly, I merged this because it's relatively self-contained, and wanted to send a signal that I'm trying to make at least some small amount of progress :sweat_smile:

PR #7149, opened 2021-05-19
shrink telemetry modal on smaller viewports

### LINKED ISSUE #6897: telemetry dialog cut off on tablets with wide aspect ratio
### Expected Behavior

When the telemetry dialog is displayed, it should be possible to see all the modal's content. In particular, it should be possible to:
* read all the text
* choose either option
* press the "Close" button

Example:

image

### Actual Behavior

On tablets with an unusually wide aspect ratio, such as the SM-T510, the bottom of the dialog, including the "Close" button, is not visible and there's no way to move or scroll the dialog. The same thing can happen in a windowed environment (Mac, Windows, ChromeOS) if the user resizes the window, which is a handy way to reproduce this issue.

On the SM-T510, for reference, the screen is 1920x1200 with a 36-pixel status bar at the top and a 72-pixel navigation bar at the bottom, for an effective screen size of 1920x1092, or 960x546 at double DPI.

Example from SM-T510:

Screenshot_20210309-144855_Scratch

### Steps to Reproduce

Two options:

1. Use a Samsung SM-T510 or similar
2. Clear all data for the Scratch app, or uninstall & reinstall it, to ensure that the telemetry modal appears
3. Launch the Scratch app
4. Create a new project
5. Wait for the editor to load

-or-

1. Use the Scratch app on Mac, Windows, or ChromeOS
2. Start the app and open the telemetry dialog
   * On Mac or Windows, use the "Data Settings" option in the "?" menu
   * On ChromeOS, clear the app data or uninstall & reinstall it
3. Resize the window, shrinking its height to about 600px or below

This issue affects all forms of the desktop or Android app, but it's most important in the Android app where the user cannot resize the window.

cwillisf:
### Resolves

Resolves #6897 

### Proposed Changes

Add two media queries to the telemetry modal's CSS:

- If the viewport height is smaller than 660, shrink the vertical margin around the telemetry modal.
- If the viewport height is smaller than 540, also remove the image banner at the top of the telemetry modal.

### Reason for Changes

Prior to these changes, a viewport height below about 660 causes the bottom of the telemetry dialog to be clipped, and below about 600 the "Close" button starts to disappear. Some tablets have even lower viewport height - for example, the SM-T510's effective viewport height is 600 minus the height of any navigation and status bars, typically resulting in the "Close" button being pushed completely off the screen. Making these changes means that the telemetry modal stays usable even on devices with such low screen heights.

These screenshots demonstrate the effect of this change on an SM-T510.

Before:
image

After:
image

### Test Coverage

Tested locally in the desktop app and Android app. The telemetry modal is not displayed by the web version of Scratch.

### DIFF
--- a/src/components/telemetry-modal/telemetry-modal.css
+++ b/src/components/telemetry-modal/telemetry-modal.css
@@ -148,3 +148,16 @@ $ui-gray: hsla(0, 0%, 95%, 1);
     box-shadow: none;
     opacity: 0.25;
 }
+
+@media screen and (max-height: 660px) {
+    .modal-content {
+        margin: 5vh auto;
+        width: 90%;
+    }
+}
+
+@media screen and (max-height: 540px) {
+    .illustration {
+        display: none;
+    }
+}


kchadha (approved):
LGTM!

PR #7565, opened 2021-08-25
fix BOOST & CDM combo

cwillisf:
### Proposed Changes

Add support for the `USER_PICKED_PERIPHERAL` event in `AutoScanningStep`, similar to #6273.

### Reason for Changes

When using the Android CDM, the Android system handles peripheral selection. When we find out which one the user selected, our CDM support code fires the `USER_PICKED_PERIPHERAL` event to simulate the UI event that would be generated by the user picking a peripheral in our regular, non-CDM UI. The `USER_PICKED_PERIPHERAL` event was being handled correctly in [`scanning-step.jsx`](https://github.com/LLK/scratch-gui/blob/native/src/containers/scanning-step.jsx) (used by micro:bit and others) but it was not handled here in `auto-scanning-step.jsx` at all (used by LEGO BOOST and others).

### Test Coverage

We don't currently have a way to simulate Android CDM events, so this was tested manually.

### Browser Coverage

Tested with LEGO BOOST on a Samsung Chromebook Pro running Android 9.0.

### DIFF
--- a/src/containers/auto-scanning-step.jsx
+++ b/src/containers/auto-scanning-step.jsx
@@ -10,7 +10,6 @@ class AutoScanningStep extends React.Component {
         bindAll(this, [
             'handlePeripheralListUpdate',
             'handlePeripheralScanTimeout',
-            'handleUserPickedPeripheral',
             'handleStartScan',
             'handleRefresh'
         ]);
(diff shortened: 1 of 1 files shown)


src/containers/auto-scanning-step.jsx
```diff
@@ -10,6 +10,7 @@ class AutoScanningStep extends React.Component {
         bindAll(this, [
             'handlePeripheralListUpdate',
             'handlePeripheralScanTimeout',
+            'handleUserPickedPeripheral',
```
chrisgarrity: Is this really needed? There isn't a `handleUserPickedPeripheral` method in the container for this to be bound to.

src/containers/auto-scanning-step.jsx
```diff
@@ -39,12 +40,16 @@ class AutoScanningStep extends React.Component {
     bindPeripheralUpdates () {
         this.props.vm.on(
             'PERIPHERAL_LIST_UPDATE', this.handlePeripheralListUpdate);
+        this.props.vm.on(
```
chrisgarrity, line 42: This does look weird in the "auto"scanning step. It might be worth a comment that in some configurations, (e.g. CDM on Android), the selection always appears to have been user selected.

chrisgarrity (approved):
I can't test, so I think it's worth building a version for @BryceLTaylor to try on his BOOST. 

A couple of questions, but overall looks good.

src/containers/auto-scanning-step.jsx
cwillisf: Good catch! This is a remnant from an earlier approach. Thanks!

src/containers/auto-scanning-step.jsx
cwillisf: I added a comment explaining the situation, but that prompted me to think about it some more and I'd like to further refine this in a future change. For now, this gets the extension working at least.

The further refinement I'd like to make is that we should treat the `USER_PICKED_PERIPHERAL` message as the "final answer" in both `auto-scanning-step.jsx` _and_ `scanning-step.jsx`, even in the cancel case. That is: in both cases, if the user picks a peripheral we'll get a list of one item and we should immediately connect. As of this PR, that's done in both files. However, if the user cancels then we'll get a list of zero items and we should immediately stop trying to connect, skipping the timeout -- that's not the case in either file, so I'd like to do that as a follow-up change.

There's another potential can of worms in the scan timeout, which might not make sense at all in the CDM world. I'll have to check more thoroughly what the CDM's behavior is when the user doesn't pick anything for a long time, and when the CDM doesn't find any devices for a long time. If those both give us a callback, we might not want our own scan timeout at all. Or, we might want a timeout that's guaranteed to be a little longer than the CDM's timeout.

PR #7745, opened 2021-10-22
CDM connection flow updates

cwillisf:
### Resolves

Progress on LLK/scratch-android#340

### Proposed Changes

Bonus change:

* Add comments explaining the difference between `ScanningStep` and `AutoScanningStep`
   * `AutoScanningStep` automatically chooses the first device discovered by a scan. It also supports a `prescan` phase, which displays a message to the user and waits for them to press a button, and `pressbutton` phase, which displays a message during the scan while waiting for the first scan result
   * `ScanningStep` allows the user to choose a peripheral device from a list managed by `scratch-gui` and does not support the `prescan` or `pressbutton` phases.

Real changes:

* Add a new flag `useExternalPeripheralList` which tells the extension connection flow to assume that peripheral selection will happen through an externally-managed UI, as with the Android Companion Device Manager (CDM)
* When `useExternalPeripheralList` is true, force all extensions to use `AutoScanningStep`
* Allow extensions to override the messages shown during the `prescan` and `pushbutton` phases of the connection flow
   * The message for the `prescan` phase is called `prescanMessage`.
   * The message for the `pushbutton` phase is called `scanBeginMessage` since for some devices it doesn't involve pushing any button.
* Add `prescanMessage` and `scanBeginMessage` overrides for all hardware extensions
* On the `prescan` phase, replace the big, static "radar" image with an extension-specific image (generally an illustration of the peripheral device)

### Reason for Changes

When running with the Android CDM, the CDM's list of scan results displays over the top of our GUI components for the connection flow. This makes it impossible for the user to see any help we might be giving them in the connection flow UI. Adding the `prescan` phase, and waiting to activate the CDM until the user presses the button, makes sure we can provide information and context for the list before it pops up.

Using the `AutoScanningStep` lets us have the `prescan` phase, but also means we get the `pressbutton` phase. Prior to these changes, the `pressbutton` phase would display a message asking the user to "Press the button on your device" -- a message that only fits some of our extensions. Adding an override lets extensions customize that text for each peripheral device's needs and features.

In theory the WeDo 2.0 and BOOST extensions don't need overrides since the "Press the button on your device" message works for them, but it's nice to have more detail. The generic messages are still available as fallbacks.

The icon change is a nice-to-have which helps the user be sure of which extension is trying to connect right now.

### Test Coverage

Tested on a Samsung Chromebook Pro

### DIFF
--- a/src/components/connection-modal/auto-scanning-step.jsx
+++ b/src/components/connection-modal/auto-scanning-step.jsx
@@ -21,22 +21,41 @@ const PHASES = keyMirror({
     notfound: null
 });
 
+const defaultPrescanMessage = (<FormattedMessage
+    defaultMessage="Have your device nearby, then begin searching."
+    description="Prompt for beginning the search"
+    id="gui.connection.auto-scanning.prescan"
+/>);
+
+const defaultScanBeginMessage = (<FormattedMessage
+    defaultMessage="Press the button on your device."
+    description="Prompt for pushing the button on the device"
+    id="gui.connection.auto-scanning.scanBeginMessage"
+/>);
+
 const AutoScanningStep = props => (
     <Box className={styles.body}>
         <Box className={styles.activityArea}>
             <div className={styles.activityAreaInfo}>
                 <div className={styles.centeredRow}>
                     {props.phase === PHASES.prescan && (
-                        <React.Fragment>
+                        props.connectionIconURL ? (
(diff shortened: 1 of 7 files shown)


chrisgarrity (approved):
I did a careful read through, and given our conversations this all looks good. Subject to testing of course

PR #7768, opened 2021-10-25
Added ASL tutorial

zoebentley:
Changes: 
- adds thumbnail file for Getting Started ASL to "thumbnails" folder
- adds wistia link to translate-video.JS
- links the thumbnail file in "index.jsx"
- adds the Getting Started ASL tutorial to the tutorials in "index.jsx"
- makes it so that the original Getting Started video points to "Getting Started ASL" as one if its 2 follow-up tutorial suggestions


src/lib/libraries/decks/index.jsx
```diff
@@ -105,14 +108,36 @@ export default {
             image: 'introGreenFlag'
         }, {
             deckIds: [
-                'add-a-backdrop',
+                'getting-started-ASL',
```
seotts:
```suggestion
                'intro-getting-started-ASL',
```

seotts: If I understand how this file works, I think it should be this.

src/lib/libraries/decks/index.jsx
```diff
@@ -105,14 +108,36 @@ export default {
...
         }
         ],
         urlId: 'getStarted'
     },
 
+    'intro-getting-started-ASL': {
+        name: (
+            <FormattedMessage
+                defaultMessage="Getting Started - ASL"
+                description="Name for the 'Getting Started ASL' how-to"
```
seotts: I wonder if it would be helpful for the translators to have ASL written out as "American Sign Language" in the description text?

seotts (changes requested):
This is awesome. I found one small (but consequential) bug, and also made one other minor suggestion.

src/lib/libraries/decks/index.jsx
zoebentley: Sounds great

src/lib/libraries/decks/index.jsx
zoebentley: Oh - thanks so much for the catch!

PR #7961, opened 2022-01-13
Circle add workflow for PRs that don't cache

BryceLTaylor:
### Resolves

Using too much storage in Circle CI

Adds a workflow that runs a single job to run tests and build.  This doesn't use any caches.  By running this workflow on PRs we should reduce the overall storage.

Similar to https://github.com/LLK/scratch-www/pull/6376 on scratch-www

### Reason for Changes

To reduce the amount of storage used in CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ aliases:
         - v3-npm-
   - &defaults
     docker:
-      - image: cimg/node:10-browsers
+      - image: cimg/node:16.13-browsers
     working_directory: ~/repo
 
 jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ jobs:
       - run:
           name: Integration
           environment:
-              JEST_JUNIT_OUTPUT_NAME: integration-results
+              JEST_JUNIT_OUTPUT_NAME: integration-results.xml
           command: |
             export TESTFILES=$(circleci tests glob "test/integration/*.test.js" | circleci tests split --split-by=timings)
             $(npm bin)/jest ${TESTFILES} --reporters="default" --reporters="jest-junit" --runInBand
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,9 +70,9 @@ jobs:
       NODE_ENV: production
       NODE_OPTIONS: --max-old-space-size=4000
     steps:
-      <<: *setup_steps
-      <<: *test_steps
-      <<: *build_steps
+    <<: *setup_steps
+    <<: *test_steps
+    <<: *build_steps
   setup:
     <<: *defaults
     steps:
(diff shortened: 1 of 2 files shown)


BryceLTaylor:
The build with no cache ran in about 7.5 minutes which is faster than the 8.5-12 minutes it took when it ran in separate jobs

.circleci/config.yml
```diff
@@ -44,6 +44,39 @@ aliases:
...
+          command: npm run test:unit -- --reporters="default" --reporters="jest-junit" --coverage --coverageReporters=text --coverageReporters=lcov --maxWorkers="2"
+      - run:
+          name: Build
+          environment:
+            NODE_ENV: production
+          command: npm run build
+      - run:
+          name: Integration
+          environment:
+              JEST_JUNIT_OUTPUT_NAME: integration-results
```
rschamp: These results aren't listed as being uploaded in the job, I think because it's missing the `.xml` piece.
```suggestion
              JEST_JUNIT_OUTPUT_NAME: integration-results.xml
```

.circleci/config.yml
```diff
@@ -44,6 +44,39 @@ aliases:
...
+          command: |
+            export TESTFILES=$(circleci tests glob "test/integration/*.test.js" | circleci tests split --split-by=timings)
+            $(npm bin)/jest ${TESTFILES} --reporters="default" --reporters="jest-junit" --runInBand
```
rschamp: Since we can't parallelize the single job, this can be simplified
```suggestion
            $(npm bin)/jest test/integration/*.test.js --reporters="default" --reporters="jest-junit" --runInBand
```

rschamp (changes requested):
Thanks for working though this! If the uncached version is faster overall, maybe we should simplify the rest of the config to use the single job too?  Just a couple of things I noticed with the integration tests that could be better.

.circleci/config.yml
```diff
@@ -44,6 +44,38 @@ aliases:
...
+          name: Build
+          environment:
+            NODE_ENV: production
+          command: npm run build
+      - run:
+          name: Integration
+          environment:
+              JEST_JUNIT_OUTPUT_NAME: integration-results.xml
+          command: |
+            $(npm bin)/jest test/integration/*.test.js --reporters="default" --reporters="jest-junit" --runInBand
```
rschamp:
```suggestion
          command: npm run test:integration -- --reporters="default" --reporters="jest-junit"
```

rschamp (approved):
This is most definitely a nit but might be worth including since it makes the integration step more similar to the rest, now that we're not trying to parallelize it.  Don't think it's necessary to include but might be nice.

.circleci/config.yml
BryceLTaylor: That makes sense.  It's probably worth it to do now.

PR #7999, opened 2022-02-18
ENA-119 Place z-index direction picker above other stage elements, so it displays correctly

### LINKED ISSUE #7467: Direction gets hidden by list
Whilst exploring on Scratch, I discovered that the list appears on top of the direction selector

https://user-images.githubusercontent.com/66112471/128066652-5fb83ee4-14ca-40ac-a043-50c3b1c1f913.mp4

matschaffer:
### Resolves

- Resolves #7467

### Proposed Changes

Adds a z-index for the direction picker above other stage element z-indices.

### Reason for Changes

The direction picker isn't part of the stage so it's unexpected to have stage elements z-indexed behind it.

### Test Coverage

I suspect z-indices aren't tested, but let me know if they are and I can add something.

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/direction-picker/direction-picker.css
+++ b/src/components/direction-picker/direction-picker.css
@@ -1,4 +1,9 @@
 @import "../../css/colors.css";
+@import "../../css/z-index.css";
+
+.popover {
+    z-index: $z-index-direction-picker;
+}
 
 .button-row {
     display: flex;
--- a/src/components/direction-picker/direction-picker.jsx
+++ b/src/components/direction-picker/direction-picker.jsx
@@ -56,6 +56,7 @@ const DirectionPicker = props => (
         text={directionLabel}
     >
         <Popover
+            className={styles.popover}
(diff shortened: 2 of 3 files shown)


chrisgarrity:
@matschaffer while I don't think we have examples of z-index tests, it would be great to add something.

matschaffer:
It would, though I think we might need https://github.com/LLK/scratch-gui/issues/8034 for that.

I can’t run the integration tests locally since our web driver version won’t use latest chrome.

matschaffer:
I'm sure this is pretty stale at this point, so closing down.

PR #8108, opened 2022-03-15
run integration tests in parallel

BryceLTaylor:
Run 4 tests at a time. 

### Proposed Changes

Replace --runInBand with --maxWorkers=4
This will run 4 test suites at a time rather than 1.
Running more at a time has historically caused errors, but using limited parallelization like this worked in WWW

### Reason for Changes

Speed up the tests 

### Test Coverage

I ran the tests locally and they dramatically sped up without causing failures.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "i18n:src": "rimraf ./translations/messages/src && babel src > tmp.js && rimraf tmp.js && build-i18n-src ./translations/messages/src ./translations/ && npm run i18n:push",
     "start": "webpack-dev-server",
     "test": "npm run test:lint && npm run test:unit && npm run build && npm run test:integration",
-    "test:integration": "jest --runInBand test[\\\\/]integration",
+    "test:integration": "jest --maxWorkers=4 test[\\\\/]integration",
     "test:lint": "eslint . --ext .js,.jsx",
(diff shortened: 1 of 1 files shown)


matschaffer:
On https://jestjs.io/docs/cli#--maxworkersnumstring I see this:

> this defaults to the number of the cores available on your machine minus one for the main thread.

Have you tried just removing `--runInBand` ? You might see a similar speed up.

Either way unofficial ✅  from me on running in parallel.

matschaffer:
I checked https://app.circleci.com/pipelines/github/LLK/scratch-gui/26898/workflows/a15a95ac-aabb-4a09-a237-256afbd5eea6/jobs/154260 against https://app.circleci.com/pipelines/github/LLK/scratch-gui/26874/workflows/7cf3bb5d-eb8f-4612-9dc6-95fe08cb3f6a/jobs/154200 - looks like the integration time got cut roughly in half which is awesome :)

kchadha (approved):
LGTM Thanks!

BryceLTaylor:
@matschaffer Though it looks like the defaults for running without --runInBand should just work it doesn't really.  Tests will fail.  I think there is still some competition for resources from the threads.  Running locally I first tried --maxWorkers=5 and I still had some test failures, so I dropped it to 4 and it worked great.

Issue #7880, opened 2021-11-29
CircleCI Docker image will be deprecated Dec 31


rschamp:
CircleCI is deprecating the Docker image this project is using on Dec 31: https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034

We need to replace it with an equivalent [`cimg/`-style image, like scratch-www uses](https://github.com/LLK/scratch-www/blob/e0f144bc8365a2381f040de7e3aa791441110986/.circleci/config.yml#L6).  We can probably just use exactly this image.

PR #8146, opened 2022-03-22
Update circle docker image

BryceLTaylor:
### Proposed Changes

Uses the new circleci docker images "cimg/..."
Updates to use node 12
Implements the browser-tools orb to install chrome and matching chromedriver
Updates chromedriver that will be installed locally to version 99
In circle we will print the version of chrome and chromderiver used when running the tests.

### Reason for Changes

This uses the new and more likely to be updated docker images in circle
It will make sure we will always have a matching chrome and chromedriver

### Test Coverage

This does not change any of the code, but failures in the build process should get caught by automated tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,6 @@
 version: 2.1
+orbs:
+    browser-tools: circleci/browser-tools@1.2.4
 aliases:
   - &save_git_cache
     save_cache:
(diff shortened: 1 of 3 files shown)


kchadha (approved):
LGTM! Thanks for making all these updates!

benjiwheeler:
This resolves https://github.com/LLK/scratch-gui/issues/7880

PR #8182, opened 2022-03-29
CircleCI use dockerhub-credentials context in each job

BryceLTaylor:
### Resolves

There is a warning on all jobs that we are not supplying dockerhub credentials

### Proposed Changes

Uses our dockerhub credentials supplied by a Circle context in all jobs

### Reason for Changes

This should remove the warning in each job.  It has been implied that one day we might not be able to use CircleCIs dockerhub credentials.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,6 +43,9 @@ aliases:
   - &defaults
     docker:
       - image: cimg/node:12.22.11-browsers
+        auth:
+          username: $DOCKERHUB_USERNAME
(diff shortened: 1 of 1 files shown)


cwillisf (approved):
Wow... it sure seems like there should be a way to specify a context at a higher level, but I looked at the docs just now and didn't see any way to do that. It's a shame that it has to be repeated in so many places.

BryceLTaylor:
@cwillisf yeah, I tried setting it at a workflow level and you just can't.  It also doesn't work to set it in `defaults`

PR #8234, opened 2022-04-19
move [skip ci] in deploy script

BryceLTaylor:
### Resolves

we still run a followup build after deploying to gh-pages even though we attempt to instruct it to [skip ci] in the commit message

### Proposed Changes

Moves [skip ci] to the beginning of the commit message.  This worked in paint.

### Reason for Changes

After deploying we commit to the gh-pages branch, which kicks off another build.  We would like to not kick off another build


cwillisf (approved):
Oops - lost track of this one. Looks good!

PR #8258, opened 2022-04-28
Update gh-pages to use published package

rschamp:
The fork we were using is now integrated into the package published to NPM.

### Resolves

Doesn't resolve any issue, but removes dependency between this repo and the @rschamp account.

### Proposed Changes

Updates the gh-pages dependency from a github dependency to the up to date NPM dependency for that package.

### Reason for Changes

We weren't getting any updates for gh-pages while it was sitting on my fork. And now the reason for using the fork is no longer true: The published package has the `-e` option.

### Test Coverage

I pushed this branch to LLK, so we can see if it pushes to gh-pages after building

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4389,12 +4389,12 @@
       "dev": true
     },
     "async": {
-      "version": "2.1.2",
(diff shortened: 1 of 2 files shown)


BryceLTaylor (approved):
Looks good

PR #8269, opened 2022-05-05
adds fastly token to project call

tomlum:
### Resolves

Adds Fastly Token to project json calls

Depends on this www PR https://github.com/LLK/scratch-www/pull/6773

### DIFF
--- a/src/lib/storage.js
+++ b/src/lib/storage.js
@@ -35,8 +35,11 @@ class Storage extends ScratchStorage {
     setProjectHost (projectHost) {
         this.projectHost = projectHost;
     }
+    setProjectToken (projectToken) {
+        this.projectToken = projectToken;
+    }
     getProjectGetConfig (projectAsset) {
-        return `${this.projectHost}/${projectAsset.assetId}`;
+        return `${this.projectHost}/${projectAsset.assetId}?token=${this.projectToken}`;
     }
     getProjectCreateConfig () {
         return {
@@ -46,7 +49,7 @@ class Storage extends ScratchStorage {
     }
     getProjectUpdateConfig (projectAsset) {
         return {
-            url: `${this.projectHost}/${projectAsset.assetId}`,
+            url: `${this.projectHost}/${projectAsset.assetId}?token=${this.projectToken}`,
             withCredentials: true
         };
     }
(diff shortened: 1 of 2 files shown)


src/lib/storage.js
```diff
@@ -35,8 +35,11 @@ class Storage extends ScratchStorage {
     setProjectHost (projectHost) {
         this.projectHost = projectHost;
     }
+    setProjectToken (projectToken) {
+        this.projectToken = projectToken;
+    }
     getProjectGetConfig (projectAsset) {
-        return `${this.projectHost}/${projectAsset.assetId}`;
+        return `${this.projectHost}/${projectAsset.assetId}?token=${this.projectToken}`;
```
rschamp: Not all environments will need the token, so may not supply it. So maybe this would be better:
```suggestion
        const path = `${this.projectHost}/${projectAsset.assetId}`;
        const qs = this.projectToken ? `?token=${this.projectToken}` : '';
        return path + qs;
```

src/lib/storage.js
```diff
@@ -46,7 +49,7 @@ class Storage extends ScratchStorage {
     }
     getProjectUpdateConfig (projectAsset) {
         return {
-            url: `${this.projectHost}/${projectAsset.assetId}`,
+            url: `${this.projectHost}/${projectAsset.assetId}?token=${this.projectToken}`,
```
rschamp: I don't think this is necessary, since the token will only be required on fetches.
```suggestion
            url: `${this.projectHost}/${projectAsset.assetId}`,
```

rschamp (changes requested):

rschamp (approved):

Scratchfangs:
Oh no

Scratchfangs:
Please dont do this

GoldcatGames:
Positives: Can Secure Projects Better For Less Data Leaks, Also Can Improve IP Blocks
Negatives: Stops Teams/Groups Of Scratchers Doing Collabs And Giant Projects Filled With Hard Work, Makes Performance A Bit Slower.

### **Just My PERSONAL Positives And Negatives, I Hope You Take Them Into Consideration.**

PR #8322, opened 2022-05-19
Update to storage v2.0.1 & handle storage.load possibly returning null

kchadha:
### Proposed Changes

- Update to storage v2.0.1 (LLK/scratch-storage#388 and LLK/scratch-storage#389) which includes changes for handling missing assets (storage now resolves null if the asset is missing and also no longer swallows errors during attempts to fetch the asset).

- Update sound library to handle storage resolving null for missing sounds. Don't attempt to play the sound if this happens.

### Test Coverage

Tested locally in the browser, and tested error and non-error path in the sound library.


cwillisf (approved):
👍

PR #8453, opened 2022-08-11
CE-137 adds dompurify to file upload

tomlum:
### Resolves

Issue where SVGs with external hrefs could be uploaded.

### Proposed Changes

Intercepts file uploads with sanitizing function from https://github.com/LLK/scratch-svg-renderer/pull/336

### DIFF
--- a/src/lib/file-uploader.js
+++ b/src/lib/file-uploader.js
@@ -102,7 +102,7 @@ const costumeUpload = function (fileData, fileType, storage, handleCostume, hand
     let assetType = null;
     switch (fileType) {
     case 'image/svg+xml': {
-
+        // run svg bytes through scratch-svg-renderer's sanitization code
         fileData = sanitizeSvg.sanitizeByteStream(fileData);
 
         costumeFormat = storage.DataFormat.SVG;
--- a/src/lib/file-uploader.js
+++ b/src/lib/file-uploader.js
@@ -2,6 +2,31 @@ import {BitmapAdapter} from 'scratch-svg-renderer';
 import randomizeSpritePosition from './randomize-sprite-position.js';
 import bmpConverter from './bmp-converter';
 import gifDecoder from './gif-decoder';
+import DOMPurify from 'dompurify';
+
+let _TextDecoder;
+let _TextEncoder;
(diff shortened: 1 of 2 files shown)


apple502j:
Question, is this performed server-side as well? A little bit silly if we can directly send the asset to the asset server so that we can bypass the check. (The `sanitizeSvg` function to be introduced in SVG renderer should be enough, as all assets go through it.)

src/lib/file-uploader.js
```diff
@@ -2,6 +2,31 @@ import {BitmapAdapter} from 'scratch-svg-renderer';
...
 import bmpConverter from './bmp-converter';
 import gifDecoder from './gif-decoder';
+import DOMPurify from 'dompurify';
+
+let _TextDecoder;
+let _TextEncoder;
+if (typeof TextDecoder === 'undefined' || typeof TextEncoder === 'undefined') {
+    // Wait to require the text encoding polyfill until we know it's needed.
+    // eslint-disable-next-line global-require
+    const encoding = require('fastestsmallesttextencoderdecoder');
```
apple502j: This needs to be added to package.json. (That said, the only supported browser which requires this code path is Old Edge - which is discontinued. Therefore it might make sense to just use the global.) GUI currently uses `text-encoding`, but I believe that library is unnecessarily big and should be replaced with this one.

benjiwheeler: Thanks for pointing this out, @apple502j. We're still using `fastestsmallesttextencoderdecoder` but we've moved the bulk of this code to scratch-svg-renderer https://github.com/LLK/scratch-svg-renderer/pull/336

benjiwheeler:
> Question, is this performed server-side as well? A little bit silly if we can directly send the asset to the asset server so that we can bypass the check. (The `sanitizeSvg` function to be introduced in SVG renderer should be enough, as all assets go through it.)

I don't want to get into too much detail about what we're doing server-side, but you make a very good point that we need to make sure we are doing similar sanitization there.

cwillisf (approved):
Assuming the tests pass after merging LLK/scratch-svg-renderer#336, this looks good to me!

PR #8634, opened 2022-11-04
feat: Disable cloud vars when video sensing

aoneill01:
This is paired with LLK/scratch-www#7165.

### Resolves

Resolves LLK/scratch-ops#565

### Proposed Changes

Disable cloud variables when camera is active. 

### Reason for Changes

The logic is in `shouldConnect` rather than `canUseCloud` because we don't want to change the editor's experience. This way they still have the option to create cloud variables in the editor.

### Test Coverage

Added test to cloud-manager-hoc to cover this scenario.

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/lib/cloud-manager-hoc.jsx
+++ b/src/lib/cloud-manager-hoc.jsx
@@ -57,7 +57,7 @@ const cloudManagerHOC = function (WrappedComponent) {
         shouldConnect (props) {
             return !this.isConnected() && this.canUseCloud(props) &&
                 props.isShowingWithId && props.vm.runtime.hasCloudData() &&
-                props.canModifyCloudData;
+                props.canModifyCloudData && !props.vm.extensionManager.isExtensionLoaded('videoSensing');
         }
         shouldDisconnect (props, prevProps) {
             return this.isConnected() &&
--- a/test/unit/util/cloud-manager-hoc.test.jsx
+++ b/test/unit/util/cloud-manager-hoc.test.jsx
@@ -50,6 +50,9 @@ describe('CloudManagerHOC', () => {
         vm.runtime = {
             hasCloudData: jest.fn(() => true)
         };
+        vm.extensionManager = {
+            isExtensionLoaded: jest.fn(() => false)
+        };
         CloudProvider.mockClear();
         mockCloudProviderInstance.requestCloseConnection.mockClear();
     });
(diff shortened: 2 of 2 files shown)


aoneill01:
Tests are failing because of a chromedriver version issue. Tests should pass locally with matching chrome+chromedriver. I don't want this to stop reviewing of the PR.

aoneill01:
Rebased onto develop to include some changes to fix the chromedriver version issue. However, there is still a known problem that is causing the first integration test to fail. That failure should be unrelated to this PR.

src/lib/cloud-manager-hoc.jsx
```diff
@@ -57,7 +57,7 @@ const cloudManagerHOC = function (WrappedComponent) {
         shouldConnect (props) {
             return !this.isConnected() && this.canUseCloud(props) &&
                 props.isShowingWithId && props.vm.runtime.hasCloudData() &&
-                props.canModifyCloudData;
+                props.canModifyCloudData && !props.vm.extensionManager.isExtensionLoaded('videoSensing');
```
benjiwheeler, line 60: My only question with this is, which logic belongs in `shouldConnect()` and which in `mapStateToProps()`, where `canModifyCloudData` is defined? Maybe that calls for a larger code change, but it seems like we've left the logic scattered, instead of either a) in one place, or b) separated into distinct groups of related concerns.

benjiwheeler (approved):
LGTM, with one comment -- something maybe to consider in future

src/lib/cloud-manager-hoc.jsx
benjiwheeler: One more thought -- let's test this and make sure it seems to stop updating cloud vars if the video sensing extension becomes loaded. Maybe we need a change in the shouldDisconnect() function too?

benjiwheeler (changes requested):
Sorry, on second thought, I might need to actually test this to see if adding video sensing after load causes cloud vars to stop?

benjiwheeler (approved):
LGTM again -- I think it’s a strict improvement over the status quo, so we should go ahead with it. But I think there’s still a risk with adding video to a project with cloud vars, even if we’re just talking about stopping someone from sharing their own personally identifiable information/image.

src/lib/cloud-manager-hoc.jsx
aoneill01: I didn't think about putting it in `mapStateToProps()`; that seems like a good idea. I think I will merge as-is so that this can move along, but I'll try to get that in a later update.

cwillisf:
Congrats on making your first changes in Scratch!
image

PR #8638, opened 2022-11-10
ci: use ⌘ or Ctrl based on whether platform is Mac

aoneill01:
### Resolves

This was blocking ENA-108 from being deployed.

### Proposed Changes

- Conditionally use command or control in the "Keyboard shortcuts" integration test based on whether the platform is Mac or not.

### Reason for Changes

One of the integration tests was using key combinations to select-all, undo, and redo. The test was failing in CircleCI with: 

```
ECONNREFUSED connect ECONNREFUSED 127.0.0.1:51720
```

The test started failing after `chromedriver` updated to version 107. 

### Test Coverage

*N/A*

### Browser Coverage

Not particularly relevant since this is an integration test being run with `chromedriver`. Local development was done on a Mac, and it is also being run in CircleCI.

### DIFF
--- a/test/integration/sounds.test.js
+++ b/test/integration/sounds.test.js
@@ -169,16 +169,18 @@ describe('Working with sounds', () => {
     });
 
     test('Keyboard shortcuts', async () => {
+        const cmdCtrl = process.platform.includes('darwin') ? Key.COMMAND : Key.CONTROL;
+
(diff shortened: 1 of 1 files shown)


cwillisf (approved):
Looks good to me, and I see that tests passed on CI.
Thanks for digging into this!

PR #8683, opened 2022-11-29
feat: disconnect from cloud variables immediately after adding video

aoneill01:
### Resolves

- Resolves [ENA-173](https://scratchfoundation.atlassian.net/browse/ENA-173)

### Proposed Changes

- listens for `EXTENSION_ADDED` event and disconnects from cloud variable server if it is the `videoSensing` extension
- moves logic to `mapStateToProps` based on previous [PR feedback](https://github.com/LLK/scratch-gui/pull/8634#discussion_r1018116505)

### Reason for Changes

This covers an edge case not handled in ENA-108. We should prevent users from encoding and sending webcam data over cloud variables. To accomplish this, we disconnect from the cloud variable server when the video sensing extension is used. This handles the scenario when a user first adds the video sensing extension to a project. 

Scenario:

1. User A creates a project that can encode/decode image data using cloud variables. Video sensing extension is not included in the project yet.
2. User A shares this project.
3. User B runs the project.
4. User A edits the project, including using the video sensing extension to capture and send their webcam to the cloud variable. User A does not leave the editing view. (_The cloud variable connection will be correctly disabled if they were to leave this page and re-run the project._)
5. Without this change, user B is able to see images from user A’s webcam.

### Test Coverage

Manual testing.

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/lib/cloud-manager-hoc.jsx
+++ b/src/lib/cloud-manager-hoc.jsx
@@ -99,6 +99,8 @@ const cloudManagerHOC = function (WrappedComponent) {
             }
         }
         handleExtensionAdded (categoryInfo) {
+            // Note that props.vm.extensionManager.isExtensionLoaded('videoSensing') is still false
+            // at the point of this callback, so it is difficult to reuse the canModifyCloudData logic.
             if (categoryInfo.id === 'videoSensing' && this.isConnected()) {
                 this.disconnectFromCloud();
             }
--- a/src/lib/cloud-manager-hoc.jsx
+++ b/src/lib/cloud-manager-hoc.jsx
@@ -25,10 +25,12 @@ const cloudManagerHOC = function (WrappedComponent) {
             super(props);
             this.cloudProvider = null;
             bindAll(this, [
-                'handleCloudDataUpdate'
+                'handleCloudDataUpdate',
+                'handleExtensionAdded'
             ]);
 
             this.props.vm.on('HAS_CLOUD_DATA_UPDATE', this.handleCloudDataUpdate);
+            this.props.vm.on('EXTENSION_ADDED', this.handleExtensionAdded);
         }
         componentDidMount () {
             if (this.shouldConnect(this.props)) {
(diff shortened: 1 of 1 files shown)


src/lib/cloud-manager-hoc.jsx
```diff
@@ -96,6 +98,11 @@ const cloudManagerHOC = function (WrappedComponent) {
                 this.connectToCloud();
             }
         }
+        handleExtensionAdded (categoryInfo) {
```
benjiwheeler, line 101: I think this approach should work great. Or... now that you've moved the `props.vm.extensionManager.isExtensionLoaded` check into `props.canModifyCloudData`, maybe we can get away with just waiting for `componentDidUpdate()` to be called? I wonder, what could we do so that adding an extension would trigger this HOC to update?

benjiwheeler (commented):
This looks good, but I have one question about a possibility that could, maybe, simplify it?

src/lib/cloud-manager-hoc.jsx
aoneill01: *Updating with a separate converation*

I investigated some options with triggering a re-render. There were some complications. One interesting issue is that `this.props.vm.extensionManager.isExtensionLoaded('videoSensing')` is still false at the point of the `EXTENSION_ADDED` callback being called, making it challenging to use `canModifyCloudData`. I will add a comment to the code to this effect.

benjiwheeler (approved):
LGTM!

benjiwheeler:
For posterity, here was a discussion we had elsewhere:

> I've never used this before, but there is a forceUpdate() for class components, so my first idea was just calling that on the EXTENSION_ADDED event. I ran into a problem: this doesn't recompute the canModifyCloudData because it is re-rendering CloudManager, not the react-redux connect hoc. The only way I can picture this recomputing canModifyCloudData is if something changes the redux state. This does eventually happen when the user interacts with the project. I can see it disconnecting even if I don't listen for EXTENSION_ADDED. (Not directly related, but another complication: this.props.vm.extensionManager.isExtensionLoaded('videoSensing') is still false in the extension added handler. The code to make this true happens right after the handler is complete.)

reply:

> Hm, that makes sense that forceUpdate() wouldn’t produce a mapStateToProps run.
this.props.vm.extensionManager.isExtensionLoaded('videoSensing') still being false is very relevant — that makes me think that your current approach probably makes sense.

PR #8685, opened 2022-11-30
fix: update extension monitors after locale change

### LINKED ISSUE #8499: ENA-142 Translation language monitor can be made to lag behind language selection
### Expected Behavior

Monitor label of extension blocks should be re-rendered in the new language when a new language is selected.

### Actual Behavior

Monitor label of extension blocks is re-rendered in the previous language when a new language is selected.
If changed value because a new language is selected, re-rendered in the new language.

### Steps to Reproduce

Assume the current language is English (en).

1. Add the Translate extension.
2. Check the language block.
3. Switch language to 日本語 (ja).
4. The language label are in 日本語 (ja). 
5. Switch language to にほんご (ja-Hira).
6. The language label are still in 日本語 (ja).

*If dose not reproduce, switch language a few times.

aoneill01:
### Resolves

- Resolves #8499
- Resolves [ENA-142](https://scratchfoundation.atlassian.net/browse/ENA-142)

### Proposed Changes

- Use changes to the `toolboxXML` from the Redux state to trigger rendering of monitors.

### Reason for Changes

When the locale changes, the virtual machine updates the extension blocks asynchronously. It is possible, and experimentally seems likely, that the React components that update the monitors will re-render before the extension blocks are done updating. This means the monitor label will display for the previous language. Most of the time the value of the language also changes. This causes a later re-render of the language monitor making everything look correct. However, when switching between `ja` and `ja-Hira`, the language value does not change.

The toolboxXML is updated after the virtual machine updates the locale. Since this is stored in the Redux state, it is easy to use changes to this value to trigger re-rendering monitors at the correct time. 

### Test Coverage

`monitor-list.test.jsx` was updated to include fake `toolboxXML` in the state, but most testing was manual.

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Safari
 
Windows
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


src/containers/monitor.jsx
```diff
@@ -250,6 +250,7 @@ Monitor.propTypes = {
     resizeMonitorRect: PropTypes.func.isRequired,
     spriteName: PropTypes.string, // eslint-disable-line react/no-unused-prop-types
     targetId: PropTypes.string,
+    toolboxXML: PropTypes.string, // eslint-disable-line react/no-unused-prop-types
```
cwillisf, line 253: Just a curiosity question: having an unused property seems like a sign that we're not doing something correctly. Does React (in this version or a newer version) have an official way to say that the rendering of an element depends on a property or event in a way that React can't analyze?

cwillisf (changes requested):
Is there a way to write a test that will fail without this change? I think that would be especially valuable since this is such an indirect issue: future-us might see that unused property and try to remove it as an optimization or cleanup without realizing that'll reintroduce this issue. (Hopefully they'd `git blame` or similar first, but I'm not sure I can trust future me...)

I'd also add "Resolves #8499" to the description of the PR

src/containers/monitor.jsx
aoneill01: I am not aware of anything exactly like that. I suppose if we were using hooks we could do something like:

```js
const toolboxXML = useSelector(...);
const label = useMemo(functionToCalculateLabel, [toolboxXML, otherDependencies]);
```

That way `toolboxXML` is used in the dependencies array, but I think there is a eslint test that could complain about the fact that `toolboxXML` was not actually used in the memoized function.

I agree having an unused prop feels a bit wrong. Seems like the heart of the problem is that we are computing labels based on some internal state that the vm is holding. React is good at rendering updates when props or state changes. I'm thinking about the idea of the monitor labels being stored in a React or Redux state similar to (or combined with) `state.scratchGui.monitorLayout` or `state.scratchGui.monitors`. I can't come up with a solution I like more than using `toolboxXML` updates.

 ---

Regarding adding a test that will fail without this change - With some minor tweaks, I could test something like:

```js
expect(Monitor.propTypes).toHaveProperty('toolboxXML');
```

That would make sure someone doesn't remove that propType.

Or we could test that when you render the component without the `toolboxXML` prop it has a warning. Unfortunately it looks like this is only displayed in the console, so I've seen examples of mocking `console.error` and trying to match the error message. Listening to `console.error` seems like a weird solution, but I can't find a better suggestion.

Neither stops someone from breaking things by removing line 270. Not a test, but another option is to just move the comment from line 269 to above this line. You'd probably notice the comment before attempting to clean it up. What do you think about those ideas, or is there another one you are thinking of?

apple502j: Hmm, can't we depend on the locale (`state.locales.locale`) instead of `toolboxXML`? That does change between ja and ja-hira. And having a big string as a props, doesn't sound great IMO.

aoneill01: Totally agree that depending on `locale` seems like a better option, and I tried that first. However, the virtual machine does not seem to refresh the locale synchronously when it changes (`extension-manager.js : refreshBlocks()`). This means that Monitor can re-render before the virtual machine has updated all of the blocks with the new locale. When this happens, the monitor still has the label for the previous locale, not the new one.

I do want to be careful about the size of `toolboxXML`. I may be missing something, but I can't spot a problem. I don't think the component will be creating a copy of the `toolboxXML` contents from the redux store, just a reference to it. React-redux will check to see if the value changed, but it should be using [strict equality for that comparison](https://react-redux.js.org/using-react-redux/connect-mapstate#return-values-determine-if-your-component-re-renders). I can't think of a scenario where this prop is using significant memory or computation time.

aoneill01: @cwillisf - I added an integration test for this issue.

cwillisf (approved):
Looks good to me!
Thanks for adding that test!

PR #8725, opened 2023-01-05
ci: do not run commitlint on develop branch

aoneill01:
### Resolves

- Resolves [ENA-199](https://scratchfoundation.atlassian.net/browse/ENA-199)

### Proposed Changes

- Skip `commitlint` on the following branches:
  - `master`
  - `develop`
  - `hotfix/*`

### Reason for Changes

We do not want to re-write commit history on these "main" branches. Commit messages that do not follow conventional commit should be caught in other branches. However, this way the CI won't fail if they are merged to one of these branches. 

### Test Coverage

- Ran `circleci config validate` to validate the configuration

### Browser Coverage

n/a

[ENA-199]: https://scratchfoundation.atlassian.net/browse/ENA-199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -219,6 +219,12 @@ workflows:
     jobs:
       - commitlint/lint:
           target-branch: develop
+          filters:
+            branches:
+              ignore:
+                - master
+                - develop
+                - /^hotfix\/.*/
   push-translations:
     triggers:
       - schedule:


.circleci/config.yml
```diff
@@ -219,6 +219,12 @@ workflows:
     jobs:
       - commitlint/lint:
           target-branch: develop
+          filters:
```
aoneill01, line 222: Went with `filters` rather than a `when` condition on the workflow because this followed the pattern of other jobs in this file. Also for consistency, I did not include `release` branches since no other jobs in this file referenced a release branch. Seems like we would want to make multiple changes to this config if we want to support `release` branches.

cwillisf (approved):

PR #8733, opened 2023-01-08
Add padding to Save Now to increase clickable area

### LINKED ISSUE #8571: Increase the clickable area of the "Save Now" button
### The problem
The current clickable area of the "Save Now" button in the top-right corner of the editor is only the text itself. Other buttons in the editor like Tutorials or See Project Page are like an actual button, whereas "Save Now" seems to be more like a link that does an action. That means you have to click _right on the text_ if you want to save your project. If you're using a mouse to do that, some people may briefly have trouble getting the mouse exactly on the button.

### What we can do about it
I would simply increase the size of the Save Now textbox. Not the font size, just the textbox area, or something like that.

Current size:
The hit area is only on top of the text, not very forgiving

Increased size:
The hit area stretches slightly above and below the text for a bigger clickable button area

As a result, the clickable area of the button would be larger and more forgiving.

DNin01:
### Resolves

- Resolves #8571

### Proposed Changes

Adds padding to both the top and bottom of the Save Now element next to the My Stuff button, which is clickable. This increases the clickable area of the Save Now button

### Reason for Changes

The clickable area of Save Now is a little small and hard to click every time. I just wanted it to be slightly more forgiving.

### Test Coverage

I don't use Git or npm, so I could not "test" my changes, but I hope it works. It was very easy.

### Browser Coverage
Check the OS/browser combinations tested (At least 2) 😕

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

(Tested by adding the style rules to the Save Now element in the browser's developer tools.)
(We use parental controls and it only works in Edge, so I don't have any other browsers to test these changes in, but I'm pretty confident that it will work across browsers.)

### DIFF
--- a/src/components/menu-bar/save-status.css
+++ b/src/components/menu-bar/save-status.css
@@ -1,4 +1,6 @@
 .save-now {
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
     cursor: pointer;
+    padding-top: 10px;
+    padding-bottom: 10px;
 }


BroJac5246:
Tested and working on:
Chrome 109 (beta, 32-bit) on Chromebook
Firefox ESR 102.6.0esr (64-bit) on Chromebook
Chrome 108 (stable, 64-bit) on Windows 10

owenkaplinsky:
The size of the Save Now button is much larger than in this PR in Scratch right now. I think this has already been added.

BroJac5246:
> The size of the Save Now button is much larger than in this PR in Scratch right now. I think this has already been added.

Do you by chance use Scratch Addons? We automatically added a temporary solution until the Scratch Team hopefully adds this.

https://github.com/ScratchAddons/ScratchAddons/pull/6611

owenkaplinsky:
Oh, yes I do, I wasn't aware of that. Thanks for telling me.

PR #8751, opened 2023-01-27
fix: remove Sentry and add ErrorBoundary tests

cwillisf:
### Resolves

Half of ENA-224

### Proposed Changes

This change removes all references to Sentry throughout `scratch-gui`. Sentry support was tightly integrated with the `ErrorBoundary` component, which would report an error to Sentry and then display a crash message that includes the Sentry event ID. Now, `ErrorBoundary` skips sending the error to Sentry but still displays the crash message.

### Reason for Changes

We're no longer planning to use Sentry in this way. There were previously no tests for `ErrorBoundary` here in `scratch-gui`, but there are tests for `scratch-www`'s version of `ErrorBoundary`, so I made similar tests here. This test sequence relies on calling Enzyme's `simulateError`, which wasn't available in the version of Enzyme adapter we were using previously, so I upgraded that dependency.

### Test Coverage

A new unit test verifies that `ErrorBoundary` switches from displaying its children to displaying a crash message when it catches an error.

You can test locally by running `npm start`, then visiting `http://localhost:8601/#9999999999999` (or some other nonexistent project ID).

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


src/containers/error-boundary.jsx
```diff
@@ -10,37 +10,39 @@ class ErrorBoundary extends React.Component {
...
-            errorId: window.Sentry ? window.Sentry.lastEventId() : null
-        });
-
-        // Log error locally for debugging as well.
-        log.error(`Unhandled Error: ${error.stack}\nComponent stack: ${info.componentStack}`);
+        // report every error in the console
+        log.error([
+            `Unhandled Error with action='${this.props.action}': ${error.stack}`,
+            `Component stack: ${errorInfo.componentStack}`
+        ].join('\n'));
```
aoneill01, line 45: 👍 I like how readable this is with each line on its own row and `join`ing them.

aoneill01 (approved):
Looks good. Nice touch with adding the jsdoc.

PR #8753, opened 2023-02-10
feat: store color theme in state

aoneill01:
_Note that this is not merging into `develop`. Instead we will likely have a long-running feature branch for the color theme epic._

### Resolves

- Resolves [ENA-236](https://scratchfoundation.atlassian.net/browse/ENA-236)

### Proposed Changes

- Add a theme name to the Redux state
- Create a framework for converting a theme name to a set of colors
- Pass the colors from the selected theme to `scratch-blocks`

### Reason for Changes

This is a first step towards the larger epic of creating high contrast and dark mode color themes for the editor. Eventually the user will be able to choose the color theme through the UI and see all of the editor components correctly themed. This change keeps the selected theme name in memory and converts it to a set of colors that will later be used to change styling. 

### Test Coverage

- Added unit tests for logic that maps a theme name to a set of colors
- Manual testing. Since there is not a user interface for changing themes yet, developer tools can be used to change the Redux state. I recommend using the Redux extension and dispatching the following payloads. (Or change the default initial theme in `scratch-gui/src/reducers/theme.js`.)

```json
{
  type: 'scratch-gui/theme/SET_THEME',
  theme: 'high-contrast'
}
```

```json
{
  type: 'scratch-gui/theme/SET_THEME',
  theme: 'standard'
}
```

```json
{
  type: 'scratch-gui/theme/SET_THEME',
  theme: 'dark-mode'
}
```

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

[ENA-236]: https://scratchfoundation.atlassian.net/browse/ENA-236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -18,6 +18,7 @@ import {BLOCKS_DEFAULT_SCALE, STAGE_DISPLAY_SIZES} from '../lib/layout-constants
 import DropAreaHOC from '../lib/drop-area-hoc.jsx';
 import DragConstants from '../lib/drag-constants';
 import defineDynamicBlock from '../lib/define-dynamic-block';
+import {getColorsForTheme, themes} from '../lib/themes';
 
 import {connect} from 'react-redux';
 import {updateToolbox} from '../reducers/toolbox';
@@ -94,7 +95,7 @@ class Blocks extends React.Component {
         const workspaceConfig = defaultsDeep({},
             Blocks.defaultOptions,
             this.props.options,
-            {rtl: this.props.isRtl, toolbox: this.props.toolboxXML}
+            {rtl: this.props.isRtl, toolbox: this.props.toolboxXML, colours: getColorsForTheme(this.props.theme)}
         );
         this.workspace = this.ScratchBlocks.inject(this.blocks, workspaceConfig);
 
@@ -351,7 +352,8 @@ class Blocks extends React.Component {
             return makeToolboxXML(false, target.isStage, target.id, dynamicBlocksXML,
                 targetCostumes[targetCostumes.length - 1].name,
                 stageCostumes[stageCostumes.length - 1].name,
-                targetSounds.length > 0 ? targetSounds[targetSounds.length - 1].name : ''
+                targetSounds.length > 0 ? targetSounds[targetSounds.length - 1].name : '',
+                getColorsForTheme(this.props.theme)
             );
         } catch {
             return null;
@@ -620,22 +622,11 @@ Blocks.propTypes = {
             wheel: PropTypes.bool,
             startScale: PropTypes.number
         }),
-        colours: PropTypes.shape({
-            workspace: PropTypes.string,
-            flyout: PropTypes.string,
-            toolbox: PropTypes.string,
-            toolboxSelected: PropTypes.string,
-            scrollbar: PropTypes.string,
-            scrollbarHover: PropTypes.string,
-            insertionMarker: PropTypes.string,
-            insertionMarkerOpacity: PropTypes.number,
-            fieldShadow: PropTypes.string,
-            dragShadowOpacity: PropTypes.number
-        }),
         comments: PropTypes.bool,
         collapse: PropTypes.bool
     }),
     stageSize: PropTypes.oneOf(Object.keys(STAGE_DISPLAY_SIZES)).isRequired,
+    theme: PropTypes.oneOf(themes),
     toolboxXML: PropTypes.string,
     updateMetrics: PropTypes.func,
     updateToolboxState: PropTypes.func,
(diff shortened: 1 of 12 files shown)


src/lib/themes/index.js
```diff
@@ -0,0 +1,28 @@
...
+import highContrast from './high-contrast';
+import defaultColors from './default-colors';
+
+const mergeWithDefaults = colors => defaultsDeep({}, colors, defaultColors);
+
+const themeMap = {
+    'dark-mode': mergeWithDefaults(darkMode),
+    'high-contrast': mergeWithDefaults(highContrast),
+    'standard': defaultColors
+};
```
aoneill01, line 13: I'm open to better names for our themes (ex standard / original / default / ???). This is easy to modify later too.

Right now the color data structure match the payload that is passed to `scratch-blocks`. I could see this being augmented later if we have some properties that are part of the editor but not in the blocks.

ilikecereal1legacy:
Tested on iPad Safari!
Dark mode:
0478EF74-34FF-47EF-B48F-A8A24FFF2394
High contrast:
3B3B6378-A1BB-47C3-8338-D01B1CFEC372
Standard:
49B8DDF8-56A4-45D2-B38E-5F8B0E09FDF0

src/containers/blocks.jsx
```diff
@@ -656,18 +657,6 @@ Blocks.defaultOptions = {
...
-        flyout: '#F9F9F9',
-        toolbox: '#FFFFFF',
-        toolboxSelected: '#E9EEF2',
-        scrollbar: '#CECDCE',
-        scrollbarHover: '#CECDCE',
-        insertionMarker: '#000000',
-        insertionMarkerOpacity: 0.2,
-        fieldShadow: 'rgba(255, 255, 255, 0.3)',
-        dragShadowOpacity: 0.6
-    },
```
cwillisf, line 670: Does removing this mean there should be a `.isRequired` on `Blocks.propTypes.options.colours`?

src/lib/themes/index.js
cwillisf: I think these names are fine for now, until/unless we settle on different user-facing names later.

> I could see this being augmented later if we have some properties that are part of the editor but not in the blocks.

Makes sense! Seems like it would be pretty easy to turn it into `theme = { blocksColors: {}, otherStuff: {} }` if necessary.

src/lib/make-toolbox-xml.js
```diff
@@ -1,17 +1,18 @@
...
 /* eslint-disable no-unused-vars */
-const motion = function (isInitialSetup, isStage, targetId) {
+const motion = function (isInitialSetup, isStage, targetId, colors) {
     const stageSelected = ScratchBlocks.ScratchMsgs.translate(
         'MOTION_STAGE_SELECTED',
         'Stage selected: no motion blocks'
     );
     return `
-    <category name="%{BKY_CATEGORY_MOTION}" id="motion" colour="#4C97FF" secondaryColour="#3373CC">
+    <category name="%{BKY_CATEGORY_MOTION}" id="motion" colour="${colors.primary}" secondaryColour="${colors.tertiary}">
```
cwillisf, line 16: Is `colors.tertiary` always used to fill in a "secondary" color?

cwillisf (commented):
Looking good so far! I have some questions, but as far as overall structure & strategy, this makes sense to me.

src/lib/make-toolbox-xml.js
aoneill01: For the category styling, yes. 

These aren't all of the uses, but for the block colors, `scratch-blocks` uses:
- primary for the block background color
- secondary for the dropdown background color
- tertiary for the border color

These categories only have two colors for the circle icon: background (primary) and border (secondary). It is a bit unfortunate that the block's tertiary matches up to the category's secondary. Any suggestions to make this less confusing?

src/containers/blocks.jsx
aoneill01: Good point... Technically if it is left off the default `scratch-blocks` colors are used, which is nice. But make-toolbox-xml now expects certain colors to be provided. I've gone a different direction that I think accomplishes the same goal. I've passed the theme name to `<Blocks />`, including a default if it is not provided. This should ensure all of the colors the editor uses will be available to this component. See the latest commit.

src/lib/make-toolbox-xml.js
cwillisf: Gotcha, that makes sense. I probably knew that at some point in the past...

I think it could make sense to add a comment about this if you see a good place for it, but once it's established it'll go from "hmm, I wonder if this is a mistake" to "hmm, it's a bit odd that it worked out this way" -- less of a concern :)

aoneill01: 👍 Added some comments to the code.

Crispydogs101:
How does the mergewithdeafault work in a way it can process the dark mode code?

aoneill01:
`mergeWithDefaults` allows you to define just the colors that are different from the default theme, using [defaultsDeep](https://lodash.com/docs/4.17.15#defaultsDeep). My assumption is that some themes will share a lot of settings with the default theme. We'll see if that holds true. 🤞

Crispydogs101:
Alright tak!

LuigiUI238:
This sounds awesome! I want to help with this dark mode test, too.

PaperMarioFan-2022:
I think this is will be a great feature!

PR #8756, opened 2023-02-16
feat: color theming for monitors based on block category

aoneill01:
_Part of the color contrast epic. During development this will live on the `feature/color-contrast` branch._

_There is not a way to change the color theme in the UI yet. If you want to see this in action, you can change the `initialState` in `src/reducers/theme.js` or use Redux developer tools to dispatch the set theme action._

### Resolves

- Resolves [ENA-243](https://scratchfoundation.atlassian.net/browse/ENA-243)

### Proposed Changes

- Use the color theme to style the visible monitors. The value is displayed on a background matching the category's primary color and the blocks' text color. There are various modes for monitors that were modified, including:
  - standard
  - large value without label
  - slider
  - lists

The actual colors may change, and dark mode may eventually result in styling outside of the value block, but here are some screenshots:

#### Standard
monitor-standard

#### High contrast
monitor-contrast

#### Dark mode
monitor-dark

### Reason for Changes

The monitors were styled based on the category block colors. Now that those colors can change, the monitors should also update based on the theme.

### Test Coverage

Added a test for the monitor component to verify colors were correctly being passed to children components.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

[ENA-243]: https://scratchfoundation.atlassian.net/browse/ENA-243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/monitor/list-monitor-scroller.jsx
+++ b/src/components/monitor/list-monitor-scroller.jsx
@@ -41,7 +41,10 @@ class ListMonitorScroller extends React.Component {
                 <div
                     className={styles.listValue}
                     dataIndex={index}
-                    style={{background: this.props.categoryColor}}
+                    style={{
+                        background: this.props.categoryColor.background,
+                        color: this.props.categoryColor.text
+                    }}
                     onClick={this.props.draggable ? this.handleEventFactory(index) : null}
                 >
                     {this.props.draggable && this.props.activeIndex === index ? (
@@ -51,6 +54,7 @@ class ListMonitorScroller extends React.Component {
                                 autoComplete={false}
                                 className={classNames(styles.listInput, 'no-drag')}
                                 spellCheck={false}
+                                style={{color: this.props.categoryColor.text}}
                                 type="text"
                                 value={this.props.activeValue}
                                 onBlur={this.props.onDeactivate}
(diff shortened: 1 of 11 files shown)


test/unit/components/monitor.test.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import {shallow} from 'enzyme';
+import DefaultMonitor from '../../../src/components/monitor/default-monitor';
+import Monitor from '../../../src/components/monitor/monitor';
+
+jest.mock('../../../src/lib/themes/default-colors');
+
+describe('Monitor Component', () => {
```
benjiwheeler, line 9: It could make sense to add at least one more test of a different theme

src/components/monitor/list-monitor-scroller.jsx
```diff
@@ -41,7 +41,7 @@ class ListMonitorScroller extends React.Component {
                 <div
                     className={styles.listValue}
                     dataIndex={index}
-                    style={{background: this.props.categoryColor}}
+                    style={{background: this.props.categoryColor.background, color: this.props.categoryColor.text}}
```
benjiwheeler: tiny suggestion
```suggestion
                    style={{
                        background: this.props.categoryColor.background,
                        color: this.props.categoryColor.text
                    }}
```

benjiwheeler (approved):
Looks good to me -- I noted two small suggestions, but I'm OK with leaving the decision to take them or not in your hands.

test/unit/components/monitor.test.jsx
aoneill01: I like both of these ideas. Pushed some changes. I don't feel these new changes need to be reviewed. But if you do, you should be aware that I rebase and squashed this PR into a single commit. That makes pulling the changes if you've already pulled them a bit tricky.

PR #8757, opened 2023-02-17
feat: apply color theme to extension categories and blocks

aoneill01:
_Part of the color contrast epic. During development this will live on the `feature/color-contrast` branch._

_If you want to test this locally, there are also some changes in `scratch-blocks` on the `feature/color-contrast` branch. This would involve building `scratch-blocks` and using `npm link scratch-blocks`. This is not strictly necessary, but some of the block colors won't look identical to the screenshots._

_There is not a way to change the color theme in the UI yet. If you want to see this in action, you can change your operating system appearance and accessibility settings, change the `initialState` in `src/reducers/theme.js`, or use Redux developer tools to dispatch the set theme action._

### Resolves

- Resolves [ENA-242](https://scratchfoundation.atlassian.net/browse/ENA-242)

### Proposed Changes

- `blocks.jsx` receives extension categories and blocks from `scratch-vm` and passes them on to `scratch-blocks`. Modify these data structures to apply the color theme.
- Added `system-preferences-hoc.jsx` to apply color theme based on operating system preferences (dark mode, high contrast accessibility setttings). This is just for ease of testing. I don't think we need to be overly critical of this code right now as it will either be removed or significantly refactored when the user can select color themes.

### Reason for Changes

The color theme for extensions needs to be applied somewhere. Since the definition for extensions currently comes from the virtual machine, one option would be to somehow support color theming in the virtual machine. However, I felt it would be nice if the virtual machine did not take on that responsibility.

Since all of the extensions currently have icons, here is a screenshot of high contrast mode with an icon-less extension:

Screenshot 2023-02-17 at 4 24 58 PM

### Test Coverage

- Added unit tests for `injectExtensionBlockColors`
- Manual testing for `injectExtensionCategoryColors`. It uses `XMLSerializer` which is a browser API.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

[ENA-242]: https://scratchfoundation.atlassian.net/browse/ENA-242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/blocks.jsx
+++ b/src/containers/blocks.jsx
@@ -18,7 +18,8 @@ import {BLOCKS_DEFAULT_SCALE, STAGE_DISPLAY_SIZES} from '../lib/layout-constants
 import DropAreaHOC from '../lib/drop-area-hoc.jsx';
 import DragConstants from '../lib/drag-constants';
 import defineDynamicBlock from '../lib/define-dynamic-block';
-import {getColorsForTheme, themes} from '../lib/themes';
+import {DEFAULT_THEME, getColorsForTheme, themes} from '../lib/themes';
+import {injectExtensionBlockColors, injectExtensionCategoryColors} from '../lib/themes/blockHelpers';
 
 import {connect} from 'react-redux';
 import {updateToolbox} from '../reducers/toolbox';
@@ -348,7 +349,10 @@ class Blocks extends React.Component {
             const stageCostumes = stage.getCostumes();
             const targetCostumes = target.getCostumes();
             const targetSounds = target.getSounds();
-            const dynamicBlocksXML = this.props.vm.runtime.getBlocksXML(target);
+            const dynamicBlocksXML = injectExtensionCategoryColors(
+                this.props.vm.runtime.getBlocksXML(target),
+                this.props.theme
+            );
             return makeToolboxXML(false, target.isStage, target.id, dynamicBlocksXML,
                 targetCostumes[targetCostumes.length - 1].name,
                 stageCostumes[stageCostumes.length - 1].name,
@@ -432,7 +436,7 @@ class Blocks extends React.Component {
                     if (blockInfo.info && blockInfo.info.isDynamic) {
                         dynamicBlocksInfo.push(blockInfo);
                     } else if (blockInfo.json) {
-                        staticBlocksJson.push(blockInfo.json);
+                        staticBlocksJson.push(injectExtensionBlockColors(blockInfo.json, this.props.theme));
                     }
                     // otherwise it's a non-block entry such as '---'
                 });
@@ -655,7 +659,7 @@ Blocks.defaultOptions = {
 Blocks.defaultProps = {
     isVisible: true,
     options: Blocks.defaultOptions,
-    theme: 'standard'
+    theme: DEFAULT_THEME
 };
 
 const mapStateToProps = state => ({
--- a/src/lib/themes/blockHelpers.js
+++ b/src/lib/themes/blockHelpers.js
@@ -1,0 +1,61 @@
+import {DEFAULT_THEME, getColorsForTheme} from '.';
+
+// scratch-blocks colours has a pen property that scratch-gui uses for all extensions
+const getExtensionColors = theme => getColorsForTheme(theme).pen;
+
+/**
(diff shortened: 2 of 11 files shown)


src/containers/blocks.jsx
```diff
@@ -655,7 +659,7 @@ Blocks.defaultOptions = {
 Blocks.defaultProps = {
     isVisible: true,
     options: Blocks.defaultOptions,
-    theme: 'standard'
+    theme: DEFAULT_THEME
```
cwillisf, line 662: 😎

src/lib/themes/index.js
```diff
@@ -4,12 +4,13 @@ import darkMode from './dark-mode';
 import highContrast from './high-contrast';
 import defaultColors from './default-colors';
 
+const DEFAULT_THEME = 'standard';
 const mergeWithDefaults = colors => defaultsDeep({}, colors, defaultColors);
 
 const themeMap = {
     'dark-mode': mergeWithDefaults(darkMode),
     'high-contrast': mergeWithDefaults(highContrast),
```
cwillisf, line 12: Do you expect these to be `[DARK_MODE]` & `[HIGH_CONTRAST]` in the future?

test/unit/util/themes.test.js
```diff
@@ -19,4 +20,42 @@ describe('themes', () => {
...
+        const updated = injectExtensionBlockColors(blockInfoJson, 'dark-mode');
+
+        expect(updated).toEqual({
+            type: 'dummy_block',
+            colour: '#FFFFFF',
+            colourSecondary: '#EEEEEE',
+            colourTertiary: '#DDDDDD'
+        });
+        // The original value was not modified
+        expect(blockInfoJson.colour).toBe('#0FBD8C');
```
cwillisf, line 51: Awesome - thanks for including this test!

src/lib/themes/blockHelpers.js
```diff
@@ -0,0 +1,61 @@
+import {DEFAULT_THEME, getColorsForTheme} from '.';
```
cwillisf, line 1: I don't think I've seen `from '.'` before... TIL :)

src/lib/themes/blockHelpers.js
```diff
@@ -0,0 +1,61 @@
...
+ * @param {string} theme - Theme name
+ * @returns {Array.<object>} Dynamic block XML updated with colors.
+ */
+const injectExtensionCategoryColors = (dynamicBlockXML, theme) => {
+    // Don't do any manipulation for the default theme
+    if (theme === DEFAULT_THEME) return dynamicBlockXML;
+
+    const extensionColors = getExtensionColors(theme);
+    const parser = new DOMParser();
+    const serializer = new XMLSerializer();
```
cwillisf, line 21: Is it possible to polyfill these in Node.js so this stuff can be added to the automated tests? Maybe `jest.mock` can help?

cwillisf (approved):
Looks good! I put a couple questions inline, but even if they're valid I think they can be "later" things.

src/lib/themes/index.js
aoneill01: Maybe? I'm kind of hoping dark mode and high contrast won't be referenced outside of here and perhaps the dropdown choices. I'll keep an eye on this and add it if it gets referenced.

src/lib/themes/blockHelpers.js
aoneill01: VS Code autocomplete FTW!

src/lib/themes/blockHelpers.js
aoneill01: I was on the fence about polyfilling because it was one more dependency and may not be identical to the browser. I didn't even consider mocking. That's a good idea. Added a test with `XMLSerializer.serializeToString` mocked.

PR #8811, opened 2023-03-16
feat: bring back cat blocks

aoneill01:
### Resolves

- Resolves [ENA-262](https://scratchfoundation.atlassian.net/browse/ENA-262)

### Proposed Changes

- Applies `hotfix/totally-normal-2022` changes with Caturday as the only special mode in the menu

To see this locally, add the `isTotallyNormal` attribute to `<WrappepdGui>` in `scratch-gui/src/playground/render-gui.jsx`.

Screenshot 2023-03-17 at 10 31 13 AM

### Reason for Changes

This will restore cat blocks for April Fools. The `isTotallyNormal` attribute will be controlled by a flag from `scratch-www`.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

[ENA-262]: https://scratchfoundation.atlassian.net/browse/ENA-262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/gui/gui.jsx
```diff
@@ -160,6 +165,9 @@ const GUIComponent = props => {
                 dir={isRtl ? 'rtl' : 'ltr'}
                 {...componentProps}
             >
+                <OldTimeyMode />
```
benjiwheeler: You don't need these, right?

src/components/gui/gui.jsx
```diff
@@ -21,6 +21,9 @@ import MenuBar from '../menu-bar/menu-bar.jsx';
 import CostumeLibrary from '../../containers/costume-library.jsx';
 import BackdropLibrary from '../../containers/backdrop-library.jsx';
 import Watermark from '../../containers/watermark.jsx';
+import OldTimeyMode from '../oldtimey-mode/oldtimey-mode.jsx';
```
benjiwheeler: You don't need these, right? Though I guess it doesn't hurt?

benjiwheeler (approved):
LGTM -- I know there's still stuff in there for other modes (old timey, etc.), but that's fine, it doesn't hurt anything...

src/components/gui/gui.jsx
aoneill01: We may completely remove the code for the other modes in the future, but I think removing the references to them from gui.jsx is a good idea. Done.

PR #8843, opened 2023-03-23
Feature/themed icons

aoneill01:
### Resolves

- Resolves [ENA-254](https://scratchfoundation.atlassian.net/browse/ENA-254)

### Proposed Changes

- Use high contrast block icons when using that theme
  - Store the icons that are different in `static/blocks-media-high-contrast`
  - Modify webpack to combine the default icons from `node_modules/scratch-blocks/media` with `static/blocks-media-high-contrast` in a new folder
  - Update `src/lib/themes/index.js` to have a per-theme property for `blocksMediaFolder`
  - Update `gui.jsx` to choose the appropriate media folder passed to `scratch-blocks`
- Use high contrast extension icons when using that theme
  - Update the themes in `src/lib/themes/index.js` to have a property for `extensions`. This stores optional block and category icon images encoded as a data image URI.
  - `src/lib/themes/blockHelpers.js` is already being used to modify data being passed to `scratch-blocks` to change colors for extension categories and blocks. Augment this to also inject the optional block and category icon images.

### Reason for Changes

This is part of an ongoing epic to introduce a high contrast theme. This solution should be generic to support other future themes that use different icons within the Scratch blocks.

### Test Coverage

- Added unit tests for the logic that injects extension icons into categories and blocks

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

[ENA-254]: https://scratchfoundation.atlassian.net/browse/ENA-254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/record-modal/record-modal.css
```diff
@@ -106,7 +106,8 @@
     text-align: center;
 }
 
-.record-button {
+/* Increase specificity to make sure this wins out over normalize.css */
+svg.record-button {
```
aoneill01, line 110: This is unrelated to the feature. It was brought to my attention that the expanding circle when recording audio is being clipped to a square region. This appears to be because of some css normalization when the editor is used in `scratch-www`. This increases the specificity, overriding the rule in `scratch-www`. Sorry for sneaking in unrelated work, but it was a simple fix.

Screenshot of problem:
Screenshot 2023-03-23 at 10 51 06 AM

src/lib/themes/high-contrast.js
```diff
@@ -80,4 +80,30 @@ const blockColors = {
...
...Dg2MiAyNC41MDI0IDIuNTc2ODQgMjUuNzk3NCAxMC4wNDUxQzI2LjQzNDkgMTMuNzAzNSAyNi45Njk3IDE2LjMyMzEgMjcuMzQ3NSAxOC4xNzM5QzI3Ljk5NTcgMjEuMzQ5NCAyOC4xODE5IDIyLjI2MTYgMjcuNjMyNSAyMi4yNzQxQzI3LjMzMjEgMjIuMTUyIDI3LjAzMjcgMjIuMDU3OCAyNi42OTU3IDIxLjk3MjRDMjYuMDg4IDIxLjgyMjUgMjUuNDUxNiAyMS43MzgyIDI0LjgwNTMgMjEuNzM4MkMyMS43MjcxIDIxLjczODIgMTkuNjAzNiAyMy41ODkyIDIwLjA2MjEgMjUuODc0QzIwLjUxMDcgMjguMTU2OSAyMy4zNjQ2IDMwIDI2LjQzMzggMzBaIiBmaWxsPSJibGFjayIvPgo8cGF0aCBkPSJNOS40Mzg2MSAzNi4wMDAxQzEyLjUwNjUgMzYuMDAwMSAxNC42MzAyIDM0LjE0OCAxNC4xODE3IDMxLjg2NDJDMTEuMzAzMiAxOS4wMzM2IDEwLjQyOTkgMTUuODQxNiAxMC45NDk2IDE1LjU5NThDMTEuMzE2NiAxNS40MjIyIDEyLjM3ODMgMTYuNzE3NyAxMy45MTkgMTcuMTI2QzE3LjY0MjggMTguMTEzNiAyMy4yNzI5IDEyLjM5ODUgMTcuNTY2NyAxMy4yOTE4QzE1LjMyNDggMTMuNjM4NiAxMy4xNjIzIDEyLjIzODIgMTEuNDkyMiAxMS4xNTY5QzguOTEyMzcgOS40ODYzNiA3LjUwNzk3IDguNTc2OTggOC44MDI1MSAxNi4wNDUxQzkuNDQyOTQgMTkuNzI4NiA5Ljk3OTggMjIuMzU5NCAxMC4zNTggMjQuMjEyNUMxMC45OTEyIDI3LjMxNTQgMTEuMTc5NSAyOC4yMzg0IDEwLjY2NDQgMjguMjczMkM5Ljc4NDMyIDI3LjkyNTggOC44MTIzOSAyNy43MjkzIDcuODEwODIgMjcuNzI5M0M0LjczNCAyNy43MjkzIDIuNjExMzQgMjkuNTgwNCAzLjA1OTc4IDMxLjg2NDJDMy41MTgwOSAzNC4xNDggNi4zNzA2OSAzNi4wMDAxIDkuNDM4NjEgMzYuMDAwMVoiIGZpbGw9ImJsYWNrIi8+Cjwvc3ZnPgo='
```
cwillisf: I'm open to debate here, but I think I disagree with years-ago-me putting these data URIs into the extensions. I think we should be using Webpack for this instead, like `scratch-www` does.

To set that up, add something like this to the `rules` section of the Webpack config (this is copied straight from `www`):
```
            {
                test: /\.(png|jpg|gif|eot|svg|ttf|woff)$/,
                loader: 'url-loader'
            }
```

then you can `import` or `require` your icon and that'll give you the data URI string. That way if there's an icon update we can just replace the files in the repo without manually re-encoding them.

It looks like `scratch-gui` doesn't list `url-loader` in `package.json` but it's already in the dependency tree, so adding it shouldn't be a big deal.

cwillisf: Oh, and as for storing the icons with the theme vs. with the extension: I think either way is fine, but this seems like less work. Since the VM's part of the extension system isn't thoroughly "theme aware" I think it would be unnecessary work to extend the extension system(heh)  to deal with this. We can do that if/when we add user extensions, and re-debate whether to associate these icons with the theme or each extension.

cwillisf (commented):
One suggestion regarding the icon URIs, but overall this looks great!

src/lib/themes/high-contrast.js
aoneill01: Great idea! This cascaded into some other changes as well. 

- I wanted to keep the icon svg files close to the theme, so I decided to create a folder per-theme in `src/lib/themes`.
- Moved the high contrast blocks-media stuff in there too.

I could even see this going a step further in the future, perhaps more convention than configuration. Each theme has its own folder, and based on sub-folders / file names it could automatically override blocks-media, extension icons, and more.

I also realized everything seems to work if I didn't modify webpack and just used the existing `file-loader` rather than the `url-loader`.  scratch-blocks doesn't seem to need it to be a data uri. But it felt like a good thing to use the `url-loader` across all of `scratch-gui` if the file size was small. I arbitrarily picked 2,048 bytes as the limit. If the asset is smaller than that it uses the data uri. If it is larger, it falls back to serving it from `static/assets`. Locally I saw a difference of <1% increase in `scratch-gui.js` when using these settings. I don't think it will adversely affect download time for the js.

cwillisf (approved):
Awesome! I like the `limit` option on `url-loader` -- I wasn't aware of that. I also like how you've organized `themes/`.

PR #8929, opened 2023-05-09
Allow typescript

cwillisf:
### Proposed Changes

Main change: enable TypeScript support in this repository

Supporting changes:

* Convert `Button` component to TypeScript as an example
* Upgrade `eslint` and `jest`
* Consolidate `eslintrc` files: the ones in `src/` and `test/` still exists but are smaller than before and share more with the config in the repository root
* MANY lint fixes
* A few test fixes

### Reason for Changes

This configuration allows us to use TypeScript for new development while not requiring us to convert our code to TypeScript all at once.

After upgrading Jest, some existing issues were exposed. The most surprising to me is that `test/unit/util/audio-context.test.js` has been failing in such a way that the failure was not "noticed" by Jest, so the build was passing anyway. I believe this is because we were rejecting a Promise without a `catch` clause (or `await`), and maybe because the older Jest was less Promise-aware. There were also several timing issues related to assuming that `process.nextTick` would be enough to finish `storage.load` or a whole project load, which is not necessarily guaranteed. Those tests now wait for a Promise to resolve.

### Test Coverage

Updated existing tests

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


.eslintrc.cjs
```diff
@@ -0,0 +1,65 @@
...
+            '.ts',
+            '.tsx',
+        ],
+        'import/resolver': {
+            webpack: {
+                config: path.resolve('webpack.config.js')
+            }
+        },
+        'react': {
+            version: '16.2' // Prevent 16.3 lifecycle method errors
```
cwillisf, line 39: Previously, we only specified the React version in `src/`, which meant that using the React plugin for eslint in `test/` would generate a big warning.

src/components/button/button.tsx
```diff
@@ -4,18 +4,26 @@ import React from 'react';
...
+        disabled: boolean,
+        iconClassName: string,
+        iconSrc: string,
+        onClick: React.MouseEventHandler,
+        children: React.ReactNode
+    }) => {
 
     if (disabled) {
-        onClick = function () {};
+        onClick = function () { /* do nothing */ };
```
cwillisf, line 26: It's now a warning to have an empty function, but adding a comment to indicate that it's intentionally empty satisfies that warning.

src/components/connection-modal/dots.jsx
```diff
@@ -5,6 +5,20 @@ import classNames from 'classnames';
...
+            {
+                [styles.inactiveStepDot]: props.type === 'inactive',
+                [styles.activeStepDot]: props.type === 'active',
+                [styles.successDot]: props.type === 'success',
+                [styles.errorDot]: props.type === 'error'
+            }
+        )}
+    />
+);
+
```
cwillisf, line 21: I moved this up to fix a use-before-definition complaint

src/components/button/button.tsx
```diff
@@ -4,18 +4,26 @@ import React from 'react';
...
+        children,
+        ...props
+    }: {
+        className: string,
+        disabled: boolean,
+        iconClassName: string,
+        iconSrc: string,
+        onClick: React.MouseEventHandler,
+        children: React.ReactNode
+    }) => {
```
hyperobject, line 23: Could be worth thinking about moving this type to a `ButtonProps` interface defined in this file, might make this a little easier to read?

test/helpers/enzyme-setup.js
```diff
@@ -1,3 +1,4 @@
+import 'regenerator-runtime/runtime';
```
cwillisf, line 1: Newer versions don't automatically include this

test/helpers/promise-barrier.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Make a Promise-based "barrier" that can be resolved or rejected externally.
+ * @example
+ * const foo = makePromiseBarrier();
+ * setTimeout(() => foo.resolve(42), 10000);
+ * const fortyTwo = await foo;
+ * @returns A promise with 'resolve' and 'reject' methods added.
+ */
+const makePromiseBarrier = function (): Promise<unknown> {
```
cwillisf, line 9: I'm open to suggestions for a better name

test/unit/components/__snapshots__/button.test.jsx.snap
```diff
@@ -3,11 +3,9 @@
...
 <span
   className=""
-  onClick={[Function]}
+  onClick={[MockFunction]}
   role="button"
 >
-  <div
-    className={undefined}
-  />
+  <div />
```
cwillisf, line 9: Snapshots in newer versions of Jest no longer list `undefined` properties and make a distinction between a mocked function vs. a non-mocked function.

test/unit/components/error-boundary-hoc.test.jsx
```diff
@@ -1,13 +1,13 @@
 import React from 'react';
 import {Provider} from 'react-redux';
-const {mountWithIntl} = require('../../helpers/intl-helpers.jsx');
+import {mountWithIntl} from '../../helpers/intl-helpers.jsx';
```
cwillisf, line 3: `import/no-commonjs` now applies to `test/` which seems nice for consistency

test/unit/components/error-boundary-hoc.test.jsx
```diff
@@ -1,13 +1,13 @@
...
 
 import configureStore from 'redux-mock-store';
 
 import CrashMessageComponent from '../../../src/components/crash-message/crash-message.jsx';
 import ErrorBoundary from '../../../src/containers/error-boundary.jsx';
 
-const ChildComponent = () => <div>hello</div>;
+const ChildComponent = () => <div>{'hello'}</div>;
```
cwillisf, line 10: `react/jsx-no-literals`

src/components/connection-modal/dots.jsx
```diff
@@ -5,6 +5,20 @@ import classNames from 'classnames';
...
 
+const Dot = props => (
+    <div
+        className={classNames(
+            styles.dot,
+            {
+                [styles.inactiveStepDot]: props.type === 'inactive',
+                [styles.activeStepDot]: props.type === 'active',
+                [styles.successDot]: props.type === 'success',
+                [styles.errorDot]: props.type === 'error'
```
hyperobject, line 16: not essential, but could be nice to type `props.type` as a string literal union or enum so typos are easier to catch

test/unit/util/audio-context.test.js
```diff
@@ -9,17 +13,14 @@ describe('Shared Audio Context', () => {
...
-        document.dispatchEvent(event);
-        expect(sharedAudioContext).toMatchObject(audioContext);
+    // TODO: support both instead of either/or (see shared-audio-context)
+    // then make this two separate tests
+    test('returns AudioContext when mousedown/touchstart is triggered', () => {
+        const sharedAudioContext1 = new SharedAudioContext();
+        expect(sharedAudioContext1).toMatchObject({});
+        document.dispatchEvent(new Event('mousedown'));
+        document.dispatchEvent(new Event('touchstart'));
+        const sharedAudioContext2 = new SharedAudioContext();
```
cwillisf, line 23: This test wasn't working before, but the Jest upgrade made it so that Jest now notices the failure 😅

The `SharedAudioContext` watches for either `mousedown` or `touchstart` (not both), depending on the characteristics of the browser, so one of the two tests should have failed in the previous configuration.

Also, `new SharedAudioContext()` grabs the value at the time of construction (function call) so neither test should have worked before.

Also also, we weren't resetting the state between tests, so the `mousedown` event could cause the `touchstart` test to pass or vice versa, depending on the order and timing of execution.

Ideally, `SharedAudioContext` should watch for both events, and we should reset the state (with `jest.resetModules()` or similar) in each test. That's extensive enough that I didn't want to make it part of this already-too-big PR.

test/unit/util/detect-locale.test.js
```diff
@@ -2,6 +2,10 @@ import {detectLocale} from '../../../src/lib/detect-locale.js';
 
 const supportedLocales = ['en', 'es', 'pt-br', 'de', 'it'];
 
+Object.defineProperty(window,
+    'location',
+    {value: new URL('https://example.org/'), configurable: true}
+);
```
cwillisf, line 8: Newer JSDOM no longer defines this property by default

test/unit/util/project-fetcher-hoc.test.jsx
```diff
@@ -55,10 +57,8 @@ describe('ProjectFetcherHOC', () => {
             storage.AssetType.Project, '100', storage.DataFormat.JSON
         );
         storage.load = originalLoad;
-        // nextTick needed since storage.load is async, and onFetchedProject is called in its then()
```
cwillisf, line 58: `storage.load` is not guaranteed to finish before the `process.nextTick` callback... it's just guaranteed to not finish before the next tick. The `await projectDidLoad` removes the timing ambiguity.

hyperobject (approved):
Nice. 😎 Tested and everything works and looks good, just a couple comments on some potential readability/dev ergonomics improvements!

hyperobject:
Would also be good at some point in the future (maybe once everything is able to be updated to a more modern react version) to ensure that typescript configs are shared across all our projects!

src/components/button/button.tsx
cwillisf: Makes sense. I'm relatively new to TypeScript so I went pretty minimal with this change, but if it's going to be an example then it should probably be a high-quality example ;)

cwillisf:
I agree! My not-so-secret plan is to work toward a Scratch editor monorepo where we'd have a natural, built-in way to share configs. I also think we're due for a major revision (both literally and semver-ly) in `eslint-config-scratch` to include some of this stuff. I think it'll make sense to do that as a next step.

cwillisf:
I tried to use this for some further work and discovered that linting is not set up quite right. I'm going to make this a draft PR for now, until I can get back to it and try to figure out how to fix the linting setup.

ghost:
Does this mean you could type instead of coding blocks?

CST1229:
> Does this mean you could type instead of coding blocks?

No, this is just a thing for Scratch's developers (and modifications).

ghost:
Ok.

PR #8956, opened 2023-05-17
Feature/settings menu

aoneill01:
### Resolves

- Resolves [ENA-311](https://scratchfoundation.atlassian.net/browse/ENA-311)

### Proposed Changes

- Moves language and theme menu items into submenus under Settings
- Update menu components and reducers to make submenus reusable
- Support RTL for submenus
- Hide menu labels when browser width is <= 1024
- Updates styling of File and Edit to match new Settings menu

Screenshot 2023-05-17 at 1 00 02 PM

Screenshot 2023-05-17 at 12 59 45 PM

### Reason for Changes

Some of the reasons include:
- The left side of the menu bar is a prime location. Language and theme will probably be changed infrequently after they are initially set.
- The settings menu could be expanded in the future with more options.

### Test Coverage

- Existing integration tests were update to handle the new styling for the File menu and new locations for the language and theme menus.
- New integration tests created around Settings submenu toggling and hiding of menu labels at small browser widths.

### Browser Coverage

Mac
 * [x] Chrome 
 * [x] Firefox 
 * [x] Safari
 
Windows
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [x] Safari

Android Tablet
* [ ] Chrome


src/components/menu/menu.css
```diff
@@ -44,3 +46,54 @@
...
+    border-radius: 8px;
+}
+
+.menu-item.expanded > .submenu {
+    display: block;
+}
+
+:not(:has(> .menu-item.expanded)) > .menu-item:hover > .submenu {
+    display: block;
+}
```
aoneill01, line 83: This styling is meant to expand the submenus when clicked or hovered. [:has() support](https://caniuse.com/css-has) does not cover all of our supported browsers. (Firefox is the obvious omission.) It wouldn't be impossible to circumvent this, perhaps adding a class to the parent of `.menu-item` based on whether any children are expanded. However, the hovering behavior is not critical. Users can still navigate the submenus by clicking. I could be convinced that identical behavior across browsers is more important than code simplicity...

src/components/menu-bar/settings-menu.css
```diff
@@ -0,0 +1,67 @@
...
+    padding-right: 1rem;
+}
+
+[dir="rtl"] .language-menu-item {
+    padding-left: 1rem;
+}
+
+.language-submenu > ul {
+    max-height: calc(100vh - 5rem); /* Fallback if dvh not supported */
+    max-height: calc(100dvh - 5rem);
```
aoneill01, line 60: My first time using [`dvh` units](https://developer.mozilla.org/en-US/docs/Web/CSS/length#relative_length_units_based_on_viewport). This worked better on Safari for iPad to keep the language options from going past the bottom of the page.

Secret-chest:
Care to make the custom dropdowns highlight the first item beginning with a letter when that letter is pressed?

aoneill01:
> Care to make the custom dropdowns highlight the first item beginning with a letter when that letter is pressed?

Good idea. This just came up in an internal discussion as well. We need to hammer out the details on the exact behavior, trying to mimic best practices with expected accessibility behavior. Right now the plan is to make it a separate task outside of this pull request.

src/components/menu-bar/menu-bar.css
```diff
@@ -241,8 +214,16 @@
     vertical-align: middle;
 }
 
+.collapsible-label {
+    margin: 0 .5rem 0 .25rem;
+}
+
+[dir="rtl"] .collapsible-label {
```
cwillisf, line 221: I feel like there should be some sort of annotation to say:
"instead of `top right bottom left`, treat this one as `top end bottom start`"
:P

src/components/menu-bar/menu-bar.jsx
```diff
@@ -900,17 +868,15 @@ MenuBar.propTypes = {
...
-    onChangeTheme: PropTypes.func,
-    modeMenuOpen: PropTypes.bool,
-    modeNow: PropTypes.bool,
-    mode220022BC: PropTypes.bool,
     mode1920: PropTypes.bool,
     mode1990: PropTypes.bool,
     mode2020: PropTypes.bool,
+    mode220022BC: PropTypes.bool,
+    modeMenuOpen: PropTypes.bool,
+    modeNow: PropTypes.bool,
```
cwillisf, line 879: 😎

src/reducers/menus.js
```diff
@@ -3,36 +3,99 @@ const CLOSE_MENU = 'scratch-gui/menus/CLOSE_MENU';
...
+        this.children.push(menu);
+        menu.parent = this;
+        return this;
+    }
+
+    descendants () {
+        return this.children.flatMap(child => [child, ...child.descendants()]);
+    }
+
+    sibblings () {
```
cwillisf:
```suggestion
    siblings () {
```

test/integration/menu-bar.test.js
```diff
@@ -124,4 +122,51 @@ describe('Menu bar settings', () => {
...
+        expect(await (await findByText('Esperanto', scope.menuBar)).isDisplayed()).toBe(false);
+
+        await clickText('Language', scope.menuBar);
+
+        // Only language options visible
+        expect(await (await findByText('High Contrast', scope.menuBar)).isDisplayed()).toBe(false);
+        expect(await (await findByText('Esperanto', scope.menuBar)).isDisplayed()).toBe(true);
+    });
+
+    test('Menu labels hidden when width is equal to 1024', async () => {
```
cwillisf, line 147: 🔬

cwillisf (approved):
Good stuff! I like shifting more toward XPath for tests, too.
I agree with your comments on hover. It seems like a nice-to-have.
I left one suggestion for a spelling change, but I don't think that'd need a re-review.

src/components/menu-bar/menu-bar.css
aoneill01: Totally! I think [`margin-inline`](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline) is basically what I want, but it looks like the browser support doesn't go far enough back.

src/reducers/menus.js
aoneill01: Good catch!

PR #8991, opened 2023-05-30
in-editor micro:bit update

cwillisf:
### Resolves

Resolves ENA-320

### Proposed Changes

Add to the Scratch editor the ability to update a micro:bit with the latest Scratch-compatible HEX file.

- Add an NPM `install` script to download the latest HEX file ZIP, extract the HEX file, and generate a Webpack-ready wrapper for it
- Implement micro:bit-specific update process using WebUSB through [DAP.js](https://www.npmjs.com/package/dapjs)
- Add an optional "update peripheral" path in the connection modal
- Hook up the micro:bit-specific update process to the "update peripheral" path
- Make the peripheral update flow pretty 🦋

### Reason for Changes

The folks at micro:bit report that an increased number of users are encountering this problem: https://support.microbit.org/support/solutions/articles/19000145576

They let us know that updating through WebUSB seems to work around the issue, so offering a way to do that within the Scratch editor should help people who encounter the issue.

It's also a nice convenience feature for everyone who uses Scratch with micro:bit, especially if they also use other programs with their micro:bit and need to repeatedly update their micro:bit back and forth.

### Test Coverage

Since this code requires Scratch Link as well as a physical micro:bit, it is not covered by tests. In fact, the connection modal in general is not tested, other than the error message indicating that Scratch Link is not available.

Someday, we should mock Scratch Link (and now DAP.js, I guess) so we can test all this stuff.

### Browser Coverage

NOTE: WebUSB is currently only supported in Chrome-family browsers (including Edge), so testing in other browsers is about making sure the update option is _not_ offered and that the UI still looks correct.

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [x] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


ghost:
Does this mean you could code a micro:bit using a USB?

CST1229:
> Does this mean you could code a micro:bit using a USB?

I don't think so.

ghost:
Ok.

cwillisf:
This is specifically about offerring an alternative to the "Getting Started" steps on this page: https://scratch.mit.edu/microbit
The way that a running Scratch project interacts with the micro:bit won't be affected by this change.

package-lock.json
```diff
@@ -7,10 +7,13 @@
     "": {
       "name": "scratch-gui",
       "version": "1.8.107",
+      "hasInstallScript": true,
```
cwillisf, line 10: Since this `hasInstallScript` line is adjacent to the version number, which is repeatedly changing in the `develop` branch, this PR will report merge conflicts most of the time. If this PR is approved, I'll resolve the conflict one last time before merging the PR.

ghost:
Ok.

package.json
```diff
@@ -14,6 +14,7 @@
     "build": "npm run clean && webpack --colors --bail",
     "clean": "rimraf ./build && mkdirp build && rimraf ./dist && mkdirp dist",
     "deploy": "touch build/.nojekyll && gh-pages -t -d build -m \"[skip ci] Build for $(git log --pretty=format:%H -n1)\"",
+    "install": "node scripts/install.mjs",
```
benjiwheeler: so does running `npm install` run this, and then also run the typical npm install stuff? Or if this is specifically for microbit, could a more specific command be used here, instead of the general-sounding "install"?

benjiwheeler:
Hm, am I doing something wrong, or looking in the wrong place? I'm not seeing an "update peripheral" path in the connection modal
image

benjiwheeler:
Ah, it just takes a while for that modal's state to give way to this

image

benjiwheeler:
Sweet! I was able to get the process to work!

The trickiest point was confirming that I'd installed the right firmware; the website seemed to suggest I should expect version 0255, the download file had 0249 in the name, but the MAINTENANCE drive's txt file had 0234 inside it after I installed... the MICROBIT drive's txt file had 0249, and I wasn't positive that was the right one, but it worked!

scripts/install.mjs
```diff
@@ -0,0 +1,121 @@
+// Install any non-NPM dependencies in this script.
```
benjiwheeler: I get how this works locally, and successfully ran it. But I don't quite understand how this works in production, when gui is being included as a static library?

src/components/connection-modal/auto-scanning-step.jsx
```diff
@@ -21,8 +24,13 @@ const PHASES = keyMirror({
     notfound: null
 });
 
-const AutoScanningStep = props => (
-    <Box className={styles.body}>
+const AutoScanningStep = props => {
+    // Offer to update both during scan and after a failed scan, as long there's an update function.
```
benjiwheeler, line 28: I didn't see an offer to update during the scan -- should I have?

Or do you mean, offer to update after a successful scan or a failed scan?

image

src/components/connection-modal/scanning-step.jsx
```diff
@@ -3,17 +3,21 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
 
+import BalancedFormattedMessage from '../../containers/balanced-formatted-message.jsx';
 import Box from '../box/box.jsx';
 import PeripheralTile from './peripheral-tile.jsx';
 import Dots from './dots.jsx';
 
+import enterUpdateIcon from './icons/enter-update.svg';
```
benjiwheeler, line 11: I took the liberty of running the new svg's through the tool svgo, which reduced their size about 30%

src/components/connection-modal/update-peripheral-step.jsx
```diff
@@ -0,0 +1,262 @@
...
+    renderResults () {
+        let resultsContent;
+        if (this.state.err === null) {
+            resultsContent = (<FormattedMessage
+                defaultMessage="Update successful!"
+                description="Message to indicate that the peripheral update was successful"
+                id="gui.connection.updatePeripheral.updateSuccessful"
+            />);
+        } else if (this.state.err.message === 'No valid interfaces found.') {
+            // this is a special case where the micro:bit's communication firmware is too old to support WebUSB
```
benjiwheeler, line 145: FWIW, I hit this step when I had taken one of our Micro:bit 1's and installed a MakeCode hex file onto it. (I guess the issue was the underlying firmware version, not the MakeCode hex)

webpack.config.js
```diff
@@ -131,9 +140,9 @@ module.exports = [
         },
         plugins: base.plugins.concat([
             new webpack.DefinePlugin({
-                'process.env.NODE_ENV': '"' + process.env.NODE_ENV + '"',
+                'process.env.NODE_ENV': `"${process.env.NODE_ENV}"`,
```
benjiwheeler, line 143: thanks, much clearer

benjiwheeler (approved):
LGTM! A few notes here and there

package.json
cwillisf: This runs automatically after (kind of "as part of") `npm install`

scripts/install.mjs
cwillisf: Hmm... you have a point. I should double-check whether that works as expected.

src/components/connection-modal/auto-scanning-step.jsx
cwillisf: The `AutoScanningStep` and the `ScanningStep` work a little differently for this. The `ScanningStep` makes the offer only after the scan is done, so you can update if the results don't include the one you're looking for. The `AutoScanningStep` will automatically connect to anything it finds, so using the same strategy there would mean that if there's another micro:bit nearby, you'll never get the opportunity to update yours. Because of that, it offers the update before the scan is done.

src/components/connection-modal/update-peripheral-step.jsx
cwillisf: Yep, I have a v1 micro:bit that I've intentionally kept on the old communications firmware just to test this :)
I don't believe this can happen with a v2

scripts/install.mjs
cwillisf: I moved this from `install` to `prepublish` based on the "Use Cases" section of the "Prepare and Prepublish" docs here: https://docs.npmjs.com/cli/using-npm/scripts, plus some experimentation. Thanks for catching this!

src/components/connection-modal/auto-scanning-step.jsx
benjiwheeler: Ah, that makes sense, thanks!

scripts/install.mjs
benjiwheeler: Sounds good! We should check on staging that this works as expected once it's merged and up there -- I want to make sure fastly passes the request through correctly, and doesn't need its settings revised.

benjiwheeler:
LGTM! See my note about testing the prepublished file on staging, to make sure Fastly is handling the request correctly!

benjiwheeler (approved):
LGTM! See my note about testing the prepublished file on staging, to make sure Fastly is handling the request correctly!

PR #9105, opened 2023-07-24
fix: update prepublish script to use correct path separator

aoneill01:
### Resolves

- Related to [ENA-304](https://scratchfoundation.atlassian.net/browse/ENA-304)

### Proposed Changes

- Replace `\` with `/` in generated `require` import for micro:bit hex updater

### Reason for Changes

I ran into this building the Windows version of `scratch-desktop`. Without this change, the generated `require` in `microbit-hex-url.cjs` looks like: 

```js
module.exports = require('./..\..\static\microbit\scratch-microbit-1.2.0.hex');
```

This causes the build to fail since the backslashes are not escaped.

### Test Coverage

Manual

### Browser Coverage

Mac
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


scripts/prepublish.mjs
```diff
@@ -93,13 +93,17 @@ const downloadMicrobitHex = async () => {
     const absoluteGeneratedDir = path.join(basePath, relativeGeneratedDir);
     fs.mkdirSync(absoluteGeneratedDir, {recursive: true});
     const absoluteGeneratedFile = path.join(basePath, relativeGeneratedFile);
+    const requirePath = `./${path
+        .relative(relativeGeneratedDir, relativeHexFile)
+        .split(path.win32.sep)
+        .join(path.posix.sep)}`;
```
aoneill01, line 99: `.replaceAll(path.win32.sep, path.posix.sep)` would have been cleaner in my opinion, but that was introduced in Node 15. Pretty sure part of our CI is using Node 14.

cwillisf (approved):
Looks good!
I feel like there should be a way to use `path.posix` to resolve this in a "path" sort of way instead of in a "string" sort of way, but I don't see an obvious solution.

scripts/prepublish.mjs
apple502j: Or, maybe just `.replace(/[\\/]g/, '/')` 😸

aoneill01: I just noticed that @xxtlichao did exactly that in #9033!

PR #9110, opened 2023-08-01
Replace react-contextmenu

cwillisf:
### Resolves

Resolves ENA-350, part of ENA-207.

### Proposed Changes

- Build new `useContextMenu` custom hook to implement a context menu in terms of React hooks
- Replace use of `react-contextmenu` with `useContextMenu` in the sprite tiles and monitors
- Update `selenium-helper` with the class names used by `useContextMenu` instead of those used by `react-contextmenu`
- Remove `react-contextmenu`

### Reason for Changes

The `react-contextmenu` library is no longer maintained and is not compatible with React 18.

### Test Coverage

Tests are currently broken due to the dependency on Enzyme, but I've updated the tests that I believe will be affected.

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


src/components/context-menu/context-menu.jsx
```diff
@@ -1,41 +1,129 @@
...
+ *     <MenuItem onClick={...}>Item 2</MenuItem>
+ * </>);
+ * return (<div onContextMenu={handleContextMenu}>
+ *     {"Right-click on this to open a menu"}
+ *     {contextMenu}
+ * </div>);
+ */
+const useContextMenu = contextMenuContent => {
+    const [contextMenuPosition, setContextMenuPosition] = useState(null);
+    const closeContextMenu = useCallback(() => setContextMenuPosition(null), [setContextMenuPosition]);
```
aoneill01, line 58: No change required, just a note in case you weren't aware. 

The `setContextMenuPosition` dependency is not required, and I think the linter is smart enough for it to be left out. 

[React - Is useState 's setter function able to change?](https://stackoverflow.com/questions/57180731/react-is-usestate-s-setter-function-able-to-change)

I can't find where the setter being stable is documented in the latest React docs, but I do see [code examples](https://react.dev/reference/react/useCallback#updating-state-from-a-memoized-callback).

src/components/monitor/monitor.jsx
```diff
@@ -40,97 +38,87 @@ const getCategoryColor = (theme, category) => {
...
-            </ContextMenu>
-        ), document.body)}
-    </ContextMenuTrigger>
-
-);
+            {React.createElement(modes[props.mode], {
+                categoryColor: getCategoryColor(props.theme, props.category),
+                ...props
+            })}
+            {contextMenu}
```
aoneill01, line 118: I am seeing this console warning the first time I right-click on a reporter:

> Warning: Failed prop type: Invalid prop `children` supplied to `Box`, expected a ReactNode.

Did some digging and I think this is because `prop-types` [can't validate a portal as a node](https://github.com/facebook/prop-types/issues/167). Looks like one workaround is to wrap it in a fragment like `<>{contextMenu}</>`. It is just a warning that will go away in production. I'm on the fence as to whether it is worth changing since technically nothing is wrong.

aoneill01 (approved):
Love how simple the context menu hook turned out. As usual, your comments are useful. Left a couple of comments that I'd be fine with as-is.

src/components/context-menu/context-menu.jsx
cwillisf: Interesting! I didn't know that. I got this pattern from an example -- I guess I should have dug into it a little more. Thanks!

Odd that this is in the [legacy hooks reference docs](https://legacy.reactjs.org/docs/hooks-reference.html#usestate) and the [hooks FAQ](https://legacy.reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often) but not the most recent version of the hooks documentation.

src/components/monitor/monitor.jsx
cwillisf: Aha! Thanks for that. I went ahead and applied that workaround to reduce warning noise... especially since the warning is 50+ lines long!

PR #9192, opened 2023-09-13
fix: adjust autosave interval

cwillisf:
### Resolves

Resolves ENG-74

### Proposed Changes

Adjust autosave interval from 120 to 600 seconds.

### Reason for Changes

This should still support our users well while reducing server load.

### Test Coverage

Tested locally, but testing in staging will be the "real" test.

### DIFF
--- a/src/lib/project-saver-hoc.jsx
+++ b/src/lib/project-saver-hoc.jsx
@@ -398,7 +398,7 @@ const ProjectSaverHOC = function (WrappedComponent) {
         vm: PropTypes.instanceOf(VM).isRequired
     };
     ProjectSaverComponent.defaultProps = {
-        autoSaveIntervalSecs: 120,
+        autoSaveIntervalSecs: 600, // 10 minutes = 600 seconds
         onRemixing: () => {},
         onSetProjectThumbnailer: () => {},
         onSetProjectSaver: () => {},


src/lib/project-saver-hoc.jsx
```diff
@@ -398,7 +398,7 @@ const ProjectSaverHOC = function (WrappedComponent) {
         vm: PropTypes.instanceOf(VM).isRequired
     };
     ProjectSaverComponent.defaultProps = {
-        autoSaveIntervalSecs: 120,
+        autoSaveIntervalSecs: 600, // 10 minutes = 600 seconds
```
cwillisf, line 401: The comment is mainly here to make it easier to find this line by text search

AlexisGoodfellow: How dare you make things easy to find? 🤣

AlexisGoodfellow (approved):
MAKE IT SO!

Secret-chest:
Any thoughts on using `localStorage` for backups and uploading once 15 minutes or so?

cwillisf:
@Secret-chest yes! I can't promise a timeline, of course, but we're certainly thinking about using local storage to support working offline, recovering from browser crashes, and other scenarios like that.

PR #9242, opened 2023-10-02
fix: disable metadata for now

cwillisf:
### Proposed Changes

Don't set metadata for now.

### Reason for Changes

Effectively backs out the storage changes without relying on `scratch-storage` version 2.3.0 or higher. Those newer versions are causing build problems with `scratch-gui` so this is a very temporary workaround.

See also: #9241

### Test Coverage

Existing tests


colbygk (approved):
LGTM

PR #9283, opened 2023-10-14
Improve test debugging

cwillisf:
### Proposed Changes

Various fixups and debugging aids for the tests in this repository:

- Fixed an error in two unit tests related to the mock web audio driver we use
- Lots of work in `selenium-helper.js` for readability and making sure we `await` properly
- Removed unnecessary steps in some tests that probably improved race condition reliability with missing `await`
- Added gobs of debug information to exceptions coming from `selenium-helper.js`

### Reason for Changes

This came from spending way too long debugging `scratch-gui` tests before figuring out I just needed scratchfoundation/scratch-storage#421

The extra debugging information helps compensate for the fact that `regenerator-runtime` loses call stack information when throwing from inside a Promise or `async` function. At first I wanted to ditch `regenerator-runtime` but that seems difficult when using Jest + React.

These changes might also help with GHA migration...

### Test Coverage

I've only tested the final version of `enhanceError` locally so far, but here's an example from some very similar code:
https://app.circleci.com/pipelines/github/scratchfoundation/scratch-gui/33151/workflows/f46ec0b5-b890-40ba-8849-c22d74839a7e/jobs/170297

I recommend scrolling to the top of the integration test output.

### DIFF
--- a/test/helpers/selenium-helper.js
+++ b/test/helpers/selenium-helper.js
@@ -4,7 +4,7 @@ import bindAll from 'lodash.bindall';
 import 'chromedriver'; // register path
 import webdriver from 'selenium-webdriver';
 
-const {By, until, Button} = webdriver;
+const {Button, By, until} = webdriver;
 
 const USE_HEADLESS = process.env.USE_HEADLESS !== 'no';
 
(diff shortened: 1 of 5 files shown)


delasare (approved):
LGTM! :shipit:

PR #9443, opened 2024-01-05
fix(deps): Use terser-webpack-plugin instead of uglifyjs-webpack-plugin

towerofnix:
### Proposed Changes

This PR does three things:

* Mark `terser-webpack-plugin@1.4.5` (an existing dependency, via webpack) as a saved dev dependency.
* Remove dev dependency `uglifyjs-webpack-plugin`.
* Replace `UglifyJsPlugin` with `TerserPlugin` in Webpack config.

### Reason for Changes

[`uglifyjs-webpack-plugin` is deprecated.](https://github.com/webpack-contrib/uglifyjs-webpack-plugin#readme) The docs indicate to use [`terser-webpack-plugin`](https://webpack.js.org/plugins/terser-webpack-plugin/) instead.

The version `terser-webpack-plugin@1.4.5` is selected as it's the version which was previously installed (as a dependency of `webpack@4.46.0`, matching `^1.4.3`). This means we don't actually add or update any dependencies: note how `package-lock.json` changes only include removing dependencies.

Marking `terser-webpack-plugin` as saved at all [is required](https://webpack.js.org/plugins/terser-webpack-plugin/):

> Webpack v5 comes with the latest `terser-webpack-plugin` out of the box. If you are using Webpack v5 or above and wish to customize the options, you will still need to install `terser-webpack-plugin`. Using Webpack v4, you have to install `terser-webpack-plugin` v4.

**Removing uglifyjs-webpack-plugin is necessary to update to Webpack v5.** uglifyjs-webpack-plugin has not been updated since 2019 and strictly requires Webpack v4. Terser is the replacement Webpack recommends, and is supported on both v4 and v5.

Webpack v4 itself is deprecated as well, and due to a compatibility change, prevents readily running on Node above 16. Node 16 left maintenance LTS support in October 2023, so the development environment is now depending on a wholly unsupported version of Node, partially due to Webpack (and consequently `uglifyjs-webpack-plugin`).

### Test Coverage

All existing tests pass(\*\), the build succeeds, and `npm start` is also functioning as normal.

(\*) I couldn't get most of the Jest tests to run and succeed, but suspect this is a device issue. The same tests fail to run for me building on latest `develop`.


cwillisf (approved):
Thanks again, @towerofnix !

PR #9536, opened 2024-03-14
Use scratch webpack config

cwillisf:
### Proposed Changes

Use our new `scratch-webpack-configuration` to standardize the way `scratch-gui` builds and fix the kinds of problems that are blocking scratchfoundation/scratch-gui#9533 and other updates.

### Reason for Changes

General dependency updates, especially in `scratch-svg-renderer`, and especially especially things related to `jsdom`, mean we need to change some things about the way we webpack. Some of it's related to webpack 5, but not all of it. I suspect we'll soon need to move to ESM for everything... this set of changes can probably be seen as a precursor to that.

I took the opportunity to make further changes. The main one is setting it up so that webpack can build "deeply" -- that is, when webpack builds (for example) `scratch-gui`, it'll access the source code of `scratch-vm` and `scratch-svg-renderer` (and more in the future) instead of building their bundled forms. This should give us a more optimal bundle overall. In particular, it will hopefully prevent problems where we're embedding multiple versions of the same Scratch library (like `scratch-svg-renderer`) into `scratch-gui` or `scratch-www`. That kind of issue was a major component of the "gray question mark issue" from a while back.

Some of the reasoning for the particulars of the webpack configuration is available here: https://github.com/scratchfoundation/scratch-webpack-configuration

There's some further cleanup to do, such as moving some of these settings into the shared configuration, but I think this is a good (and working!) time to pause for feedback.

### Test Coverage

Affects code already under test. Does not increase or decrease test coverage.

### Browser Coverage

_Coming soon_

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome


.babelrc
```diff
@@ -7,7 +7,7 @@
             "messagesDir": "./translations/messages/"
         }]],
     "presets": [
-        ["@babel/preset-env", {"targets": {"browsers": ["last 3 versions", "Safari >= 8", "iOS >= 8"]}}],
+        "@babel/preset-env",
```
cwillisf, line 10: Moved to `.browserslistrc`

package.json
```diff
@@ -12,7 +12,7 @@
   "main": "./dist/scratch-gui.js",
   "scripts": {
     "build": "npm run clean && webpack",
-    "clean": "rimraf ./build && mkdirp build && rimraf ./dist && mkdirp dist",
+    "clean": "rimraf ./build ./dist",
```
cwillisf, line 20: webpack makes its own directories, so we don't need to use `mkdirp` to do that.

package.json
```diff
@@ -155,7 +155,7 @@
       "src/test.js"
     ],
     "moduleNameMapper": {
-      "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/test/__mocks__/fileMock.js",
+      "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)\\??$": "<rootDir>/test/__mocks__/fileMock.js",
```
cwillisf, line 163: We were already telling `jest` to mock imports for these file types. This tells `jest` to keep mocking them even if they include an optional question mark at the end (`\??`, but with an extra layer of escaping).

src/lib/backpack/sound-payload.js
```diff
@@ -1,5 +1,5 @@
 // eslint-disable-next-line import/no-unresolved
-import soundThumbnail from '!base64-loader!./sound-thumbnail.jpg';
+import soundThumbnail from '!base64-loader!./sound-thumbnail.jpg?';
```
cwillisf, line 2: This makes me a bit sad. What I really wanted was a way to say, "hey webpack, please use the new fancy Asset Modules feature with all assets EXCEPT those handled by some other loader." That way, things like `import foo from '!base64-loader!./foo.png'` could use `base64-loader` and things like `import foo from './foo.png'` would use Asset Modules. I tried a few ways to hook that up and couldn't get it working, so instead I went with this: imports with certain file extensions (including `jpg` and `png`, etc.) are handled as Asset Modules only if they explicitly say so (with a query like `?url`) or have no query at all (not even `?`, an empty query).

src/lib/define-dynamic-block.js
```diff
@@ -1,7 +1,6 @@
 // TODO: access `BlockType` and `ArgumentType` without reaching into VM
 // Should we move these into a new extension support module or something?
-import ArgumentType from 'scratch-vm/src/extension-support/argument-type';
-import BlockType from 'scratch-vm/src/extension-support/block-type';
+import {ArgumentType, BlockType} from 'scratch-vm';
```
cwillisf, line 3: Reaching into a module to grab some arbitrary script is no longer allowed, so I adjusted `scratch-vm` to export these through the main entry point. In the future, I might make these named, non-default exports in `scratch-vm`'s `index.js` (once that's ESM) and/or make them separately-specified entry points in the `exports` section of `scratch-vm`'s `package.json`.

test/unit/util/themes.test.js
```diff
@@ -36,9 +36,13 @@ describe('themes', () => {
...
             serializeToString = jest.fn(() => 'mocked xml');
 
-            global.XMLSerializer = () => ({
-                serializeToString
-            });
+            class XMLSerializer {
+                constructor () {
+                    this.serializeToString = serializeToString;
+                }
+            }
```
cwillisf, line 43: Fixes an "is not a constructor" error

webpack.config.js
```diff
@@ -1,64 +1,56 @@
...
+                type: 'umd2'
             }
         },
-        {
-            test: /\.css$/,
-            use: [{
+        resolve: {
+            fallback: {
+                Buffer: require.resolve('buffer/'),
+                stream: require.resolve('stream-browserify')
```
cwillisf, line 34: These are no longer automatically polyfilled in webpack 5.

webpack.config.js
```diff
@@ -79,192 +72,154 @@ const base = {
...
+// build the examples and debugging tools in `build/`
+const buildConfig = baseConfig.clone()
+    .merge({
+        devServer: {
+            client: {
+                progress: true
             },
-            runtimeChunk: {
-                name: 'lib.min'
-            }
```
cwillisf: The previous version expressed more control over the way webpack would chunk shared code, but letting webpack figure it out leads to more optimal results. Also, using a single output (`lib.min`) as the chunk for the webpack runtime _and_ the `splitChunks` target _and_ a named entry point (old line 130) is no longer allowed by webpack 5. The new configuration is simpler and, IMO, better.

webpack.config.js
```diff
@@ -79,192 +72,154 @@ const base = {
...
+            {
+                from: 'extensions/**',
+                to: 'static',
+                context: 'src/examples'
+            }
+        ]
+    }));
+
+// Skip building `dist/` unless explicitly requested
+// It roughly doubles build time and isn't needed for `scratch-gui` development
```
cwillisf, line 186: On my main dev computer, building each target takes a little over 30 seconds, or about 1 minute if you do both (saves like 2-3 seconds).

webpack.config.js
```diff
@@ -1,64 +1,56 @@
-const defaultsDeep = require('lodash.defaultsdeep');
```
cwillisf, line 1: This file is slightly easier to view if you set "Hide whitespace" (or add `&w=1` to the URL)

PR #9649, opened 2024-08-26
ci: ensure tests use build output and simplify CI/CD workflow

cwillisf:
Thanks, @takaokouji, for making me take a closer look at the caching behavior in this workflow. Subtle and dangerous!

### Resolves

- Closes #9643 

### Proposed Changes

In the previous version, the `build/` directory was cached with this key:
```yaml
key: ${{ runner.os }}-build-${{ hashFiles('package-lock.json') }}
```

Later workflow steps would restore the build output with this same key.

The change proposed in this PR combines the steps together, eliminating the need to save and restore caches between steps.

### Reason for Changes

As @takaokouji observed in #9643, this means that if a commit or PR doesn't change `package-lock.json`, the cache key will not change. Two unrelated builds, possibly with very different content in `src/` or elsewhere, could share the same build cache. In this case, when the later testing steps of the two builds each restore the cache, they'll both be testing the same build output. One of them won't be testing what you expect!

Luckily, the deploy steps also used the same cache keys, so they also didn't deploy what you expect. I say "luckily" because the alternative is to release something that didn't run any tests.

All the caching and retrieval steps also took a lot of time, enabled only a tiny amount of extra parallelism (in the deploy steps), and added a lot of complexity to the workflow. Removing all of that means a faster, easier-to-understand workflow with no sneaky gremlins making us test or deploy the wrong thing.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -90,6 +90,7 @@ jobs:
               run: npx --no -- semantic-release
             - name: Deploy playground to GitHub Pages
               uses: peaceiris/actions-gh-pages@373f7f263a76c20808c831209c920827a82a2847 # v3
+              if: github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main'
               with:
                 github_token: ${{ secrets.GITHUB_TOKEN }}
                 publish_dir: ./build
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1,8 +1,10 @@
 name: CI/CD
 
 on:
-    workflow_dispatch:
-    push:
+    pull_request: # Runs whenever a pull request is created or updated (including from another fork)
+    push: # Runs whenever a commit is pushed to the repository...
+        branches: [master, develop, hotfix/*] # ...on any of these branches
+    workflow_dispatch: # Allows you to run this workflow manually from the Actions tab
 concurrency:
     group: "${{ github.workflow }} @ ${{ github.head_ref || github.ref }}"
     cancel-in-progress: true
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -14,7 +14,7 @@ permissions:
     pull-requests: write # comment on released pull requests
 
 jobs:
-    setup:
+    ci-cd:
         runs-on: ubuntu-latest
         env:
             DETECT_CHROMEDRIVER_VERSION: "true"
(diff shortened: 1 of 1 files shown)


cwillisf:
For my first test, this new workflow took about 10 minutes to run. The previous workflow was more like 20 minutes.

cwillisf:
OK, sorry for the churn. This is genuinely ready now, I think :sweat_smile: 

Commit dea503f was because I realized that the build & test workflow didn't run on this PR, and we do want to test PRs!

Commit 7bb299b was because, once the build & test workflow did run on this PR, I realized that it was attempting to run the deploy steps. That's good for `semantic-release` since deploying from the "wrong" branch causes a dry run, effectively a config test, but it's bad for the GH Pages deploy since that fails with a permission error.

.github/workflows/ci-cd.yml
```diff
@@ -135,55 +65,20 @@ jobs:
...
             - name: Run Integration Tests
+              env:
+                  JEST_JUNIT_OUTPUT_NAME: results.txt
+                  JEST_JUNIT_OUTPUT_DIR: test-results/integration
               run: npm run test:integration -- --reporters="default" --reporters="jest-junit"
             - name: Store Integration Test Results
               uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3
               with:
                 name: integration-test-output
                 path: ./test-results/*
```
MiroslavDionisiev: With the the env variables as they are the integration tests artifact would contain both unit and integration tests result.

MiroslavDionisiev (approved):

.github/workflows/ci-cd.yml
cwillisf: Ah, good point - thanks!

MiroslavDionisiev:
I think it is safe to cache `node-modules` as if `package-lock.json` is not updated we would not expect any change there.

cwillisf:
The `npm ci` command deletes everything in the `node_modules` directory, so caching that ends up being ineffective. The `setup-node` action caches the files used by `npm ci`:
https://github.com/actions/setup-node?tab=readme-ov-file#caching-global-packages-data

.github/workflows/ci-cd.yml
cwillisf: I combined the two steps for storing test results, so now there's one artifact that contains any/all output from both testing steps. I set it to run "always" so that (for example) if the unit tests have output but the build fails, the unit test output is still available. That also means any output from failed tests will be available, which probably should have been the way we set it up to begin with!

PR #9753, opened 2024-11-26
resizable stage

DanFessler:
https://github.com/user-attachments/assets/10e5f8d4-1a68-41b7-b147-0b9a246d8182

### Proposed Changes

This PR implements a drag-resizable stage panel which is also used as the "large" stage toggle. This is a purely additive quality-of-life feature that shouldn't change any behaviors users expect.

### Reason for Changes

With higher DPI displays becoming more common on PCs and Tablets, the stage size options are a less-than-ideal "small" or "tiny." Often this means users rely on full-screen mode to check their work in more detail, but at the cost of losing the context of your code editor making iteration slow and annoying.

The solution is to have a more responsive user-customizable approach to displaying the stage.  Users can set the stage size to whatever makes sense for their working environment.  We can preserve the toggle functionality for users by allowing the "large" setting to be the user-defined draggable size.

This refactors some old code to be more flexible using react hooks to observe the size of the container elements to determine the necessary scale whereas the prior approach had a lot of hard-coded assumptions about the two size modes.

### Resolves

While this PR isn't meant to be a bug-fix, it does address some undocumented issues:
- Fixes dragging monitors in the "small" stage mode
- Fixes the stage size when in full-screen mode to adapt to screen when resized

### Browser Coverage

Mac
 * [ ] Chrome 
 * [ ] Firefox 
 * [ ] Safari
 
Windows
 * [x] Chrome 
 * [x] Firefox 
 * [x] Edge
 
Chromebook
 * [ ] Chrome
 
iPad
* [ ] Safari

Android Tablet
* [ ] Chrome

### DIFF
--- a/src/components/box/resizeableBox.jsx
+++ b/src/components/box/resizeableBox.jsx
@@ -1,0 +1,107 @@
+import PropTypes from 'prop-types';
+import React, {useEffect, useRef, useState, useCallback} from 'react';
+
+import Box from './box.jsx';
+
+const useMouseDrag = (ref, onDrag, onDragEnd) => {
+    useEffect(() => {
+        if (!onDrag) return;
+        let isDragging = false;
+        const onMouseMove = e => {
+            e.preventDefault();
+            if (!isDragging) return;
+            if (onDrag) onDrag(e);
+        };
+        const onMouseUp = e => {
+            isDragging = false;
+            if (onDragEnd) onDragEnd(e);
+        };
+        const onMouseDown = e => {
+            e.preventDefault();
+            isDragging = true;
+        };
+        ref.current.addEventListener('mousedown', onMouseDown);
+        window.addEventListener('mousemove', onMouseMove);
+        window.addEventListener('mouseup', onMouseUp);
+        return () => {
+            ref.current.removeEventListener('mousedown', onMouseDown);
+            window.removeEventListener('mousemove', onMouseMove);
+            window.removeEventListener('mouseup', onMouseUp);
+        };
+    }, [onDrag]);
+};
+
+
+const ResizableBox = props => {
+    const {
+        children,
+        minWidth = 420,
+        maxWidth = 480,
+        defaultWidth = 480,
+        onResize,
+        initialSize,
+        ...rest
+    } = props;
+
+    const boxRef = useRef(null);
+    const handleRef = useRef(null);
+    const [width, setWidth] = useState(defaultWidth);
+
+    // whenever the initial size changes, lets override whatever the current width is
+    useEffect(() => {
+        setWidth(initialSize);
+    }, [initialSize]);
+
+    // This is a workaround to force other elements to resize that
+    // are listening to the window resize event (such as the editor).
+    useEffect(() => {
+        window.dispatchEvent(new Event('resize'));
+    }, [width]);
+
+    const onDrag = useCallback(e => {
+        if (!boxRef.current) return;
+        const rect = boxRef.current.getBoundingClientRect();
+        const newWidth = Math.max(Math.min(rect.width - e.movementX, maxWidth), minWidth);
+        onResize(newWidth);
+        setWidth(newWidth);
+    }, [minWidth, maxWidth]);
+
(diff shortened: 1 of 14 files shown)


src/components/box/resizeableBox.jsx
DanFessler, line 1: This is a new "Box" wrapper which gives a drag-resizable handle on the left side of the container

src/components/gui/gui.css
```diff
@@ -207,9 +207,6 @@
     /* pad entire wrapper to the left and right; allow children to fill width */
     padding-left: $space;
     padding-right: $space;
-
-    /* this will only ever be as wide as the stage */
-    flex-basis: 0;
```
DanFessler, line 212: needed to get rid of this to have dynamic sizing

src/components/monitor-list/monitor-list.jsx
```diff
@@ -8,15 +8,32 @@ import {stageSizeToTransform} from '../../lib/screen-utils';
...
+            width: 480,
+            height: 360,
+            position: 'absolute',
+            top: 0,
+            left: 0,
+            visibility: 'hidden'
+        }}
+    />
+);
+
```
DanFessler, line 27: react-draggable really did not like using a scaled container as the bounding reference, but we needed the scaling both visually as well as for proper drag deltas for the monitors.  This dummy div of a constant width is a hacky solution that makes react-draggable happy.  A better solution would likely be way more involved, perhaps replacing react-draggable entirely.

src/components/monitor/monitor.jsx
```diff
@@ -52,6 +52,7 @@ const MonitorComponent = props => (
             defaultClassNameDragging={styles.dragging}
             disabled={!props.draggable}
             onStop={props.onDragEnd}
+            scale={props.scale}
```
DanFessler, line 55: drilling this scale prop to the draggable element corrects the drag deltas for different stage sizes

src/components/sprite-info/sprite-info.jsx
DanFessler, line 1: My changes to this file are purely superficial. now that the stage panel can get extra wide, it just looked silly to have the input fields spread across the entire width, so I made containers that act as a max-width for the controls

src/components/stage/stage.jsx
DanFessler, line 1: most of this work (and it's corresponding css) just enables the stage canvas to be the maximum size it can without overflowing its container.

Here we're also using the new useCanvasSize hook to get the dimensions we need

src/containers/stage.jsx
DanFessler, line 1: we need to update the canvas resolution when we resize the stage

src/lib/screen-utils.js
DanFessler, line 1: replacing getStageDimensions with a more flexible react hook

DanFessler:
I have read the CLA Document and I hereby sign the CLA

DanFessler:
Any feedback on this, one way or the other?

raynecloudy:
i have a few questions.
- what's the maximum size the stage can be?
- is it possible to hide the entire sprite pane / code editor?
- what happens when you resize the window? does the stage resize accordingly?

raynecloudy:
it gets a little glitchy at the bottom of the block editor when i'm resizing
Screenshot 2025-01-16 102002
i was able to hide the entire sprite pane
Screenshot 2025-01-16 102038
Screenshot 2025-01-16 102529
costume editor camera moves in weird ways when resizing
Screenshot 2025-01-16 102149
stage gets much smaller than usual and breaks the UI with small windows
Screenshot 2025-01-16 102331
turbo mode with smallest stage setting breaks UI a little
Screenshot 2025-01-16 102746
but this is a really cool idea and you've done better than i could do and if these bugs get fixed this could be a fantastic addition to scratch

mxmou:
> turbo mode with smallest stage setting breaks UI a little
> Screenshot 2025-01-16 102746

That's an existing issue - here's Scratch Desktop v3.29.1 in small stage mode:

image

